PR #49, opened 2019-04-14
Add support for descriptions

### LINKED ISSUE #20: Add support for description in fields
```python
@strawberry.type
class Query:
    @strawberry.field
    async def todos(self, info) -> typing.List[str]:
        """A list of TODOs"""
        return ["a", "b"]
```

```python
@strawberry.type
class Query:
    @strawberry.field(description="A list of TODOs")
    async def todos(self, info) -> typing.List[str]:
        return ["a", "b"]
```

We also need a way to define descriptions on simple fields

```python
@strawberry.type
class TODO:
    id: strawberry.ID
    """This would be unique"""
```

```python
@strawberry.type
class TODO:
    id: str = strawberry.field(description="This would be unique")
```

See:
- https://www.python.org/dev/peps/pep-0257/
- https://stackoverflow.com/a/20529741/169274
- http://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoattribute

patrick91:
Closes #20

### DIFF
--- a/strawberry/enum.py
+++ b/strawberry/enum.py
@@ -7,7 +7,7 @@ from .exceptions import NotAnEnum
 from .type_converter import REGISTRY
 
 
-def _process_enum(cls, name="", description=""):
+def _process_enum(cls, name=None, description=None):
     if not isinstance(cls, EnumMeta):
         raise NotAnEnum()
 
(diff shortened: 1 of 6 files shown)


patrick91:
Merging! Thanks for the help :)

Also, thanks to @paulmonk for suggesting to use docstrings <3

PR #51, opened 2019-04-24
Automatic Deployment using CircleCI

marcoacierno:
## How does this work

Every time a commit is pushed to CircleCI the pipeline checks if a file named `RELEASE.md` exists and if it follows the correct format.

The first line of a RELEASE.md should be the release type, followed by the changelog.

Example of a `RELEASE.md`:

```
Release type: patch

Here you can insert all the changes you want. Multiple lines supported, markdown supported too.
```

Release type can be of the following values: `patch`, `minor` and `major`.

The release type is used to decide which part of Strawberry version to bump (major => x.0.0, minor => y.x.0, patch => y.y.x) (where `x` will be the incremented version, and `y` the current).

The changelog part (everything below the release type) is copied and inserted into `CHANGELOG.md` with the version and current date.

If everything is correct, the library is built and a new tag with the current version (example `0.5.6`) is created together with a new GitHub's release, where the changelog will used as a description of the release.

A new commit with the title `Release xxx` (where `xxx` is the new version) is created with the changes to the CHANGELOG.md, pyproject.toml and removing the RELEASE.md file.

Once everything completed with success, we release the library on PyPi using `poetry publish` aaand it's over :)

## TODO

We need to decide:

- [x] Username and email of the git commit

## How to configure CircleCI:

- [x] Generate an SSH key for CircleCI using `ssh-keygen -m pem`.
- [x] Add the public key in the `GitHub` > `Settings` > `Deploy keys` section
- [x] Add the private key in the CircleCI `SSH Permissions`. Using `github.com` as `HOST`.
- [ ] NOT REQUIRED: `add_ssh_keys` adds by default all the keys, we can manually add one key by passing the fingerprint as parameter
```      
- add_ssh_keys:
     fingerprints:
       - "54:7e:55:49:d3:5b:67:d6:12:d0:a0:51:ed:d7:81:8c"
```
- [x] On CircleCI add the following env. variables:
   - `GITHUB_TOKEN`: Used to create the release on GitHub
   - `PYPI_USERNAME`: Used to publish the library
   - `PYPI_PASSWORD`: Used to publish the library


marcoacierno:
I wonder if we want to also implement a bot that checks all the PR merging to `master` to see if the `RELEASE.md` file exists and is valid, what do you think?

.circleci/config.yml
```diff
@@ -66,6 +66,19 @@ jobs:
...
+    steps:
+      - add_ssh_keys
+      - checkout
+      - run: make check-has-release
+      - run: make install-deploy-deps
+      - run: make prepare-release
+      - run: poetry build
+      - run: make create-github-release
+      - run: make publish-changes
+      - run: poetry publish --username $PYPI_USERNAME --password $PYPI_PASSWORD
```
patrick91: doesn't look like we can publish using a token, so I guess we should create a new user for this :)

scripts/base.py
```diff
@@ -0,0 +1,80 @@
...
+    subprocess.check_output(["git", "rev-parse", "--show-toplevel"])
+    .decode("ascii")
+    .strip()
+)
+PROJECT_NAME = "Strawberry"
+
+# Files configuration
+
+RELEASE_FILE_NAME = "RELEASE.md"
+RELEASE_FILE = os.path.join(ROOT, RELEASE_FILE_NAME)
```
patrick91: nitpick, maybe it would be nicer to use [Pathlib](https://docs.python.org/3/library/pathlib.html) since we are on python 3 :)

scripts/base.py
```diff
@@ -0,0 +1,80 @@
...
+GIT_USERNAME = "USERNAME_HERE"
+GIT_EMAIL = "EMAIL_HERE"
+
+
+def run_process(popenargs):
+    return subprocess.check_output(popenargs).decode("ascii").strip()
+
+
+def git(popenargs):
+    # avoid decode ascii for the commit message
```
patrick91: what does this mean? :)

scripts/base.py
```diff
@@ -0,0 +1,80 @@
...
+PROJECT_TOML_FILE_NAME = "pyproject.toml"
+PROJECT_TOML_FILE = os.path.join(ROOT, PROJECT_TOML_FILE_NAME)
+
+CHANGELOG_FILE_NAME = "CHANGELOG.md"
+CHANGELOG_FILE = os.path.join(ROOT, CHANGELOG_FILE_NAME)
+
+# Git configuration
+
+GIT_USERNAME = "USERNAME_HERE"
+GIT_EMAIL = "EMAIL_HERE"
```
patrick91: do we need environment variables for these?

scripts/has_release.py
```diff
@@ -0,0 +1,12 @@
...
+
+sys.path.append(os.path.dirname(__file__))  # noqa
+
+from base import run_process, RELEASE_FILE
+
+
+if __name__ == "__main__":
+    if not os.path.exists(RELEASE_FILE):
+        print("Not releasing a new version because there isn't a RELEASE.md file.")
+        run_process(["circleci", "step", "halt"])
```
patrick91: TIL :) will this fail the build, or just stop it?

marcoacierno: The build will stop with a success status

scripts/base.py
marcoacierno: It depends, I think we can keep those in the code as they will be public anyway?

scripts/base.py
marcoacierno: If we want to use (for example) emojis in the commit message we cannot use `run_process` because the output is converted to ascii using `decode('ascii')`, hence the new utility only for git

scripts/base.py
marcoacierno: sure, will do it soon

scripts/base.py
patrick91: good point :)

scripts/base.py
patrick91: got it :)

tswicegood:
Came across this repo from GraphQL Weekly and thru some random browsing came across this PR. This is some pretty awesome code. Any thoughts about taking this and making it a helper package that could be shared amongst multiple repos?

scripts/base.py
marcoacierno: I will write a better comment now

patrick91:
@tswicegood sounds good actually :) I guess that's up to @marcoacierno to decide as the code has been written by him, but I like the idea of having something that can be reused in other repositories :)

marcoacierno:
@tswicegood Yes I like the idea :) I will have a look at this weekend, we will need to make it generic to use other stuff other than poetry. Thank you :)

tswicegood:
Sweet! Feel free to ping me @marcoacierno if ya want an extra set of eyes on the code. I'm a big fan of Circle CI. 😄

scripts/base.py
marcoacierno: done!

PR #59, opened 2019-05-10
Add support for passing a resolver function when declaring a field

patrick91:
TODO: 

- [x] Check if it has correct annotations (return and arguments)
- [x] Support for arguments (now they are ignored)

### DIFF
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -37,18 +37,18 @@ class LazyFieldWrapper:
         self.kwargs = kwargs
 
         if callable(self._wrapped_obj):
-            self._check_has_annotations()
+            self._check_has_annotations(self._wrapped_obj)
 
-    def _check_has_annotations(self):
+    def _check_has_annotations(self, func):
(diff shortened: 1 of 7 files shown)


patrick91:
Going to do some refactoring while adding support for this, bear with me :)

strawberry/field.py
```diff
@@ -103,19 +118,46 @@ class strawberry_field:
...
+        self.field_description = description
+        self.field_resolver = resolver
         self.is_subscription = is_subscription
-        self.description = kwargs.get("description", None)
-        self.name = kwargs.pop("name", None)
-        self.kwargs = kwargs
+        self.is_input = is_input
+
+        super().__init__(
+            # TODO:
```
marcoacierno: What is this TODO left here?

patrick91: I need to support default values when using field, forgot about it :D

patrick91:
@marcoacierno I think I can merge this, do you want to have one final look? :)

PR #82, opened 2019-06-06
Add support for optional arguments when a default value is provided

jaydenwindle:
This PR adds support for marking an argument as optional when a default value is provided, resolving #9.

A couple of comments on this approach:
1) It relies on the return value of `typing.get_type_hints(...)` maintaining the order of the arguments when iterated through. I haven't been able to find this behavior documented, but it seems to be the case in the testing I've done. 
2) It makes use of the `force_optional` argument in `get_graphql_type_for_annotation`. I noticed some comments mentioning that `force_optional` might have some edge cases that would lead to unwanted behavior. Is there a better way to mark a field as optional?

### DIFF
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -202,6 +202,7 @@ def _get_field(
     name = wrap.__name__
 
     annotations = typing.get_type_hints(wrap, None, REGISTRY)
+    default_argument_start_index = len(annotations) - len(wrap.__defaults__) - 1
     field_type = get_graphql_type_for_annotation(annotations["return"], name)
 
     arguments_annotations = {
@@ -211,8 +212,10 @@ def _get_field(
     }
 
     arguments = {
-        to_camel_case(name): get_graphql_type_for_annotation(annotation, name)
-        for name, annotation in arguments_annotations.items()
+        to_camel_case(name): get_graphql_type_for_annotation(
+            arguments_annotations[name], name, index >= default_argument_start_index
+        )
+        for index, name in enumerate(arguments_annotations.keys())
     }
 
     def resolver(source, info, **args):
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -1,3 +1,4 @@
+import inspect
 import typing
 
 import dataclasses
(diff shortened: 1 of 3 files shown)


strawberry/field.py
```diff
@@ -210,10 +210,20 @@ def _get_field(
         if key not in ["info", "return"]
     }
 
-    arguments = {
-        to_camel_case(name): get_graphql_type_for_annotation(annotation, name)
-        for name, annotation in arguments_annotations.items()
-    }
+    if wrap.__defaults__:
```
patrick91: oh, I didn't know about this :D

strawberry/field.py
```diff
@@ -210,10 +210,20 @@ def _get_field(
...
+        default_argument_start_index = len(annotations) - len(wrap.__defaults__) - 1
```
patrick91: what do you think of extracting this into an utility function? Feels a bit hard to understand right now, we could have something like: `def is_argument_optional(func: Callable, argument_name: str) -> bool: ...`, it should also prevent the duplication of code :)

Have you considered using `inspect.signature`? I did a quick test and seems also another way of checking if an argument has a default value:

image

tests/test_field.py
```diff
@@ -29,6 +29,24 @@ def hello(self, info, id: int) -> str:
     assert hello.field.args["id"].type.of_type.name == "Int"
 
 
+def test_field_default_arguments_are_optional():
+    @strawberry.field
+    def hello(self, info, test: int, id: int = 1, asdf: str = "hello") -> str:
```
patrick91: well done! I'd also add a test where the default value is `None`, just in case :D

jaydenwindle: Good call!

strawberry/field.py
jaydenwindle: Oh cool, I didn't know about this method! Ill re-write the logic using `inspect.signature` - it looks like it would be a lot cleaner.

jaydenwindle:
@patrick91 I've updated this PR with the changes you mentioned, and I've added a test case for default arguments with custom types. Let me know what you think :)

patrick91:
@jaydenwindle that's really neat! thanks for that, sorry I should have mentioned this before (and document it somewhere), if you add a `RELEASE.md` file as soon as we merge the PR we'll get a release out on PyPI :)

Here's an example for this: https://github.com/strawberry-graphql/strawberry/pull/76/files#diff-781a53e648f3df8d16a08ec083b04bf4

I think we can release this as a minor as it is a new feature, what do you think? :)

🎉

jaydenwindle:
@patrick91 Sounds great, will do!

PR #85, opened 2019-06-07
Add support for extensions + Add Apollo Tracing support

### LINKED ISSUE #40: Support for Apollo Tracing
This should allow tracing performance problems in APIs made with Strawberry:

https://github.com/apollographql/apollo-tracing

jaydenwindle:
This PR adds [Apollo Tracing](https://github.com/apollographql/apollo-tracing) support. It builds on the tracing middleware implementation that was proposed in `graphql-core` (see https://github.com/ekampf/graphql-core/tree/feature/tracing)

Fixes #40 

EDIT(Patrick):

I've changed this PR to have a generic support for extensions, that you can pass to the schema, like this:

```python
@strawberry.type
class Person:
    name: str = "Jess"

@strawberry.type
class Query:
    @strawberry.field
    def person(self, info) -> Person:
        return Person()

schema = strawberry.Schema(query=Query, extensions=[ApolloTracingExtension()])
```

an extensions needs to follow this protocol:

```python
class Extension:  # pragma: no cover
    def on_request_start(self):
        ...

    def on_request_end(self):
        ...

    def on_validation_start(self):
        ...

    def on_validation_end(self):
        ...

    def on_parsing_start(self):
        ...

    def on_parsing_end(self):
        ...

    def on_resolver_start(self):
        ...

    def on_resolver_end(self):
        ...

    def resolve(self, _next, root, info, *args, **kwargs):
        ...

    def get_results(self):
        return {}
```


patrick91:
@jaydenwindle this is so cool! I haven't done a PR yet but I'm working on a new structure for the ASGI app, would be ok in "porting" this PR to the new implementation when done? I'll ping you here when I'm done :)

jaydenwindle:
Sure, happy to!

strawberry/contrib/starlette/app/graphql_app.py
```diff
@@ -69,16 +72,38 @@ def __call__(self, scope: Scope) -> ASGIInstance:
 
         background = BackgroundTasks()
         context = {"request": request, "background": background}
+        additional_arguments = {}
+        extensions = {}
+
+        if self.tracing:
```
patrick91: what if we allow users to pass the list of middlewares when instantiating the GraphQL app?

patrick91: ok, this might not be feasible since we need the tracing.start and .end

strawberry/middleware.py
```diff
@@ -0,0 +1,98 @@
+import asyncio
+import time
+
+
+class TracingMiddleware(object):
```
patrick91: let's remove (object) ;)

strawberry/middleware.py
```diff
@@ -0,0 +1,98 @@
...
+    DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
+
+    def __init__(self):
+        self.resolver_stats = list()
```
patrick91: [] instead of list? :)

strawberry/middleware.py
```diff
@@ -0,0 +1,98 @@
...
+        return self.end_time - self.start_time
+
+    @property
+    def tracing_dict(self):
+        result = dict(
+            version=1,
+            startTime=self.start_time_str,
+            endTime=self.end_time_str,
+            duration=self.duration,
+            execution=dict(resolvers=self.resolver_stats),
```
patrick91: we could use {} instead of dict()

same for the ones below :)

strawberry/middleware.py
```diff
@@ -0,0 +1,98 @@
...
+
+            stat = {
+                "path": info.path,
+                "parentType": str(info.parent_type),
+                "fieldName": info.field_name,
+                "returnType": str(info.return_type),
+                "startOffset": self.now() - self.start_time,
+                "duration": elapsed_ns,
+            }
+            self.resolver_stats.append(stat)
```
patrick91: since we love dataclasses, shall we use them instead of the dicts we have here? it will make the code much nicer :D

strawberry/contrib/starlette/app/graphql_app.py
patrick91: I'm wondering if we should expose a way for users to be able to this themselves, I think for now let's keep this hard-coded, and we can improve it later ;)

jaydenwindle: Ok, sounds good!

botberry:

```markdown
Release type: patch

```

strawberry/asgi/http.py
```diff
@@ -43,15 +48,36 @@ async def get_http_response(
...
+        query,
+        variables=variables,
+        context=context,
+        operation_name=operation_name,
+        **additional_arguments
     )
 
+    if tracing_enabled:
+        tracing.end()
+        extensions["tracing"] = tracing.tracing_dict
```
patrick91, line 73: I'm not enjoying these ifs, might be worth having some hooks for these things

patrick91:
Added support for extensions, will re-check the code tomorrow. I need to update ASGI, Flask and Django to support the new way of using extensions (I think I'll remove the middleware option for now)

strawberry/extensions/tracing/apollo.py
```diff
@@ -0,0 +1,146 @@
...
+            "validation": self.validation.to_json(),
+            "parsing": self.parsing.to_json(),
+        }
+
+
+class ApolloTracingExtension(Extension):
+    def __init__(self):
+        self._resolver_stats: typing.List[ApolloResolverStats] = []
+
+    def on_request_start(self):
```
patrick91, line 87: add params here (ie. query, operation name)

patrick91: context as well

strawberry/extensions/__init__.py
```diff
@@ -0,0 +1,92 @@
+import functools
+import typing
+from contextlib import contextmanager
+
+from graphql import MiddlewareManager
+
+
+class Extension:  # pragma: no cover
```
BryceBeagle, line 8: Should this be abstract?

patrick91: In terms of extending the abc base class?

BryceBeagle: yeah

patrick91: Probably! Will update the code tomorrow, was going to write some docs for this first, but I'm too tired now ☺️

patrick91: @BryceBeagle I changed the class to be used as baseclass, the empty methods are fine now, they shouldn't be abstract as they are always called :)

patrick91 (approved):

PR #93, opened 2019-06-21
Add coverage report when running tests

patrick91:


patrick91:
Not sure why it doesn't show the link to coverage, but it is here: https://codecov.io/gh/strawberry-graphql/strawberry/commit/e2b2a813257db23845afc1f4b7f2c09236cbe15c

patrick91:
@marcoacierno will do that with another PR :D

PR #95, opened 2019-06-24
Pass query type as root value

ethe:
In current, the type of Query can not use instance method in resolver cause the `self` always be `None`.

```python
import strawberry


@strawberry.type
class User:
    name: str
    age: int


@strawberry.type
class Query:
    @strawberry.field
    def user(self, info) -> User:
        # raise exception here
        return self.get_user()

    def get_user(self):


```

Pass Query type as root value may solve this problem I think.

### DIFF
--- a/strawberry/contrib/starlette/app/graphql_app.py
+++ b/strawberry/contrib/starlette/app/graphql_app.py
@@ -99,4 +99,5 @@ class GraphQLApp(BaseApp):
             variable_values=variables,
             operation_name=operation_name,
             context_value=context,
+            root_value=self.schema.query_type
         )


strawberry/contrib/starlette/app/graphql_app.py
```diff
@@ -99,4 +99,5 @@ def __call__(self, scope: Scope) -> ASGIInstance:
             variable_values=variables,
             operation_name=operation_name,
             context_value=context,
+            root_value=self.schema.query_type
```
patrick91: I'm wondering if we should allow users to pass a custom root value 🤔 
what do you think?

jaydenwindle:
I was able to reproduce this error locally. Here is the error that gets thrown:
```json
{
  "error": {
    "data": null,
    "errors": [
      {
        "message": "'NoneType' object has no attribute 'get_user'",
        "locations": [
          [
            2,
            3
          ]
        ],
        "path": [
          "user"
        ]
      }
    ]
  }
}
```

It appears that when `root_value` is not specified in the call to `execute`, its value is `None` and it is passed into the resolver function as `self`, causing the error. 

To fix this issue, the default behavior of the `strawberry.field` or `strawberry.type` decorator (I'm not sure which one handles this logic) should be changed so that the parent `strawberry.type` class is passed to the resolver as `self` if `root_value` is `None`.

strawberry/contrib/starlette/app/graphql_app.py
jaydenwindle: It looks like this will pass in the `Query` type as a root value for all requests, so using `self` in mutation or subscription resolvers would likely still throw an error.

patrick91:
@jaydenwindle ah, I see. Not sure how I'd handle this, maybe we should copy graphene's behaviour and allow to pass a custom root value, like here:

https://github.com/graphql-python/graphene-django/blob/master/graphene_django/views.py#L72

ethe:
I find that it is really hard to walk around with graphene-core-next, so I just create another GraphQL implementation: https://github.com/ethe/pygraphy

patrick91:
@ethe that's really cool, wondering if we could work together in future, instead of having two similar libraries, I've been using graphql core next mainly to do things quickly :)

ethe:
@patrick91 Yeah, but I can not find the best way to match the method of Model instance and graphql-core's lambda resolver. Directly translated from graphql.js is not a good idea I think.

Pygraphy based on Python metaclass to build the whole type system. Therefore, we can use class-based model without any pain.

patrick91:
@ethe I see! well keep me posted with progress (even via email, it's on my profile) :)

I'll close this PR for now ;)

ethe:
@patrick91 Thank you, I will pay attention to Strawberry. Although using graphql-core can move fast currently, there is something we can not easily walk around those problems in the end, such as `self` passing. The root cause is, graphql-core using class-based type system, and the type instance must be a Python object. Therefore, the resolver of a GraphQL object can only be a lambda function.

Making type system based on Python metaclass, then every type instance becomes a common Python class. Finally, we can use the GraphQL object as a common Python object.

If you have any suggestions about this idea, please contact me. I am glad to have a discuss.

ethe:
I have invited you to be a collaborator of Pygraphy, please do not forget to check your mailbox.

PR #103, opened 2019-07-09
Add Support for Field Permissions

### LINKED ISSUE #37: Support for field permissions & visibility
Would be nice to have a simple way to require permission for specific fields and also hide them from the schema (visibility).

Permission example (borrowed from DRF)

```py
import strawberry

    def has_permission(self, info) -> bool:
        return False


@strawberry.type
class User:
    name: str
    age: int


@strawberry.type
class Query:
    @strawberry.field(permission_classes=[IsAuthenticated])
    def me(self, info) -> Optional[User]:
```

Visibility example, what this should do is to completely hide the field from the schema if the client is not able to see it (so not even shown when doing introspection).

```py
import strawberry

class IsAdminClient(BasePermission):
    def has_permission(self, info) -> bool:
        return False


@strawberry.type
class User:
    name: str
    age: int


@strawberry.type
class Query:
    @strawberry.field(visiblity_classes=[IsAdminClient])
    def me(self, info) -> Optional[User]:
```

johnchuks:
This PR adds support for creating permission classes in strawberry. It takes inspiration from DRF's `BasePermission` and `permission_classes`

Fixes #37

### DIFF
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -48,7 +48,7 @@ class LazyFieldWrapper:
         self.field_name = name
         self.field_resolver = resolver
         self.field_description = description
-        self.permission_classes = permission_classes
+        self.permission_classes = permission_classes or []
 
         if callable(self._wrapped_obj):
             self._check_has_annotations(self._wrapped_obj)
@@ -92,23 +92,17 @@ class LazyFieldWrapper:
         Gets all permissions defined in the permission classes
         >>> strawberry.field(permission_classes=[IsAuthenticated])
         """
-        if self.permission_classes:
-            return [permission() for permission in self.permission_classes]
+        return (permission() for permission in self.permission_classes)
 
     def _check_permissions(self, info):
         """
         Checks if the permission should be accepted and
         raises an exception if not
         """
-        if not self._get_permissions():
-            return
         for permission in self._get_permissions():
             if not permission.has_permission(info):
                 message = getattr(permission, "message", None)
-                return self.permission_denied(message=message)
-
-    def permission_denied(self, message=None):
-        raise PermissionError(message)
+                raise PermissionError(message)
 
     @lazy_property
     def field(self):
@@ -268,6 +262,8 @@ def _get_field(
         if is_subscription:
 
             def _resolve(event, info):
+                if check_permission:
+                    check_permission(info)
                 return event
 
             field_params.update({"subscribe": resolver, "resolve": _resolve})
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -86,7 +86,7 @@ class LazyFieldWrapper:
 
     def __call__(self, *args, **kwargs):
         return self._wrapped_obj(self, *args, **kwargs)
-    
+
     def _get_permissions(self):
         """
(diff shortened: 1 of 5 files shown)


strawberry/field.py
```diff
@@ -40,13 +40,15 @@ def __init__(
...
         description=None,
+        permission_classes=None
     ):
         self._wrapped_obj = obj
         self.is_subscription = is_subscription
         self.is_input = is_input
         self.field_name = name
         self.field_resolver = resolver
         self.field_description = description
+        self.permission_classes = permission_classes
```
patrick91:
```suggestion
        self.permission_classes = permission_classes or []
```

if we default it to a list we can skip some ifs below :)

strawberry/field.py
```diff
@@ -85,6 +87,29 @@ def __getattr__(self, attr):
     def __call__(self, *args, **kwargs):
         return self._wrapped_obj(self, *args, **kwargs)
 
+    def _get_permissions(self):
+        """
+        Gets all permissions defined in the permission classes
+        >>> strawberry.field(permission_classes=[IsAuthenticated])
+        """
+        if self.permission_classes:
+            return [permission() for permission in self.permission_classes]
```
patrick91: if we default permission_classes to an empty list we can remove this if
and I'd suggest to change this to a generator so it is only evaluated when using it :)

```python
return (permission() for permission in self.permission_classes)
```

strawberry/field.py
```diff
@@ -85,6 +87,29 @@ def __getattr__(self, attr):
...
+
+    def _check_permissions(self, info):
+        """
+        Checks if the permission should be accepted and
+        raises an exception if not
+        """
+        if not self._get_permissions():
```
patrick91: same as above 😄

strawberry/field.py
```diff
@@ -85,6 +87,29 @@ def __getattr__(self, attr):
...
+            return
+        for permission in self._get_permissions():
+            if not permission.has_permission(info):
+                message = getattr(permission, "message", None)
+                return self.permission_denied(message=message)
```
patrick91: can we inline the raise instead of calling another method?

strawberry/field.py
```diff
@@ -220,8 +255,9 @@ def _get_field(
     }
 
     def resolver(source, info, **args):
+        if check_permission:
```
patrick91: 🎉  can we add this to the _resolve function below for the subscriptions? :)

strawberry/permission.py
```diff
@@ -0,0 +1,9 @@
+class BasePermission(object):
```
patrick91:
```suggestion
class BasePermission:
```

strawberry/field.py
johnchuks: Awesome! - that sounds like a good plan, it will be bring in the concept of `lazy` loading which beneficial to this scenario

strawberry/field.py
johnchuks: Alright will do

strawberry/field.py
johnchuks: Sure will do so as well.

PR #110, opened 2019-07-18
Fix Union usage as return type for a mutation

marcoacierno:


patrick91:
@marcoacierno let's add the release file and then we can merge this :)

Related to #4

PR #123, opened 2019-08-16
Add Django Support

### LINKED ISSUE #84: Add Django Support
I'd love to see strawberry support Django as a first class citizen via a contrib package. To start, it could be as simple as adding a Django view that exposes a GraphQL endpoint and the GraphQL Playground. From there, a lot of other neat features could be added, such as:

- Django authentication integration
- Out of the box subscription support using Django channels
- Automatic type generation from Django models
- Automatic CRUD query / mutation / subscription generation for Django models

Is this something that strawberry would be interested in supporting?

jaydenwindle:
This PR adds a basic Django view which exposes a strawberry schema for querying.

Fixes #84


.circleci/config.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - restore_cache:
           key: deps-{{ .Branch }}-{{ checksum "poetry.lock" }}
       - run:
-          command: poetry install
+          command: poetry install --extras=django
```
patrick91: TIL how to use extras with poetry :D

pyproject.toml
```diff
@@ -36,6 +38,10 @@ black = {version = "^19.3b0",allows-prereleases = true}
 isort = "^4.3"
 pytest-asyncio = "^0.10.0"
 pytest-cov = "^2.7"
+pytest-django = {version = "^3.5"}
+
+[tool.poetry.extras]
+django = ["django","pytest-django","asgiref"]
```
patrick91: pytest django is only installed when installing dev dependencies, right?

pytest.ini
```diff
@@ -1,3 +1,3 @@
 [pytest]
 addopts = -xs --emoji
-testpaths=tests/
\ No newline at end of file
+DJANGO_SETTINGS_MODULE=strawberry.contrib.django.tests.django_settings
```
patrick91: don't we still need testpaths?

strawberry/contrib/django/tests/test_views.py
```diff
@@ -0,0 +1,41 @@
...
+    hello: str = "strawberry"
+
+
+schema = strawberry.Schema(query=Query)
+
+
+def test_playground_view():
+    factory = RequestFactory()
+
+    request = factory.get("/graphql/", HTTP_ACCEPT="text/html")
```
patrick91: can we use django's reverse here?

strawberry/contrib/django/tests/test_views.py
```diff
@@ -0,0 +1,41 @@
...
+
+    response = GraphQLView.as_view(schema=schema)(request)
+    body = response.content.decode()
+
+    assert "GraphQL Playground" in body
+    assert f'endpoint: "{request.get_full_path()}"' in body
```
patrick91: ✨

strawberry/contrib/django/views.py
```diff
@@ -0,0 +1,70 @@
...
+        try:
+            query = data["query"]
+            variables = data.get("variables")
+            operation_name = data.get("operationName")
+        except KeyError:
+            return HttpResponseBadRequest("No GraphQL query found in the request")
+
+        context = {"request": request}
+
+        graphql_sync = async_to_sync(graphql)
```
patrick91: ohh, nice! this means that we can use async resolvers directly, right?

if so I think it might be worth adding a test for it :)

strawberry/contrib/django/views.py
```diff
@@ -0,0 +1,70 @@
...
+            query,
+            variable_values=variables,
+            context_value=context,
+            operation_name=operation_name,
+        )
+
+        response_data = {"data": result.data}
+
+        if result.errors:
+            print(result.errors)
```
patrick91: let's remove this print!

strawberry/contrib/django/views.py
```diff
@@ -0,0 +1,70 @@
...
+from graphql.error import format_error as format_graphql_error
+from graphql.type.schema import GraphQLSchema
+
+
+class GraphQLView(View):
+    schema = None
+
+    def __init__(self, schema=None):
+        assert schema, "You must pass in a schema to GraphQLView"
+        assert isinstance(
```
marcoacierno: I am not a big fan of using `assert` for this kind of validation because they can be disabled if python is run with `-O` but it might not matter much after all? What do you think?

marcoacierno:
Do you know if there is a way to specify which django version we support (maybe with a range? like from django 2.0 to 2.2) in poetry, so that if someone tries to install strawberry-django with the wrong version it fails?

Maybe we just need to update the dependency with a range?

```
django = {version = "^2.2",optional = true}
```

pyproject.toml
jaydenwindle: As far as I know, but I'm not entirely sure. I'm new to this whole poetry thing 😆

strawberry/contrib/django/tests/test_views.py
jaydenwindle: We could, but since we're passing the request object directly to the view instead of performing a request using the test client it doesn't really matter. The path could be any arbitrary string and the behavior of the test will be the same.

strawberry/contrib/django/tests/test_views.py
jaydenwindle: 😊

strawberry/contrib/django/views.py
jaydenwindle: I'm indifferent - will remove the assertions :)

strawberry/contrib/django/views.py
jaydenwindle: As far as I know, yes. Although this will block until the async resolver returns a value. Will add a test case :)

strawberry/contrib/django/views.py
jaydenwindle: 🤦‍♂

.circleci/config.yml
jaydenwindle: Haha same!

jaydenwindle:
@marcoacierno I think we can change the dependency to `django = {version = "^2.0",optional = true}`, which should allow users to install any version `>=2.0.0 <3.0.0` (if I'm reading the [docs](https://github.com/sdispater/poetry#caret-requirement) correctly).

patrick91:
@jaydenwindle sorry about the failure with the release check, going to rebase this now :)

PR #128, opened 2019-08-31
Fix Lists being marked as Optional

### LINKED ISSUE #127: Lists being marked as Optional
When defining a list the resulting schema marks the list as optional (or nullable in GraphQL terms) even if it wasn't wrapped in `typing.Optional`, we should fix that :)

sciyoshi:
Closes #127.


sciyoshi:
It seems like the new `RELEASE.md` check isn't working correctly, there's some error inside the GitHub action.

PR #129, opened 2019-08-31
Fix conversion of Enum types in resolvers, arguments, and input types

sciyoshi:
Currently, a field that is an Enum type will fail to resolve properly with an error inside `graphql-core`:

```python
>>> @strawberry.enum
... class IceCreamFlavour(enum.Enum):
...     VANILLA = "vanilla"
...     STRAWBERRY = "strawberry"
...     CHOCOLATE = "chocolate"
...
>>> @strawberry.type
... class Query:
...     flavour: IceCreamFlavour
...
>>> schema = strawberry.Schema(query=Query)
>>> await graphql(schema, '{ flavour }', root_value=Query(flavour=IceCreamFlavour.STRAWBERRY))
ExecutionResult(data=None, errors=[GraphQLError("Expected a value of type 'IceCreamFlavour' but received: <IceCreamFlavour instance>", locations=[SourceLocation(line=1, column=3)], path=['flavour'])])
```

This is because the resolver is expecting to see an Enum value (in this case, `"strawberry"`) rather than an instance of the Enum itself. A similar thing happens in arguments and input types as well. Enum handling has been a big pain point for me in Graphene, and I think Strawberry has an opportunity to get it right since it has better typing information available.

This PR adds some test cases showing the failures and implements the conversion. I haven't tested `Optional[Enum]` annotations yet for arguments and input types, which may still face a similar issue. The code path handling those conversions is separate, and doesn't seem to properly handle `List`s either yet. (I feel that `convert_args` and `dict_to_type` should be combined into one).

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+Fixes the conversion of Enums in resolvers, arguments and input types.
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -1,3 +1,4 @@
+import enum
 import inspect
 import typing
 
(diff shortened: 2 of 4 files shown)


OlmoDalco:
Optional Enums with default values will still fail to resolve properly!
Optional[t] requires a single type. Got uom_input(distance=<distance_unit.km: 'km'>)

patrick91:
@OlmoDalco, can you provide a small code example?

PR #130, opened 2019-08-31
Add a mypy plugin that enables typechecking Strawberry types

sciyoshi:
This PR also adds `mypy` as a dev dependency and re-enables the mypy CircleCI task, which seems to be succeeding :man_shrugging:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,7 +48,7 @@ jobs:
             - "/usr/local/lib/python3.7/site-packages"
       - run:
           command: |
-            poetry run mypy
+            poetry run mypy --ignore-missing-imports .
   flake8:
     docker:
       - image: circleci/python:3.7.2
@@ -91,8 +91,7 @@ workflows:
   main:
     jobs:
       - test
-      # TODO: need typings to be fixed
-      # - mypy
+      - mypy
       - flake8
   deploy:
     jobs:
--- a/README.md
+++ b/README.md
@@ -52,6 +52,17 @@ Open the debug server by clicking on the follwing link:
 
 This will open a GraphQL playground where you can test the API.
 
+### Type-checking
+
+Strawberry comes with a [mypy] plugin that enables statically type-checking your GraphQL schema. To enable it, add the following lines to your `mypy.ini` configuration:
+
+```ini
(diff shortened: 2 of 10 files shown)


tests/test_union.py
```diff
@@ -124,7 +123,6 @@ class B:
     class Mutation:
         @strawberry.mutation
         def hello(self, info) -> Union[A, B]:
-            # TODO: mypy is unable to understand that Outside is a dataclass
             return Outside(c=5)  # type:ignore
```
sciyoshi: Mypy now understands that `Outside` is a dataclass, but then fails since it isn't a member of `Union[A, B]`, which is to be expected.

README.md
```diff
@@ -52,6 +52,17 @@ Open the debug server by clicking on the follwing link:
...
+### Type-checking
+
+Strawberry comes with a [mypy] plugin that enables statically type-checking your GraphQL schema. To enable it, add the following lines to your `mypy.ini` configuration:
+
+```ini
+[mypy]
+plugins = strawberry.ext.mypy_plugin
+```
+
+[mypy]: http://www.mypy-lang.org/
```
patrick91: TIL: didn't know about this markdown syntax :D

README.md
```diff
@@ -52,6 +52,17 @@ Open the debug server by clicking on the follwing link:
 
 This will open a GraphQL playground where you can test the API.
 
+### Type-checking
```
patrick91: thanks for adding this to the README :)

.circleci/config.yml
```diff
@@ -91,8 +91,7 @@ workflows:
   main:
     jobs:
       - test
-      # TODO: need typings to be fixed
-      # - mypy
+      - mypy
```
patrick91: ✨

strawberry/ext/mypy_plugin.py
```diff
@@ -0,0 +1,21 @@
+from typing import Callable, Optional
+
+from mypy.plugin import ClassDefContext, Plugin
+from mypy.plugins import dataclasses
+
+
+class StrawberryPlugin(Plugin):
```
patrick91: 👍

PR #136, opened 2019-09-05
Fix for arguments conversion

patrick91:
Our convert_args function is currently broken, as you can see from the tests. It might be time to refactor it a bit, @sciyoshi would like to take a look at this?

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,5 @@
+Release type: patch
+
+Fix issue with converting arguments with optional fields.
+
+Thanks to [@sciyoshi](https://github.com/sciyoshi) for the fix!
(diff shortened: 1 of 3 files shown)


botberry:

```markdown
Release type: patch

```

sciyoshi:
Sure - I pushed a fix for this specific issue, which was simply that the annotation was marked as optional so the type wasn't being checked correctly. Would still be worth looking into how to clean this up further though!

patrick91:
@sciyoshi thanks! that was quick :) I'll merge now :)

Feel free to submit PR to refactor the code, there's quite a bit of things I'd like to improve, but I'm currently focusing a bit on PyCon Italia (we are building the new website with Strawberry) :)

PR #137, opened 2019-09-06
Add automatic type generation for Django models

jaydenwindle:
This PR adds automatic GraphQL type generation for Django models.

Usage:
```python
# models.py
from django.db import models

class Todo(models.Model):
    name = models.CharField(max_length=250)
    done = models.BooleanField(default=False)

# types.py
from todo.models import Todo
from strawberry.contrib.django.type import model_type

@model_type(model=Todo, fields=['id', 'name', 'done'])
class TodoType:
    pass

# OR

@model_type
class TodoType:
    class Meta:
        model = Todo
        fields = ['id', 'name', 'done']

# Generated type:
type TodoType {
    id: ID!
    name: String!
    done: Boolean!
}
```


patrick91:
Awesome! I'm going to take a look at this pretty soon, thank you so much for this!

I think we might have to spend a bit of time on choices, which seems to be a pain point in graphene (but this can be done, and probably should, in another PR)

botberry:

```markdown
Release type: patch

```

jaydenwindle:
@patrick91 Happy to contribute :)

Can you expand a little on what needs to be done as far as choices are concerned?

patrick91:
@jaydenwindle I think it might be useful to convert django choices to enums, but we need to investigate this a bit more (might not be worth the troubles)

See also: https://github.com/graphql-python/graphene-django/issues/185#issuecomment-388639988

jaydenwindle:
@patrick91 Got it, that makes sense. 

Auto conversion to enums would definitely be useful. I'll log an issue for adding choices support and we can brainstorm there.

strawberry/contrib/django/tests/test_converter.py
```diff
@@ -0,0 +1,51 @@
...
+
+from django.db import models
+
+import strawberry
+from strawberry.contrib.django.converter import convert_django_field
+
+
+def test_converter_returns_correct_type():
+
+    for FieldType in [
```
patrick91: I think it might be better to use pytest fixtures for this :)

and I'd one test for each type (str, int, bool, etc)

patrick91: sorry I meant parametrize http://doc.pytest.org/en/latest/example/parametrize.html

strawberry/contrib/django/tests/test_model_type.py
```diff
@@ -0,0 +1,167 @@
...
+
+import pytest
+
+import strawberry
+from strawberry.contrib.django.tests.models import TestModel
+from strawberry.contrib.django.type import model_type
+from strawberry.graphql import execute
+
+
+@pytest.mark.asyncio
```
patrick91: do we need to test this with async?

strawberry/contrib/django/tests/test_model_type.py
```diff
@@ -0,0 +1,167 @@
...
+    """,
+        schema=schema,
+    )
+
+    assert response.data["modelType"]["name"] == "modeltype"
+    assert response.data["modelType"]["extraInfo"] == "more info"
+
+
+@pytest.mark.asyncio
+async def test_model_type_doesnt_override_type_fields():
```
patrick91: what about changing the name to:

```suggestion
async def test_model_type_can_override_type_fields():
```

?

strawberry/contrib/django/tests/test_model_type.py
```diff
@@ -0,0 +1,167 @@
...
+    """,
+        schema=schema,
+    )
+
+    assert response.data["modelType"]["name"] == "joe"
+
+
+@pytest.mark.asyncio
+async def test_model_type_exclude_fields():
+    @model_type(model=TestModel, exclude_fields=["secret"])
```
patrick91: 👍

strawberry/contrib/django/tests/test_model_type.py
```diff
@@ -0,0 +1,167 @@
...
+    class TestModelType:
+        pass
+
+    assert "secret" not in TestModelType.field.fields
+
+
+@pytest.mark.asyncio
+async def test_model_type_only_fields():
+    @model_type(model=TestModel, only_fields=["name"])
```
patrick91: 👍

strawberry/contrib/django/tests/test_model_type.py
```diff
@@ -0,0 +1,167 @@
...
+
+    @strawberry.type
+    class Query:
+        hello: str = "world"
+
+    @strawberry.type
+    class Mutation:
+        @strawberry.field
+        def create_test_model(self, info, test: TestModelInputType) -> TestModelType:
+            return TestModel.objects.create(**test.__dict__)
```
patrick91: uhm, I'm not really convinced by this, doesn't seem user friendly.

I'll think of some alternatives, maybe @marcoacierno has a suggestion.

I guess we also need to come up with a solution for FK and M2M fields (but that can be in another PR)

strawberry/contrib/django/type.py
```diff
@@ -0,0 +1,74 @@
+import strawberry
+from strawberry.contrib.django.converter import convert_django_field
+from strawberry.type import type as strawberry_type
+
+
+def model_type(
```
patrick91: I wondering if we can find a nicer name for this? :D

Also, have you considered doing a django approach, like

```
class ABC(ModelType):
    class Meta:
        ...
```

to be consistent with django's approach? I prefer decorators but it could be worth thinking about this :) 

Pinging @adamchainz if he has some suggestion (and time to look at this) ;)

strawberry/contrib/django/tests/test_converter.py
jaydenwindle: Oh cool - didn't know this was a thing! Will update.

strawberry/contrib/django/type.py
```diff
@@ -0,0 +1,74 @@
...
+    cls=None,
+    *,
+    model=None,
+    exclude_fields=[],
+    only_fields=[],
```
adamchainz: I would say for security's sake: please do not add an `exclude_fields` option and support only `fields`, the list of included fields. So often internal fields are added to models and get accidentally shown on a Django ModelForm  or a DRF API endpoint, because they used their "all fields except these" mode. Please only support a list of fields, and if users want to show them on the API make them go the small extra step of listing them explicitly.

strawberry/contrib/django/type.py
adamchainz: A meta class might work but I'm not sure about the strawberry internals so I can't really advise on it. I guess it'd be not too much work to convert this class decorator since you just need to reorder it so strawberry's `ModelType` executes the current code at the end of class definition time (on python 3.6+ you can use `__init_subclass__` instead of a metaclass).

strawberry/contrib/django/tests/test_model_type.py
jaydenwindle: Can you expand a bit on what isn't convincing about this?

Using `**test.__dict__` is just a shorthand for the tests right now. I wouldn't recommend passing data directly from an input type to `objects.create` like this in most cases, since it will only work with some field types. I can change this test to be more verbose if you'd like.

The goal of this test is to show that a mutation resolver can access data from an input `model_type` correctly. As of right now, this PR only support generation of type classes and their usage in resolver definitions. This is definitely not as user friendly as I would like, since most mutation resolvers defined by users will end up have a lot of boilerplate code. I am planning on opening another PR later on to add query / mutation generation for model types (a la DRFs `ModelViewSet `), but that's a little ways off right now. 

We should definitely come up with an automatic solution for FK and M2M fields, but right now that can be achieved with custom resolvers on the `model_type` decorated class. I'll add some tests to show that functionality.

strawberry/contrib/django/type.py
jaydenwindle: Sure! Any ideas? Maybe `django_object_type` to keep it consistent with Graphene lingo, or `django_model_type` or `django_type`?

I had thought of trying to use a meta class early on, but decided to go with a decorator to keep things consistent with the rest of the library. I also wanted to be able to tap into the power of strawberry's existing `@strawberry.type` decorator, and this seemed like the simplest way to do that. 

I'm happy to experiment with metaclasses if you think it would be beneficial to have them though :)

strawberry/contrib/django/tests/test_model_type.py
jaydenwindle: I believe so, since it uses `async`/`await`. Otherwise pytest skips it.

strawberry/contrib/django/tests/test_model_type.py
patrick91: >  I am planning on opening another PR later on to add query / mutation generation for model types (a la DRFs ModelViewSet ), but that's a little ways off right now.

Oh awesome!

> Using **test.__dict__ is just a shorthand for the tests right now. I wouldn't recommend passing data directly from an input type to objects.create like this in most cases, since it will only work with some field types. I can change this test to be more verbose if you'd like.

Yes please! Sometime people use tests to see how to achieve something they are doing, so better to show the best practises ;)

strawberry/contrib/django/type.py
patrick91: don't know, my thinking was around consistency with what django users are familiar with, but again, this is something that can be added later :)

Let's also stick with model_type for now, should be easy to change later if needed :)

strawberry/contrib/django/tests/test_model_type.py
jaydenwindle: Sounds good! I'll update the tests :)

jaydenwindle: Done :)

strawberry/contrib/django/type.py
jaydenwindle: Good call. I've modified the `model_type` decorator to match the way that Django handles form fields ([docs](https://docs.djangoproject.com/en/2.2/topics/forms/modelforms/#selecting-the-fields-to-use)). By adding a `field` parameter which specifies a list of fields to use, along with a special value of `'__all__'` that will expose all fields, we mitigate the security risk of accidentally exposing fields by default while still allowing developers to expose all fields if they prefer.

strawberry/contrib/django/type.py
jaydenwindle: @patrick91 @adamchainz After thinking this over and experimenting with both approaches, I think we should stick with a decorator. Using a decorator maximizes compatibility with the rest of the library and keeps our developer experience consistent.

However, I think we should try to keep the developer experience consistent with Django's patterns wherever we can as well. To accomplish this, I've updated the `model_type` decorator to allow developers to use either decorator arguments or `Meta` class attributes to configure a model type. Developers can now define model types using either of these syntaxes:

```python
@model_type(model=Todo, fields=['id', 'name', 'done'])
class TodoType:
    pass

# OR

@model_type
class TodoType:
    class Meta:
        model = Todo
        fields = ['id', 'name', 'done']
```

This seems like a happy medium to me. Would love your thoughts on this approach :)

strawberry/contrib/django/converter.py
```diff
@@ -0,0 +1,67 @@
...
+
+from django.core.exceptions import ImproperlyConfigured
+from django.db import models
+
+import strawberry
+
+
+@singledispatch
+def convert_django_field(field):
+    raise ImproperlyConfigured(
```
patrick91: is this the right exception? we could change this to a custom one, maybe in another pr :)

strawberry/contrib/django/tests/test_converter.py
```diff
@@ -0,0 +1,67 @@
...
+    assert convert_django_field(models.fields.BooleanField()) == bool
+    assert convert_django_field(models.fields.NullBooleanField()) == Optional[bool]
+
+
+def test_converter_returns_id():
+    assert convert_django_field(models.fields.AutoField()) == strawberry.ID
+    assert (
+        convert_django_field(models.fields.AutoField(null=True))
+        == Optional[strawberry.ID]
+    )
```
patrick91: well tested! :)

strawberry/contrib/django/tests/test_model_type.py
patrick91: we have graphql sync we can use ;) https://github.com/strawberry-graphql/strawberry/blob/master/tests/test_schema.py#L9

strawberry/contrib/django/tests/test_model_type.py
```diff
@@ -0,0 +1,195 @@
...
+from strawberry.contrib.django.type import model_type
+from strawberry.graphql import execute
+
+
+@pytest.mark.asyncio
+async def test_model_type():
+    @model_type(model=TestModel, fields="__all__")
+    class TestModelType:
+        @strawberry.field
+        def extra_info(self, info) -> str:
```
patrick91: 👍

strawberry/contrib/django/tests/test_model_type.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+@pytest.mark.asyncio
+@pytest.mark.django_db
+async def test_model_input_type():
+    @model_type(model=TestModel, fields="__all__")
+    class TestModelType:
+        extra: str
+
+    @model_type(model=TestModel, is_input=True, fields="__all__")
```
patrick91: ah nice! maybe we could have a decorator that does this (I think it would be more explicity), something like `@input_model_type`

strawberry/contrib/django/tests/test_model_type.py
```diff
@@ -0,0 +1,195 @@
...
+
+    @strawberry.type
+    class Query:
+        hello: str = "world"
+
+    @strawberry.type
+    class Mutation:
+        @strawberry.field
+        def create_test_model(self, info, test: TestModelInputType) -> TestModelType:
+            name = getattr(test, "name")
```
patrick91: mmh, why getattr here? :)

strawberry/contrib/django/type.py
```diff
@@ -0,0 +1,68 @@
+import strawberry
+from strawberry.contrib.django.converter import convert_django_field
+from strawberry.type import type as strawberry_type
+
+
+def model_type(cls=None, *, model=None, fields=[], is_input=False, is_interface=False):
```
patrick91: do we test is interface? Do we need it? I guess we could check if a model is abstract, if that's what it was meant for :)

strawberry/contrib/django/converter.py
```diff
@@ -0,0 +1,67 @@
...
+        "Don't know how to convert the Django form field %s (%s) "
```
patrick91: should this be model field? :)

patrick91: (let's use f-strings as well ;) )

strawberry/contrib/django/converter.py
jaydenwindle: @patrick91 I think this is fine for now. It's the same error [graphene-django](https://github.com/graphql-python/graphene-django/blob/865c9535b969816a818e841c7ca219c992cfbf68/graphene_django/forms/converter.py#L15) throws when it can't properly convert a field.

strawberry/contrib/django/tests/test_model_type.py
jaydenwindle: Will do - will add a `model_input_type` and `model_interface` decorator.

strawberry/contrib/django/tests/test_model_type.py
jaydenwindle: `mypy` complains that `"TestModelInputType" has no attribute "name"` if it's accessed directly. Do I need to make some adjustments to our `mypy` plugin for this to work?

strawberry/contrib/django/type.py
jaydenwindle: It could be useful for abstract types. I'll add a test for it.

strawberry/contrib/django/tests/test_model_type.py
patrick91: will check this tomorrow, probably!

strawberry/contrib/django/converter.py
```diff
@@ -10,8 +10,8 @@
 @singledispatch
 def convert_django_field(field):
     raise ImproperlyConfigured(
-        "Don't know how to convert the Django form field %s (%s) "
-        "to type" % (field, field.__class__)
+        f"Don't know how to convert the Django model "
+        + "field {field} ({field.__class__}) to type"
```
denizdogan: I'm not sure this will work as you want it to, you need `f` on the second line as well!

```
>>> def foo():
...     x = 1
...     print(f"foo" +
...           "{x}") # no f-string

>>> foo()
foo{x}

>>> def foo():
...     x = 1
...     print(f"foo" +
...           f"{x}") # f-string!

>>> foo()
foo1
```

jaydenwindle: Good catch!

strawberry/contrib/django/tests/test_model_type.py
jaydenwindle: @patrick91 After looking into this, it seems like `mypy` support might be difficult to add here. The `model_type` decorator adds attributes to the class dynamically using `setattr`, but this isn't compatible with `mypy`'s static analysis. A [similar issue](https://github.com/python/mypy/issues/5719#issuecomment-426427083) in the mypy repo notes that:

> ... mypy doesn't run your code, it only reads it, and it only looks at the types of attributes, not at actions like setattr().

We could potentially write a custom plugin for this (as the poster of that issue did), but that seems overly complex in this situation.

sarendsen:
I would like to see this merged. Whats needed to push this forward? Anything I can do to help?

patrick91:
@sarendsen hey there! we are currently focusing on making strawberry a bit more stable, so we parked this for a bit. 

The last time I was thinking about this I wasn't sure about the API.

```
# models.py
from django.db import models

class Todo(models.Model):
    name = models.CharField(max_length=250)
    done = models.BooleanField(default=False)

# types.py
from todo.models import Todo
from strawberry.contrib.django.type import model_type

@model_type(model=Todo, fields=['id', 'name', 'done'])
class TodoType:
    pass
```

This looks inline with what strawberry does, but I think we need to find a way to handle additional fields, ie:

```
@model_type(model=Todo, fields=['id', 'name', 'done'])
class TodoType:
    name: str # what happens here? is this name coming from the instance?

    @strawberry.field
    def example(self, info) -> str:
        return self.name.upper() # <--- what about this? does it make sense to
                                 # have self as the current instance?
                                 # maybe we can use self.instance.name instead
```

And then there's also to talk about providing data for editing/querying data.

Would like to discuss these? We have a discord channel, I think it might be easier to chat/talk there :) https://discord.gg/3uQ2PaY

Rocamonde:
Hi guys, just posed asking about this on another issue but found this one later. I see that there is already some work on adding support for Django models. I'd like to contribute to this. Do you think you can bring me up to speed? Does this support QuerySets? 

I think this is not in the Docs, as before coming here I did not find anything. I think a lot of docs are also missing in general. Happy to contribute writing docs too. 

Let me know how I can be helpful,

Best

patrick91:
@Rocamonde we put this on hold for the time being, as we are finishing up working on strawberry and getting it stable first. I saw you joined the discord channel, maybe we can discuss the next steps there. I'm finishing this PR https://github.com/strawberry-graphql/strawberry/pull/352, it should unblock this :)

patrick91:
I'll close this for now, we have a discusson on pydantic here: https://github.com/strawberry-graphql/strawberry/discussions/444

That will probably shape how we handle the django integration too in future :)

PR #138, opened 2019-09-06
Fix conversion of enums when value is falsy

patrick91:
Just found out this small bug :)

There's an edge case that might be worth checking more in depth, don't know if this is something that's common, but if we use None as a value of one of the enums, the check that we have breaks. I guess we can leave like it is and maybe add a check in future that disallows None as a value. What do you think @sciyoshi?

```
@strawberry.enum
class IceCreamFlavour(Enum):
    VANILLA = ""
    STRAWBERRY = 0
    EXAMPLE = None
```

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+Fixed conversion of enum when value was falsy.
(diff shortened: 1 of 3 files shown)


botberry:



---

Fixed conversion of enum when value was falsy.

PR #140, opened 2019-09-06
Don't try to convert optional arguments to dataclasses

patrick91:
This fixes an issue when trying to convert optional arguments to a type, we were always defaulting the value of the dataclass to a dict, but this wasn't ideal when the argument was optional.


botberry:



---

Fixed issue when trying to convert optional arguments to a type

PR #146, opened 2019-09-10
Add flake8-bugbear to catch likely bugs

sciyoshi:
In particular, this will catch mutable default arguments, which are a common antipattern in Python.


botberry:

```markdown
Release type: patch

```

strawberry/schema.py
```diff
@@ -7,7 +7,7 @@
 
 
 class Schema(GraphQLSchema):
-    def __init__(self, query, mutation=None, subscription=None, directives=[]):
```
patrick91: 👍

sciyoshi:
It should be automatically loaded as part of flake8.

sciyoshi:
Should have been added as a dev dependency however, hang on...

sciyoshi:
Rebased and set flake8-bugbear as a dev dependency :+1:

PR #147, opened 2019-09-10
Fix mypy plugin

patrick91:
Not sure why when running mypy in our library the fullname returns "strawberry.type.type"

while in userland it returns "schema.strawberry.type" 🤷‍♂️

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+Fixes MYPY plugin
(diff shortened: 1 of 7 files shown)


botberry:



---

Fixes MYPY plugin

patrick91:
I'll see if I can add a test by following this article: https://sobolevn.me/2019/08/testing-mypy-types :)

PR #148, opened 2019-09-10
Move CI tests to GitHub actions

patrick91:
This should make it easier to test under multiple python versions (ie 3.8 beta)


botberry:

```markdown
Release type: patch

```

patrick91:
This is blocked by GitHub not having beta python versions :( https://github.com/actions/setup-python/issues/15

marcoacierno:
Can we merge it if it works and then update it once Github supports 3.8?

patrick91:
@marcoacierno I think we should try to add 3.8a on Circle CI and then when it is supported on GH Actions we can migrate :)

patrick91:
This can be merged when uvloop 0.14 gets released (as it is broken on py38)

PR #150, opened 2019-09-13
Update graphql-core

davidroeca:
`graphql-core-next` is being renamed to `graphql-core` and has had a few alpha releases. See https://github.com/graphql-python/graphql-core-next/releases/tag/v3.0.0a2 as well as https://github.com/graphql-python/graphene/blob/master/setup.py#L83. 

Also see [this comment](https://github.com/graphql-python/graphql-core-next/issues/53#issuecomment-528069398) explaining the use of `alpha`; the interface is considered stable.

At the very least, this would require a minor version bump because it may break things if users are relying on older versions of `graphql-core-next`.


botberry:

```markdown
Release type: patch

```

PR #165, opened 2019-09-29
Remove custom __repr__

### LINKED ISSUE #141: Types representation is not helpful
Currently the representations for types and enums aren't really helpful, especially while debugging:

```
>>> import strawberry
>>> @strawberry.input
... class Example:
...     arg: str
...
>>> Example
<class '__main__.Example'>
>>> x = Example("abc")
>>> x
input Example {
  arg: String!
}
```

I think we should move the current representation to be used only when using the class and not an instance, for the instances we can keep what dataclasses is giving us:

    >>> @dataclasses.dataclass()
    ... class Example:
    ...     arg: str
    ...
    >>> Example("abc")
    Example(arg='abc')

patrick91:
Closes #141 :)

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,4 @@
+Release type: patch
+
+Removed custom representation for Strawberry types,
+this should make using types much nicer.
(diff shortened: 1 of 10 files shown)


botberry:



---

Removed custom representation for Strawberry types,
this should make using types much nicer.

PR #168, opened 2019-10-01
Add docs on how to use Django view

matmunn:


botberry:

```markdown
Release type: patch

```

README.md
```diff
@@ -63,6 +63,18 @@ plugins = strawberry.ext.mypy_plugin
 
 [mypy]: http://www.mypy-lang.org/
 
+### Django Integration
```
patrick91: cool! can you also add an introduction specifying that this for now includes "only" a django view for serving the api?

I also think we need to add the app to the installed apps so the template can be found :)

matmunn:
Finally got time to add it, sorry for the delay.

PR #183, opened 2019-10-14
Fix execution of async resolvers.

### LINKED ISSUE #142: ASGI app broken with asyncio
Our ASGI app is currently broken with async io stuff. The implementation I made needs a bit of love :) We might have a look at how Ariadne is doing this :)

Here's a test that fails:

```python
import typing

import pytest

import strawberry
from starlette.testclient import TestClient
from strawberry.asgi import GraphQL


@pytest.fixture
def async_schema():
    @strawberry.type
    class Query:
        @strawberry.field
        async def hello(self, info, name: typing.Optional[str] = None) -> str:
            return f"Hello {name or 'world'}"

    return strawberry.Schema(Query)


@pytest.fixture
def test_client(async_schema):
    app = GraphQL(async_schema)

    return TestClient(app)


def test_simple_query(schema, test_client):
    response = test_client.post("/", json={"query": "{ hello }"})

    assert response.json() == {"data": {"hello": "Hello world"}}
```

jgosmann:
Fixes #142 and adds a test for it.

This is my first PR in this project. Let me know if a more elaborate PR description is required or any changes to the PR are required.

### DIFF
--- a/strawberry/graphql.py
+++ b/strawberry/graphql.py
@@ -48,7 +48,7 @@ async def execute(
         context_value=context_value,
     )
     if isawaitable(result):
-        result = await result
+        result = await typing.cast(typing.Awaitable[ExecutionResult], result)
     return result
 
 
--- a/strawberry/graphql.py
+++ b/strawberry/graphql.py
@@ -1,4 +1,5 @@
 import typing
+from inspect import isawaitable
 
 from graphql import (
     ExecutionResult,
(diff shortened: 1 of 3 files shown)


botberry:

```markdown
Release type: patch

```

strawberry/graphql.py
```diff
@@ -38,14 +39,17 @@
...
+    result = graphql_excute(
         schema,
         parse(query),
         middleware=[DirectivesMiddleware()],
         variable_values=variable_values,
         operation_name=operation_name,
         context_value=context_value,
     )
+    if isawaitable(result):
+        result = await typing.cast(typing.Awaitable[ExecutionResult], result)
```
jgosmann: Not sure if this is the best way to convince MyPy that `result` is of the correct type. I haven't used static typing in any of my Python projects so far.

patrick91: I think this is the right way for doing so!

PR #187, opened 2019-10-15
Adding pull request and issue templates

### LINKED ISSUE #186: Add pull request and issue templates

Kerl1310:
Fixes #186

### DIFF
--- a/.github/ISSUE_TEMPLATES/bug_report.md
+++ b/.github/ISSUE_TEMPLATES/bug_report.md
@@ -1,0 +1,21 @@
+---
+name: Bug report
+about: Create a bug report to help us improve the Strawberry GraphQL library
+---
+<!-- Provide a general summary of the bug in the title above. -->
+
+<!--- This template is entirely optional and can be removed, but is here to help both you and us. -->
+<!--- Anything on lines wrapped in comments like these will not show up in the final text. -->
+
+## Describe the Bug
+
+<!-- A clear and concise description of what the bug is. -->
+
+## System Information
+
+ - Operating system:
+ - Strawberry version
+
+## Additional Context
+
+<!-- Add any other relevant information about the problem here. -->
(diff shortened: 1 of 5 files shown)


botberry:

```markdown
Release type: patch

```

.github/ISSUE_TEMPLATES/bug_report.md
```diff
@@ -0,0 +1,21 @@
...
+<!--- Anything on lines wrapped in comments like these will not show up in the final text. -->
+
+## Describe the Bug
+
+<!-- A clear and concise description of what the bug is. -->
+
+## System Information
+
+ - Operating system:
+ - Strawberry version
```
patrick91:
```suggestion
 - Strawberry version:
```

.github/ISSUE_TEMPLATES/bug_report.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Additional Context
+
+<!-- Add any other relevant information about the problem here. -->
```
patrick91: Shall we also add something like this:

```py
import strawberry

def test_abc():
    assert False
```

This might help with getting reproduceable test cases :) I have a WIP action that runs test from an issue: https://github.com/patrick91/action-tests/issues/4

Kerl1310: I'm assuming you mean requesting that the reporter writes a reproduceable test case in the bug report? My only concern with making that a default thing is that it could put off a number of people who just want to report an issue they've experienced, meaning that some bugs don't get reported

patrick91: uhm, good point, maybe we can write something to tell them we have this bot. But let's think about it when we add the bot/action :)

patrick91:
@Kerl1310 thanks for this PR :) Send me an email for the stickers :)

PR #191, opened 2019-10-19
Add support for custom scalars

### LINKED ISSUE #114: Custom scalars
Sometimes it's useful to define custom scalar types in your schema, the examples I can think of right now are DateTime and Date.

That's how I implemented them in a project

```python
DateTime = NewType("DateTime", str)


def serialize_datetime(value):
    return value.isoformat()


def parse_value_datetime(value):
    return parse_datetime(value)


def parse_literal_datetime(ast, _variables=None):
    if not isinstance(ast, StringValueNode):
        return INVALID

    try:
        return parse_datetime(ast.value)
    except ValueError:
        return INVALID


REGISTRY[DateTime] = GraphQLScalarType(
    name="DateTime",
    description="Date with time (isoformat)",
    serialize=serialize_datetime,
    parse_value=parse_value_datetime,
    parse_literal=parse_literal_datetime,
)
```

It would be cool to have a decorator like `strawberry.scalar` that allows you to automatically define and register a new type + scalar:

```python

@strawberry.scalar
class DateTime:
    ???
```

I'm not sure how the interface might look to define `serialize`, `parse_value` and `parse_literal`.

This is how it's implemented in Graphene: https://github.com/graphql-python/graphene/blob/master/graphene/types/datetime.py

I feel like `parse_literal` can be implemented on a strawberry level by simply doing something like:

```
if typing == str:
   # Check that the AST is a StringValue and then call parse_value
elif typing == int:
   # Check that the AST is a int type and then call parse_value
```

While `serialize` and `parse_value` have to be implemented custom in the scalar

```python

@strawberry.scalar
class DateTime:
    @staticmethod
    def serialize(value):
        return value.isoformat()

    @staticmethod
    def parse_value(value):
       return parse_datetime(value)
```

To make it easier to understand we might want to call them `to_python` and `from_python`. But.. it doesn't really look good :)

We also need to change the type resolution system to support a class as scalar

```python
@strawberry.scalar
class Always5(int):
    @staticmethod
    def from_python(value):
        return 5

    @staticmethod
    def to_python(value):
        if value != 5:
            return None

        return 5


@strawberry.type
class Query:
    hey: Always5 = 10
```

jgosmann:
## Description

This adds support for custom GraphQL scalars. The usage is a bit different from what was proposed in #114: the serialize/parse functions are passed as argument to the decorator instead of being defined on the type itself. This allows to use an existing type without any changes to the type itself as custom scalar.

Usage examples can be found in the docstring of the `strawberry.scalar` decorator and the accompanying tests.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

* Closes #114.

## Checklist

- [x] My code follows the code style of this project. (as far as I can tell)


botberry:



---

Add support for custom GraphQL scalars.

strawberry/custom_scalar.py
```diff
@@ -0,0 +1,56 @@
...
+
+def scalar(cls=None, *, description=None, serialize, parse_value, parse_literal=None):
+    """Annotates a class or type as a GraphQL custom scalar.
+
+    Example usages:
+
+    >>> strawberry.scalar(
+    >>>     datetime.date,
+    >>>     serialize=lambda value: value.isoformat(),
+    >>>     parse_value=datetime.parse_date)
```
patrick91:
```suggestion
    >>>     parse_value=datetime.parse_date
    >>> )
```

strawberry/custom_scalar.py
```diff
@@ -0,0 +1,56 @@
...
+
+    >>> Base64Encoded = strawberry.scalar(
+    >>>     NewType("Base64Encoded", bytes),
+    >>>     serialize=base64.b64encode,
+    >>>     parse_value=base64.b64decode)
```
patrick91:
```suggestion
    >>>     parse_value=base64.b64decode
    >>> )
```

strawberry/custom_scalar.py
```diff
@@ -0,0 +1,56 @@
...
+
+    >>> @strawberry.scalar(
+    >>>     serialize=lambda value: ",".join(value.items),
+    >>>     parse_value=lambda value: CustomList(value.split(",")))
```
patrick91:
```suggestion
    >>>     parse_value=lambda value: CustomList(value.split(","))
    >>> )
```

strawberry/custom_scalar.py
```diff
@@ -0,0 +1,56 @@
...
+        description=description,
+        serialize=serialize,
+        parse_value=parse_value,
+        parse_literal=parse_literal,
+    )
+
+    return cls
+
+
+def scalar(cls=None, *, description=None, serialize, parse_value, parse_literal=None):
```
patrick91: what if we add a default serializer and parse value? This could make it easy to implement scalars that are just aliases of normal types, like:

```
Currency = strawberry.scalar(int, description="Currency in cents.")
Code = strawberry.scalar(str, description="Conference code.")
```

tests/test_custom_scalar.py
```diff
@@ -0,0 +1,68 @@
...
+    parse_value=base64.b64decode,
+)
+
+
+@strawberry.scalar(serialize=lambda x: 42, parse_value=lambda x: Always42())
+class Always42:
+    pass
+
+
+@pytest.mark.asyncio
```
patrick91: is there any particular reason why this test is async? I think we can use graphql sync here ;)

tests/test_custom_scalar.py
```diff
@@ -0,0 +1,68 @@
...
+    @strawberry.type
+    class Query:
+        @strawberry.field
+        def decode_base64(self, info, encoded: Base64Encoded) -> str:
+            return bytes(encoded).decode("ascii")
+
+    schema = strawberry.Schema(Query)
+
+    encoded = base64.b64encode(b"decoded").decode("ascii")
+    result = await execute(schema, f'{{ decodeBase64(encoded: "{encoded}") }}')
```
patrick91: you can also pass variables to execute, so you don't have to use the double brackets example :)

tests/test_custom_scalar.py
jgosmann: The `execute` function is defined as `async`. I could wrap the call in an `asyncio.run`. Not sure if that is better?

tests/test_custom_scalar.py
jgosmann: True, I made a commit to pass it as variable. It's not really any shorter, but maybe more readable? Let me know what you think.

strawberry/custom_scalar.py
jgosmann: Note that the could would need to actually look like this (if you don't want to overwrite the handling of all ints and strings):
```python
Currency = strawberry.scalar(NewType("Currency", int), description="Currency in cents.")
Code = strawberry.scalar(NewType("Code", str), description="Conference code.")
```

I added a commit implementing this.

jgosmann:
Added custom scalars for `Date`, `DateTime`, and `Time`. They can't go into `scalars.py` because that leads to a cyclic import (because `strawberry.scalar` needs to be imported there, which imports `type_converter.py` which again imports `scalars.py`).

jgosmann:
Seems like `aniso8601` isn't included in the dependencies? Should it be added or should we rely on the Python `fromisoformat` methods (which apparently cannot parse arbitrary ISO 8601 strings).

tests/test_custom_scalar.py
patrick91: There's also graphql sync we could use, maybe we should make execute sync function 😊 
https://github.com/strawberry-graphql/strawberry/blob/master/tests/test_schema.py#L9

strawberry/custom_scalar.py
```diff
@@ -0,0 +1,69 @@
...
+    >>> class CustomList:
+    >>>     def __init__(self, items):
+    >>>         self.items = items
+
+    """
+
+    if serialize is None:
+
+        def serialize(x):
+            return x
```
patrick91: do we have to create a function everytime? we could use a generic one, no?

tests/test_custom_scalar.py
```diff
@@ -0,0 +1,90 @@
...
+        def decode_base64(self, info, encoded: Base64Encoded) -> str:
+            return bytes(encoded).decode("ascii")
+
+    schema = strawberry.Schema(Query)
+
+    encoded = Base64Encoded(base64.b64encode(b"decoded"))
+    query = """query decode($encoded: Base64Encoded!) {
+        decodeBase64(encoded: $encoded)
+    }"""
+    result = await execute(schema, query, variable_values={"encoded": encoded})
```
patrick91: 👍

strawberry/datetime_scalar.py
```diff
@@ -0,0 +1,29 @@
...
+    description="Date with time (isoformat)",
+    serialize=_serialize_isoformatted,
+    parse_value=aniso8601.parse_datetime,
+)
+Time = scalar(
+    datetime.time,
+    description="Time (isoformat)",
+    serialize=_serialize_isoformatted,
+    parse_value=aniso8601.parse_time,
+)
```
patrick91: can we add some tests for these? :) also I'd put them inside `strawberry/types/datetime.py`

patrick91: do you think we need to add parse_literal as well? https://github.com/graphql-python/graphene/blob/master/graphene/types/datetime.py

jgosmann: I'll add tests later this week.

I'm not sure about `parse_literal`. It is unclear to me under which conditions this method gets invoked (and I find neither the graphql-core nor the Graphene documentation very helpful on this). [There also seems to be a default implemention](https://github.com/graphql-python/graphql-core-next/blob/712ef8435bbc3066e73b69da96b09c13de945080/src/graphql/type/definition.py#L394), but not sure if that one is appropriate.

tests/test_custom_scalar.py
jgosmann: I changed the test to use `graphql_sync`.

strawberry/custom_scalar.py
jgosmann: Yes, I changed it.

jgosmann:
I update the PR.

* I added `aniso8601` as an optional dependency. This dependency is not installed on the CI system (because it is optional). I need to figure out how to fix that.
* Also, the pytest-mypy-plugins 1.1 version is breaking the tests for me, so I fixed the version to `~1.0` (I guess fixing the issue would be preferrable, but I'd consider that out of the scope of this PR).

The question regarding `parse_literal` is still open.

patrick91:
@jgosmann turns out that it is used when parsing arguments passed directly, instead of using variables, see:

image

poetry.lock
```diff
@@ -645,16 +652,18 @@ version = "0.6.0"
 more-itertools = "*"
 
 [extras]
+datetime = ["aniso8601"]
```
patrick91: maybe we should add this as default dependency :)

pyproject.toml
```diff
@@ -39,10 +40,11 @@ pytest-cov = "^2.8"
 mypy = "^0.730"
 flake8-bugbear = "^19.8"
 pytest-django = {version = "^3.5", optional = true}
-pytest-mypy-plugins = "^1.0"
+pytest-mypy-plugins = "~1.0"
```
patrick91: is there any reason for this change?

poetry.lock
jgosmann: I was thinking that someone who doesn't need the DateTime type, might not want to install this dependency. But maybe it is common enough to warrant making this a hard dependency.

patrick91: @jgosmann I'd add this as default, graphene does that too (with another library) :)

pyproject.toml
jgosmann: Yes, see [this comment](https://github.com/strawberry-graphql/strawberry/pull/191#issuecomment-546462128).

jgosmann:
I'll take a look at `parse_literal` later this week. Thanks, for figuring that out!

poetry.lock
jgosmann: Did that.

pyproject.toml
jgosmann: I merged master into this branch and there it has been bumped to `^1.1` and is working. :)

strawberry/datetime_scalar.py
jgosmann: So I think I finally understand `parse_literal` (or at least better than before): the GraphQL query is parsed into an AST, then `parse_literal` is used to convert the node representing the value of a scalar into a native Python object, and then `parse_value` converts that Python object to the final object that we want for the scalar.

The date/time objects will be serialized and deserialized as simple strings. Thus, the default `parse_literal` converting the `StringValueNode` from the AST to a Python string is sufficient because we then take that string in `parse_value` to convert it into the proper object. I also added a test verifying the parsing of literal date/time objects.

Looking at [Graphene](https://github.com/graphql-python/graphene/blob/master/graphene/types/scalars.py) the `parse_literal` could be used to ensure that the AST node if of the expected type. But the `parse_value` would also fail (with a reasonable exception) if we do not get a `StringValueNode`.

patrick91: That makes sense! Thanks for spending time on this!

PR #193, opened 2019-10-21
type: add is_type_of for GraphlQLObjectType

### LINKED ISSUE #190: Add test for/document interface type resolution
I switched over from graphene for a small test and have a hard time figuring out how to solve the

```
Abstract type NodeIface must resolve to an Object type at runtime for field Root.nodes with value <Node instance>, received 'None'. Either the NodeIface type should provide a \"resolve_type\" function or each possible type should provide an \"is_type_of\" function.
```

error message.

What I have so far is:

```python
@strawberry.interface
class NodeIface:
    uuid: strawberry.ID
    ctime: str
    mtime: str
    label: str

@strawberry.type
class Node(NodeIface):
    pass

@strawberry.type
class Calculation(NodeIface):
    attributes: CalculationAttributes
```

if I do something like:

```python
@strawberry.type
class Root:
    @strawberry.field
    def nodes(self, info) -> typing.List[typing.Union[Calculation, Node]]:
        # do a dispatch to the correct type here based on DB data
        def dispatch(dbobj):
            if dbobj.is_calculation:
                return Calculation(**dbobj)
            return Node(**dbobj)
        return [dispatch(dbobj) for dbobj in db.all()]
```

I can write a query like:

```graphql
{
  nodes {
    __typename
    ... on Calculation {
      attributes {
        processState
      }
    }
  }
}
```

but the following fails:

```graphql
{
  nodes {
    uuid
    __typename
    ... on Calculation {
      attributes {
        processState
      }
    }
  }
}
```

with

```
{
  "error": {
    "data": null,
    "errors": [
      {
        "message": "Cannot query field 'uuid' on type 'nodes'. Did you mean to use an inline fragment on 'NodeIface', 'Calculation' or 'Node'?",
        "locations": [
          {
            "line": 3,
            "column": 5
          }
        ],
        "path": null
      }
    ]
  }
}
```

On the other hand, when I use the interface as return type:

```
    @strawberry.field
    def nodes(self, info) -> typing.List[NodeIface]:
        # do a dispatch to the correct type here based on DB data
        def dispatch(dbobj):
            if dbobj.is_calculation:
                return Calculation(**dbobj)
            return Node(**dbobj)
        return [dispatch(dbobj) for dbobj in db.all()]
```

I get:

```
{
  "error": {
    "data": null,
    "errors": [
      {
        "message": "Abstract type NodeIface must resolve to an Object type at runtime for field Root.nodes with value <Node instance>, received 'None'. Either the NodeIface type should provide a \"resolve_type\" function or each possible type should provide an \"is_type_of\" function.",
        "locations": [
          {
            "line": 2,
            "column": 3
          }
        ],
        "path": [
          "nodes",
          0
        ]
      }
    ]
  }
}
```

but I am a bit at a loss here on where/how to specify that information. Especially since I'm under the impression that with the typing info this should in principle work out-of-the box.

dev-zero:
fixes #190

## Description

Currently when returning types for an interface I get the error

```
Abstract type NodeIface must resolve to an Object type at runtime for field Root.nodes with value <Node instance>, received 'None'. Either the NodeIface type should provide a \"resolve_type\" function or each possible type should provide an \"is_type_of\" function.
```

this change adds a `is_type_of` for all `GraphQLObjectType` to return `True` when given the wrapped dataclass and forwards the `types` argument for the `GraphQLSchema`.

I don't know atm how to write a test for this :(

## Types of Changes

- [ ] Core
- [x] Bugfix
- [x] New feature
- [ ] Documentation

* #190

## Checklist

### DIFF
--- a/strawberry/type.py
+++ b/strawberry/type.py
@@ -79,6 +79,8 @@ def _process_type(cls, *, is_input=False, is_interface=False, description=None):
 
     extra_kwargs = {"description": description or cls.__doc__}
 
+    wrapped = dataclasses.dataclass(cls)
+
     if is_input:
         TypeClass = GraphQLInputObjectType
     elif is_interface:
@@ -92,7 +94,10 @@ def _process_type(cls, *, is_input=False, is_interface=False, description=None):
             if hasattr(klass, IS_STRAWBERRY_INTERFACE)
         ]
 
-    wrapped = dataclasses.dataclass(cls)
+        # the user might want to register other classes,
+        # but for now let's just add our dataclass
+        extra_kwargs["is_type_of"] = lambda obj, _: isinstance(obj, wrapped)
+
     wrapped.field = TypeClass(name, lambda: _get_fields(wrapped), **extra_kwargs)
 
     return wrapped
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -459,3 +459,48 @@ def test_can_declare_directives():
     '''
 
     assert repr(schema) == textwrap.dedent(expected_schema).lstrip()
+
+
+def test_query_interface():
+    """Test that a query on an interface correctly detects instances of the dataclasses
+       as one of the inherited types, given that types are registered in addition to the
+       root type (since there is no field for them)"""
+
+    @strawberry.interface
+    class Cheese:
+        name: str
+
+    @strawberry.type
+    class Swiss(Cheese):
+        canton: str
+
+    @strawberry.type
+    class Italian(Cheese):
+        province: str
+
+    @strawberry.type
+    class Root:
+        @strawberry.field
+        def assortment(self, info) -> typing.List[Cheese]:
+            return [
+                Italian(name="Asiago", province="Friuli"),
+                Swiss(name="Tomme", canton="Vaud"),
+            ]
+
+    schema = strawberry.Schema(query=Root, types=[Swiss, Italian])
+
+    query = """{
+        assortment {
+            name
+            ... on Italian { province }
+            ... on Swiss { canton }
+        }
+    }"""
+
+    result = graphql_sync(schema, query)
+
+    assert not result.errors
+    assert result.data["assortment"] == [
+        {"name": "Asiago", "province": "Friuli"},
+        {"canton": "Vaud", "name": "Tomme"},
+    ]
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -11,5 +11,5 @@ known_pytest=pytest
 known_future_library=future
 known_standard_library=types,requests
 default_section=THIRDPARTY
-known_third_party = base,click,config,github_release,graphql,httpx,hupper,mypy,pygments,release,starlette,uvicorn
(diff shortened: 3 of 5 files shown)


botberry:

```markdown
Release type: patch

```

patrick91:
@dev-zero what about adding a test like these? https://github.com/strawberry-graphql/strawberry/blob/master/tests/test_schema.py

Thanks a lot for the contribution! And sorry for the late reply, I'm not doing well these days! :)

dev-zero:
will do, thanks for the feedback

dev-zero:
@patrick91 done. Please consider putting a simple documentation skeleton structure in place in case you want people to document.

tests/test_schema.py
```diff
@@ -459,3 +459,44 @@ def uppercase(value: str, example: str):
     '''
 
     assert repr(schema) == textwrap.dedent(expected_schema).lstrip()
+
+
+def test_query_interface():
+    """Test that a query on an interface correctly detects instances of the dataclasses
```
patrick91: 😍

tests/test_schema.py
```diff
@@ -459,3 +459,44 @@ def uppercase(value: str, example: str):
...
+            return [
+                Italian(name="Asiago", province="Friuli"),
+                Swiss(name="Tomme", canton="Vaud"),
+            ]
+
+    schema = strawberry.Schema(query=Root, types=[Swiss, Italian])
+
+    query = (
+        "{ assortment { name ... on Italian { province } ... on Swiss { canton } } }"
+    )
```
patrick91: nit: could be a triple quote string ;)

strawberry/type.py
```diff
@@ -92,7 +94,10 @@ def _get_fields(wrapped):
             if hasattr(klass, IS_STRAWBERRY_INTERFACE)
         ]
 
-    wrapped = dataclasses.dataclass(cls)
+        # the user might want to register other classes,
+        # but for now let's just add our dataclass
+        extra_kwargs["is_type_of"] = lambda obj, _: isinstance(obj, wrapped)
```
patrick91: can wee expand a bit on this? not sure what you mean with that comment :)

tests/test_schema.py
dev-zero: done

strawberry/type.py
dev-zero: Basically the user could return other datatypes (like an instance of an ORM class, a named tuple) instead of explicitly wrapping the data in an instance of the dataclass, right?
But this would not work for interfaces unless `is_type_of` called on GraphQLObjectType returns `True` for the used class. This could be done in various ways, from additional "meta" attributes on the dataclass to arguments to the decorator and there either as another lambda function or as a list of types. Since I didn't want to make this decision yet I only wanted to add a comment...

Would the following explain it better?
```
-        # the user might want to register other classes,
-        # but for now let's just add our dataclass
+        # make sure instances of our wrapped dataclass
+        # are recognized by the GraphQLObjectType as types with the correct interface 
+        # TODO: user might want to add custom types (like ORM classes) at some point
```

patrick91: ah I see! I think we can keep the comment like that, the context on this PR should be enough for now :)

PR #196, opened 2019-10-24
Refactor CLI Folder Structure (#12)

### LINKED ISSUE #12: Refactor CLI folder structure

bryanculver:
- Creates separate command folder structure

- Refactored cli __init__ to import commands explicitly

## Description

Supports breaking out CLI commands into separate files, following Click's complex example: https://github.com/pallets/click/tree/master/examples/complex/complex

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* Closes #12

## Checklist


botberry:

```markdown
Release type: patch

```

RELEASE.md
```diff
@@ -0,0 +1,3 @@
+Release type: minor
```
patrick91: can we change this to patch? :)

strawberry/cli/__init__.py
```diff
@@ -1,48 +1,11 @@
...
-from starlette.middleware.cors import CORSMiddleware
-
-import importlib
-
-import uvicorn
-
-import hupper
-
-from strawberry.asgi import GraphQL
+from .commands.cmd_server import server as cmd_server
```
patrick91: I think we can drop the `cmd_` from the filename ;)

RELEASE.md
bryanculver: Was just following the recommendations in SEMVER:
> MAJOR version when you make incompatible API changes,
> MINOR version when you add functionality in a backwards compatible manner, and
> PATCH version when you make backwards compatible bug fixes.

Considering it seemed like I was adding functionality yet maintaining backwards compatibility and it wasn't a bugfix, I figured it was a minor release. But if you feel it should be a patch I'd be happy to change it.

tests/cli/test_server.py
```diff
@@ -0,0 +1,38 @@
...
+    # Mock to prevent the reloader from kicking in
+    monkeypatch.setattr(hupper, "start_reloader", Mock(wraps=mock_hupper_start_reloader))
+    runner = CliRunner()
+
+    result = runner.invoke(cmd_server, ['tests.cli.helpers.sample_schema'], "")
+    assert result.exit_code == 0
+
+    # We started the reloader
+    hupper.start_reloader.assert_called_once
+    uvicorn.run.assert_called_once
```
bryanculver: Eventually would suggest further mocking to assert uvicorn was called with the right GraphQL object but that's a lot of hoisting and mocking.

Since GraphQL's "contract" with uvicorn is extensively tested elsewhere we can reasonably expect it to be successful.

Likely the server command would need to be refactored to allow more function responses to be hooked in to.

bryanculver:
> This is quite nice, I added a couple of comments, as soon as they are resolve we can merge! Thanks a lot :)

Responded to one, resolved the other. 👍 

> PS. do you think we can test the CLI somehow? not super familiar with click :)

Wrote a somewhat simple tester to start with without refactoring the server command.

RELEASE.md
patrick91: I'd say we are not following semver until we release version 1.0.0 :D

But that's ok, no need for a change ;)

tests/cli/test_server.py
patrick91: uhm, I usually tend to use mocker (we might need to add it to the deps) with pytest and then do something like this:

```
def test_cli_cmd_server(mocker):
   uv_patch = mocker.patch('path.server.uvicorn.run')
   uv_patch.side_effect = ... # or uv_patch.return_value = 
```

would you be ok with trying this approach?

In addition to that I think we need to call `assert_called_once` like this `uvicorn.run.assert_called_once()`

That fixed we can merge this! :)

tests/cli/test_server.py
```diff
@@ -0,0 +1,32 @@
...
+    hupper_reloader_patch = mocker.patch('hupper.start_reloader')
+    hupper_reloader_patch.return_value = MockReloader()
+    runner = CliRunner()
+
+    result = runner.invoke(cmd_server, ['tests.cli.helpers.sample_schema'], "")
+    assert result.exit_code == 0
+
+    # We started the reloader
+    assert hupper.start_reloader.call_count == 1
+    assert uvicorn.run.call_count == 1
```
patrick91: 👍

PR #202, opened 2019-11-01
Implement `resolve_type` on interfaces

patrick91:
Implement `resolve_type` on interfaces

When specifying `is_type_of` we aren't able to return a type
other than the field one, this prevents for example to return
a django model instead of the strawberry type.

So now instead of implementing `is_type_of` on every type that
implements an interface we only implement `resolve_type` on the
interface itself.

See: https://github.com/pythonitalia/pycon/pull/1216#issuecomment-548923736

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,4 @@
+Release type: patch
+
+Set `is_type_of` only when the type implements an interface,
+this allows to return different (but compatibile) types in simple cases.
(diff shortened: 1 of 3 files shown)


botberry:



---

Set `is_type_of` only when the type implements an interface,
this allows to return different (but compatibile) types in simple cases.

PR #231, opened 2019-12-09
Fix boolean arguments conversion

patrick91:


botberry:



---

Fix missing support for booleans when converting arguments

PR #240, opened 2019-12-20
Remove support for async resolvers in django view

patrick91:
🤷‍♂️ I didn't realise the database doesn't work in a normal async_to_sync wrapper, so for now we'll remove support for async resolvers in django.


botberry:



---

This releases removes support for async resolver in django
as they causes issues when accessing the databases.

PR #243, opened 2020-01-02
Update permissions to receive source and args

patrick91:
This PR adds source and arguments in `has_permission`. It also fixes a bug where you had to use `is_input` when declaring field with `a: str = strawberry.field()` :)

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,33 @@
+Release type: minor
+
+This release improves support for permissions (it is a breaking change).
+Now you will receive the source and the arguments in the `has_permission` method,
+so you can run more complex permission checks. It also allows to use permissions 
+on simple fields, here's an example:
+ 
+
+```python
+import strawberry
+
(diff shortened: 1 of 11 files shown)


botberry:



---

This release improves support for permissions (it is a breaking change).
Now you will receive the source and the arguments in the `has_permission` method,
so you can run more complex permission checks. It also allows to use permissions 
on simple fields, here's an example:
 

```python
import strawberry

class IsAdmin(BasePermission):
    message = "You are not authorized"

    def has_permission(self, source, info):
      return source.name.lower() == "Patrick" or _is_admin(info)


@strawberry.type
class User:
    name: str
    email: str = strawberry.field(permission_classes=[IsAdmin])


@strawberry.type
class Query:
    @strawberry.field(permission_classes=[IsAdmin])
    def user(self, info) -> str:
      return User(name="Patrick", email="example@email.com")
```

PR #248, opened 2020-01-14
Add error logging in Django GraphQLView

estyxx:
## Description
Add a method in the Django GraphQLView to logging the error if a logger object is passed to the view. 
So in your `urls.py` when you define the graphql endpoint you only have to pass the additional argument `logger`

```
import logging

urlpatterns = [
    path(
        "graphql", GraphQLView.as_view(schema=schema, logger=logging.getLogger()), name="graphql"
    ),
    ...
] 
````
In this should be easier to inherit the `GraphQLView` in order to customize the log (for example if you are using an external system like Sentry...) 

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* 

## Checklist

### DIFF
--- a/strawberry/django/views.py
+++ b/strawberry/django/views.py
@@ -14,8 +14,9 @@ from graphql.type.schema import GraphQLSchema
 
 class GraphQLView(View):
     schema = None
+    logger = None
 
-    def __init__(self, schema=None):
+    def __init__(self, schema=None, logger=None):
         if not schema:
             raise ValueError("You must pass in a schema to GraphQLView")
 
@@ -23,6 +24,7 @@ class GraphQLView(View):
             raise ValueError("You must pass in a valid schema to GraphQLView")
 
         self.schema = schema
+        self.logger = logger
 
     def get_root_value(self):
         return None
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by botberry** [comment]

```markdown
Release type: patch

```

tests/django/test_views.py
```diff
@@ -138,3 +139,55 @@ def test_returns_errors_and_data():
...
+
+class ListHandler(logging.Handler):
+    def __init__(self, *args, **kwargs):
+        super(ListHandler, self).__init__(*args, **kwargs)
+
+        for level in logging._nameToLevel.keys():
+            setattr(self, level.lower(), [])
+
+    def emit(self, record):
+        msg = record.getMessage()
```
marcoacierno, line 152: Maybe we can mock `log_errors` or `self.logger.error` in the tests instead of creating a custom logger like this?

strawberry/django/views.py
```diff
@@ -14,15 +14,17 @@
 
 class GraphQLView(View):
     schema = None
+    logger = None
 
-    def __init__(self, schema=None):
+    def __init__(self, schema=None, logger=None):
```
marcoacierno, line 19: question: do/can we add type hints to `schema` and `logger`?

strawberry/django/views.py
```diff
@@ -62,10 +64,19 @@ def dispatch(self, request, *args, **kwargs):
...
         return JsonResponse(response_data)
+
+    def log_errors(self, errors):
+        if not self.logger:
+            return
+        for error in errors:
+            if hasattr(error, "original_error") and error.original_error:
+                self.logger.error(error.original_error)
+            else:
+                self.logger.error(error)
```
patrick91, line 82: sorry @estyxx, I completely forgot to reply to this. I was thinking about how to make this different, as right I'm not sure this is ideal. Will thinking about it over the next few days 😊

Issue #251, opened 2020-01-14
Autogenerated `return` directive argument bug


yngvarsh:
Having the following code:

```python
import strawberry
from graphql import DirectiveLocation


@strawberry.type
class User:
    name: str


@strawberry.type
class Query:
    @strawberry.field
    def user(self, info) -> User:
        return User(name="Patrick")


@strawberry.directive(locations=[DirectiveLocation.FIELD])
def upper(value: str) -> str:
    return value.upper()


schema = strawberry.Schema(query=Query, directives=[upper])
```

And the following query:

```
{
  user {
    name @upper
  }
}
```

We'll receive this response:

```
{
  "data": null,
  "errors": [
    {
      "message": "Directive '@upper' argument 'return' of type 'String!' is required, but it was not provided.",
      "locations": [
        {
          "line": 3,
          "column": 10
        }
      ],
      "path": null
    }
  ]
}
```

Issue #250, opened 2020-01-14
Include and skip directives bug


yngvarsh:
Having the following code:

```python
import strawberry


@strawberry.type
class User:
    name: str
    age: int


@strawberry.type
class Query:
    @strawberry.field
    def user(self, info) -> User:
        return User(name="Patrick", age=100)


schema = strawberry.Schema(query=Query)
```

And the following query:

```graphql
query user($nameRequired: Boolean!){
  user {
    name @include(if: $nameRequired)
    age
  }
}
```

with variables:

```
{
    "nameRequired": true
}
```

We'll receive this response:

```
{
  "data": null,
  "errors": [
    {
      "message": "'VariableNode' object has no attribute 'value'",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "path": [
        "user",
        "name"
      ]
    }
  ]
}
```

PR #252, opened 2020-01-14
Fix directives bugs

yngvarsh:
## Description

- Fix autogenerated `return` argument bug

- Fix include and skip failure bug

## Types of Changes

- [ ] Core
- [X] Bugfix
- [ ] New feature
- [ ] Documentation

* #250
 
* #251 

## Checklist


botberry:



---

Fix directives bugs:

- Fix autogenerated `return` argument bug

- Fix include and skip failure bug

strawberry/middleware.py
```diff
@@ -1,13 +1,19 @@
...
+
 class DirectivesMiddleware:
     def resolve(self, next_, root, info, **kwargs):
         result = next_(root, info, **kwargs)
 
         for directive in info.field_nodes[0].directives:
             directive_name = directive.name.value
 
+            if directive_name in SPECIFIED_DIRECTIVES:
+                continue
```
patrick91: can we add a comment on why we do this? 

Do you have any other possible names for SPECIFIED_DIRECTIVES? Not sure what it means right now ;)

tests/test_directives.py
```diff
@@ -12,3 +12,4 @@ def uppercase(value: str) -> str:
     assert uppercase.directive.__class__ == GraphQLDirective
     assert uppercase.directive.name == "uppercase"
     assert uppercase.directive.description == "Sample description"
+    assert uppercase.directive.args == {}
```
patrick91: 👍

tests/test_execute.py
```diff
@@ -45,17 +45,21 @@ def replace(value: str, old: str, new: str):
 
     schema = strawberry.Schema(query=Query, directives=[uppercase, replace])
 
-    query = """{
+    query = """query persons($identified: Boolean!){
```
patrick91:
```suggestion
    query = """query People($identified: Boolean!){
```

strawberry/directive.py
```diff
@@ -12,11 +12,13 @@
 
 def _get_arguments(func):
     annotations = func.__annotations__
-    name = func.__name__
```
patrick91: 👍

strawberry/middleware.py
yngvarsh: `DirectivesMiddleware` tries to find resolve function for `skip` and `include` directives,  but `DIRECTIVE_REGISTRY` is empty when app startup. `graphql-core-next` has own logic for these specified directives [here](https://github.com/graphql-python/graphql-core-next/blob/master/src/graphql/execution/execute.py#L513), so we can skip them.

I used this name because `graphql-python` [uses](https://github.com/graphql-python/graphql-core-next/blob/master/src/graphql/type/directives.py#L200) it. Maybe we can use better name, I'm not sure about this. Also, probably we can dynamically create this set, like:

```
from graphql.type.directives import specified_directives

SPECIFIED_DIRECTIVES = {directive.name for directive in specified_directives}
```

patrick91: ah good, point 😊 let's keep it like that for now!

PR #259, opened 2020-01-28
Add apollo federation support 

patrick91:
⚠️ The code is quite bad at the moment, this PR is an exploration of supporting apollo federation: https://www.apollographql.com/docs/apollo-server/federation/introduction/ ⚠️

Tasks:

- [x] Add custom federation field and type to support extending a type, adding keys, supports, provides and external directives
- [x] Add custom printer that knows how to print new directives
- [x] Add entities type
- [x] Add Any scalar type
- [x] Add service type and resolver
- [x] Add entities resolver
- [x] Remove duplicated print code
- [ ] Add e2e federation test

There are a few hacks and also code inconsistencies that I want to fix before merging, but this seems to be working ok at the moment :)


botberry:

```markdown
Release type: patch

```

strawberry/enum.py
```diff
@@ -22,6 +22,8 @@ def _process_enum(cls, name=None, description=None):
         values=[(item.name, GraphQLEnumValue(item.value)) for item in cls],
         description=description,
     )
+    # :D
+    cls.field._strawberry_type = cls
```
patrick91: this is needed to fetch the strawberry type from the graphql field, we can find a way to improve it (maybe our own type map)

strawberry/federation.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def type(cls=None, *args, **kwargs):
+    def wrap(cls):
+        keys = kwargs.pop("keys", [])
+        extend = kwargs.pop("extend", False)
+
+        wrapped = _process_type(cls, *args, **kwargs)
+        wrapped._federation_keys = keys
+        wrapped._federation_extend = extend
```
patrick91: not sure about this

strawberry/federation.py
```diff
@@ -0,0 +1,85 @@
...
+            if hasattr(graphql_type, "_strawberry_type"):
+                if graphql_type._strawberry_type and getattr(
+                    graphql_type._strawberry_type, "_federation_keys", []
+                ):
+                    federation_key_types.append(graphql_type)
+
+        if federation_key_types:
+            types += [GraphQLUnionType("_Entity", federation_key_types)]
+
+        return types
```
patrick91: This is quite ugly, but it does add the required types and resolvers for federation

strawberry/field.py
```diff
@@ -110,6 +110,8 @@ def field(self):
             check_permission=self._check_permissions,
         )
 
+        return field
```
patrick91: need to revert this 😊

strawberry/printer.py
```diff
@@ -1,7 +1,215 @@
-from graphql.utilities.schema_printer import print_type as graphql_print
+from itertools import chain
```
patrick91: this has been copy and pasted from graphql core, we can remove the duplicated functions, but we did change some of them :)

strawberry/printer.py
```diff
@@ -1,7 +1,215 @@
...
+
+def print_federation_field_directive(field, metadata):
+    out = ""
+
+    if metadata and "federation" in metadata:
+        federation = metadata["federation"]
+
+        provides = federation.get("provides", "")
+        requires = federation.get("requires", "")
+        external = federation.get("external", False)
```
patrick91: this can probably be a utility function :)

strawberry/printer.py
```diff
@@ -1,7 +1,215 @@
...
+
+    def _get_metadata(field_name):
+        return next(
+            (
+                f.metadata
+                for f in strawberry_fields
+                if (getattr(f, "field_name", None) or f.name) == field_name
+            ),
+            None,
+        )
```
patrick91: we need this to map strawberry fields to graphql fields. Didn't find a way that didn't involve changing the whole print again (which we might want to do in future)

strawberry/type.py
```diff
@@ -38,8 +38,10 @@ def _resolver(root, info):
     return _resolver
 
 
-def _process_type(cls, *, is_input=False, is_interface=False, description=None):
-    name = cls.__name__
+def _process_type(
+    cls, *, name=None, is_input=False, is_interface=False, description=None
+):
+    name = name or cls.__name__
```
patrick91: never noticed that we didn't have a way to rename a type. This should be done in another PR :)

remorses:
Awesome work 👍

if you need a gateway to test with other services i made this docker container https://github.com/remorses/apollo-federation-gateway so that you don't have to make a nodejs server yourself

patrick91:
@remorses wish I had seen that earlier 😊 I have some e2e test locally already :D

strawberry/federation.py
```diff
@@ -0,0 +1,136 @@
...
+
+    field = strawberry_field(*args, **kwargs)
+
+    return field
+
+
+def field(wrap=None, *args, **kwargs):
+    field = strawberry_federation_field(*args, **kwargs)
+
+    if wrap is None:
```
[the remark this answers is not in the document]
patrick91: This allows to use the decorator with our without the parenthesis:

```python

@strawberry.federation.field
def x(...): ...

@strawberry.federation.field(arg=123)
def x(...): ...

```

strawberry/federation.py
```diff
@@ -0,0 +1,136 @@
...
+        )
+
+        self.type_map["_Any"] = Any
+        self.type_map["_Service"] = Service.field
+        self.type_map[self.query_type.name] = self.query_type
+
+    def _get_entity_type(self):
+        federation_key_types = []
+
+        for graphql_type in self.type_map.values():
```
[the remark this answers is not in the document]
patrick91: Yup, the code is really bad to be honest, I just hacked it together to make sure this could work, thanks for the suggestion, I’ll probably copy it :D

strawberry/printer.py
patrick91: From what I remember GraphQL-core tries to stick to the same API as GraphQL.js, so I don’t think they will add this feature.

Also I think it might be worth writing our own printer, as maybe in future we might migrate away from GraphQL-core!

tests/federation/e2e/reviews.py
```diff
@@ -0,0 +1,45 @@
...
+    body: str
+
+REVIEWS_PER_PRODUCT = {
+    "ABC123": [Review("1", "Awesome product!")],
+    "DEF456": [Review("2", "Not really enjoying it.")],
+}
+
+@strawberry.federation.type(extend=True, keys=["upc"])
+class Product:
+    upc: str = strawberry.federation.field(external=True)
```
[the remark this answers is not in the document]
patrick91: I think it might be good, yes!

strawberry/federation.py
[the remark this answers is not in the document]
Ambro17: Nice

strawberry/type.py
```diff
@@ -78,6 +79,8 @@ def _get_fields(wrapped, types_replacement_map=None):
             if class_field.default not in (dataclasses.MISSING, None):
                 fields[field_name].default_value = class_field.default
 
+            fields[field_name]._strawberry_type = class_field
```
marcoacierno, line 82: What is this?

patrick91: 🔪  gone

Issue #175, opened 2019-10-08
Add pagination support


un33k:
[ x ] Feature Request
[ x ] Documentation 

If pagination is supported, be nice to include it in documentations.  If there is no plan to support it, it'd be nice to provide an example for developers to role their own.  Otherwise, let's keep this issue around to track the pagination feature.

jaydenwindle:
Right now there are no specific utilities designed to support pagination (as far as I know). It would definitely be neat to support this though!

One way to implement basic pagination would be to use something like this:
```python
@strawberry.type
class Query:

    @strawberry.field
    def numbers(self, info, first: int = None, skip: int = None) -> typing.List[int]:
        numbers = range(1000)
        return numbers[skip:first]
```

This approach could be abstracted into a decorator for use in multiple queries as well.

If you'd like to add support for pagination using this approach, or write some documentation for it, a PR would be welcome :)

patrick91:
@jaydenwindle thanks for answering 😊 at PyCon DE someone asked for Relay support, maybe we could write some helpers for that, I kinda this something here:

https://github.com/strawberry-graphql/strawberry-swapi/blob/master/utils.py

I think Graphene has something for arrays already, so we could import that. I don't really have much experience with Relay, so I'd happy to see someone working on this :)

denizdogan:
Are there any simple examples of how to implement a decorator that's meant to be used in combination with `@strawberry.field`? I gave it a half-assed shot, but quickly got lost in type annotation jungle. I only recently started trying out this library, but it seems that type annotations are very central to the Strawberry library?

Any help would be appreciated, even just a simple example :) Sorry if this is the wrong place to ask.

patrick91:
@denizdogan what did you try already? I'd like to know mainly because I'd like to make the library as user friendly as possible :)

I'll try to come up with an example tomorrow :)

denizdogan:
@patrick91 As I said, I didn't really try very hard :) But I slapped together a simple example which shows where I get stuck:

```py
def strawberry_limit(limit=100):
    def inner(func):
        def replacement(*args, **kwargs):
            output = inner(*args, **kwargs)
            return output[:limit]
        return replacement
    return inner

@strawberry.type
class UserType:
    uuid: str

@strawberry.type
class Query:
    @strawberry.field
    @strawberry_limit(limit=10)  # does order matter?
    def users(self, info) -> List[UserType]:
        return User.objects.all()
```

The above style gives me the following error:

[...]

PR #289, opened 2020-04-14
Add support for generic types

patrick91:
This PR adds support for generic types, allowing to reuse them in an easy and pythonic way:

```python
T = typing.TypeVar("T")

@strawberry.type
class Edge(typing.Generic[T]):
    cursor: strawberry.ID
    node: T

@strawberry.type
class Query:
    @strawberry.field
    def int_edge(self, info, **kwargs) -> Edge[int]:
        return Edge(cursor=strawberry.ID("1"), node=1)
```

This will allow us to add support for pagination: #175, since now we should be able to create generic types that can be reused (see above example for a simplified relay Edge type).

### DIFF
--- a/strawberry/type.py
+++ b/strawberry/type.py
@@ -1,6 +1,7 @@
+import dataclasses
+import typing
 from functools import partial
 
-import dataclasses
 from graphql import GraphQLInputObjectType, GraphQLInterfaceType, GraphQLObjectType
 
 from .constants import IS_STRAWBERRY_FIELD, IS_STRAWBERRY_INPUT, IS_STRAWBERRY_INTERFACE
@@ -42,12 +43,15 @@ def _process_type(cls, *, is_input=False, is_interface=False, description=None):
     name = cls.__name__
     REGISTRY[name] = cls
 
-    def _get_fields(wrapped):
+    def _get_fields(wrapped, types_replacement_map=None):
         class_fields = dataclasses.fields(wrapped)
 
         fields = {}
 
         for class_field in class_fields:
+            if isinstance(class_field.type, typing.TypeVar):
+                class_field.type = types_replacement_map[class_field.type.__name__]
+
             field_name = getattr(class_field, "field_name", None) or to_camel_case(
                 class_field.name
             )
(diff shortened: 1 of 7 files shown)


botberry:



---

This release adds support for generic types, allowing
to reuse types easily, here's an example:

```python
T = typing.TypeVar("T")

@strawberry.type
class Edge(typing.Generic[T]):
    cursor: strawberry.ID
    node: T

@strawberry.type
class Query:
    @strawberry.field
    def int_edge(self, info, **kwargs) -> Edge[int]:
        return Edge(cursor=strawberry.ID("1"), node=1)
```

strawberry/type.py
```diff
@@ -110,7 +114,20 @@ def _get_fields(wrapped):
...
+        types_replacement_map = dict(
+            zip([param.__name__ for param in wrapped.__parameters__], types)
+        )
+        copied_name = "".join([type.__name__.capitalize() for type in types]) + name
+
+        return TypeClass(
+            copied_name,
+            lambda: _get_fields(wrapped, types_replacement_map=types_replacement_map),
+            **extra_kwargs
+        )
```
patrick91, line 127: this part is getting quite complex, might be worth refactoring this soon :)

strawberry/type_converter.py
```diff
@@ -34,7 +34,10 @@ def get_graphql_type_for_annotation(
     # TODO: this might lead to issues with types that have a field value
     is_field_optional = force_optional
 
-    if hasattr(annotation, "field"):
+    if is_generic(annotation):
+        # TODO: now we are assuming we have this function, but we should check
+        graphql_type = annotation.copy_with_type(*annotation.__args__)
```
patrick91, line 39: not sure we need this

strawberry/utils/typing.py
```diff
@@ -40,3 +40,16 @@ def get_optional_annotation(annotation):
...
+    if isinstance(annotation, typing._GenericAlias):
+        bases = [
+            getattr(b, "__origin__", None)
+            for b in getattr(annotation.__origin__, "__orig_bases__", [])
+        ]
+
+        if any(base is typing.Generic for base in bases):
+            return True
+
+    return False
```
patrick91, line 55: not sure what I have done here 🤔 will double check tomorrow, wish python had more utilities around typings

strawberry/type.py
```diff
@@ -110,7 +114,20 @@ def _get_fields(wrapped):
...
+
     wrapped.field = TypeClass(name, lambda: _get_fields(wrapped), **extra_kwargs)
+    wrapped.copy_with_type = copy_with_type
```
patrick91, line 130: we should forbid using the type directly, and only add copy with type when extending a generic

patrick91:
Useful resource: https://stackoverflow.com/questions/50563546/validating-detailed-types-in-python-dataclasses

PR #291, opened 2020-04-17
Allow to omit info and/or root params in resolvers

### LINKED ISSUE #257: Only pass info when resolver asks for it
I was thinking of making the API slightly easier for resolvers

so instead of having:

```python
@strawberry.type
class Query:
    @strawberry.field
    def user(self, info) -> User:
```

we could have this:

```python
@strawberry.type
class Query:
    @strawberry.field
    def user(self) -> User:
```

the difference is minimal, but it should make the library slightly more user-friendly 😊

patrick91:
This PR adds the possibility of omitting info and/or root params in resolvers.

I think this feature makes the code a bit nicer when not using root and info:

```python
def function_resolver() -> str:
    return "I'm a function resolver"

def function_resolver_with_params(x: str) -> str:
    return f"I'm {x}"

@strawberry.type
class Query:
    hello: str = strawberry.field(resolver=function_resolver)
    hello_with_params: str = strawberry.field(
        resolver=function_resolver_with_params
    )
```

I'm doing this now as I'm working on the getting started documentation and I think it will be nicer to use resolvers with these two params there.

closes #257

### DIFF
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -214,8 +214,27 @@ def _get_field(
         if check_permission:
             check_permission(source, info, **kwargs)
 
+        # the following code allows to omit info and root arguments
+        # by inspecting the original resolver arguments,
+        # if it asks for self, the source will be passed as first argument
+        # if it asks for root, the source it will be passed as kwarg
+        # if it asks for info, the info will be passed as kwarg
+
         kwargs = convert_args(kwargs, arguments_annotations)
-        result = wrap(source, info, **kwargs)
+        function_args = get_func_args(wrap)
+
+        args = []
+
+        if "self" in function_args:
+            args.append(wrap)
+
+        if "root" in function_args:
+            kwargs["root"] = source
+
+        if "info" in function_args:
+            kwargs["info"] = info
+
+        result = wrap(*args, **kwargs)
 
         # graphql-core expects a resolver for an Enum type to return
         # the enum's *value* (not its name or an instance of the enum).
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -1,11 +1,11 @@
 import textwrap
 import typing
+from dataclasses import InitVar, dataclass
 from enum import Enum
 
 import pytest
 
 import strawberry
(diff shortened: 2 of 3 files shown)


botberry:



---

This releases allows to use resolver without having 
to specify root and info arguments:

```python
def function_resolver() -> str:
    return "I'm a function resolver"

def function_resolver_with_params(x: str) -> str:
    return f"I'm {x}"

@strawberry.type
class Query:
    hello: str = strawberry.field(resolver=function_resolver)
    hello_with_params: str = strawberry.field(
        resolver=function_resolver_with_params
    )


@strawberry.type
class Query:
    @strawberry.field
    def hello(self) -> str:
        return "I'm a function resolver"

    @strawberry.field
    def hello_with_params(self, x: str) -> str:
        return f"I'm {x}"
```

This makes it easier to reuse existing functions and makes code
cleaner when not using info or root.

strawberry/field.py
```diff
@@ -215,7 +215,27 @@ def resolver(source, info, **kwargs):
             check_permission(source, info, **kwargs)
 
         kwargs = convert_args(kwargs, arguments_annotations)
-        result = wrap(source, info, **kwargs)
+
+        # allow to omit info and root arguments, we do this
+        # by inspecting the resolver arguments,
+        # if it asks for self, the source will be passed as first argument
+        # if it asks for root, the source it will be passed as kwarg
+        # if it asks for info, the info will be passed as kwarg
```
marcoacierno, line 223: I would just move the comment above `kwargs` because I was lost for a moment where kwargs was coming from 😅🤦‍♂️

tests/test_schema.py
```diff
@@ -529,3 +529,86 @@ def assortment(self, info) -> Cheese:
 
     assert not result.errors
     assert result.data["assortment"]["name"] == "Asiago"
+
+
+def test_optional_info_and_root_params_function_resolver():
```
marcoacierno, line 534: do we have a test that tries the `root` parameter?

PR #303, opened 2020-04-25
Serve debug API also on /

patrick91:
I was playing with Glitch[1] and I was thinking it would be nice to have the debug server on / as well, so the preview is not a "Not found" page :)

This PR adds the route for the debug server on / as well as /graphql. I wasn't able to find a way to serve all paths to the GraphQL app with starlette. But this should be good enough :)

[1] https://glitch.com/~starter-strawberry


botberry:



---

This releases updates the debug server to serve the API on '/'.

PR #308, opened 2020-04-29
Pass correct parameter as self

patrick91:
Fixes a bug introduced in #291 🙈

Previously we were passing wrap instead of source, which was wrong


botberry:



---

This release fixes a bug introduced by 0.24.0

PR #314, opened 2020-05-08
Fix subscriptions when they accept input arguments

marcoacierno:
Subscriptions with input arguments currently fail with

```
>       assert not result.errors
E       assert not [GraphQLError("_resolve() got an unexpected keyword argument 'name'", locations=[SourceLocation(line=1, column=16)], path=['example'])]
E        +  where [GraphQLError("_resolve() got an unexpected keyword argument 'name'", locations=[SourceLocation(line=1, column=16)], path=['example'])] = ExecutionResult(data=None, errors=[GraphQLError("_resolve() got an unexpected keyword argument 'name'", locations=[SourceLocation(line=1, column=16)], path=['example'])]).errors
```

because the resolver doesn't accept kwargs.

This PR adds support for it and sends the arguments to the `check_permissions` function

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+Fix error when a subscription accepted input arguments
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -250,9 +250,9 @@ def _get_field(
 
         if is_subscription:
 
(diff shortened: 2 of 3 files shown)


botberry:

```markdown
Release type: patch

```

Issue #306, opened 2020-04-29
Consistent handling of input types and args across optionals, nullables, and defaults.


coady:
Input args are using default values to indicate a nullable field, but not setting the `default_value` on the GraphQL object:
```python
In []: @strawberry.type 
    ...: class Query: 
    ...:     @strawberry.field 
    ...:     def hello(self, info, name: str = 'World') -> str: 
    ...:         return f"Hello {name}" 
    ...:                                                                                                                                                                                                                                        

In []: print(strawberry.printer.print_type(Query))                                                                                                                                                                                            
type Query {
  hello(name: String): String!
}
```
Whereas input types are using `Optional` to indicate a nullable field, and default values are ignored.
```python
In []: @strawberry.input 
    ...: class Options: 
    ...:     required: int 
    ...:     nullable: Optional[int] 
    ...:     omittable: int = 0 
    ...:     both: Optional[int] = 0 
    ...:                                                                                                                                                                                                                                        

In []: print(strawberry.printer.print_type(Options))                                                                                                                                                                                          
input Options {
  required: Int!
  nullable: Int
  omittable: Int!
  both: Int
}
```
But in GraphQL, being nullable is orthogonal to having a default.  All 4 combinations are supported on types and args.
```python
In []: schema = graphql.build_schema(''' 
    ...: type Query { 
    ...:     hello(name: String! = "World"): String! 
    ...: } 
    ...: input Options {  
    ...:     required: Int!  
    ...:     nullable: Int  
    ...:     omittable: Int! = 0  
    ...:     both: Int = 0  
    ...: }     
    ...: ''')
```
So I think the consistent way to handle this is to translate `Optional` to nullable, and defaults to `default_value`.

coady:
Looking at this a bit more, I have a simpler proposal.  Based on the observations:
* `String! = null` isn't allowed, which makes sense.
* `String = null` is allowed, but seems useless. A default `null` doesn't communicate anything to the client.
* `String = ""` is allowed, but an edge case.  It would imply that an explicit `null` actually behaves differently than a missing field.  And it's a code smell on the resolver side, because `Optional[str] = ""` doesn't look like it would handle an explicit `None` correctly.

Given all of that, maybe it would be better to ignore `Optional` for now (it can always be argued for later).  And support the 3 useful cases:
* `str` -> `String!`
* `str = None` -> `String`
* `str = ""` -> `String! = ""`

patrick91:
[...]

PR #324, opened 2020-05-19
Default values make input arguments nullable when the default is None.

coady:
## Description
The next step for #306.  A default of `None` still makes an input arg implicitly nullable as always, but a non-`None` keeps the arg non-nullable.  This matches input type behavior, and the arg is still "optional", just not nullable.

Although technically not backwards-compatible, it should have minimal impact because users who want a nullable field typically use `None` as a default.

## Types of Changes
- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* #306 

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,14 @@
+Release type: patch
+
+Default values make input arguments nullable when the default is None.
+```python
+class Query:
+     @strawberry.field
+     def hello(self, i: int = 0, s: str = None) -> str:
+         return s
+```
+```graphql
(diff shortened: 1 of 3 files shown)


botberry:



---

Default values make input arguments nullable when the default is None.
```python
class Query:
     @strawberry.field
     def hello(self, i: int = 0, s: str = None) -> str:
         return s
```
```graphql
type Query {
  hello(i: Int! = 0, s: String): String!
}
```

PR #326, opened 2020-05-21
Argument conversion doesn't populate missing args with defaults.

coady:
## Description
This allows resolvers to use a non-`None` sentinel to distinguish missing from explicitly null.
Input types also use `None` to imply optional, matching the function behavior.

## Types of Changes
- [x] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #306 

## Checklist

### DIFF
--- a/tests/test_enum.py
+++ b/tests/test_enum.py
@@ -247,7 +247,7 @@ def test_enum_falsy_values():
     @strawberry.input
     class Input:
         flavour: IceCreamFlavour
-        optionalFlavour: typing.Optional[IceCreamFlavour]
+        optionalFlavour: typing.Optional[IceCreamFlavour] = None
 
     @strawberry.type
     class Query:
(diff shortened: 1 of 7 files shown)


botberry:

```markdown
Release type: patch

```

tests/test_enum.py
```diff
@@ -247,7 +247,7 @@ class IceCreamFlavour(Enum):
     @strawberry.input
     class Input:
         flavour: IceCreamFlavour
-        optionalFlavour: typing.Optional[IceCreamFlavour]
```
patrick91, line 250: uhm, why this change?

coady: The default allows it to be instantiated with a missing arg.  (and to test that being explicitly `Optional` still works)

patrick91:
Testing this internally and will merge and release soon :)

tests/test_enum.py
patrick91: I'm wondering if it is ok to not set none as default automatically, guess explicit is better 🤔

PR #333, opened 2020-05-30
Fix asgi view to allow multiple subscriptions

marcoacierno:
Right now we cannot send multiple blocking subscriptions to the server because the ASGI hangs on the connection until the subscription is complete

- [ ] Make sure all the tasks are correctly cleaned up when the connection is closed both via Graphql's "STOP" and by force
- [ ] Add tests

Test schema:

```python
import asyncio
import typing
import strawberry


@strawberry.type
class User:
    name: str
    age: int


@strawberry.type
class Subscription:
    @strawberry.subscription
    async def sub1(self) -> typing.AsyncGenerator[User, None]:
        count = 0

        while True:
            count = count + 1
            yield User(name="A", age=count)
            await asyncio.sleep(5)

    @strawberry.subscription
    async def sub2(self) -> typing.AsyncGenerator[User, None]:
        count = 0

        while True:
            count = count + 1
            yield User(name="B", age=count)
            await asyncio.sleep(5)


@strawberry.type
class Query:
    @strawberry.field
    def user(self, info) -> User:


schema = strawberry.Schema(query=Query, subscription=Subscription)
```

### DIFF
--- a/strawberry/asgi/__init__.py
+++ b/strawberry/asgi/__init__.py
@@ -3,7 +3,7 @@ import typing
 
 from starlette.requests import Request
 from starlette.types import Receive, Scope, Send
-from starlette.websockets import WebSocket, WebSocketState
+from starlette.websockets import WebSocket, WebSocketDisconnect, WebSocketState
 
 from graphql import GraphQLError
 from graphql.error import format_error as format_graphql_error
@@ -39,7 +39,7 @@ class GraphQL:
         self.graphiql = graphiql
         self.keep_alive = keep_alive
         self.keep_alive_interval = keep_alive_interval
-        self._keep_alive_task = None
+        self._keep_alive_task: typing.Optional[asyncio.Task[typing.Any]] = None
         self.debug = debug
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send):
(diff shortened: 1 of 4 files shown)


botberry:

```markdown
Release type: patch

```

sarendsen:
@marcoacierno do you have time to add some tests to get that code coverage up? If not i could lend a hand with that

marcoacierno:
Thanks for the interest @sarendsen! I think the unit tests we will need are:

- Unit test that all the tasks are closed
- One unit test (or integration test/end2end) that starts two subscriptions (with different operation id), it should send and receive messages from both

I think the last one is the more complicated one 🤔

I am busy at work during the week but maybe in the weekend I can have a look or support you on discord/etc

sarendsen:
@marcoacierno don't think Ill have time this weekend but good to hear its close. I probably can move on it next week.
One thing what's missing I think is allowing queries and mutations over websockets. Only subscriptions are supported currently.

marcoacierno:
I think to add support for queries/mutations over Websocket it's probably best to open another PR

sarendsen:
Fine by me

patrick91 (approved):

strawberry/asgi/__init__.py
```diff
@@ -59,38 +59,70 @@ async def get_context(
...
-            elif message_type == GQL_START:
-                await self.start_subscription(
-                    message.get("payload"), operation_id, websocket
-                )
-            elif message_type == GQL_STOP:
-                await websocket.close()
+        try:
+            while (
+                websocket.client_state != WebSocketState.DISCONNECTED
+                and websocket.application_state != WebSocketState.DISCONNECTED
```
BryceBeagle, line 83:
```suggestion
                websocket.client_state is not WebSocketState.DISCONNECTED
                and websocket.application_state is not WebSocketState.DISCONNECTED
```

strawberry/asgi/__init__.py
```diff
@@ -59,38 +59,70 @@ async def get_context(
         return {"request": request}
 
     async def handle_keep_alive(self, websocket):
-        if websocket.application_state == WebSocketState.DISCONNECTED:
+        if (
+            websocket.application_state == WebSocketState.DISCONNECTED
```
BryceBeagle, line 63:
```suggestion
            websocket.application_state is not WebSocketState.DISCONNECTED
```

strawberry/asgi/__init__.py
```diff
@@ -129,12 +165,11 @@ async def start_subscription(self, data, operation_id: str, websocket: WebSocket
...
 
-        await self._send_message(websocket, GQL_COMPLETE, None, operation_id)
-
-        if self._keep_alive_task:
-            self._keep_alive_task.cancel()
-
-        await websocket.close()
+        if (
+            websocket.client_state != WebSocketState.DISCONNECTED
+            and websocket.application_state != WebSocketState.DISCONNECTED
```
BryceBeagle, line 170:
```suggestion
            websocket.client_state is not WebSocketState.DISCONNECTED
            and websocket.application_state is not WebSocketState.DISCONNECTED
```

strawberry/types/fields/resolver.py
```diff
@@ -79,7 +79,9 @@ def type(self) -> Type[T]:
 
     @cached_property
     def is_async(self) -> bool:
-        return iscoroutinefunction(self.wrapped_func)
+        return iscoroutinefunction(self.wrapped_func) or isasyncgenfunction(
+            self.wrapped_func
+        )
```
BryceBeagle, line 84:
```suggestion
        return (
            iscoroutinefunction(self.wrapped_func)
            or isasyncgenfunction(self.wrapped_func)
        )
```
Nitpicky code cleanup

strawberry/asgi/__init__.py
```diff
@@ -59,38 +59,70 @@ async def get_context(
...
-                    asyncio.create_task(self.handle_keep_alive(websocket))
-            elif message_type == GQL_CONNECTION_TERMINATE:
-                await websocket.close()
-            elif message_type == GQL_START:
-                await self.start_subscription(
-                    message.get("payload"), operation_id, websocket
-                )
-            elif message_type == GQL_STOP:
-                await websocket.close()
+        try:
```
BryceBeagle, line 80: Any chance this `try/catch` could be refactored a bit? It's quite long and daunting to look at? I think a good start would be to move each `message_type` switch/case to a dedicated function?

BryceBeagle (commented):
I know the merge was already performed, but I had a few comments/suggestions so I figured I'd do a review anyways. Feel free to ignore since already merged.

strawberry/types/fields/resolver.py
patrick91: what if I do
```suggestion
        return any(
             iscoroutinefunction(self.wrapped_func),
             isasyncgenfunction(self.wrapped_func)
        )
```

Just curios to know what you think 😊, btw will amend these when I fix the other subscription bug ;)

BryceBeagle: I like that even better

PR #336, opened 2020-06-03
Fix issue when using unions and generics in the same type

patrick91:
We weren't handling lists when trying to get the original type for a generic.

This PR fixes that, at least in the basic case, where we only have one
type per list :)


botberry:



---

Fixes a bug when using unions and lists together

PR #350, opened 2020-06-11
Add __call__ method on result of strawberry.union

patrick91:
This fixes an issue when using named unions with generic types.

Error that was fixed:
image

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,21 @@
+Release type: patch
+
+This release fixes an issue when using named union types in generic types,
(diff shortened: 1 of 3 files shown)


botberry:



---

This release fixes an issue when using named union types in generic types,
for example using an optional union. This is now properly supported:


```python
@strawberry.type
class A:
    a: int

@strawberry.type
class B:
    b: int

Result = strawberry.union("Result", (A, B))

@strawberry.type
class Query:
    ab: Optional[Result] = None
```

Issue #340, opened 2020-06-08
Forward references with non-builtin types


BryceBeagle:
I see that in #17, support for forward references was added, but is this fully implemented? 

If I enable PEP563 and define a field that uses a non-builtin type, I get an exception. Here I took the example from the Getting Started page, imported the annotations future feature, and changed age to an Optional field. This exception doesn't occur when the future feature is not imported.

```python
from __future__ import annotations

import typing

import strawberry


@strawberry.type
class User:
    name: str
    age: typing.Optional[int]


@strawberry.type
class Query:
    @strawberry.field
    def user(self, info) -> User:
        return User(name="Patrick", age=100)


schema = strawberry.Schema(query=Query)
```

```
Traceback (most recent call last):
  File "/home/ignormies/.local/share/virtualenvs/gql-bf-XGX4szKA-py3.8/lib/python3.8/site-packages/graphql/type/definition.py", line 735, in fields
    fields = resolve_thunk(self._fields)
  File "/home/ignormies/.local/share/virtualenvs/gql-bf-XGX4szKA-py3.8/lib/python3.8/site-packages/graphql/type/definition.py", line 293, in resolve_thunk
    return thunk() if callable(thunk) else thunk
  File "/home/ignormies/.local/share/virtualenvs/gql-bf-XGX4szKA-py3.8/lib/python3.8/site-packages/strawberry/type.py", line 133, in <lambda>
    lambda types_replacement_map=None: _get_fields(wrapped, types_replacement_map),
  File "/home/ignormies/.local/share/virtualenvs/gql-bf-XGX4szKA-py3.8/lib/python3.8/site-packages/strawberry/type.py", line 76, in _get_fields
    fields[field_name] = field(
  File "/home/ignormies/.local/share/virtualenvs/gql-bf-XGX4szKA-py3.8/lib/python3.8/site-packages/strawberry/utils/lazy_property.py", line 8, in _lazy_property
    setattr(self, attr_name, fn(self))
  File "/home/ignormies/.local/share/virtualenvs/gql-bf-XGX4szKA-py3.8/lib/python3.8/site-packages/strawberry/field.py", line 105, in graphql_type
    return _get_field(
  File "/home/ignormies/.local/share/virtualenvs/gql-bf-XGX4szKA-py3.8/lib/python3.8/site-packages/strawberry/field.py", line 197, in _get_field
    annotations = typing.get_type_hints(wrap, None, get_registered_types())
[...]

PR #368, opened 2020-07-07
Fix schema-basics.md example

mfigurski80:
- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

Example on line 56 does not in fact work.

Issue #340 - given that importing `__future__.annotations` is not yet supported,
I refactored the example to include @patrick91 suggestion of wrapping types in quotes:

> I mean in general. Personally I haven’t used them much blush
> 
> The alternative would be to use quotes around types, which I think should work

## Checklist

### DIFF
--- a/docs/01_general/schema-basics.md
+++ b/docs/01_general/schema-basics.md
@@ -60,16 +60,18 @@ import strawberry
 @strawberry.type
 class Book:
   title: str
-  author: Author
+  author: 'Author'
 
 @strawberry.type
 class Author:
   name: str
-  books: typing.List[Book]
+  books: typing.List['Book']
 ```
 
 As you can see the code maps almost one to one with the schema, thanks to
-python’s type hints feature.
+python’s type hints feature. However, note that we need to wrap our custom
+strawberry types with quotes in order to avoid a 'name Author not defined' 
+TypeError.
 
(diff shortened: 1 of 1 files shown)


botberry:

```markdown
Release type: patch

```

patrick91:
@mfigurski80 thanks for this, I'm wondering if can make a better example that doesn't require quotes, as it might be confusing for a gettings started 🤔

mfigurski80:
I agree, the quotes might make it a bit confusing.
However, while this example did trip me up a bit, it was still super useful because it's currently the only example of 'recursive' strawberry types. I would hate to lose that aspect because it was really important to my own understanding of this.

Ideally I suppose this example would be buried in a more specific part of the documentation so it's not out front

patrick91:
uh, good point, ok, let's merge it. I'm doing a rewrite of the internals of strawberry, maybe that will fix the issue with the future annotations (even though it might be a python issue IIRC). 

Thank you!

BryceBeagle:
This has already been merged, but it looks like the second usage of the exact same example in the same file was forgotten:
https://github.com/mfigurski80/strawberry/blame/26a5f2ff88c865ce1d13141e2132cec45a7d276f/docs/01_general/schema-basics.md#L118

PR #394, opened 2020-07-27
Add support for using lazy types in generics

patrick91:


botberry:



---

Add support for using lazy types inside generics.

PR #395, opened 2020-07-28
Fix issue when using generic types with federation

patrick91:
Fixes this bug when using generics and federation together: 

image


botberry:



---

Fix issue when using generic types with federation

PR #403, opened 2020-08-01
Fix support for using interfaces

### LINKED ISSUE #402: Error adding more than one implementation of an interface
**Observed Behaviour**: When i try to add two implementations of an interface, i get a duplicated type name exception

**Expected Behaviour**: Instead of trying to recreate the interface type again, reuse it.

**Steps to reproduce**:
1. Create an interface
2. Create two types which implement the interface
3. Launch `strawberry server app`
4. See it fails with ` Schema must contain uniquely named types but contains multiple types named '<InterfaceName>'`

**Snippet to reproduce the issue**
````python
from typing import List, Optional, Union
import strawberry
from strawberry import field


@strawberry.interface
class Person:
    name: str
    email: str


@strawberry.type
class Speaker(Person):
    job: str 


@strawberry.type
class Attendee(Person):
    interests: List[str]


def get_people_by_name(name: str): 
    return []


@strawberry.type
class Query:
    searchPeopleByName: List[Union[Speaker, Attendee]] = field(resolver=get_people_by_name)

````
**Full traceback:**
```
  File "/mnt/c/Users/<User>/code/nerdearla/test_app.py", line 30, in <module>
  File "/home/crow/.virtualenvs/venv/lib/python3.8/site-packages/strawberry/schema/schema.py", line 42, in __init__
    self._schema = GraphQLSchema(
  File "/home/crow/.virtualenvs/venv/lib/python3.8/site-packages/graphql/type/schema.py", line 240, in __init__
    raise TypeError(
TypeError: Schema must contain uniquely named types but contains multiple types named 'Person'.
```

### PR DESCRIPTION
**Pull request by patrick91**
Closes #402

My bad for this bug, I forgot to uncomment a test and I also add a TODO about this 🤦‍♂️

Fixed here though, thanks @Ambro17 for the report!

### DIFF
--- a/strawberry/schema/types/object_type.py
+++ b/strawberry/schema/types/object_type.py
@@ -12,32 +12,43 @@ def _get_object_type_for_type_definition(
     type_definition: TypeDefinition, type_map: TypeMap
 ) -> GraphQLType:
 
-    TypeClass: Type = GraphQLObjectType
+    if type_definition.name not in type_map:
+        TypeClass: Type = GraphQLObjectType
 
-    kwargs = {}
+        kwargs = {}
 
-    if type_definition.is_input:
-        TypeClass = GraphQLInputObjectType
-    elif type_definition.is_interface:
-        TypeClass = GraphQLInterfaceType
+        if type_definition.is_input:
+            TypeClass = GraphQLInputObjectType
+        elif type_definition.is_interface:
+            TypeClass = GraphQLInterfaceType
 
-    if type_definition.interfaces:
-        kwargs["interfaces"] = [
-            _get_object_type_for_type_definition(interface, type_map)
-            for interface in type_definition.interfaces
-        ]
+        if type_definition.interfaces:
+            kwargs["interfaces"] = [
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by botberry** [comment]



---

Fix issue when reusing an interface

File: strawberry/schema/types/object_type.py [hunk 1]
```diff
@@ -12,32 +12,43 @@ def _get_object_type_for_type_definition(
...
-    return TypeClass(
-        name=type_definition.name,
-        fields=lambda: {
-            field.name: get_field(field, type_definition.is_input, type_map)
-            for field in type_definition.fields
-        },
-        description=type_definition.description,
-        **kwargs,
-    )
+    if type_definition.name not in type_map:
```
Ambro17, line 15: I wonder what would happen now if i define two classes but different with different fields. Should we issue a warning that the duplicate type will be ignored?

patrick91: yes, good point, we should add a check for that :)

https://github.com/strawberry-graphql/strawberry/issues/404

PR #427, opened 2020-08-17
Fix check for unset values

patrick91:
This PR fixes how we change if a value is unset. Turns out in some cases the object is not the one we instanciated, for example when checking after a type has been converted to a dict using `dataclasses.todict`. Didn't check the source code but I guess they do a copy of the value.


botberry:



---

This releases fixes the check for unset values.

PR #428, opened 2020-08-18
Add process_result hook for django, flask and ASGI views

### LINKED ISSUE #424: Allow to process response in our views
In https://github.com/strawberry-graphql/strawberry/pull/423 we improve our views to support for passing context and root value to the schema, an additional improvement we can do is to allow for customising the response, most likely for customising and logging errors.

We do this at work to hide some internal errors. We currently do this by overriding the dispatch method in the django view. But we could provide a method to override for this :)

https://github.com/strawberry-graphql/strawberry/blob/master/strawberry/django/views.py#L79-L84

patrick91:
This hook can be useful to customise the GraphQL response and also to process errors :)

Closes #424
Closes #248

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,4 +1,4 @@
-Release type: patch
+Release type: minor
 
 Add `process_result` to views for Django, Flask and ASGI. They can be overriden
 to provide a custom response and also to process results and errors.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,47 @@
+Release type: patch
+
+Add `process_result` to views for Django, Flask and ASGI. They can be overriden
+to provide a custom response and also to process results and errors.
+
+Django example:
+
+```python
+# views.py
+from strawberry.django.views import GraphQLView as BaseGraphQLView
+from strawberry.http import GraphQLHTTPResponse
+from strawberry.schema import ExecutionResult
+
+class GraphQLView(BaseGraphQLView):
+    def process_result(self, result: ExecutionResult) -> GraphQLHTTPResponse:
+        return {"data": result.data, "errors": result.errors or []}
+
+```
+
+Flask example:
+
+```python
+# views.py
+from strawberry.flask.views import GraphQLView as BaseGraphQLView
+from strawberry.http import GraphQLHTTPResponse
+from strawberry.schema import ExecutionResult
+
+class GraphQLView(BaseGraphQLView):
+    def process_result(self, result: ExecutionResult) -> GraphQLHTTPResponse:
+        return {"data": result.data, "errors": result.errors or []}
+
+```
+
+ASGI example:
+
+```python
+from strawberry.asgi import GraphQL as BaseGraphQL
+from strawberry.http import GraphQLHTTPResponse
+from strawberry.schema import ExecutionResult
+
+from .schema import schema
+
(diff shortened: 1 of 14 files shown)


botberry:



---

Add `process_result` to views for Django, Flask and ASGI. They can be overriden
to provide a custom response and also to process results and errors.

Django example:

```python
# views.py
from strawberry.django.views import GraphQLView as BaseGraphQLView
from strawberry.http import GraphQLHTTPResponse
from strawberry.schema import ExecutionResult

class GraphQLView(BaseGraphQLView):
    def process_result(self, result: ExecutionResult) -> GraphQLHTTPResponse:
        return {"data": result.data, "errors": result.errors or []}

```

Flask example:

```python
# views.py
from strawberry.flask.views import GraphQLView as BaseGraphQLView
from strawberry.http import GraphQLHTTPResponse
from strawberry.schema import ExecutionResult

class GraphQLView(BaseGraphQLView):
    def process_result(self, result: ExecutionResult) -> GraphQLHTTPResponse:
        return {"data": result.data, "errors": result.errors or []}

```

ASGI example:

```python
from strawberry.asgi import GraphQL as BaseGraphQL
from strawberry.http import GraphQLHTTPResponse
from strawberry.schema import ExecutionResult

from .schema import schema

class GraphQL(BaseGraphQLView):
    async def process_result(self, result: ExecutionResult) -> GraphQLHTTPResponse:
        return {"data": result.data, "errors": result.errors or []}

```

RELEASE.md
```diff
@@ -0,0 +1,47 @@
...
+Django example:
+
+```python
+# views.py
+from strawberry.django.views import GraphQLView as BaseGraphQLView
+from strawberry.http import GraphQLHTTPResponse
+from strawberry.schema import ExecutionResult
+
+class GraphQLView(BaseGraphQLView):
+    def process_result(self, result: ExecutionResult) -> GraphQLHTTPResponse:
```
jkimbo, line 15: It might be useful to pass the request here as well?

patrick91: Very good point! Will change 😄

patrick91: done! @jkimbo I've removed the request param from the flask view, since request is a global there 😊

RELEASE.md
```diff
@@ -0,0 +1,52 @@
...
+```python
+# views.py
+from django.http import HttpRequest
+from strawberry.django.views import GraphQLView as BaseGraphQLView
+from strawberry.http import GraphQLHTTPResponse
+from strawberry.schema import ExecutionResult
+
+class GraphQLView(BaseGraphQLView):
+    def process_result(self, request: HttpRequest, result: ExecutionResult) -> GraphQLHTTPResponse:
+        return {"data": result.data, "errors": result.errors or []}
```
marcoacierno, line 20: Hmmmmmm I wonder if there is a way to do it that does not require writing a custom view anyway? Like `path('graphql', GraphQLView(process_result=custom_method_here))`? I still feel that it's more complicated than it needs to be?

patrick91: yeah, we could do that too, but it is not a huge difference, I'd keep this way :)

patrick91:
I'll force myself to write docs about the views before merging it :D

jkimbo (approved):

Issue #396, opened 2020-07-29
Improve how we fetch the type for fields with resolvers


patrick91:
To define a type for a field we use the approach used by dataclasses:

```python
@strawberry.type
class Q:
    x: int
```

But we also allow to use `strawberry.field` to pass a resolver:

```python
def get_x() -> int:
    return 123

@strawberry.type
class Q:
    x: int = strawberry.field(resolver=get_x)
```

Now, we could allow to omit the type on the field, like this:

```python
def get_x() -> int:
    return 123

@strawberry.type
class Q:
    x = strawberry.field(resolver=get_x)
```

And we'll use the type coming from the resolver, like we do when using `@strawberry.field` as a decorator.

This shouldn't be too difficult to achieve, I think the only complexity might be that removing the type might hide the field from the dataclasses fields, thus having to find it manually here: https://github.com/strawberry-graphql/strawberry/blob/master/strawberry/types/type_resolver.py#L224-L235

Currently we also allow to use resolvers without type hints:

```python
def get_x():
    return 123

@strawberry.type
class Q:
    x: int = strawberry.field(resolver=get_x)
```

I think we should keep this, and maybe just log a warning, which might be useful for debugging.

We also have to consider these two cases where things might be wrong:

1. Both places are missing type hints

```python
def get_x():
    return 123

@strawberry.type
class Q:
    x = strawberry.field(resolver=get_x)
```

I'm not 100% what happens in this case, I guess the field isn't shown in the schema currently.

2. Type hints don't match

```python
def get_x() -> str:
    return '456'

@strawberry.type
class Q:
    x = strawberry.field(resolver=get_x)
```

Here we need to throw an error and let the user know that something is wrong with the type hints :)

BryceBeagle:
You might also want to consider what will happen with 

```python
def get_x() -> int:
    return 456

@strawberry.type
class Q:
    x: float = strawberry.field(resolver=get_x)
```

patrick91:
@BryceBeagle That was supposed to be my second example, ironically I forgot the type hint 😂 

I'd throw an error in this case

BryceBeagle:
Well I specifically chose `int` and `float` because of [PEP484](https://www.python.org/dev/peps/pep-0484/#the-numeric-tower)

> when an argument is annotated as having type float, an argument of type int is acceptable

patrick91:
[...]

PR #429, opened 2020-08-19
Single resolver argument for strawberry.field

BryceBeagle:
## Description
Previously, `strawberry.field` had redundant arguments for the resolver, one for when `strawberry.field` was used as a decorator, and one for when it was used as a function. This PR seeks to unify them.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

#438 

## Checklist


botberry:

```markdown
Release type: patch

```

strawberry/field.py
```diff
@@ -9,13 +9,15 @@
 from .utils.str_converters import to_camel_case
 
 
-def check_return_annotation(field_definition: FieldDefinition):
-    f = cast(Callable, field_definition.base_resolver)
-    name = cast(str, field_definition.name)
+def get_return_annotation(field_definition: FieldDefinition) -> Type:
+    resolver = cast(Callable, field_definition.base_resolver)
+    name = resolver.__name__
```
BryceBeagle, line 14: Discussion: This is no longer the camelcase-d resolver name. I figure this is actually preferable, as this error will occur at startup. The camelcase name would only be helpful if this was a runtime error

strawberry/arguments.py
```diff
@@ -55,7 +53,7 @@ def get_arguments_from_resolver(
 
     if len(arguments_missing_annotations) > 0:
         raise MissingArgumentsAnnotationsError(
-            field_name, arguments_missing_annotations
+            resolver.__name__, arguments_missing_annotations
```
BryceBeagle, line 56: Discussion: This is no longer the camelcase-d resolver name. I figure this is actually preferable, as this error will occur at startup. The camelcase name would only be helpful if this was a runtime error

This also means this function now only needs one argument

strawberry/types/type_resolver.py
```diff
@@ -222,60 +222,181 @@ def _resolve_types(fields: List[FieldDefinition]) -> List[FieldDefinition]:
...
+        # Check if there is a matching Type #1 field:
+        if field_name in type_1_fields:
+            # Stop tracking the Type #1 field, an explict strawberry.field was
+            # defined
+            type_1_fields.pop(field_name)
+        # Otherwise, ensure that a resolver has been specified
+        else:
+            field_definition: FieldDefinition = field._field_definition
+            if field_definition.base_resolver is None:
+                raise WindowsError(...)
```
BryceBeagle, line 281: This actually doesn't get hit. An exception (a TypeError with reasonable detail) is thrown in `type_resolver._get_fields` when processing the dataclass. 

Should we just use that exception, or do we want to convert it to a Strawberry exception?

strawberry/types/type_resolver.py
```diff
@@ -260,22 +313,25 @@ def _get_fields(cls: Type) -> List[FieldDefinition]:
                 default_value=getattr(cls, field.name, undefined),
```
BryceBeagle, line 313: Why not just use `dataclasses.MISSING`?

strawberry/types/type_resolver.py
BryceBeagle: (note that WindowsError is a fill-in, and won't be in the final commit)

strawberry/field.py
```diff
@@ -33,6 +35,29 @@ def __init__(self, field_definition: FieldDefinition):
...
+
+        field_definition.origin_name = resolver_name
+
+        field_definition.origin = resolver
+        field_definition.base_resolver = resolver
+        field_definition.arguments = get_arguments_from_resolver(resolver)
+
+        field_definition.type = get_return_annotation(field_definition)
+
+        resolver._field_definition = field_definition  # type: ignore
```
BryceBeagle, line 57: This breaks when `resolver` is a classmethod. Apparently they are immutable? Not sure how to address...

strawberry/types/type_resolver.py
patrick91: I didn't think of that, we can change it 😊

tests/schema/test_resolvers.py
```diff
@@ -207,7 +208,7 @@ class User:
         age: int
 
         @classmethod
-        def get_users(cls):
+        def get_users(cls) -> "List[User]":
```
patrick91, line 214: we need to add tests to make sure that we use the type defined on the class, and that we throw when the types are different (between resolvers and the class field) :)

strawberry/field.py
```diff
@@ -33,6 +35,36 @@ def __init__(self, field_definition: FieldDefinition):
...
+
+        field_definition.origin_name = resolver_name
+
+        field_definition.origin = resolver
+        field_definition.base_resolver = resolver
+        field_definition.arguments = get_arguments_from_resolver(resolver)
+
+        field_definition.type = get_return_annotation(field_definition)
+
+        try:
```
BryceBeagle, line 57: We probably want to use something along the lines of
```python
if inspect.ismethod(resolver) and resolver.__self__ is cls:
    ...
```

There are other things that could, in theory, cause an AttributeError?

BryceBeagle: Also, does this work with the `hasattr(field, "_field_definition")` in `type_resolver._get_fields`?

patrick91: > There are other things that could, in theory, cause an AttributeError?

maybe, not sure when it would happen though :)

> Also, does this work with the hasattr(field, "_field_definition") in type_resolver._get_fields?

yup :) 

image

BryceBeagle: Actually it [looks like](https://stackoverflow.com/questions/7891277/why-does-setattr-fail-on-a-bound-method) `setattr` fails on _all_ bound methods, not just classmethods... so I think all we need is 
```python
inspect.ismethod(resolver)
```

BryceBeagle: > yup :)

huh very interesting... I didn't know there was a passthrough getattr of sorts on bound methods

patrick91: I wonder if we should check if a method is classmethod though, if we assign a bound that's not static or a classmethod then we might have a misusage, no?

BryceBeagle: I think someone could pass a bound instance method as the resolver, right?

```python
class ResolverClass:
    _state: SomeState
    def resolve_func(self, info, arg) -> int:
        return self.do_something(self._state, info, arg)

resolver_class = ResolverClass()
```
```python
@strawberry.type
class Query:
    resolver_class.resolve_func here is a bound method
    query: int = strawberry.field(resolver=resolver_class.resolve_func)
```

patrick91: oh, good point. I wonder what happens in that case with self though. This is something for later I guess (we can add an issue and a test for this), not sure it is going to be common anyway 💯

strawberry/types/type_resolver.py
BryceBeagle: Will address in another PR

tests/schema/test_resolvers.py
BryceBeagle: How strict do we want to be? Should any of these be valid?
```python
@strawberry.type
class Query:

    @strawberry.interface
    class Cheese:
        name: str
    @strawberry.type
    class Swiss(Cheese):
        canton: str

    def int_resolver() -> int: ...
    def swiss_resolver() -> Swiss: ...
    def uuid_resolver() -> UUID: ...
    
    my_float: float = strawberry.field(resolver=int_resolver)
    my_cheese: Cheese = strawberry.field(resolver=swiss_resolver)
    my_id: strawberry.ID = strawberry.field(resolver=uuid_resolver)
```

I think we already discussed the float/int example and determined that it should throw an exception. What about the others?

strawberry/types/type_resolver.py
```diff
@@ -222,36 +226,129 @@ def _resolve_types(fields: List[FieldDefinition]) -> List[FieldDefinition]:
...
+                # This should be caught by _wrap_dataclass in type.py, but just
+                # in case, we'll check again
+                raise MissingResolverError(field_name)
 
-    # get all the fields from the dataclass
-    dataclass_fields = dataclasses.fields(cls)
+            # resolver with @strawberry.field decorator must be typed
+            if field_definition.type is None:
+                resolver_name = field_definition.base_resolver.__name__
+                raise MissingReturnAnnotationError(resolver_name)
```
BryceBeagle, line 325: This check, and the one below, are currently swallowed by graphql.GraphQLSchema, which just returns a simple TypeError with our error's message.

I wonder if it would be best/possible to attempt calling `type_resolver._get_fields` _before_ building the schema?

tests/schema/test_resolvers.py
BryceBeagle: Discussed on Discord. Current plan is to do the type matching verification in another PR. Will start by enforcing _exact_ matches, potentially being more lenient afterwards.

strawberry/types/type_resolver.py
patrick91: oh really? that's annoying. I think maybe we can move the check when we run the decorator?

BryceBeagle: Well the issue is that `strawberry.field` doesn't know if it's being used as a decorator, or just with `resolver` as the only argument

As far as the `strawberry.field` can tell, these are equivalent, but only the first is valid:

```python
@strawberry.type
class Query:
    def my_func(): ...
    my_field: int = strawberry.field(resolver=my_func)

    @strawberry.field
    def my_func():
        ...
```

It doesn't know if the typing will/will not be provided externally on the field

strawberry/types/type_resolver.py
BryceBeagle: Created issue #439

strawberry/types/type_resolver.py
BryceBeagle: Issue already tracked in #396 . Will address in future PR

strawberry/field.py
BryceBeagle: I added a test for bounded instance method resolvers... Is there anything further that we wanted to investigate in another issue?

strawberry/types/type_resolver.py
```diff
@@ -222,36 +226,129 @@ def _resolve_types(fields: List[FieldDefinition]) -> List[FieldDefinition]:
...
+    for field_name, field in type_2_fields.items():
+
+        field_definition: FieldDefinition = field._field_definition
+
+        # Check if there is a matching Type #1 field:
+        if field_name in type_1_fields:
+            # Make sure field and resolver types are the same if both are
+            # defined
+            # TODO: Implement
+            # assert field.type == resolver.type
```
patrick91, line 304: can we add a reference to the issue for this? 😊

strawberry/types/type_resolver.py
```diff
@@ -260,22 +357,25 @@ def _get_fields(cls: Type) -> List[FieldDefinition]:
...
+                # Make sure field isn't already accounted for (occurs when using
+                # interfaces)
+                if field.name not in field_definitions
+            }
 
-    return fields
+            # Add base's field definitions to cls' field definitions
+            field_definitions = {**field_definitions, **fields}
+
+    return list(field_definitions.values())
```
patrick91, line 381: I'll double check this with fresh eyes tomorrow, but so far really good!

patrick91 (commented):
Added a first review, this is great work 👏

strawberry/type.py
```diff
@@ -22,6 +24,22 @@ def _get_interfaces(cls: Type) -> List[TypeDefinition]:
...
+        message: str
+        [message] = exc.args
+
+        pattern = re.compile(
+            r"'(?P<field_name>\w*)' is a field but has no type annotation"
+        )
+        match = pattern.match(message)
+        if match:
+            field_name = match.group("field_name")
+            raise MissingFieldAnnotationError(field_name) from exc
```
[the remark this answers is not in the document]
BryceBeagle: Oops and this doesn't even work if the pattern doesn't match -- I forgot to re-raise 🤦‍♂️

tests/types/test_resolvers.py
```diff
@@ -58,3 +77,20 @@ def hello2(self, info, query, limit) -> str:
...
+
+def test_raises_error_when_missing_annotation_and_resolver():
+    with pytest.raises(MissingFieldAnnotationError) as e:
+
+        @strawberry.type
+        class Query:
+            missing = strawberry.field(name="annotation")
+
+        # Avoid unused local warning
+        _ = Query
```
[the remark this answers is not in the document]
BryceBeagle: It's actually PyCharm... but I didn't want to leave a PyCharm-specific pragma to ignore it. Turns out I can ignore the flake8 error code and PyCharm understands it.

I do something similar [here](https://github.com/strawberry-graphql/strawberry/blob/3a4db59859ff6e4ae5900c5e820afab7f355235b/tests/schema/test_resolvers.py#L251), but that's to get PyCharm to quit complaining that the function could be static. There's no flake8 analog afaik. In hindsight, it may be too much...

Thoughts?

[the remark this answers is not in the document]
patrick91: yeah, I wonder why pychamr complains about self not being used, feels a bit strange, since it is a required param on classmethod :)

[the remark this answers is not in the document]
BryceBeagle: It's not exactly complaining that it's unused... moreso saying that the function should be static _because_ it's not used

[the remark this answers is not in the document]
BryceBeagle: For this test, I'm _specifically_ testing against instance methods, so it felt kinda off to have PyCharm insist that the function become static

BryceBeagle:
Should be ready for review again

strawberry/types/type_resolver.py
```diff
@@ -222,36 +226,128 @@ def _resolve_types(fields: List[FieldDefinition]) -> List[FieldDefinition]:
...
+            # Stop tracking the Type #1 field, an explict strawberry.field was
+            # defined
+            type_1_fields.pop(field_name)
+
+        # Otherwise, ensure that a resolver has been specified
+        else:
+            if field_definition.base_resolver is None:
+                # This should be caught by _wrap_dataclass in type.py, but just
+                # in case, we'll check again
+                raise MissingFieldAnnotationError(field_name)
```
patrick91, line 319: when would this happen? if it doesn't I'd skip this check 😊 
we can add a `assert field_definition.base_resolver` instead, so it is clear there was a bug :)

tests/types/test_resolvers.py
```diff
@@ -24,17 +25,35 @@ def name(self) -> str:
...
-    with pytest.raises(MissingReturnAnnotationError) as e:
-
+    @strawberry.type
+    class Query1:
         @strawberry.field
         def hello(self, info):
             return "I'm a resolver"
 
+    with pytest.raises(MissingReturnAnnotationError) as e:
+        strawberry.Schema(Query1)
```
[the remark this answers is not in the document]
patrick91: I'm going to check this tonight or tomorrow :)

tests/types/test_resolvers.py
```diff
@@ -58,3 +77,18 @@ def hello2(self, info, query, limit) -> str:
...
+# noinspection PyUnusedLocal
+def test_raises_error_when_missing_annotation_and_resolver():
+    with pytest.raises(MissingFieldAnnotationError) as e:
+
+        @strawberry.type
+        class Query:  # noqa: F841
+            missing = strawberry.field(name="annotation")
+
+    [message] = e.value.args
+    assert message == (
```
patrick91, line 91: nit: This reads better I think :)

```suggestion
    assert e.value.args[0] == (
```

patrick91 (approved):
Awesome work on this! I added a couple of minor comments, but nothing major.

I just want to check if we can raise `MissingReturnAnnotationError` as early as possible.

I think we can merge this after updating the minor comments 😊

strawberry/type.py
```diff
@@ -33,10 +64,12 @@ def _process_type(
 ):
     name = name or to_camel_case(cls.__name__)
 
-    wrapped = dataclasses.dataclass(cls)
+    wrapped = _wrap_dataclass(cls)
 
     interfaces = _get_interfaces(wrapped)
 
+    # TODO: check for missing types here
```
patrick91, line 71: @BryceBeagle what if we add some safety checks here? At least to check if all fields have return types. Would be amazing to check that before we do the schema generation, as the schema generation shouldn't have to deal with missing types :)

tests/types/test_resolvers.py
BryceBeagle: I dunno.... I find the single-item tuple syntax to be pretty bad...

I'll change it though to be consistent

patrick91: I just release in other places we check the whole tuple, which looks nicer 😊

BryceBeagle: Sorry, but I'm not sure what this comment is supposed to mean 😅

strawberry/types/type_resolver.py
BryceBeagle: If something else in the future ever calls `_get_fields` I figured it would be helpful to catch that issue here too

I'm a little torn here... The [entire `else` clause](https://github.com/strawberry-graphql/strawberry/blob/246d23d6998a33342186e9f7caa6ef96c422ec9d/strawberry/types/type_resolver.py#L315-L324) this is in currently does nothing but identify problems with the FieldDefinition. The [other item in here](https://github.com/strawberry-graphql/strawberry/blob/246d23d6998a33342186e9f7caa6ef96c422ec9d/strawberry/types/type_resolver.py#L321-L324) is also a check, but this time it _can_ currently get hit. It feels weird to have one be an `assert` and the other raise an exception, does it not?

tests/types/test_resolvers.py
patrick91: this: https://github.com/strawberry-graphql/strawberry/blob/3a4b6ec0effd3c86db781059c77618651f9d71dc/tests/types/test_resolvers.py#L34-L36 :)

as opposed to sue my suggestion above with args[0] :)

strawberry/types/type_resolver.py
patrick91: maybe, I just think they are different cases so it might be ok to have different ways to fail. One happens as a user error while using resolver, the other one, not sure :)

tests/types/test_resolvers.py
BryceBeagle: Oh I didn't even notice that the suggestion you proposed _didn't_ do that, cause you did it everywhere else and I guess I just glossed over it. That's why I mentioned the single-item tuple syntax -- that trailing comma is really easy to miss haha.

To clarify, you're suggesting I use the literal tuple (second example), instead of the indexing pattern (first example), right? I'll change it to whichever you want 👍.

strawberry/types/type_resolver.py
BryceBeagle: up to you 👍

tests/types/test_resolvers.py
patrick91: let's do this then 😊 https://github.com/strawberry-graphql/strawberry/blob/3a4b6ec0effd3c86db781059c77618651f9d71dc/tests/types/test_resolvers.py#L34-L36

strawberry/types/type_resolver.py
patrick91: let's keep it for now, I think if we do this: https://github.com/strawberry-graphql/strawberry/pull/429#discussion_r481392751

we might be good :)

PR #434, opened 2020-08-26
Always return 200 in Flask view

patrick91:
Fix for flask returning 400 when there were errors in the GraphQL response. 

Thanks @Ambro17 for the report!


botberry:



---

This release fixes the Flask view that was returning 400 when there were errors
in the GraphQL results. Now it always returns 200.

Issue #328, opened 2020-05-23
Documentation meta issue


patrick91:
## General

- [x] Getting started (https://strawberry.rocks/docs/)
- [x] Schema Basic (https://strawberry.rocks/docs/schema-basics)
- [x] Query Basics @jaydenwindle (https://strawberry.rocks/docs/queries)
- [x] Mutation Basics (https://strawberry.rocks/docs/mutations)
- [x] Subscriptions Basics 
- [x] Why strawberry (code first, batteries included, dataclasses) #603 

## Types

- [ ] Fields (#450)
	- [x] Private fields (https://github.com/strawberry-graphql/strawberry/pull/479)
- [ ] Input types
- [x] Interfaces
- [x] Generics
- [x] Custom scalars
- [x] [Enums](https://github.com/strawberry-graphql/strawberry/blob/master/docs/02_types/enums.md)
- [x] Union
- [ ] Schema directives
- [ ] Resolvers

## Integrations @jaydenwindle 

- [x] Django support
- [x] Flask support
- [x] ASGI Support

## Features

- [x] Error handling
	- > 1. what happens when there's an exception in the code (and how to debug if we manage to improve things there)
	  > 2. validation errors, ie when sending bad queries/inputs
      > 3. expected errors vs unexpected errors
- [x] File uploads
- [ ] Authentication
- [x] Pagination
- [x] Permissions
- [x] Federation (experimental)
- [x] Extensions (https://strawberry.rocks/docs/feature/extensions)

## Operation

- [ ] Tracing / profiling (https://strawberry.rocks/docs/ops/tracing)
- [ ] Testing
- [ ] Deployment

## Demos

- [ ] SWAPI 
- [ ] Subscription
- [ ] More

---

## Examples of good documentation

* https://docs.djangoproject.com/en/3.1/
* https://graphql.org/learn/

## Useful resources

* https://developers.google.com/tech-writing
* https://www.writethedocs.org/
* https://documentation.divio.com/

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/strawberry-graphql) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/strawberry-graphql/strawberry/issues/328">
<picture>
[...]

PR #450, opened 2020-09-07
(WIP) Field documentation

BryceBeagle:
Add documentation for strawberry.field

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation

* #328 (partial)

## Checklist


botberry:

```markdown
Release type: patch

```

patrick91:
We should also document `InitVar`s :)

image

jkimbo:
@BryceBeagle @patrick91 this is ready to review

docs/types/fields.md
```diff
@@ -0,0 +1,210 @@
...
+
+# Fields
+
+## Usage
+
+GraphQL fields can be declared in Strawberry using either
+[dataclass][dataclass_docs]-style fields, or using the `strawberry.field`
+function. The function can be used as a decorator around a resolver for the
+field, or directly provided to the function directly using its `resolver`
+argument.
```
patrick91, line 13: we can also use without resolver, for example to change the name of the field. Should we mention that?

docs/types/fields.md
```diff
@@ -0,0 +1,210 @@
...
+Strawberry must be able to infer the type of the field—either from the field
+itself, or from the return type of the resolver.
+
+```python
+@strawberry.type
+class CoolType:
+    my_field: int = strawberry.field()
+```
+
+If these two sources of typing differ, an exception is thrown.
```
patrick91, line 171: do we do this already?

docs/types/fields.md
jkimbo: I think the whole sentence needs rewording actually...

docs/types/fields.md
```diff
@@ -0,0 +1,210 @@
...
+        ...
+
+    my_field: int = strawberry.field(resolver=float_resolver)
+```
+
+## Private fields
+
+`strawberry.Private` allows you to pass data to a Strawberry type without it
+being exposed in the GraphQL schema. Sometimes this is useful so that a resolver
+on the type can return the correct value.
```
patrick91, line 187: mmh, not sure this is easy to understand. can we rephrase it?

docs/types/fields.md
```diff
@@ -0,0 +1,210 @@
...
+    @strawberry.field
+    def get_my_cool_type(self) -> CoolType:
+        return CoolType(private_stuff={"my_field": "Hi"})
+```
+
+<!--
+## Exceptions
+
+TODO
+TODO: What happens when the field has no resolver or default value
```
patrick91, line 209: we allow no resolvers (the default is getattr :))

docs/types/fields.md
jkimbo: Hmm just tried it in a shell and apparently not...

patrick91: ok! that's something we need to enable at some point:) 

https://github.com/strawberry-graphql/strawberry/issues/1149

PR #451, opened 2020-09-07
Fetch fields from bases before fetching all the other fields

patrick91:
This PR fixes an issue when extending a type:

```python
def name_resolver(
    id: strawberry.ID, argument: str, optional_argument: Optional[str]
) -> str:
    return "Name"

@strawberry.type
class NameQuery:
    name: str = strawberry.field(resolver=name_resolver)

@strawberry.type
class Query(NameQuery):
    pass
```

Previously the schema would look almost correct but the name field would be missing the arguments, since its field definition would be created twice

### DIFF
--- a/strawberry/types/type_resolver.py
+++ b/strawberry/types/type_resolver.py
@@ -280,22 +280,36 @@ def _get_fields(cls: Type) -> List[FieldDefinition]:
     """
     field_definitions: Dict[str, FieldDefinition] = {}
 
-    # Type #1 fields
+    # before trying to find any fields, let's first add the fields defined in
+    # parent classes, we do this by checking if parents have a type definition
+    for base in cls.__bases__:
+        if hasattr(base, "_type_definition"):
+            base_field_definitions = {
+                field.name: field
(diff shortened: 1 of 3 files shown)


botberry:



---

This releases fixes an issue when extending types, now
fields should work as they were working before even 
when extending an exising type.

strawberry/types/type_resolver.py
```diff
@@ -280,22 +280,36 @@ def _get_fields(cls: Type) -> List[FieldDefinition]:
...
+            base_field_definitions = {
+                field.name: field
+                # TODO: we need to rename _fields to something else
+                for field in base._type_definition._fields  # type: ignore
+            }
+
+            # Add base's field definitions to cls' field definitions
+            field_definitions = {**field_definitions, **base_field_definitions}
+
+    # then we can proceed with finding the fields for the current class
```
patrick91, line 296: I'm surprised this worked before, this should have been always done before find all the other fields

patrick91:
@BryceBeagle I'll merge this, but let me know if you have any thoughts on the changes :)

PR #453, opened 2020-09-08
Add experimental support for Pydantic

### LINKED ISSUE #302: Pydantic support?
Pydantic[1] seems to be gaining popularity (also thanks to Fast API).

It would be interesting to support Pydantic for input types, I suspect it might already work for basic cases. I wonder if we need to support advanced features from the start[2].

[1] https://pydantic-docs.helpmanual.io/
[2] For example: https://pydantic-docs.helpmanual.io/usage/model_config/

patrick91:
This PR adds (basic) support for Pydantic. Full support will likely be done with multiple PRs. For now the focus will be in being able to generate GraphQL types from a Pydantic model and use it.

So far we have a conversion being implemented, just for str and int (but we can easily add more scalars), we also have a way to reference other registered pydantic models, though the implementation is not really nice.

For this PR do be done we need:

- [x] Support for all the scalars
- [x] Support for extending types
- [ ] Proper resolvers and __init__ method (see explaination below).
- [x] Force users to declare the list of fields that should be exposed in GraphQL

Closes #302 
Solves #444


botberry:

```markdown
Release type: patch

```

AndrewIngram:
Sorry if i've missed this being discussed elsewhere. Given the example above, assuming you have a field that should return `UserType`, presumably you can't just return an instance of `User` (which is good, because one of the big issues with graphene-django is confusion about what your return values should be). If this is the case, `UserType` probably also needs some way of constructing an instance from `User`, e.g `UserType.from(user_instance)`.

strawberry/pydantic/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+
+            return cls(**{**instance_kwargs, **kwargs})
+
+        def to_pydantic(self) -> Any:
+            instance_kwargs = self.__dict__
+
+            return model(**instance_kwargs)
+
+        cls.from_pydantic = staticmethod(from_pydantic)
+        cls.to_pydantic = to_pydantic
```
AndrewIngram, line 86: these look like good ideas

patrick91: I know, right? ;) 

Thanks for the suggestions, I'm quite liking this API now 😊

tests/beta/pydantic/schema/test_basic.py
```diff
@@ -0,0 +1,85 @@
...
+
+    query = "{ user { friendNames } }"
+
+    result = schema.execute_sync(query)
+
+    assert not result.errors
+    assert result.data["user"]["friendNames"] == ["A", "B"]
+
+
+def test_basic_type_excluding_fields():
```
AndrewIngram, line 58: Semantically this feels backwards. Since the API is strictly about declaring which fields to include, rather than exclude. Perhaps instead have a test that asserts that not providing any fields to include is an error?

tests/beta/pydantic/test_basic.py
```diff
@@ -0,0 +1,151 @@
...
+        age: int
+        password: Optional[str]
+
+    @strawberry.beta.pydantic.type(User, fields=["age", "password"])
+    class UserType:
+        pass
+
+    definition: TypeDefinition = UserType._type_definition
+
+    assert definition.name == "UserType"
```
AndrewIngram, line 22: If there's a way to print the SDL for a single type, i'd add an assertion for that too. Primarily because developers aren't 'really going to be manually interrogating their types for `name`, `type` and `is_optional` very often, they're most concerned with what the end schema looks like.

strawberry/beta/pydantic/fields.py
```diff
@@ -0,0 +1,66 @@
...
+        if issubclass(type_, pydantic.ConstrainedInt):
+            return int
+        if issubclass(type_, pydantic.ConstrainedStr):
+            return str
+
+    if type_ in FIELDS_MAP:
+        type_ = FIELDS_MAP.get(type_)
+
+        if type_ is None:
+            raise UnsupportedTypeError()
```
jkimbo, line 64: Would be nice to include the type that is unsupported and ideally the field that its assigned to.

tests/beta/pydantic/schema/test_mutation.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @strawberry.type
+    class Query:
+        h: str
+
+    @strawberry.type
+    class Mutation:
+        @strawberry.mutation
+        def create_user(self, input: CreateUserInput) -> UserType:
+            return UserType(input.name)
```
jkimbo, line 26: Don't you need to run `data = input.to_pydantic()` here?

jkimbo (commented):
This is looking good 👍 Added some small comments

docs/03_features/pydantic.md
```diff
@@ -0,0 +1,153 @@
...
+import strawberry
+
+from .models import User
+
+@strawberry.beta.pydantic.input(model=User, fields=[
+    'id',
+    'name',
+    'friends'
+])
+class UserType:
```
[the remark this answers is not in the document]
AndrewIngram: Another point here, if the name of the class is what it ends up being in the schema, I don't think the docs should include examples with names like `FooType` and `FooInputType`, should just be `Foo` and `FooInput`. If it clashes with an existing namespace name like in this case, we should alias the original.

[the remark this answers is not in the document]
jkimbo: Not sure I agree with that @AndrewIngram . I think adding `Type` to the end of the type name is fine. It's not ideal but it's better than having to deal with any confusion around which thing you're dealing with.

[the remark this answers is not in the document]
AndrewIngram: People have a tendency to copy and adapt examples from docs because they see them as a fountain of "correct" examples (the React team is trying to tackle this problem right now), so I fear that this is going to result in people new to GraphQL end up with `FooTypes` littered throughout their schema and not realise that it's not really idiomatic GraphQL.

pyproject.toml
```diff
@@ -33,6 +33,7 @@ typing_extensions = "^3.7.4"
 opentelemetry-api = {version = "^0.13b0",optional = true}
 opentelemetry-sdk = {version = "^0.13b0",optional = true}
 python-dateutil = "^2.7.0"
+pydantic = {version = "^1.6.1", optional = true}
```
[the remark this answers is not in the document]
marcoacierno: no idea, @patrick91? I think it doesn't include pydantic by default as it's defined in "extras" below

docs/03_features/pydantic.md
```diff
@@ -0,0 +1,153 @@
...
+    pass
```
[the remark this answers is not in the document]
marcoacierno: I think we should allow adding additional fields

strawberry/beta/pydantic/fields.py
marcoacierno: I am not sure how we can go from the "value" of the type alias, to the name... @patrick91 ?

tests/beta/pydantic/schema/test_mutation.py
marcoacierno: `to_pydantic` is needed if you also want to run validation

docs/03_features/pydantic.md
```diff
@@ -0,0 +1,153 @@
...
+import pydantic
+import strawberry
+
+class User(BaseModel):
+    id: int
+    name: pydantic.constr(min_length=2)
+    signup_ts: Optional[datetime] = None
+    friends: List[int] = []
+
+@strawberry.beta.pydantic.input(model=User, fields=[
```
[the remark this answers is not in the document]
marcoacierno: where do we say it's an error type? can't find it to remove it, it seems to me the documentation is explaining the input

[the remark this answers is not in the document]
marcoacierno: found it I think 👌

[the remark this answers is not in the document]
marcoacierno: should be better now

docs/03_features/pydantic.md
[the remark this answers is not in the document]
marcoacierno: 👌

docs/03_features/pydantic.md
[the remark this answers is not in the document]
marcoacierno: this is now done & updated the doc

docs/03_features/pydantic.md
```diff
@@ -0,0 +1,153 @@
...
+@strawberry.beta.pydantic.type(model=User, fields=[
+    'id',
+    'name',
+])
+class UserType:
+    pass
+
+instance = User(id='123', name='Jake')
+
+data = UserType.from_pydantic(instance)
```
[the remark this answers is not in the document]
marcoacierno: this is now done

docs/03_features/pydantic.md
```diff
@@ -0,0 +1,232 @@
...
+
+The generated type will look like this:
+
+```python
+@strawberry.type
+class UserError:
+    id: List[str]
+    name: List[str]
+    friends: List[List[str]]
+```
```
jkimbo, line 103: I think it makes more sense to show the SDL here rather than the equivalent Strawberry type. You can even use the new `python+schema` code block to show them side by side (https://github.com/strawberry-graphql/strawberry.rocks/pull/23)

e.g.
```suggestion
```python+schema
import pydantic
import strawberry

class User(BaseModel):
    id: int
    name: pydantic.constr(min_length=2)
    signup_ts: Optional[datetime] = None
    friends: List[int] = []
    
@strawberry.beta.pydantic.error_type(model=User, fields=[
    'id',
    'name',
    'friends'
])
class UserError:
    pass
---
type UserError {
  id: [String!]!
  name: [String!]!
  friends: [[String!]!]!
}
```
```

File: docs/03_features/pydantic.md [hunk 12]
```diff
@@ -0,0 +1,232 @@
...
+
+The generated type will look like this:
+
+```python
+@strawberry.type
+class User:
+    id: int
+    name: str
+    age: str
+```
```
jkimbo, line 138: Same as above

strawberry/beta/pydantic/fields.py
jkimbo: I think you'll need to pass the field to this function so that you can extract the name from it.

tests/beta/pydantic/schema/test_mutation.py
jkimbo: 👍

tests/beta/pydantic/test_basic.py
jkimbo: There is a `print_type` function available from graphql-core: https://github.com/graphql-python/graphql-core/blob/7d826f0ec0a447cb869dfb891a755dbe8ea9b66f/tests/utilities/test_build_ast_schema.py#L922-L927

jkimbo (commented):
Some small suggestions for the docs but they aren't blocking

jkimbo:
Also would it make more sense for the docs to live under the "Integrations" section rather than "Features"?

docs/03_features/pydantic.md
marcoacierno: will change

marcoacierno:
I was thinking of having it in some section like `Experimental` to further separate it from the main documentation, what do you all think?

patrick91:
> I was thinking of having it in some section like Experimental to further separate it from the main documentation, what do you all think?

Haven't managed yet to do a review, but I'd say on this, let's keep the docs in the right place, but let's add a banner on the navigation item and on top of the documentation page 😊

strawberry/beta/pydantic/fields.py
marcoacierno: the problem is that we won't get `pydantic.NoneStr` but `Union[str, None]` which doesn't look great 🤔

tests/beta/pydantic/test_basic.py
marcoacierno: I will leave this test as it's right now as it's how we test the fields definition structure, and instead test the printed schema in the schema directory tests

docs/03_features/pydantic.md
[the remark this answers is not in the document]
patrick91: nice one! I was thinking of not doing it for the first release, mostly because I was lazy :D

pyproject.toml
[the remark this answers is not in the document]
patrick91: the idea is that we don't require pydantic by default, so that if you're not using you don't need to install (same for django, flask, etc). 

You could install either manually or by doing `pip install strawberry-graphql[pydantic]` (and this would install both strawberry and pydantic at once :)

strawberry/__init__.py
```diff
@@ -16,6 +16,7 @@
 
 __all__ = [
     "BasePermission",
+    "beta",
```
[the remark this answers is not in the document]
patrick91: yeah, I'm going to be honest, the more I think about this more I don't like beta

tests/beta/pydantic/test_basic.py
patrick91: yeah, I think we'll provide our own custom print type in future, especially if we add better support for federation (now we had to hack the printer for showing the directives)

strawberry/beta/pydantic/fields.py
patrick91: I'm not suggesting this, but we could inspect the frame and check :D

jkimbo: Oh I see, thats annoying. Well at least the field on the Pydantic model could be included in the error message? An error message like `Unsupported type for field "{field_name}" on Pydantic model "{pydantic_model_name}"` would make finding the offending field much easier.

docs/05_integrations/pydantic.md
```diff
@@ -0,0 +1,229 @@
+---
+title: Pydantic
+path: /docs/integrations/pydantic
+experimental: true
+---
+
+# Pydantic support
+
+Strawberry comes with support for
+[Pydantic](https://pydantic-docs.helpmanual.io/). This allows to create
```
BryceBeagle, line 10: `This allows the creation of Strawberry types`

BryceBeagle (changes requested):

PR #454, opened 2020-09-08
Add workflow to preview docs

patrick91:


botberry:

```markdown
Release type: patch

```

.github/workflows/preview-docs.yml
```diff
@@ -0,0 +1,60 @@
...
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  build:
+    name: Deploy docs preview
+    if:
+      github.event.issue.pull_request != '' &&
+      contains(github.event.comment.body, '/preview')
```
BryceBeagle, line 12: maybe this should be `startsWith`? I don't see it impossible that a comment could just happen to contain `/preview` for an unrelated reason. Someone could link a url that contains `/preview` in it, for example

patrick91: Good point! Will update that tomorrow, I still need to find a way to test this though 😅 not sure it will work, we need the secrets from GitHub but not sure they are sent to actions triggered on comments. Will check on a new repo tomorrow:)

patrick91:
aaa

Issue #361, opened 2020-06-29
Add schema visibility fillters


Ambro17:
Ruby has its way of hiding fields
https://graphql-ruby.org/authorization/visibility.html

Java has its way too
https://www.graphql-java.com/documentation/v12/fieldvisibility/

Javascript too
https://www.apollographql.com/docs/apollo-server/features/schema-transforms/

The main goal of this is to hide experimental or in progress types from clients, but to allow to include them for testing or staging for example

There seems to be two valid approaches for visibility filters.
- Per request filters based on user authorization and other context information. This should account for filtering types from introspection too
- Per schema filters that build a different schema based on rules upon types registering that don't change after the schema is built.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/strawberry-graphql) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/strawberry-graphql/strawberry/issues/361">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/361/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/361/pledge.svg">
</picture>
</a>

patrick91:
I'll mark this as blocked as I'm doing a rewrite on the internals :)

Should be easier to implement this after that 🤞

patrick91:
@Ambro17 I'm wondering if we should go with something similar to our permission:

```python
import strawberry


class IsPaidAPIUser(BaseVisiblity):
    message = "User is not authenticated"

    def has_permission(self, source, info):
        return False


@strawberry.type
class User:
    name: str
    age: int


@strawberry.type
class Query:
    @strawberry.field(visibility_classes=[IsPaidAPIUser])
    def user(self, info) -> User:
        return User(name="Patrick", age=100)


schema = strawberry.Schema(query=Query)
```

not really sure about the names to be honest

Ambro17:
Didn't know about s trawberry's permissions support! 
That would certainly be enough for the use case i was thinking when i opened the issue, i would keep the issue open until we get documentation on usage. Perhaps not only showing its usage as a permission based filter, but as a tool to maintain public and private parts of the schema

patrick91:
[...]

PR #455, opened 2020-09-10
[Discussion] Add kwarg to control fields visibility

Ambro17:
This PRs is incomplete and should serve as a complement of the discussion at #361 

For demonstration purposes i built a minimal schema with a hidden field that hides it from introspection and queries.
The behaviour of the visible function could be controlled by reading any relevant app config

## Example schema
```python
import strawberry


@strawberry.type
class Query:

    @strawberry.field
    def visible(self) -> str:
        return 'ABC'

    @strawberry.field(visible=lambda: False)
    def invisible(self) -> str:
        return 'ABC'


print('Schema SDL:\n', schema.as_str())
print('Introspection:\n', schema.execute_sync("""
    { 
        __schema {
            types {
                name
                fields {
                    name
                }
            }
        }
    }""")
)
print('Query Attempt:\n', schema.execute_sync('{invisible}').errors)
```
## Output
```
Schema SDL:
 type Query {
  visible: String!
}
Introspection:
 ExecutionResult(data={'__schema': {'types': [{'name': 'Query', 'fields': [{'name': 'visible'}]}, {'name': 'String', 'fields': None}, {'name': 'Boolean', 'fields': None}, {'name': '__Schema', 'fields': [{'name': 'description'}, {'name': 'types'}, {'name': 'queryType'}, {'name': 'mutationType'}, {'name': 'subscriptionType'}, {'name': 'directives'}]}, {'name': '__Type', 'fields': [{'name': 'kind'}, {'name': 'name'}, {'name': 'description'}, {'name': 'specifiedByUrl'}, {'name': 'fields'}, {'name': 'interfaces'}, {'name': 'possibleTypes'}, {'name': 'enumValues'}, {'name': 'inputFields'}, {'name': 'ofType'}]}, {'name': '__TypeKind', 'fields': None}, {'name': '__Field', 'fields': [{'name': 'name'}, {'name': 'description'}, {'name': 'args'}, {'name': 'type'}, {'name': 'isDeprecated'}, {'name': 'deprecationReason'}]}, {'name': '__InputValue', 'fields': [{'name': 'name'}, {'name': 'description'}, {'name': 'type'}, {'name': 'defaultValue'}]}, {'name': '__EnumValue', 'fields': [{'name': 'name'}, {'name': 'description'}, {'name': 'isDeprecated'}, {'name': 'deprecationReason'}]}, {'name': '__Directive', 'fields': [{'name': 'name'}, {'name': 'description'}, {'name': 'isRepeatable'}, {'name': 'locations'}, {'name': 'args'}]}, {'name': '__DirectiveLocation', 'fields': None}]}}, errors=None)
Query Attempt:
 [GraphQLError("Cannot query field 'invisible' on type 'Query'. Did you mean 'visible'?", locations=[SourceLocation(line=1, column=2)])]
```

## Alternative APIs
Pass filter function to Schema construction ([Reference](https://graphql-ruby.org/schema/limiting_visibility.html#filtering))
```python
def filter_types(gql_type) -> bool:
    pass

schema = strawberry.Schema(Query, except: filter_types)
```

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* #361 

## Checklist


botberry:

```markdown
Release type: patch

```

patrick91:
@Ambro17 interesting solution. I do like the API, but I wonder if we could do something where we can hide the fields based on the request?

I imagined this for public and private API, for example some API keys might have more fields :)

Ambro17:
As we discussed on discord, it _is_ possible to hide fields based on context, but it requires changes on graphql-core
I leave the diff link here for anyone interested to peek at the required changes
https://github.com/graphql-python/graphql-core/compare/master...Ambro17:master

Ambro17:
I think we could do something by adding some logic [here](https://github.com/strawberry-graphql/strawberry/blob/master/strawberry/schema/execute.py#L73-L74)
```
        with extensions_runner.validation():
            validation_errors = validate(schema, document)
```
If we added the context argument that is already available a few lines earlier we could send an error if any field is not visible for the given context.

What do you think about offering visibility as an extension? I'm not sure this was the intended use case when developed.. but it may work

patrick91:
> What do you think about offering visibility as an extension? I'm not sure this was the intended use case when developed.. but it may work

Sorry I missed this message, so, for schema visibility to work we need these two things to happen:

1. the schema introspection doesn't return the hidden fields
2. the schema doesn't run any resolver related to the fields 

I think they can both be done via extensions, as you suggested, but I wonder if it is the best way, as we are basically changing the behaviour of the schema, instead of changing the schema itself. It would be ideal to change the way the schema is composed before even running anything related to the execution. 

I feel like the work that @BryceBeagle is doing in https://github.com/strawberry-graphql/strawberry/pull/489 will help with this

Ambro17:
Hiding fields from the schema in build time is the easiest approach and it was what I first implemented, but i understood that you wanted to explore more granularity being able to hide them based on the context argument.
For that to work, the schema should have every field (perhaps wrapped in a custom object as the related pr proposes) and filter at execution time based on context.

In other words, if we want to dynamically hide fields, we would need a layer of abstraction on top of graphql core and perhaps the validation dance.

On the other hand, the second point may solve itself as soon as we decide on an approach on the first one.
I think it will never reach the resolver for a hidden field because a higher layer will be in charge of excluding it from a type fields

patrick91:
> In other words, if we want to dynamically hide fields, we would need a layer of abstraction on top of graphql core and perhaps the validation dance.

yup, this is something we want to achieve with #489 :)

The idea is to run the schema building with the context of the request in mind, so we can allow developers to check for things like an API key :)

Ambro17:
Oh i understand now, 
`strawberry.Schema(Query, Mutation, context={'api_scope': 'internal'})` would build schema `A` 
and
`strawberry.Schema(Query, Mutation, context={'api_scope': 'public'})` would build another schema.

Am i right?

In that case, the schema shown for each request would be the same?
Or should we update it/mask it once we receive a request?

patrick91:
> Or should we update it/mask it once we receive a request?

I would do this, I think it might be the best option, I'll see if anyone has done something like this anywhere 😊

Ambro17:
I think graphql ruby sets a good example, graphql java approach uses another approach using the visitor pattern that I didn't find really Pythonic but perhaps to disagree

patrick91:
@Ambro17 would you be ok to close this and move it to a discussion? :)

Ambro17:
No problem for me, I hadn't had the time to try passing the context to the query like ruby does not I won't have time on the near future so I think you can close it for staleness

patrick91:
@Ambro17 perfect. I think this is a feature we might support in future, will let you know!

PR #470, opened 2020-09-25
Fix issue when trying to print type with UNSET as default value

patrick91:
GraphQL core was trying to serialize UNSET to the type of the field where it was being used when printing the type.


botberry:



---

This release fixes an issue when trying to print a type
with a UNSET default value

PR #471, opened 2020-09-26
Use get_context for subscriptions as well

marcoacierno:
Right now subscriptions just receive "websocket" in the context and not what you set using `get_context`

My initial idea was to use `get_context` and send either request or websocket, right now the implementation I pushed uses `get_websocket_context` but can change it back to use one method with two parameters, but it will break current code that uses `get_context`

## ToDo

I can update the documentation once we agree on what interface we like for `get_context`

## Types of Changes

- [x] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

-- None

## Checklist

- [ ] My change requires a change to the documentation. (?)

### DIFF
--- a/strawberry/asgi/__init__.py
+++ b/strawberry/asgi/__init__.py
@@ -53,7 +53,9 @@ class GraphQL:
     async def get_root_value(self, request: Request) -> typing.Optional[typing.Any]:
         return None
 
-    async def get_context(self, request: Request) -> typing.Optional[typing.Any]:
+    async def get_context(
+        self, request: typing.Union[Request, WebSocket]
+    ) -> typing.Optional[typing.Any]:
         return {"request": request}
 
     async def handle_keep_alive(self, websocket):
@@ -98,7 +100,7 @@ class GraphQL:
         if self.debug:
             pretty_print_graphql_operation(operation_name, query, variables)
 
-        context = {"websocket": websocket}
+        context = await self.get_context(websocket)
 
         data = await self.schema.subscribe(
             query,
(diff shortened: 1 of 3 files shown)


botberry:

```markdown
Release type: patch

```

strawberry/asgi/__init__.py
```diff
@@ -54,6 +54,11 @@ async def get_root_value(self, request: Request) -> typing.Optional[typing.Any]:
     async def get_context(self, request: Request) -> typing.Optional[typing.Any]:
         return {"request": request}
 
+    async def get_websocket_context(
+        self, websocket: WebSocket
```
jaydenwindle, line 58: Since both the `Request` and `Websocket` objects are representations of a request (and have the same API for accessing request details like headers and query params), I'd recommend we stick with one context generator function instead of two.

Having both `get_context` and `get_websocket_context` as methods that users are expected to override means more effort for users. It also exposes an implementation detail of subscriptions (since theoretically they are protocol agnostic). 

Instead, we could modify `get_context` to accept a union type of `Union[Request, Websocket]` as the `request` parameter. That way, users only have to override one method to set the context for both websocket and http connections, and they can access request details using the same API for both.

patrick91: I like that! maybe in future we can also have an abstraction so we expose only one type that has the same API (if possible)

jkimbo: +1 on keeping a single method `get_context` that accepts a request or a websocket

patrick91 (approved):
Nice one!

PR #477, opened 2020-09-30
Remove type from field to prevent mypy issue

patrick91:


botberry:



---

This release fixes an issue with mypy when doing the following:

```python
import strawberry

@strawberry.type
class User:
    name: str = strawberry.field(description='Example')
```

tests/mypy/test_plugin.yml
```diff
@@ -36,3 +36,16 @@
...
+- case: test_field
+  main: |
+    import strawberry
+
+    @strawberry.type
+    class User:
+        name: str = strawberry.field(description='Example')
+
+    User(n="Patrick")
+    User(name="Patrick")
```
BryceBeagle, line 49: Does this line ever get hit?

patrick91: yes :) I've inverted the lines so it is more clear that it runs that line too :)

strawberry/field.py
```diff
@@ -70,7 +70,7 @@ def field(
     description: Optional[str] = None,
     permission_classes: Optional[List[Type[BasePermission]]] = None,
     federation: Optional[FederationFieldParams] = None
-) -> Union[Callable, StrawberryField]:
```
patrick91, line 73: @BryceBeagle I wonder if we need this in future 🤔

BryceBeagle: well hopefully it will just be `-> StrawberryField` in the future

BryceBeagle: Also, in hindsight, `StrawberryField` _is_ callable, so this is a bit redundant

BryceBeagle: what I _really_ should have done was
```python
def field(resolver: Optional[Callable[T]], ...) -> Union[Callable[T], StrawberryField]: ...
```

patrick91: I think that would still break mypy, if we don't update the plugin 😊

for reference `dataclasses.field` is not typed :)

BryceBeagle: Yeah for sure. I think the long-term solution would be to update the mypy plugin _and_ annotate here. Then we can have our 🎂 and eat it too.

PR #479, opened 2020-10-01
Add a Private class to exclude fields from the schema

jkimbo:
This PR adds `strawberry.Private` to be able to mark fields as "private" and so not show up in the GraphQL schema. This allows passing extra context to the type that can then be used in the resolvers.

## Description

When using Strawberry types you might want to pass extra data that shouldn't be exposed in the GraphQL schema but is necessary for your resolvers. `strawberry.Private` allows you to mark certain fields as "private" to allow this. Implementation is taken from the `dataclasses.InitVar` implementation. 

Example:

```python
import strawberry

@strawberry.type
class User:
	age: strawberry.Private[int]

	@strawberry.field
	def age_in_months(self) -> int:
		return self.age * 12
```

Ref: discussion on [Discord](https://discordapp.com/channels/689806334337482765/750088532479180840/761163294588272680) 

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

*

## Checklist


botberry:

```markdown
Release type: patch

```

tests/types/test_private_field.py
```diff
@@ -0,0 +1,61 @@
...
+    instance = Query(name="Luke", age=22)
+    assert instance.name == "Luke"
+    assert instance.age == 22
+
+
+def test_private_field_with_strawberry_field():
+    @strawberry.type
+    class Query:
+        name: str
+        age: strawberry.Private[int] = strawberry.field(description="🤫")
```
AndrewIngram, line 27: what's the use case for a private field being strawberry field?

jkimbo: I guess there isn't one, I just wanted to check if it was possible. It might make more sense to raise an error if you try and do this. What do you think @AndrewIngram ?

AndrewIngram: It feels like using strawberry.field contradicts the reason behind private fields, so raising an error might make sense.

patrick91: I agreed with raising an error 😊

RELEASE.md
```diff
@@ -0,0 +1,17 @@
...
+```python
+import strawberry
+
+@strawberry.type
+class User:
+	age: strawberry.Private[int]
+
+	@strawberry.field
+	def age_in_months(self) -> int:
+		return self.age * 12
```
patrick91, line 16: nit: can we use spaces here? :D

strawberry/ext/mypy_plugin.py
```diff
@@ -12,11 +12,21 @@ def lazy_type_analyze_callback(ctx: AnalyzeTypeContext) -> Type:
...
+
+
 class StrawberryPlugin(Plugin):
     def get_type_analyze_hook(self, fullname: str):
         if fullname == "strawberry.lazy_type.LazyType":
             return lazy_type_analyze_callback
 
+        if fullname == "strawberry.private.Private":
+            return private_type_analyze_callback
+
```
patrick91, line 29: hero! I'm glad this wasn't too difficult 😊

patrick91: can we add a test for this here: https://github.com/strawberry-graphql/strawberry/blob/master/tests/mypy/test_plugin.yml?

strawberry/private.py
```diff
@@ -0,0 +1,30 @@
...
+    def __init__(self, type):
+        self.type = type
+
+    def __repr__(self):
+        if isinstance(self.type, type):
+            type_name = self.type.__name__
+        else:
+            # typing objects, e.g. List[int]
+            type_name = repr(self.type)
+        return f"strawberry.Private[{type_name}]"
```
patrick91, line 27: nice!

tests/types/test_private_field.py
```diff
@@ -0,0 +1,61 @@
...
+    schema = strawberry.Schema(query=Query)
+
+    result = schema.execute_sync(
+        "query { ageInMonths }", root_value=Query(name="Dave", age=7)
+    )
+
+    assert not result.errors
+    assert result.data == {
+        "ageInMonths": 84,
+    }
```
patrick91, line 61: can we move this test in side `tests/schema/test_private_fields.py`?

patrick91 (commented):
Added a couple of minor comments, this looks pretty cool! Thanks for doing this! So quick 😊

tests/types/test_private_field.py
jkimbo: Done

RELEASE.md
jkimbo: Done

tests/types/test_private_field.py
jkimbo: Done

strawberry/ext/mypy_plugin.py
jkimbo: Done

RELEASE.md
```diff
@@ -0,0 +1,17 @@
+Release type: minor
+
+Add `strawberry.Private` type to mark fields as "private" so they don't show up in the GraphQL schema.
```
```suggestion
Added `strawberry.Private` type to mark fields as "private" so they don't show up in the GraphQL schema.
```

tests/mypy/test_plugin.yml
```diff
@@ -49,3 +49,21 @@
...
+
+        @strawberry.field
+        def age_in_months(self) -> int:
+            return self.age * 12
+
+        @strawberry.field
+        def wrong_type(self) -> int:
+            return self.age.trim()
+  out: |
+    main:13: error: "int" has no attribute "trim"
```
patrick91, line 69: I'll check why this fails locally 🤔

patrick91 (approved):
Awesome, I'll check why the tests are failing 😊

BryceBeagle:
A couple of questions:

1. Would it be a good idea to implement this using the permissions system, instead of first-party support for private members? I'm not very knowledgeable of how the PermissionClasses system exactly works, but I could imagine one could be made that provides _no_ access to a field?
2. Should we recommend (not enforce) that the private members still be prefixed with an `_` to match Python's style, i.e. have the documentation do so?

patrick91:
> Would it be a good idea to implement this using the permissions system, instead of first-party support for private members? I'm not very knowledgeable of how the PermissionClasses system exactly works, but I could imagine one could be made that provides no access to a field?

Permissioning won't hide the field from the schema. Maybe we should rename this to `HiddenField` or something else to make it clearer?

> Should we recommend (not enforce) that the private members still be prefixed with an _ to match Python's style, i.e. have the documentation do so?

We could, I guess it depends the use case too, maybe a field is hidden from GraphQL but it is not private in the code?

jkimbo:
> 1. Would it be a good idea to implement this using the permissions system, instead of first-party support for private members? I'm not very knowledgeable of how the PermissionClasses system exactly works, but I could imagine one could be made that provides _no_ access to a field?

As @patrick91 said the permissions system doesn't work here because it only applies at runtime while the "Private" field applies at schema creation time. 

> 2. Should we recommend (not enforce) that the private members still be prefixed with an `_` to match Python's style, i.e. have the documentation do so?

I don't think it's necessary to enforce that private fields/members should be prefixed with `_`. It's also technically possible for GraphQL fields to start with `_` so it might lead people to assume that you can't have fields that begin with `_` in a schema.

BryceBeagle:
> Permissioning won't hide the field from the schema. 

> As @patrick91 said the permissions system doesn't work here because it only applies at runtime while the "Private" field applies at schema creation time.

Makes sense. I'm on board with this feature.

> I don't think it's necessary to enforce that private fields/members should be prefixed with _. It's also technically possible for GraphQL fields to start with _ so it might lead people to assume that you can't have fields that begin with _ in a schema.

That's why I said _not_ enforced 🙂, just recommended. But maybe I'm being a bit too nitpicky with styling here, and that there are too many uses cases too worry about it.

> Maybe we should rename this to HiddenField or something else to make it clearer?

What about `InternalField`? I feel that that could address the slight ambiguities that exist with `PrivateField`/`HiddenField`.

BryceBeagle:
Also, I know that `InitVar` was discussed, but doesn't `typing.ClassVar` already [provide what we want](https://docs.python.org/3/library/dataclasses.html#class-variables)?

patrick91:
> Also, I know that InitVar was discussed, but does typing.ClassVar already provide what we want?

`ClassVar`s won't be added to the `__init__` method though: 

image

> What about InternalField? I feel that that could address the slight ambiguities that exist with PrivateField/HiddenField. To me, both of the latter imply that there is some mechanism that would allow exposure.

Not sure, I don't really like the `Field` suffix, as these variables are not fields anymore. I think `Private` is good as long as we document it :)

jkimbo:
> That's why I said _not_ enforced 🙂, just recommended. But maybe I'm being a bit too nitpicky with styling here, and that there are too many uses cases to worry about it.

Oh sorry I missed that! I don't have a strong opinion on whether we should recommend that private fields be prefixed with an `_`. I would say that these private fields aren't truly private because they still are available on the dataclass instance and are passed in as kwargs which leads me to the conclusion that they shouldn't be prefixed with `_`.

strawberry/private.py
```diff
@@ -0,0 +1,30 @@
+class Private:
+    """Represent a private field that won't be converted into a GraphQL field
+
+    Example:
+
+    ```
+    import strawberry
```
BryceBeagle, line 7: Other docstring examples in the repo use doctest style. May want to stay consistent?

BryceBeagle (changes requested):

strawberry/types/type_resolver.py
```diff
@@ -376,7 +382,8 @@ def _get_fields(cls: Type) -> List[FieldDefinition]:
                 default_value=getattr(cls, field.name, undefined),
             )
 
-        field_name = cast(str, field_definition.origin_name)
-        field_definitions[field_name] = field_definition
+        if not isinstance(field_definition.type, Private):
```
BryceBeagle, line 385: I would move this check inside of the `else` [here](https://github.com/strawberry-graphql/strawberry/pull/479/files#diff-23d017233da3ec1535492f9e4e36ef87R368-R383), and just `continue` if the `.type` is `Private` instead of going through the effort of creating the `FieldDefinition` and then abandoning it

BryceBeagle (changes requested):

strawberry/types/type_resolver.py
BryceBeagle: Very sad. The link doesn't appear to work. It's the `else` of the
```python
if hasattr(field, "_field_definition"): ...
else: ...
```
block right above

strawberry/private.py
jkimbo: Good point. Updated

strawberry/types/type_resolver.py
jkimbo: Done

BryceBeagle (approved):

Issue #461, opened 2020-09-17
Investigate the necessity of comparing against NoneType in UnionDefinition.types


BryceBeagle:
Currently, `UnionDefinition.types` returns all de-generified types, after filtering those that are `NoneType` (`None.__class__`)
https://github.com/strawberry-graphql/strawberry/blob/8126ec0b6a7acf0edf553752d703bcd1f6273ac4/strawberry/union.py#L18-L22

@patrick91 explained that the `NoneType` filter was added to handle `Optional[...]` (which expands to `Union[..., NoneType]`), but this may not be an issue after all.

No test fails if this check is removed. Either the check is not necessary or we have a missing test.

DoctorJohn:
Closing since the mentioned code does no longer exist in this form (especially the filtering seems to be gone).

Issue #449, opened 2020-09-06
Improve public interface of TypeDefinition


patrick91:
We internally use TypeDefinition to generate an intermediate GraphQL type what will later used to create GraphQL-core types.

This TypeDefinition should be improve in order to allow developers to programmatically create types when needed. For example this will be useful for adding support to pydantic.

Something we need to improve for sure are:

- the `_fields` init param should be `fields`
- we should find a better way to call `_resolve_types`, this is useful for schema generation as it can be used to resolve types later (when they are all defined).

https://github.com/strawberry-graphql/strawberry/blob/master/strawberry/types/types.py#L16-L54

patrick91:
Closing this as we have #489 that has better way of improving the internal structure :)

nrbnlulu:
This should probably re-opened @patrick91 ?

Issue #448, opened 2020-09-05
Fields declared with strawberry.field function with resolver from same class are not created properly


BryceBeagle:
(Issue will be introduced with the merge of PR #429)

```python
@strawberry.type
 class Query:  # noqa: F841
     def adios(self) -> int:
         return -1
 
     goodbye: int = strawberry.field(resolver=adios)
```

This fails:
```python
definition = Query._type_definition
assert len(definition.fields) == 1
```

This also fails. Only `adios` is added as a field to the schema, which is not what we want:
```python
schema =  strawberry.Schema(query=Query)
query = "{ goodbye }"
result = schema.execute_sync(query)

assert not result.errors
```

jordangarside:
This is related to not being able to use the same function for multiple resolvers correct?

```py
def get_empty_string(root, info) -> str:
    return ""

@strawberry.type
class Book:
    id: str = strawberry.field(resolver=get_empty_string)
    title: str = strawberry.field(resolver=get_empty_string)

async def get_books():
    ...

@strawberry.type
class Query:
    books: typing.List[Book] = strawberry.field(resolver=get_books)
```

Only `id` will be added to the schema.

BryceBeagle:
Yes I believe this is related. It also provides a bit clearer of a picture for the source of the problem:

https://github.com/strawberry-graphql/strawberry/blob/ebcf62a8b91130d98ef98cf795b6d251effec2cb/strawberry/field.py#L39-L49

When a resolver is added to a field (either immediately using the `strawberry.field` argument, or deferred using `StrawberryField.__call__`), what's returned is the resolver _itself_, with a monkeypatched `_field_definition` attribute.

When the resolver exists on the class, or two different fields use the same resolver, the class will end up having multiple fields that refer to the same resolver object, causing problems when generating the schema.

patrick91:
@BryceBeagle I think this is fixed, @jordangarside's example works well now too, shall we close?

BryceBeagle:
Sounds good to me 👍

Issue #447, opened 2020-09-05
Fields (with resolvers) missing both type annotation and resolver return type annotation should raise MissingFieldAnnotationError instead of MissingReturnAnnotationError


BryceBeagle:
(Issue will be introduced with the merge of PR #429)

```python
@strawberry.type
class Query2:
    def adios(self):
        return -1
 
    goodbye = strawberry.field(resolver=adios)
```

Currently this raises a `MissingReturnAnnotationError`. It should raise a `MissingFieldAnnotationError` instead

Birdi7:
Hi everyone! I found this code in `object_type.py::_check_field_annotations`. What’s the correct way to fix this? If we replace the current `MissingReturnAnnotationError` with `MissingFieldAnnotationError`, users will lose the ability to distinguish between these two cases. Is that correct?

bellini666:
Hey @Birdi7 ,

Wow, this one has been here for a while. I think the main reason why this should be `MissingFieldAnnotationError` instead of `MissingReturnAnnotationError` is because this code expects a `goodbye: <sometype> = strawberry.field(...)`

> users will lose the ability to distinguish between these two cases

Could you expand on this?

Birdi7:
> Could you expand on this?

My hypothesis on the reason behind two exceptions as of today: `MissingFieldAnnotationError` is meant to direct user to add annotation to the field definition, and `MissingReturnAnnotationError` is meant to change resolver's type annotation. I might be wrong on this. 

> because this code expects a goodbye: <sometype> = strawberry.field(...)

My current undestanding of the issue is that we should replace `raise MissingReturnAnnotationError` inside `_check_field_annotations` with `raise MissingFieldAnnotationError`. Does it sound correct? If it does, I see no other usage of `MissingReturnAnnotationError` and we should remove it completely? Or we should create something like the following?

1. Field has annotation  → all good
4. Field has no annotation and has resolver with annotation → raise `MissingFieldAnnotationError`
5. Field has no annotation and has resolver without annotation → raise `MissingFieldAnnotationError`
6. Field has no annotation and has no resolver → raise `MissingFieldAnnotationError`

bellini666:
I think so, yes. But `MissingReturnAnnotationError` would still be useful when using a resolver with a `@strawberry.field` decorator, like:

```python
@strawberry.type
class Query:
    @strawberry.field
    def goodbye(self):
        return -1
```

patrick91:
[...]

PR #489, opened 2020-10-04
(WIP) (Proposal) Convert everything to well-defined and typed objects

BryceBeagle:
## Description

(Proposal)

Convert all Strawberry-related types to well-defined objects for complete typing and enhanced extend-ability. 

This PR is meant to serve as a reference guide for potential future work on the project, but should not itself be merged in. The goal is take the changes made in this branch(/PR) and slowly integrate them into the master branch without breaking things. Eventually the two branches will merge and this PR can be closed.

This work is very much so a WIP and is very very incomplete so far. Imports are broken, tests are broken, pre-commit has been disabled, types are just missing, etc. These will be resolved slowly, but eventually.

[(Rough) UML diagram](https://viewer.diagrams.net/?highlight=0000ff&edit=_blank&layers=1&nav=1&title=Strawberry%20Types#Uhttps%3A%2F%2Fdrive.google.com%2Fuc%3Fid%3D1CBppvarzluikBW_mYQMT9eJWSpJIZH9y%26export%3Ddownload)

## Types of Changes

- [x] Core
- [x] Bugfix
- [x] New feature
- [ ] Documentation

Eventually:
* #396
* #447 
* #448
* #449
* #461 

## Checklist


botberry:

```markdown
Release type: patch

```

patrick91:
oh, wow, this is probably one of the best PRs I've ever reviewed 🙌

I really like the approach, so much better than what I came up with previously ☺️, how would you now proceed? Shall I add some comments here? Or start with the small PRs first?

❤️ 🍓

strawberry/backends/base_backend.py
```diff
@@ -0,0 +1,33 @@
...
+
+    # TODO: Other parameters
+    @abstractmethod
+    async def execute_async(self, query: str) -> ...:
+        ...
+
+    # TODO: Other parameters
+    @abstractmethod
+    async def subscribe(self, query: str) -> ...:
+        # TODO: Should you be able to make subscription queries from execute and
```
patrick91, line 30: no, subscription are only to be done here :)

strawberry/backends/base_backend.py
```diff
@@ -0,0 +1,33 @@
...
+        #       execute_async?
+        # TODO: Should there be a subscribe/subscribe_async combo?
```
patrick91, line 32: subscription are always asynchronous, as they relay on things like web sockets :)

strawberry/schema/schema.py
```diff
@@ -26,9 +50,7 @@ def __init__(
         subscription: Optional[Type] = None,
         directives=(),
         types=(),
-        extensions: Sequence[Type[Extension]] = (),
```
patrick91, line 29: any reason for removing them from here?

strawberry/types/federation/field.py
```diff
@@ -0,0 +1,35 @@
+from typing import Union
+
+from strawberry.types import StrawberryField
+
+_FIELD_FED_TYPE = Union[StrawberryField, 'StrawberryFederatedField']
```
patrick91, line 5: should this use typing.NewType?

strawberry/types/federation/object_type.py
```diff
@@ -0,0 +1,79 @@
+from typing import List, Optional, Union
+
+from strawberry.types import StrawberryObjectType, StrawberryUnion
+
+_OBJECT_FED_TYPE = Union[StrawberryObjectType, 'StrawberryObjectTypeFederation']
+
+
+# TODO: Should federation use composition or inheritance?
+#       i.e. should StrawberryObjectTypeFederation inherit from
+#       StrawberryObjectType or just hold a reference to one?
```
patrick91, line 10: federation is directives added to fields and objects, so I’d say we don’t need to extend a type, but we need a way to add directives/metadata to the field :)

strawberry/types/federation/object_type.py
```diff
@@ -0,0 +1,79 @@
...
+
+# TODO: This name is probably too long and too verbose. Think of something
+class StrawberryObjectTypeFederation:
+    """
+
+    TODO: One decorator for federation? Or multiple? This feels a bit stacked
+    >>> @strawberry.key("upc")
+    ... @strawberry.extends(OtherType)
```
patrick91, line 18: nit; extends doesn’t extend another type, it extends the current type in the federated GraphQL (based on the name) so it doesn’t need to accept a param :)

strawberry/types/federation/object_type.py
```diff
@@ -0,0 +1,79 @@
...
+    ... @strawberry.type
+    ... class ThisType:
+    ...     @strawberry.field
+    ...     def resolver(self) -> OtherType:
+
+    TODO: This is probably better?
+    >>> @strawberry.federation(extends=OtherType, key="upc")
```
patrick91, line 25: I prefer `strawberry.federation.field` :D

strawberry/types/federation/object_type.py
```diff
@@ -0,0 +1,79 @@
...
+    ...     @strawberry.field
+    ...     def resolver(self) -> OtherType:
+
+    """
+
+    def __init__(self, cls: _OBJECT_FED_TYPE):
+
+        if isinstance(cls, StrawberryObjectType):
+            ...
+        elif isinstance(cls, StrawberryObjectTypeFederation):
```
patrick91, line 43: what’s this for?

strawberry/types/fields/field.py
```diff
@@ -0,0 +1,144 @@
...
+        """The name specified explicitly, or if not set the wrapped resolver's
+        name
+        """
+        if self._name is not None:
+            return self._name
+
+        if self.resolver is not None:
+            return self.resolver.name
+
+        # TODO: Should we raise an exception instead?
```
patrick91, line 84: This case should never happen, so I’d say yes :)

strawberry/types/fields/field.py
```diff
@@ -0,0 +1,144 @@
...
+        return None
+
+    @name.setter
+    def name(self, name: str):
+        self._name = name
+
+    # TODO: Implement. What is it?
```
patrick91, line 91: I used this to attach the parent class IIRC, it is used to be able to understand the actual type class (especially when type hints are strings)

strawberry/types/fields/field.py
```diff
@@ -0,0 +1,144 @@
...
+    @property
+    def origin(self):
+        ...
+
+    @property
+    def type(self) -> StrawberryObject:
+        # TODO: This can't be set anywhere
```
patrick91, line 98: what do you mean?

strawberry/types/fields/field.py
```diff
@@ -0,0 +1,144 @@
...
+        if self._type is not None:
+            return self._type
+
+        if self.resolver is not None:
+            return self.resolver.type
+
+        # TODO: Should we raise an exception instead?
```
patrick91, line 105: yes :)

strawberry/types/fields/mutation.py
```diff
@@ -0,0 +1,6 @@
+from strawberry.types import StrawberryField
+
+
+# TODO: What advantage does a separate type provide?
+class StrawberryMutation(StrawberryField):
+    ...
```
patrick91, line 6: not much to be honest, for decorators it helps with understand what’s a mutation and what’s a field (when reading code) :)

strawberry/types/fields/query.py
```diff
@@ -0,0 +1,6 @@
+from strawberry.types import StrawberryField
+
+
+# TODO: What advantage does a separate type provide?
+class StrawberryQuery(StrawberryField):
```
patrick91, line 5: we probably don’t need this :)

strawberry/types/fields/resolver.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @cached_property
+    def arguments(self) -> Set[StrawberryArgument]:
+        # TODO: Implement
+        ...
+
+    @property
+    def description(self) -> Optional[str]:
+        # TODO: Do resolvers get descriptions?
+        return self._description
```
patrick91, line 38: no, description is only for object types, directives, fields and maybe arguments :)

strawberry/types/fields/resolver.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @property
+    def name(self) -> str:
+        # TODO: What to do if resolver is a lambda?
+        return self.wrapped_func.__name__
```
patrick91, line 43: do we need the name here?

strawberry/types/object_types/object_type.py
```diff
@@ -0,0 +1,180 @@
...
+    """Annotates a class as a GraphQL type.
+
+    Example usage:
+
+    >>> @strawberry.type:
+    >>> class X:
+    >>>     field_abc: str = "ABC"
+    """
+    StrawberryObjectType.check_dataclass(cls)
+    wrapped_dataclass = dataclasses.dataclass(cls)
```
patrick91, line 168: I wonder how we can make it easier to use pydantic here for example 🤔

strawberry/types/schema/schema.py
```diff
@@ -0,0 +1,32 @@
...
+    def __init__(self, query: StrawberryObjectType, *,
+                 mutation: Optional[StrawberryObjectType] = None,
+                 subscription: Optional[StrawberryObjectType] = None,
+                 backend: Type[StrawberryBackend] = GraphQLCoreBackend):
+        self.backend = backend(query, mutation, subscription)
+
+    # TODO: I switched from execute_sync to execute_async as that appears to be
+    #       the standard for async "overloads". Certainly up for discussion;
+    #       leaning on switching it back
+    #       https://stackoverflow.com/q/52459138/8134178
```
patrick91, line 19: I kinda think that for graphql the default should async, since you get more benefits with it, also helps with pushing async adoption I guess ^^

strawberry/types/schema/schema.py
```diff
@@ -0,0 +1,32 @@
...
+    async def execute_async(self, query: str):
+        return await self.backend.execute_async(query)
+
+    @cached_property
+    def types(self) -> Set[StrawberryType]:
+        ...
+
+    # TODO: Should this just be __str__?
+    def as_str(self) -> str:
+        ...
```
patrick91, line 32: if we have a good __repr__, yes :D

strawberry/types/types/argument.py
```diff
@@ -0,0 +1,29 @@
...
+from .object import StrawberryObject
+from .type import StrawberryType
+
+T = TypeVar("T", bound=StrawberryObject)
+
+
+class StrawberryArgument(StrawberryType[T]):
+    def __init__(self, name: str, type_: Type[T], *,
+                 description: Optional[str],
+                 # TODO: Can we use None, or do we want fields.NO_DEFAULT_VALUE?
```
patrick91, line 12: I think this is similar to default values for fields, we can’t really use None by default :)

strawberry/types/types/object.py
```diff
@@ -0,0 +1,17 @@
...
+from typing import Optional
+
+
+class StrawberryObject(ABC):
+
+    @property
+    @abstractmethod
+    def name(self) -> Optional[str]:
+        """Optional to support anonymous types"""
+        # TODO: Should it actually be optional?
```
patrick91, line 11: I’d say no :)

strawberry/backends/base_backend.py
BryceBeagle: Could the subscription call itself be synchronous and return an object that behaves asynchronously?

strawberry/types/schema/schema.py
BryceBeagle: Sounds good enough to me :+1:

strawberry/backends/base_backend.py
patrick91: uhm, don't think so, but I might be wrong :)

strawberry/schema/schema.py
BryceBeagle: Are extensions something the schema should care about? Or should they be at a higher level?

BryceBeagle: Is this the kind of extension that is being dealt with here: https://spec.graphql.org/June2018/#sec-Schema-Extension ?

strawberry/types/federation/field.py
BryceBeagle: I'm not sure I understand why? I just didn't want to have to keep writing Union[..., ...] everywhere

strawberry/types/federation/object_type.py
BryceBeagle: Could we just add a `directives` field, to which federation would add to?

strawberry/backends/base_backend.py
BryceBeagle: To me, it seems that if we ever want to _fully_ support the capabilities of GraphQL, we need to support creating a document with both queries and subscriptions at the same time. 

This could very well be a can to kick down the road, but it does make me think we'll need to create a more abstract function that handles all forms of queries/subscription/mutation documents. If we do this, I think the `execute`/`execute_sync`/`subscribe`/etc methods would need to become just convenience methods.

strawberry/types/federation/field.py
BryceBeagle: The reason each of the decorators takes both types was so that they could stack without worrying about order

strawberry/types/federation/object_type.py
BryceBeagle: but this is around a type, not a field?

strawberry/types/fields/field.py
BryceBeagle: I coudl see someone/something trying to access the name of the field, without first setting a resolver. That alone doesn't seem to me like it would warrant an exception

strawberry/types/fields/field.py
BryceBeagle: > the actual type class

I'm not sure I understand what you mean here?

strawberry/types/fields/field.py
BryceBeagle: Erroneous old todo, I think

strawberry/types/fields/mutation.py
BryceBeagle: I could see it _maybe_ making type annotation easier in the future

strawberry/types/fields/resolver.py
BryceBeagle: I use the resolver's name as the wrapping field's name if the field didn't get an explicitly set name

strawberry/schema/schema.py
patrick91: no :) extensions for us are have to add hooks to the schema, so for example to run code when the validation starts or when any resolver runs :)

strawberry/types/federation/field.py
patrick91: I was wondering if you could use this: https://docs.python.org/3/library/typing.html#newtype to declare the type, maybe it is not needed :)

strawberry/types/federation/object_type.py
patrick91: potentially! might be a clean approach :)

strawberry/types/federation/object_type.py
patrick91: right, but we also have both :)

`strawberry.federation.field` and `strawberry.federation.type`

strawberry/types/fields/field.py
patrick91: what would the usecase be?

strawberry/types/fields/field.py
patrick91: we use it here: https://github.com/strawberry-graphql/strawberry/blob/master/strawberry/types/type_resolver.py#L67

we get the origin class and then the module so we can get the correct types :)

we use it here too: https://github.com/strawberry-graphql/strawberry/blob/master/strawberry/federation.py#L119

PR #490, opened 2020-10-04
Fix extensions runner when using async resolvers

patrick91:
Had to duplicate the code because I needed to run the await inside `runner.request`. Also simplified the runner and removed the broken run_on_all_extension decorator


botberry:



---

This release fixes an issue with the extension runner and async resolvers

PR #491, opened 2020-10-05
Add support for union in mypy plugin

### LINKED ISSUE #485: Add mypy plugin for strawberry.union
We have a function to create union types and also add name and description that are used when generating the GraphQL schema[1].

Unfortunately MyPy complains when using the annotation, since it doesn't recognise it as a type, here's an example:

```python
ExampleUnion = strawberry.union(
    "ExampleUnion",
    types=(
        ABC,
        CDE,
    ),
)

@strawberry.mutation
def abc(phone_number: str) -> ExampleUnion:
    # ...
```

Error:

```
Variable "api.mutations.abc.ExampleUnion" is not valid as a type
```

[1] This is why we can't use typing.Union directly, as we can't pass arguments to it

patrick91:
This PR improves the mypy plugin to work with `strawberry.union`, this code should not raise any error:

```python
import strawberry


@strawberry.type
class User:
    name: str


@strawberry.type
class Error:
    message: str


Response = strawberry.union("Response", (User, Error))

a: Response
reveal_type(Response)


a = User('abc')
```

currently it does, but we are getting there :)

Closes #485

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,7 +11,7 @@ repos:
         additional_dependencies: ["flake8-eradicate==0.4.0"]
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.770
+    rev: v0.782
     hooks:
       - id: mypy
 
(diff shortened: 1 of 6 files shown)


botberry:

```markdown
Release type: patch

```

PR #495, opened 2020-10-05
Allow to use UNSET everywhere

patrick91:
This PR allows to use UNSET everywhere without mypy complaining :)


botberry:



---

This releases allows UNSET to be used anywhere and prevents mypy to report an error.

Issue #509, opened 2020-10-15
Feature request: Be able to add descriptions to field arguments


jkimbo:
Currently it isn't possible (I don't think) to add descriptions to field arguments. This would be a useful feature to help document a GraphQL schema.

Not sure what the api for this should look like but here is one example:

```python
@strawberry.type
class Query:
	@strawberry.field
	def user_by_id(id: str = strawberry.argument(description="The ID of the user")) -> User:
		...
```

Which would result in the following schema:

```graphql
type Query {
  userById(
    """The ID of the user"""
    id: String
  ): User!
}

**Comment by patrick91** [comment]
@jkimbo yes, I think that's the best API, fastapi does something similar too, so I think we go this way :)

**Comment by jkimbo** [comment]
After some discussion on Discord we're going use the new [`Annotated`](https://docs.python.org/3/library/typing.html#typing.Annotated) type to add extra metadata to an argument:

```python
from typing import Annotated

@strawberry.type
class Query:
	@strawberry.field
	def user_by_id(id: Annotated[str, strawberry.argument(description="The ID of the user")]) -> User:
		...
```

patrick91:
annotated is available in typing_extensions for python < 3.9, I wonder if we should expose it from strawberry too, but maybe not :) 

https://github.com/python/typing/blob/master/typing_extensions/src_py3/typing_extensions.py#L1781

jkimbo:
I don't think it's necessary to expose it directly from strawberry @patrick91 since typing_extensions is part of the dependencies anyway.

jkimbo:
#513 has been released

mikeldking:
Is there support for descriptions on static fields of an object type?

jkimbo:
@mikeldking I'm not sure I understand. Can you provide an example of what you're trying to achieve?

mikeldking:
@jkimbo I want to add descriptions to more places like such:

## Input type fields
```python
@strawberry.input
class TimeRange:
    start: Annotated[
        datetime.datetime,
        strawberry.argument(description="The start of the time range"),
    ]
    end: Annotated[
        datetime.datetime,
        strawberry.argument(description="The end of the time range. Right exclusive."),
    ]

    def is_valid(self) -> bool:
        return self.start < self.end
```
## type fields
```python
@strawberry.type
class Dataset:
    name: str
    start_time: Annotated[datetime, strawberry.field(description="The start of the time range")]
[...]

PR #513, opened 2020-10-18
Annotated arguments

jkimbo:
## Description

Add support for adding a description to field arguments using the [`Annotated`](https://docs.python.org/3/library/typing.html#typing.Annotated) type:

```python
from typing import Annotated

@strawberry.type
class Query:
	@strawberry.field
	def user_by_id(id: Annotated[str, strawberry.argument(description="The ID of the user")]) -> User:
		...
```

which results in the following schema:

```graphql
type Query {
  userById(
    """The ID of the user"""
    id: String
  ): User!
}
```

**Note:** if you are not using Python v3.9 or greater you will need to import `Annotated` from `typing_extensions`

**Another note:** pyflakes currently has a bug where it doesn't recognise the `Annotated` syntax

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* #509

## Checklist

### DIFF
--- a/strawberry/arguments.py
+++ b/strawberry/arguments.py
@@ -1,12 +1,24 @@
 import enum
 import inspect
-from typing import Any, Callable, Dict, List, Mapping, Type, cast
+from typing import Any, Callable, Dict, List, Mapping, Type, cast, Optional
+from typing_extensions import Annotated, get_origin, get_args
 
-from .exceptions import MissingArgumentsAnnotationsError, UnsupportedTypeError
+from .exceptions import (
+    MissingArgumentsAnnotationsError,
+    UnsupportedTypeError,
+    MultipleStrawberryArgumentsError,
+)
 from .scalars import is_scalar
 from .types.type_resolver import resolve_type
 from .types.types import ArgumentDefinition, undefined
 from .utils.str_converters import to_camel_case
+
+
+class StrawberryArgument:
+    description: Optional[str]
+
+    def __init__(self, description: Optional[str] = None):
+        self.description = description
 
 
 def get_arguments_from_annotations(
@@ -26,9 +38,27 @@ def get_arguments_from_annotations(
             origin_name=name,
             name=to_camel_case(name),
             origin=origin,
-            type=annotation,
             default_value=default_value,
         )
+
+        if get_origin(annotation) is Annotated:
+            argument_definition.type = get_args(annotation)[0]
+
+            argument_metadata = None
+            # Find any instances of StrawberryArgument in the Annotated metadata
+            # raising an exception if there are multiple StrawberryArguments
+            for arg in get_args(annotation)[1:]:
+                if isinstance(arg, StrawberryArgument):
+                    if argument_metadata is not None:
+                        raise MultipleStrawberryArgumentsError(
+                            field_name=origin.__name__, argument_name=name
+                        )
+                    argument_metadata = arg
+
+            if argument_metadata is not None:
+                argument_definition.description = argument_metadata.description
+        else:
+            argument_definition.type = annotation
 
         arguments.append(argument_definition)
 
@@ -111,8 +141,7 @@ def convert_argument(value: Any, argument_definition: ArgumentDefinition) -> Any
 
 
 def convert_arguments(
-    value: Dict[str, Any],
-    arguments: List[ArgumentDefinition],
+    value: Dict[str, Any], arguments: List[ArgumentDefinition],
 ) -> Dict[str, Any]:
     """Converts a nested dictionary to a dictionary of actual types.
 
@@ -132,3 +161,7 @@ def convert_arguments(
             kwargs[origin_name] = convert_argument(current_value, argument)
 
     return kwargs
+
+
+def argument(description: Optional[str] = None) -> StrawberryArgument:
+    return StrawberryArgument(description=description)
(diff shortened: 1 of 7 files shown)


botberry:

```markdown
Release type: patch

```

tests/types/test_arguments.py
```diff
@@ -200,3 +205,125 @@ class RootQuery(NameQuery, ExampleQuery):
...
+    assert argument.is_optional is True
+    assert argument.description == "This is a description"
+
+
+def test_annotated_argument_with_default_value():
+    @strawberry.type
+    class Query:
+        @strawberry.field
+        def name(  # type: ignore
+            argument: Annotated[
```
patrick91, line 267: should this have self here? 😊 (same for the other methods)

tests/schema/test_arguments.py
```diff
@@ -0,0 +1,28 @@
...
+
+    assert str(schema) == dedent(
+        '''\
+        type Query {
+          hello(
+            """Your name"""
+            name: String! = "Patrick"
+          ): String!
+        }'''
+    )
```
patrick91, line 28: nice :)

strawberry/schema/types/arguments.py
```diff
@@ -20,7 +20,11 @@ def convert_argument(
     # TODO: we could overload the function to tell mypy that it returns input types too
     argument_type = cast(GraphQLInputType, get_graphql_type(argument, type_map))
 
-    return GraphQLArgument(argument_type, default_value=default_value)
+    return GraphQLArgument(
+        argument_type,
+        default_value=default_value,
+        description=argument.description,
```
patrick91, line 26: 💯

tests/types/test_arguments.py
```diff
@@ -1,6 +1,11 @@
 from typing import List, Optional
 
+import pytest
+
+from typing_extensions import Annotated
```
patrick91, line 5: I wonder if we should do a test just for py3.9? maybe it's not worth doing it

strawberry/arguments.py
```diff
@@ -26,10 +39,28 @@ def get_arguments_from_annotations(
...
             default_value=default_value,
         )
 
+        if get_origin(annotation) is Annotated:
+            argument_definition.type = get_args(annotation)[0]
+
+            argument_metadata = None
+            # Find any instances of StrawberryArgument in the Annotated metadata
+            # raising an exception if there are multiple StrawberryArguments
+            for arg in get_args(annotation)[1:]:
```
patrick91, line 51: can we store `get_args(annotation)` in a variable?

patrick91: also I'm not sure I understand this whole check, sorry, I'm having a slow day :D

patrick91 (commented):
added some initial comments, nice work so far (and that was so quick too!)

tests/types/test_arguments.py
jkimbo: Probably a good idea, will add.

tests/types/test_arguments.py
jkimbo: Well it could but it's not necessary. TBH I find it quite misleading to have `self` on resolvers because it's not necessary an instance of the type and calling it `self` implies that it always will be. For example on the `Query` type `self` will always be `None` unless you pass a root value to `execute`:

```python
@strawberry.type
class Query:
	@strawberry.field
	def hello(self, name: str) -> str:
		assert self is None
		return f"Hi {name}"

schema.execute_sync(""" { hello } """)
```

strawberry/arguments.py
jkimbo: > can we store get_args(annotation) in a variable?

Yep can do (I was just being lazy because I couldn't think of a good name for it!)

> also I'm not sure I understand this whole check, sorry, I'm having a slow day :D

What part are you having problems understanding?

tests/types/test_arguments.py
patrick91: that's true, but I wonder if we should make it required, I feel like making self optional is quite confusing 😊

strawberry/arguments.py
```diff
@@ -26,10 +39,31 @@ def get_arguments_from_annotations(
...
+            argument_metadata = None
+            # Find any instances of StrawberryArgument in the other Annotated args,
+            # raising an exception if there are multiple StrawberryArguments
+            for arg in annotated_args[1:]:
+                if isinstance(arg, StrawberryArgument):
+                    if argument_metadata is not None:
+                        raise MultipleStrawberryArgumentsError(
+                            field_name=origin.__name__, argument_name=name
+                        )
+                    argument_metadata = arg
```
patrick91, line 60: this prevents from doing Annotated[str, argument(), argument()], right?

tests/types/test_arguments.py
```diff
@@ -200,3 +206,157 @@ class RootQuery(NameQuery, ExampleQuery):
...
+    @strawberry.type
+    class Query:
+        @strawberry.field
+        def name(  # type: ignore
+            argument: Annotated[
+                str,
+                strawberry.argument(description="This is a description"),  # noqa: F722
+            ]
+        ) -> str:
+            return "Name"
```
patrick91, line 349: nice, thanks for adding the test for py3.9 😊

patrick91 (approved):
Looks good to me!

strawberry/arguments.py
jkimbo: Yep that's right

PR #525, opened 2020-10-22
Add naive check for making sure that mypy doesn't crash in certain scenarios

patrick91:
Title, unfortunately I wasn't able to find a way to reproduce this, hopefully the check is safe. Will test on Pollen's codebase today :)


botberry:



---

This releases adds a workaround to prevent mypy from crashing in specific occasions

PR #527, opened 2020-10-23
Use field definition to check if we are returning a list or not

patrick91:
The previous check I've done would crash when returning string types that weren't real strings (like proxy strings in django).

This check should be much better :)


botberry:



---

This release improves how we handle enum values when returing lists of enums.

PR #528, opened 2020-10-23
Skip enum check when result is falsy

patrick91:
Improves how we check for enums and fixes a bug when returning None :)


botberry:



---

This release improves how we check for enums

PR #530, opened 2020-10-26
Add support for deprecation_reason on field

### LINKED ISSUE #375: Add support for field deprecation
I was looking in the code and I couldn't find a way to mark field as deprecated. I also didn't find any issues regarding this.
I know that it is handled by `deprecation_reason` in graphql's `GraphQLField` but how I mark a field as deprecated in strawberry-graphql?

marcottebear:
## Description

Added deprecation_reason to allow to mark a field as deprecated

New optional string field in `strawberry.field`

image

Example schema:

```
import typing
import strawberry


@strawberry.type
class User:
    name: str
    age: int = strawberry.field(deprecation_reason="ciao!")

@strawberry.type
class Query:
    @strawberry.field
    def user(self, info) -> User:


```

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* Closes #375

## Checklist

### DIFF
--- a/strawberry/schema/types/fields.py
+++ b/strawberry/schema/types/fields.py
@@ -40,6 +40,8 @@ def get_field(
     else:
         kwargs["args"] = convert_arguments(field.arguments, type_map)
         kwargs["resolve"] = resolver
+
+    if not is_input:
         kwargs["deprecation_reason"] = field.deprecation_reason
 
     return TypeClass(graphql_type, **kwargs)  # type: ignore
--- a/strawberry/schema/types/fields.py
+++ b/strawberry/schema/types/fields.py
@@ -40,5 +40,6 @@ def get_field(
     else:
         kwargs["args"] = convert_arguments(field.arguments, type_map)
         kwargs["resolve"] = resolver
+        kwargs["deprecation_reason"] = field.deprecation_reason
 
     return TypeClass(graphql_type, **kwargs)  # type: ignore
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: minor
+
+This release adds support to mark a field as deprecated via `deprecation_reason`
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -69,7 +69,8 @@ def field(
     is_subscription: bool = False,
     description: Optional[str] = None,
     permission_classes: Optional[List[Type[BasePermission]]] = None,
-    federation: Optional[FederationFieldParams] = None
+    federation: Optional[FederationFieldParams] = None,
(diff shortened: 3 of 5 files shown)


botberry:

```markdown
Release type: patch

```

BryceBeagle:
Could/should this be made syntactically closer to a directive by marking the deprecation using a decorator?

```python
@strawberry.type
class Query:
    @strawberry.deprecated(reason="Some problem")
    @strawberry.field
    def user(self, info) -> User:
```

patrick91:
@BryceBeagle that'd be interesting too, but I wonder how it would work for simple fields, we still have to decide on a approach here: https://github.com/strawberry-graphql/strawberry/pull/473

Maybe we can merge this for now, and change later? 🤔

marcoacierno:
I like the idea of having separate decorators but I feel it would be a bit inconsistent to have one just for deprecated? 🤔

BryceBeagle:
> we still have to decide on a approach here: #473

That discussion is about allowing _users_ to create their own decorators that affect fields' resolvers. This would be a first-party, strawberry-defined decorator that modifies the field itself, so I'd be ok with it using a different system. I'm also fine with accepting this for now and coming back to this again in the future.

> I feel it would be a bit inconsistent to have one just for deprecated

Gotta start somewhere :slightly_smiling_face: . I'm more worried about adding more and more arguments to `strawberry.field`, to the point that it's an overburdened function call.

patrick91:
> That discussion is about allowing users to create their own decorators that affect fields' resolvers. 

Yup, but we didn't find a good solution for this, right?

```python
@strawberry.type
class Query:
	user: User = strawberry.deprecated(strawberry.field(), reason="ABC")
```

vs

```python
@strawberry.type
class Query:
	user: User = strawberry.field(resolver=strawberry.deprecated(strawberry.default_resolver, reason="ABC"))
```

vs

```python
@strawberry.type
class Query:
	user: User = strawberry.field(decorators=strawberry.deprecated(reason="ABC"))
```

BryceBeagle:
Yeah I don't think we've agreed on a solution to that yet

jkimbo (approved):
This looks good to me 👍

strawberry/schema/types/fields.py
```diff
@@ -40,5 +40,6 @@ def get_field(
     else:
         kwargs["args"] = convert_arguments(field.arguments, type_map)
         kwargs["resolve"] = resolver
+        kwargs["deprecation_reason"] = field.deprecation_reason
```
patrick91, line 43: let's check if we can do this on subscriptions as well :)

PR #533, opened 2020-10-26
Bugfix: Async lists

jkimbo:
## Description

The changes introduced to enable resolvers to return a list of enums has broken support for returning lists from async resolvers. Specifically this line: https://github.com/strawberry-graphql/strawberry/blob/0ed51802b9ce3e7abe0282953384e8155a575cfa/strawberry/resolvers.py#L28 
fails with the error `'coroutine' object is not iterable`. I've added some regression tests to show the issue but I'm not sure how to solve it. Any ideas @patrick91 @BryceBeagle ?

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/strawberry/resolvers.py
+++ b/strawberry/resolvers.py
@@ -1,6 +1,6 @@
 import enum
 from inspect import iscoroutine, iscoroutinefunction
-from typing import Any, Callable, Dict, List, Tuple, cast
+from typing import Any, Awaitable, Callable, Dict, List, Tuple, Union, cast
 
 from .arguments import convert_arguments
 from .field import FieldDefinition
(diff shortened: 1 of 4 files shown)


patrick91:
@jkimbo I knew I had forgotten about something when doing that fix 🤦‍♂️

I think we can check if the resolver is async, if so we provide a different resolver, basically an async version of this: https://github.com/strawberry-graphql/strawberry/blob/0ed51802b9ce3e7abe0282953384e8155a575cfa/strawberry/resolvers.py#L46

what do you think?

patrick91:
@jkimbo I did a fix, not sure it is the best way though 😊

strawberry/resolvers.py
```diff
@@ -30,6 +31,49 @@ def convert_enums_to_values(field: FieldDefinition, result: Any) -> Any:
...
+    # by inspecting the original resolver arguments,
+    # if it asks for self, the source will be passed as first argument
+    # if it asks for root, the source it will be passed as kwarg
+    # if it asks for info, the info will be passed as kwarg
+
+    function_args = get_func_args(actual_resolver)
+
+    args = []
+
+    if "self" in function_args:
```
BryceBeagle, line 51: Unlikely edge-case, but should this only apply if `self` is the first arg?

jkimbo (commented):
Looks good @patrick91 👍 thanks!

botberry:



---

This release fixes a bug when returning list in async resolvers

strawberry/resolvers.py
patrick91: fixed 😊

patrick91 (approved):

PR #538, opened 2020-10-28
Fix issue when using namespaced variables for enums

patrick91:


botberry:



---

Fix issue when using `strawberry.enum(module.EnumClass)` in mypy

marcoacierno (approved):
Thanks a lot 🙏

PR #543, opened 2020-10-29
StrawberryResolver object

BryceBeagle:
## Description

Wrap resolvers in a custom class. This is the first step towards implementing #489.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,7 @@ typing_extensions = "^3.7.4"
 opentelemetry-api = {version = "^0.13b0",optional = true}
 opentelemetry-sdk = {version = "^0.13b0",optional = true}
 python-dateutil = "^2.7.0"
+cached-property = "^1.5.2"
 
 [tool.poetry.dev-dependencies]
 pytest = "^6.1"
--- a/strawberry/arguments.py
+++ b/strawberry/arguments.py
@@ -26,9 +26,7 @@ def get_arguments_from_annotations(
     for name, annotation in annotations.items():
         default_value = parameters[name].default
         default_value = (
-            undefined
-            if default_value == inspect._empty  # type: ignore
-            else default_value
+            undefined if default_value is inspect.Parameter.empty else default_value
         )
 
         argument_definition = ArgumentDefinition(
--- a/strawberry/arguments.py
+++ b/strawberry/arguments.py
@@ -1,14 +1,10 @@
 import enum
 import inspect
-from typing import Any, Callable, Dict, List, Mapping, Optional, Type, cast
+from typing import Any, Dict, List, Mapping, Optional, Type, cast
 
 from typing_extensions import Annotated, get_args, get_origin
 
-from .exceptions import (
-    MissingArgumentsAnnotationsError,
-    MultipleStrawberryArgumentsError,
-    UnsupportedTypeError,
-)
+from .exceptions import MultipleStrawberryArgumentsError, UnsupportedTypeError
 from .scalars import is_scalar
 from .types.type_resolver import resolve_type
 from .types.types import ArgumentDefinition, undefined
@@ -69,28 +65,6 @@ def get_arguments_from_annotations(
         resolve_type(argument_definition)
 
     return arguments
-
-
-def get_arguments_from_resolver(resolver: Callable) -> List[ArgumentDefinition]:
-    annotations = resolver.__annotations__
-    parameters = inspect.signature(resolver).parameters
-    function_arguments = set(parameters) - {"root", "self", "info"}
-
-    annotations = {
-        name: annotation
-        for name, annotation in annotations.items()
-        if name not in ["root", "info", "return", "self"]
-    }
-
-    annotated_function_arguments = set(annotations.keys())
-    arguments_missing_annotations = function_arguments - annotated_function_arguments
-
-    if len(arguments_missing_annotations) > 0:
-        raise MissingArgumentsAnnotationsError(
-            resolver.__name__, arguments_missing_annotations
-        )
-
-    return get_arguments_from_annotations(annotations, parameters, origin=resolver)
 
 
 class _Unset:
@@ -180,7 +154,6 @@ __all__ = [
     "convert_argument",
     "convert_arguments",
     "get_arguments_from_annotations",
-    "get_arguments_from_resolver",
     "is_unset",
     "undefined",
 ]
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -27,7 +27,9 @@ class StrawberryField(dataclasses.Field):
         )
 
     def __call__(self, resolver: _RESOLVER_TYPE) -> "StrawberryField":
-        """Migrate the field definition to the resolver"""
+        """Add a resolver to the field"""
+
+        # Allow for StrawberryResolvers or bare functions to be provided
         if not isinstance(resolver, StrawberryResolver):
             resolver = StrawberryResolver(resolver)
 
@@ -37,7 +39,7 @@ class StrawberryField(dataclasses.Field):
         self._field_definition.arguments = resolver.arguments
         self._field_definition.type = resolver.type
 
-        # Don't add field to __init__
+        # Don't add field to __init__ or __repr__
         self.init = False
         self.repr = False
 
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -84,11 +84,9 @@ def field(
 
     field_definition = FieldDefinition(
         origin_name=None,  # modified by resolver in __call__
-        name=name,  # modified by resolver in __call__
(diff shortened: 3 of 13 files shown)


patrick91:
We need to double check why this codepaths aren't tested anymore: https://codecov.io/gh/strawberry-graphql/strawberry/pull/543/src?path=types%2Ftypes.py#L311

tests/mypy/test_fields.yml
```diff
@@ -44,8 +44,8 @@
     main:22: note: Revealed type is 'builtins.str'
     main:23: note: Revealed type is 'builtins.str'
     main:24: note: Revealed type is 'builtins.str'
-    main:25: note: Revealed type is 'def (self: main.Example, info: Any) -> builtins.str'
-    main:26: note: Revealed type is 'def (self: main.Example, info: Any) -> builtins.str'
+    main:25: note: Revealed type is 'strawberry.field.StrawberryField'
```
BryceBeagle, line 47: Should _all_ the fields have revealed types as `StrawberryField`? I anticipate being able to make StrawberryField generic so I think it could be
```
main:21: note: Revealed type is 'strawberry.field.StrawberryField[str]'
```
for each.

strawberry/field.py
```diff
@@ -112,11 +84,9 @@ def field(
 
     field_definition = FieldDefinition(
         origin_name=None,  # modified by resolver in __call__
-        name=name,  # modified by resolver in __call__
-        type=None,  # type: ignore
-        origin=resolver,  # type: ignore
```
BryceBeagle, line 125: We were leaving `origin` as `None` unless a resolver was provided. Was this correct? Would a resolver-less field never have an origin?

tests/mypy/test_fields.yml
patrick91: that's a good question, that's only for the class? The instance types would still be 'builtins.str', right?

strawberry/field.py
patrick91: we use origin here: https://github.com/strawberry-graphql/strawberry/blob/23ee3adce2bf062fb76080030c74d6218bcf4cf5/strawberry/types/type_resolver.py#L67 

a resolver-less field should have origin equal to the class where it was defined (might not be the case now) :)

BryceBeagle: It ends up being set [here](https://github.com/strawberry-graphql/strawberry/blob/6ebbc76da0bf213256ad5695e1a370e00f011abf/strawberry/types/type_resolver.py#L366). I think my changes make sense*, as previously the constructor to `FieldDefinition` was _always_ being set to `None` or `resolver`, and then re-set to `resolver` in `__call__`. Now I don't set it in the initial constructor call and just defer it to the two other places.

*for now; at least until `FieldDefinition` goes by the wayside with the future overhaul of `StrawberryField`

strawberry/arguments.py
```diff
@@ -30,9 +26,7 @@ def get_arguments_from_annotations(
     for name, annotation in annotations.items():
         default_value = parameters[name].default
         default_value = (
-            undefined
-            if default_value == inspect._empty  # type: ignore
-            else default_value
+            undefined if default_value is inspect.Parameter.empty else default_value
```
BryceBeagle, line 29: This was using an inspect internal, but they provide a public access point to it, so I used it instead

strawberry/types/fields/resolver.py
```diff
@@ -0,0 +1,85 @@
...
+    def __call__(self, *args, **kwargs) -> T:
+        return self.wrapped_func(*args, **kwargs)
+
+    # TODO: Return StrawberryArguments instead. Maybe this should be a classmethod
+    #       there instead?
+    # TODO: Return a set instead
+    @cached_property
+    def arguments(self) -> List[ArgumentDefinition]:
+        # TODO: Move to StrawberryArgument? StrawberryResolver ClassVar?
+        SPECIAL_ARGS = {"root", "self", "info"}
```
BryceBeagle, line 31: I figured I'd turn this into a constant, but I wasn't quite sure where to put it. Not sure if it would be better within `StrawberryResolver` or the future `StrawberryArgument`

strawberry/types/type_resolver.py
```diff
@@ -230,55 +228,35 @@ def _resolve_types(fields: List[FieldDefinition]) -> List[FieldDefinition]:
 def _get_fields(cls: Type) -> List[FieldDefinition]:
```
BryceBeagle, line 228: This function is much much simpler now after these changes :tada:

strawberry/types/type_resolver.py
```diff
@@ -385,7 +313,8 @@ def _get_fields(cls: Type) -> List[FieldDefinition]:
                 default_value=getattr(cls, field.name, undefined),
             )
 
-        field_name = cast(str, field_definition.origin_name)
+        field_name = field_definition.origin_name
+        assert field_name is not None  # Not Optional by this point
         field_definitions[field_name] = field_definition
```
BryceBeagle, line 318: I think an assert is better than a cast, because if `field_name` really _is_ `None`, then we probably want to know. Mypy also knows this makes the variable know longer `Optional` so I see this as a win-win.

botberry:

```markdown
Release type: patch

```

strawberry/types/type_resolver.py
```diff
@@ -296,64 +274,19 @@ def _get_fields(cls: Type) -> List[FieldDefinition]:
             field_definitions = {**field_definitions, **base_field_definitions}
 
     # then we can proceed with finding the fields for the current class
+    for field in dataclasses.fields(cls):
+        # TODO: Importing at top of file cause cyclical import problems. Maybe there's
+        #       a better place for this?
+        from strawberry.field import StrawberryField
```
patrick91, line 280: we can do this before the for?

strawberry/types/type_resolver.py
```diff
@@ -385,7 +313,8 @@ def _get_fields(cls: Type) -> List[FieldDefinition]:
                 default_value=getattr(cls, field.name, undefined),
             )
 
-        field_name = cast(str, field_definition.origin_name)
+        field_name = field_definition.origin_name
+        assert field_name is not None  # Not Optional by this point
```
patrick91, line 317: can we add a comment for the assertion here?

```
assert field_name is not None, "This happens when..."
```

patrick91 (approved):
This look brilliant! I'm going to test this on our code base just to make sure all works fine :)

strawberry/field.py
```diff
@@ -27,29 +26,24 @@ def __init__(self, field_definition: FieldDefinition):
...
-        if not inspect.ismethod(resolver):
-            # resolver is a normal function
-            resolver._field_definition = field_definition  # type: ignore
-        else:
-            # resolver is a bound method and immutable (most likely a
-            # classmethod or an instance method). We need to monkeypatch its
-            # underlying .__func__ function
-            # https://stackoverflow.com/a/7891681/8134178
-            resolver.__func__._field_definition = field_definition  # type:ignore
+        # Don't add field to __init__ or __repr__
```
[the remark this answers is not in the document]
BryceBeagle: > this is only when fields have a decorator, right

More specifically, it's only when fields have a resolver, using either technique (parameter/decorator) of adding a resolver

strawberry/types/type_resolver.py
BryceBeagle: Oh oops. I meant to do that. I was trying to keep the import close to where it's used, but accidentally put it _too_ close.

strawberry/types/type_resolver.py
BryceBeagle: I have an in-line one. Would you like it on its own line, or are you wanting a more detailed comment?

patrick91: A more detailed comment could be better, but I was mainly referring to add a message to the assertion (I used the wrong wording 😊), something like this:

image

BryceBeagle: Ah ok, sure

pyproject.toml
```diff
@@ -33,6 +33,7 @@ typing_extensions = "^3.7.4"
 opentelemetry-api = {version = "^0.13b0",optional = true}
 opentelemetry-sdk = {version = "^0.13b0",optional = true}
 python-dateutil = "^2.7.0"
+cached-property = "^1.5.2"
```
scratchmex, line 36: is there any reason to use this library instead of `functools.cached_property`?

BryceBeagle: I'd love to use that, but we still support python 3.7. the library is basically a backport.

scratchmex: Oh, I missed that. Thanks

PR #546, opened 2020-10-29
Disable implicit reexport

jordangarside:
This disables [implicit re-export](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-reexport) of modules. This is one of the stricter rules in MyPy and if someone is using this project and that flag then they will not be able to import modules off of the top level modules.

```
import strawberry

@strawberry.field
...
```

Would throw an implicit reexport error for a user.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+This release disable implicit re-export of modules. This fixes Strawberry for you if you were using `implicit_reexport = False` in your MyPy config.
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,2 +1,3 @@
 [mypy]
 plugins = ./strawberry/ext/mypy_plugin.py
+implicit_reexport = False
--- a/strawberry/__init__.py
+++ b/strawberry/__init__.py
@@ -1,17 +1,39 @@
 __version__ = "0.1.0"
 
 
-from . import federation  # noqa
-from .arguments import argument  # noqa
-from .custom_scalar import scalar  # noqa
-from .directive import directive  # noqa
-from .enum import enum  # noqa
-from .field import field  # noqa
-from .lazy_type import LazyType  # noqa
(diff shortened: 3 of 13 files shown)


botberry:

```markdown
Release type: patch

```

BryceBeagle:
Would it be better to use `__all__`?

patrick91:
Looks like using `__all__` might be a cleaner solution:

https://mypy.readthedocs.io/en/stable/config_file.html#confval-implicit_reexport

```
# This won't re-export the value
from foo import bar
# This will re-export it as bar and allow other modules to import it
from foo import bar as bar
# This will also re-export bar
from foo import bar
__all__ = ['bar']
```

jordangarside:
Yeah I'm indifferent. The reason I didn't use `__all__` is because we have some `__init__` files that also have full classes in them. It's a bit more verbose to also add these to `__all__` instead of just explicitly re-export.

I'll update it.

patrick91:
@jordangarside I think I'd prefer to use `__all__`, seems a lot nicer, since we are reexporting a lot of things.

I wasn't able to find a linting rule for it, do you know if flake8 has it?

jordangarside:
There isn't an off-the-shelf one. It also turns out we don't only need this in `__init__` files since we also re-export from other files. I'm just going with this for now.

Just something people will have to be aware of as a potential source of confusion/issues (different exporting depending on if `__all__` is in the file or not).

jordangarside:
Updated the PR with `__all__` syntax.

patrick91 (approved):
Nice one! I honestly didn't about this flag in mypy, might try at work 😊

I'm going to push a commit to remove the __version__ as we don't really update that :)

Issue #542, opened 2020-10-29
Make camelCasing optional


jordangarside:
Would be nice if we could make camelCasing optional:

`schema = strawberry.Schema(query=Query, auto_camelcase=False)`

https://github.com/strawberry-graphql/strawberry/issues/36

peniqliotuv:
@patrick91 is there an update on this :O

patrick91:
> @patrick91 is there an update on this :O

sorry I've been quite busy in the last months, I'll make sure this happens over next week 😊

zerocewl:
Any news?

patrick91:
@zerocewl I'm aiming to finish it this week :) I made some changes to make this easier in #1044 so hopefully I won't find any blockers :)

patrick91:
Sorry this took more than expected, we want to release a big refactor before doing it! Now it is done and should be released soon!

Thanks for waiting!

PR #548, opened 2020-10-30
Make to_camel_case and capitalize_first configurable

jordangarside:
Adds `auto_camelcase` and `auto_capitalize` bools inside of `str_converters` that can be modified to stop camelCasing or Capitalizing.

```py
from strawberry.utils import str_converters

str_converters.auto_camelcase = False
```

I wasn't sure of a better way to do this, as these utilities are used in various functions that can be run at any time before a schema is generated.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* https://github.com/strawberry-graphql/strawberry/issues/542

## Checklist

There wasn't any test on these methods already and I don't think this functionality is very complicated at the moment.
Also not sure about adding to the docs since it might not be the long-term API that is wanted here.

### DIFF
--- a/strawberry/utils/str_converters.py
+++ b/strawberry/utils/str_converters.py
@@ -1,6 +1,12 @@
+auto_camel_case = True
+auto_capitalize = True
+
+
 # Adapted from this response in Stackoverflow
 # http://stackoverflow.com/a/19053800/1072990
 def to_camel_case(snake_str: str) -> str:
+    if auto_camel_case is False:
+        return snake_str
     components = snake_str.split("_")
     # We capitalize the first letter of each component except the first one
     # with the 'capitalize' method and join them together.
@@ -8,4 +14,6 @@ def to_camel_case(snake_str: str) -> str:
 
 
 def capitalize_first(name: str) -> str:
+    if auto_capitalize is False:
+        return name
     return name[0].upper() + name[1:]
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,10 @@
+Release type: patch
+
+This release adds the ability to disable to_camel_case and capitalize_first:
+
+```py
+from strawberry.utils import str_converters
+
+str_converters.auto_camelcase = False
+str_converters.auto_capitalize = False
+```


botberry:

```markdown
Release type: patch

```

strawberry/utils/str_converters.py
```diff
@@ -1,11 +1,19 @@
+auto_camelcase = True
```
BryceBeagle, line 1: Nitpick: We're being inconsistent about whether whether it's `camel_case` or `camelcase`

BryceBeagle (changes requested):

BryceBeagle:
Are global switches the best option? I think in the ideal situation, this configuration should be stored in the (each) `Schema` object, but that may be more work than its worth.

patrick91:
> Are global switches the best option? I think in the ideal situation, this configuration should be stored in the (each) Schema object, but that may be more work than its worth.

I'd prefer this too if possible, might be a bit of pain to implement though 🤔

strawberry/utils/str_converters.py
jordangarside: Yeah I was debating that here. Graphene had the inconsistency so I was keeping it the same as that. I'll update it.

AndrewIngram:
> > Are global switches the best option? I think in the ideal situation, this configuration should be stored in the (each) Schema object, but that may be more work than its worth.
> 
> I'd prefer this too if possible, might be a bit of pain to implement though 🤔

@patrick91 What in particular makes it a pain? Is it the problem of essentially having to pass around the schema/config to all the locations that may need it?

patrick91:
> @patrick91 What in particular makes it a pain? Is it the problem of essentially having to pass around the schema/config to all the locations that may need it?

@AndrewIngram yes :) also it kinda prevents to convert a type without having a schema (ie if you want to see how a type looks like in the SDL)

AndrewIngram:
Another use case for ContextVars? https://docs.python.org/3/library/contextvars.html

jordangarside:
The issue here is that you need to set `auto_camel_case` before you use any of the `field` or `type` functions in strawberry. So adding the option to `strawberry.Schema` won't actually affect the behavior (even if you use a contextVar there), because the name has already been camelCased before you even initialize the schema.

i.e. In the way strawberry is currently setup, you have to modify _some_ global value before you run any strawberry methods to make a feature like this work.

PR #550, opened 2020-10-30
Use schema to determine if we should camel case field names

patrick91:
This PR is an alternative approach to https://github.com/strawberry-graphql/strawberry/pull/548

I prefer this way as we can have multiple schemas with different settings.

Be mindful that this is stil a WIP, also it does not address auto_capitalize, yet :)


jordangarside:
Yeah this looks reasonable to me. 👍

patrick91:
closing this as we'll be doing it in #798

PR #551, opened 2020-10-31
Make field typing more specific

jordangarside:
Improves `field` typing using overloading and `StrawberryField` typing using generics.

## Description

`field` could not be used as a decorator in strict codebases as there was no way for MyPy to know whether or not it was going to return a decorator or `Any`.

https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-disallow-untyped-decorators

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -24,6 +24,7 @@ exclude_lines =
     if 0:
     if __name__ == .__main__.:
 
+    @overload
 
 
 omit =
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,4 +56,4 @@ jobs:
       - run: pip install poetry
       - name: Install dependencies
         run: poetry install
-      - run: poetry run mypy --ignore-missing-imports .
+      - run: poetry run mypy --ignore-missing-imports --config-file mypy.ini .
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+Improve typing for `field` and `StrawberryField`.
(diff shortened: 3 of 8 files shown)


botberry:



---

Improve typing for `field` and `StrawberryField`.

patrick91:
@jaydenwindle this looks good, but I want to test it before merging, just want to make sure we cover all the cases :)

what mypy config do you use?

jordangarside:
MyPy Config
```ini
[mypy]
plugins = ./strawberry/ext/mypy_plugin.py
implicit_reexport = False
disallow_untyped_decorators = True
```

Example
```py
import strawberry


def some_resolver() -> str:
    return ""


@strawberry.type
class Example:
    a: str
    b: str = strawberry.field(name="b")
    c: str = strawberry.field(name="c", resolver=some_resolver)
    d: str = strawberry.field(resolver=some_resolver)

    @strawberry.field(description="ABC")
    def e(self, info) -> str:
        return ""

    @strawberry.field(name="f")
    def f_resolver(self, info) -> str:
        return ""
```

^ this will should error on the decorators without the changes from the PR.

We can't turn `disallow_untyped_decorators` for the entire codebase because there are a few other decorators that would also need to be addressed.

patrick91:
@jordangarside I pushed some tests, but mypy is currently failing for some reason 🤷‍♂️  I'll try to understand why we get the error we are actually trying to fix:

```
error: Untyped decorator makes function "e" untyped
```

patrick91 (approved):
Nice one, I added a test, unfortunately for now we can enable disallow_untyped_decorators on our codebase (unless I split the mypy config for the tests) 😊

Just wanted to add a note, we will probably change the type of `@strawberry.field` in future, but that shouldn't affect existing code :)

Issue #553, opened 2020-11-01
[BUG] Can't create interfaces that implement other interfaces


Ambro17:
After RFC https://github.com/graphql/graphql-spec/pull/373 was merged, not its possible in graphql-js and in the python port to create interfaces that implement interfaces.
The classic example are relay `Connection` Types, but a more simple example is listed below:
```graphql
interface GenericError {
    message: String!
}

interface InternalError implements GenericError {
    message: String!
    code: Int!
}

type UnexpectedError implements InternalError {
    message: String!
    code: Int!
}

interface UsageError implements GenericError {
    message: String!
    suggestion: String!
}
```

I also have a PR with an attempt to fix this

PR #554, opened 2020-11-01
Allow interfaces to implement interfaces #553

Ambro17:
## Description
After [0c85996](https://github.com/graphql-python/graphql-core/commit/0c85996115738ab92d6d2be8a7245635446a3c59) it's possible in graphql for interfaces to implement interfaces. Before only object types could.

Before adding tests i wanted to validate if this is the correct approach or there's a better approach.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

#553 

## Checklist

## Minimal Example to test the fix
Write this into a `filename.py`
```python
import random
import strawberry


@strawberry.type
class Query:

    @strawberry.field
    def hello() -> str:
        return 'World'


@strawberry.interface
class ErrorInterface:
    message: str


@strawberry.interface
class FirstErrorInterface(ErrorInterface):
    code: int = 0


@strawberry.type
class ValueError(FirstErrorInterface):
    message: str


@strawberry.type
class KeyError(FirstErrorInterface):
    message: str


@strawberry.type
class Mutation:

    @strawberry.mutation
    def first(self) -> FirstErrorInterface:
        klass = random.choice([ValueError, KeyError])
        return klass('Wrong!')


schema = strawberry.Schema(Query, Mutation, types=[ValueError, KeyError])
print(str(schema))
```
Without this changes, it should break with
```
[...]
  File "strawberry\schema\types\object_type.py", line 39, in _get_object_type_for_type_definition
    object_type = TypeClass(
TypeError: __init__() got an unexpected keyword argument 'is_type_of'
```

With this changes it should print the schema as follows:
```graphql
interface ErrorInterface {
  message: String!
}

interface FirstErrorInterface implements ErrorInterface {
  message: String!
  code: Int!
}

type KeyError implements FirstErrorInterface {
  message: String!
  code: Int!
}

type Mutation {
  first: FirstErrorInterface!
}

type Query {
  hello: String!
}

type ValueError implements FirstErrorInterface {
  message: String!
  code: Int!
}
```

### DIFF
--- a/strawberry/schema/types/object_type.py
+++ b/strawberry/schema/types/object_type.py
@@ -14,7 +14,7 @@ def _get_object_type_for_type_definition(
 ) -> GraphQLType:
 
     if type_definition.name not in type_map:
-        TypeClass: Type = GraphQLObjectType
+        TypeClass: Type = GraphQLInterfaceType
 
         kwargs = {}
 
@@ -28,10 +28,17 @@ def _get_object_type_for_type_definition(
                 _get_object_type_for_type_definition(interface, type_map)
                 for interface in type_definition.interfaces
             ]
-            # this tells GraphQL core what the returned object's actual type is
-            kwargs["is_type_of"] = lambda obj, _: isinstance(  # type: ignore
-                obj, type_definition.origin
-            )
+
+            # After https://github.com/graphql/graphql-spec/pull/373/files
+            # Now interfaces can also implement interfaces.
+            # But `GraphQLInterfaceType` constructor doesn't accept
+            # `is_type_of` argument.
+            # So we check first, and only add it for `GraphQLInterfaceType`
+            if TypeClass == GraphQLObjectType:
+                # this tells GraphQL core what the returned object's actual type is
+                kwargs["is_type_of"] = lambda obj, _: isinstance(  # type: ignore
+                    obj, type_definition.origin
+                )
 
         assert not type_definition.is_generic
 
--- a/strawberry/schema/types/object_type.py
+++ b/strawberry/schema/types/object_type.py
@@ -33,8 +33,8 @@ def _get_object_type_for_type_definition(
             # Now interfaces can also implement interfaces.
             # But `GraphQLInterfaceType` constructor doesn't accept
             # `is_type_of` argument.
-            # So we check first, and only add it for `GraphQLInterfaceType`
-            if TypeClass == GraphQLObjectType:
+            # So we check first, and we avoid adding it for `GraphQLInterfaceType`s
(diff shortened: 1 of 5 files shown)


botberry:

```markdown
Release type: patch

```

strawberry/schema/types/object_type.py
```diff
@@ -28,10 +28,17 @@ def _get_object_type_for_type_definition(
...
-            kwargs["is_type_of"] = lambda obj, _: isinstance(  # type: ignore
-                obj, type_definition.origin
-            )
+
+            # After https://github.com/graphql/graphql-spec/pull/373/files
+            # Now interfaces can also implement interfaces.
+            # But `GraphQLInterfaceType` constructor doesn't accept
+            # `is_type_of` argument.
+            # So we check first, and only add it for `GraphQLInterfaceType`
+            if TypeClass == GraphQLObjectType:
```
patrick91, line 37: this looks good, but I'd change the check to be `if type_definition.is_interface` :)

Ambro17: You mean something like `if not type_definition.is_interface`?

patrick91: yes, sorry 😊

Ambro17: Np, just wanted to check. I'll push the changes

Ambro17:
I have two questions that i left as comments in the code.
1. Tests only pass if i inherit explicitly from both interfaces
2. Type that inherits from second interface, only has one interface implemented. Is that the reason of the first behaviour?

patrick91:
> I have two questions that i left as comments in the code.
> 
>     1. Tests only pass if i inherit explicitly from both interfaces

This is a bug yes, I think the code I wrote here is wrong: https://github.com/strawberry-graphql/strawberry/blob/master/strawberry/type.py#L13-L24

this should also check the bases' bases I think :)

 
>     2. Type that inherits from second interface, only has one interface implemented. Is that the reason of the first behaviour?
I think this is another bug 😊

Ambro17:
Do you want to tackle those 2 in this PR?

patrick91:
> Do you want to tackle those 2 in this PR? 

If it's not a big hassle yes 😊

Ambro17:
It seems to work now, tests pass after a small change

patrick91:
> It seems to work now, tests pass after a small change

@Ambro17 awesome! Will check this today 😊

patrick91 (approved):
Nice work, and well test 👏  thank you!

PR #555, opened 2020-11-02
Add implementation of dataloaders

### LINKED ISSUE #413: N + 1 problem - Support for dataloaders?
Hi 👋

How to solve N+1 problem when using strawberry?

patrick91:
Pretty much based on https://github.com/graphql/dataloader

I did this mainly so that we can add new features and do bug fixes quickly and also to make sure we actually support dataloders properly in our code :)

# TODO: 

- [x] Support for max batch items
- [x] Error checking 
- [x] Basic caching
- [x] Disabling caching
- [x] Check exceptions inside resolvers and loader

# TODO: in future

- [ ] Custom caching
- [ ] Priming/Clearing cache
- [ ] Batch schedule function

Closes #413

### DIFF
--- a/docs/03_features/dataloaders.md
+++ b/docs/03_features/dataloaders.md
@@ -3,7 +3,7 @@ title: DataLoaders
 path: /docs/features/dataloaders
 ---
 
-# DataLoader
+# DataLoaders
 
 Strawberry comes with a built-in DataLoader, a generic utility that can be used
 to reduce the number of requests to databases or third party APIs by batching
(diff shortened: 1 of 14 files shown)


strawberry/dataloader.py
```diff
@@ -0,0 +1,80 @@
...
+        future = self.loop.create_future()
+
+        batch = get_current_batch(self)
+        batch.add_task(key, future)
+
+        return future
+
+
+def should_create_new_batch(batch: Batch) -> bool:
+    return False
```
patrick91, line 46: This should return true if the batch is being scheduled or if we reached the max items per batch

docs/03_features/dataloaders.md
```diff
@@ -0,0 +1,127 @@
...
+
+loader = DataLoader(load_fn=load_user)
+
+@strawberry.type
+class Query:
+    @strawberry.field
+    async def get_user(self, id: strawberry.ID) -> User:
+        return await loader.load(id)
+
+schema = strawberry.Schema(query=Query)
```
AndrewIngram, line 96: It's common to see loaders created (or at least instantiated) as part of building the context object, because you often want to reset the cache between queries. It might be worth documenting this pattern too

strawberry/dataloader.py
```diff
@@ -0,0 +1,125 @@
...
+    def __init__(
+        self,
+        load_fn: Callable,
+        max_batch_size: Optional[int] = None,
+        cache: bool = True,
+    ):
+        self.load_fn = load_fn
+        self.max_batch_size = max_batch_size
+
+        self.loop = get_event_loop()
```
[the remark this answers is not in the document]
patrick91: yes!

botberry:

```markdown
Release type: patch

```

strawberry/dataloader.py
```diff
@@ -0,0 +1,125 @@
...
+
+class DataLoader(Generic[T, K]):
+    queue: List[LoaderTask] = []
+    batch: Optional[Batch] = None
+    cache: bool = False
+    cache_map: Dict[K, Future]
+
+    def __init__(
+        self,
+        load_fn: Callable,
```
AndrewIngram, line 41: is it possible the type of load_fn more specific? ie it's a callable that takes an array of keys `K` and returns an array of optional results `R`, then you could type the cache and the return value of `load`.

AndrewIngram: Actually, we already have K, but we're not using T from the generic above.

patrick91: Done! Added some tests too 😊

patrick91:
> Might be worth documenting how to use the DataLoader in the Django section as well?

@jkimbo 100%, I'm doing a small POC, I think I will add docs later though, when I'm sure I know most of the gotchas. 
Also I think we might need to start having subpages in the docs 😊

jkimbo (approved):

PR #556, opened 2020-11-02
Add django 3.1 async view

patrick91:
This PR adds a new view for django that is Async. It works the same as the old one, but most of the methods are async now :)

I was going to add a new test, like this:

```python
@pytest.mark.asyncio
async def test_api_with_async_client():
    query = """{ exampleAsync }"""

    client = AsyncClient()

    response = await client.post(
        reverse("graphql_async"),
        data={"query": query},
        content_type="application/json",
    )

    assert response.status_code == 200

    data = response.json()
    assert data == ...
```

but this seems to be broken in current Django, will see if I can find some workaround :)

### DIFF
--- a/docs/05_integrations/django.md
+++ b/docs/05_integrations/django.md
@@ -114,3 +114,114 @@ class MyGraphQLView(GraphQLView):
 
 In this case we are doing the default processing of the result, but it can
 be tweaked based on your needs.
+
+# Async Django
+
+Strawberry also provides an async view that you can use with Django 3.1+
+
+```python
+from django.urls import path
+
+from strawberry.django.views import AsyncGraphQLView
+
+from api.schema import schema
+
+urlpatterns = [
+    path("graphql/", AsyncGraphQLView.as_view(schema=schema)),
+]
+```
+
+You'd also need to add `strawberry.django` to the `INSTALLED_APPS` of your
+project, this is needed to provide the template for the GraphiQL interface.
+
+## Options
+
+The `AsyncGraphQLView` accepts two options at the moment:
+
+- schema: mandatory, the schema created by `strawberry.Schema`.
+- graphiql: optional, defaults to `True`, whether to enable the GraphiQL
+  interface.
+
+## Extending the view
+
+We allow to extend the base `AsyncGraphQLView`, by overriding the following methods:
+
+- `async get_context(self, request: HttpRequest) -> Any`
+- `async get_root_value(self, request: HttpRequest) -> Any`
+- `async process_result(self, request: HttpRequest, result: ExecutionResult) -> GraphQLHTTPResponse`
+
+## get_context
+
+`get_context` allows to provide a custom context object that can be used in your
+resolver. You can return anything here, by default we return a dictionary with
+the request.
+
+```python
+class MyGraphQLView(AsyncGraphQLView):
+    async def get_context(self, request: HttpRequest) -> Any:
+        return {"example": 1}
+
+
+@strawberry.type
+class Query:
+    @strawberry.field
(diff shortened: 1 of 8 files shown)


patrick91:
I think we can merge this (after review), will write another test when this PR gets released: https://github.com/django/django/pull/13632

strawberry/django/views.py
```diff
@@ -1,39 +1,35 @@
+import asyncio
 import json
 import os
 from typing import Any, Dict, Optional
 
+from django.core.exceptions import SuspiciousOperation
```
marcoacierno, line 6: sus

tests/django/test_async_view.py
```diff
@@ -0,0 +1,120 @@
...
+from strawberry.django.views import AsyncGraphQLView as AsyncBaseGraphQLView
+
+from .app.models import Example
+
+
+pytestmark = [
+    pytest.mark.asyncio,
+    pytest.mark.skipif(
+        django.VERSION < (3, 1),
+        reason="Annotated type was added in python 3.9",
```
marcoacierno, line 21: is this reason correct?

patrick91: no :D I copied and pasted it, will fix!

botberry:



---

This release adds a new AsyncGraphQLView for django.

docs/05_integrations/django.md
```diff
@@ -114,3 +114,114 @@ class MyGraphQLView(GraphQLView):
...
+
+from api.schema import schema
+
+urlpatterns = [
+    path("graphql/", AsyncGraphQLView.as_view(schema=schema)),
+]
+```
+
+You'd also need to add `strawberry.django` to the `INSTALLED_APPS` of your
+project, this is needed to provide the template for the GraphiQL interface.
```
jkimbo, line 135: BTW this isn't true because the template is loaded in using the Template loader directly: https://github.com/strawberry-graphql/strawberry/blob/7f2e848717050abc26d6b664459757761cb8f078/strawberry/django/views.py#L69

docs/05_integrations/django.md
```diff
@@ -114,3 +114,114 @@ class MyGraphQLView(GraphQLView):
...
+    @strawberry.field
+    def example(self, info) -> str:
+        return str(info.context["example"])
+```
+
+Here we are returning a custom context dictionary that contains only one item
+called "example".
+
+Then we use the context in a resolver, the resolver will return "1" in this
+case.
```
jkimbo, line 176: Using the `grapqhl+response` code block might be nice here to show the result rather than just describing it.

docs/05_integrations/django.md
```diff
@@ -114,3 +114,114 @@ class MyGraphQLView(GraphQLView):
...
+        return Query(name="Patrick")
+
+
+@strawberry.type
+class Query:
+    name: str
+```
+
+Here we are returning a Query where the name is "Patrick", so we when requesting
+the field name we'll return "Patrick" in this case.
```
jkimbo, line 197: Same as above

strawberry/django/views.py
jkimbo: not sus

jkimbo:
Looks great @patrick91 👍

docs/05_integrations/django.md
patrick91: really? :O Will check that!

jkimbo: Actually it's these lines that do it:
https://github.com/strawberry-graphql/strawberry/blob/7f2e848717050abc26d6b664459757761cb8f078/strawberry/django/views.py#L71-L79

PR #568, opened 2020-11-14
Fix order of type evaluation

patrick91:
This is a bit hard to explain or even reproduce, the gist is that
I am converting a big schema to strawberry and there's a lot of 
circular dependencies, so I have to use lazy types and forward refs.

In one case, not sure why, the LazyType annotation came back as a
string and since we where evaluating the LazyType before doing
the eval, things went wrong.

This PR changes the order and fixes that :)

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+This release fixes an issue with forward types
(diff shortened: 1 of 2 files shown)


botberry:



---

This release fixes an issue with forward types

PR #572, opened 2020-11-17
Add support for strawberry.field in mypy

patrick91:
This is a very lazy way of doing this, but we can improve the plugin later :)

Ideally we should differentiate when strawberry.field is used as decorator and we should check that the resolver type matches the field definition, but this is proving hard to do, so for now I'll just return any.


botberry:



---

This release improves mypy support for strawberry.field

PR #573, opened 2020-11-17
Fix attribute without default error in mypy

patrick91:
Pretty much the title, previously we were not able to make this code type check:

```python
import strawberry


def some_resolver() -> str:
    return ""


@strawberry.type
class Example:
    b: str = strawberry.field(name="b")
    a: str
    c: str = strawberry.field(name="c", resolver=some_resolver)
    d: str = strawberry.field(resolver=some_resolver)
```

because mypy didn't know that strawberry.field works as dataclass.field.

This PR fixes is it. In future we will improve the plugin to check for 
example if the resolver type is correct.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,5 @@
+Release type: patch
+
+This release type improves support for strawberry.field in mypy,
+now we don't get `Attributes without a default cannot follow attributes with one`
+when using strawberry.field before a type without a default.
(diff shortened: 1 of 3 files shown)


botberry:



---

This release type improves support for strawberry.field in mypy,
now we don't get `Attributes without a default cannot follow attributes with one`
when using strawberry.field before a type without a default.

PR #575, opened 2020-11-17
Add some runtime checking for unions

### LINKED ISSUE #571: Raise error when using Scalars (or non strawberry types) with Union
If we do something any of the following:

```python
X = strawberry.Union("X", (int, str))

@strawberry.type
class A:
    field: Union[int, str]


@strawberry.type
class A:
    field: Union[NonStrawberryTypeA, NonStrawberryTypeB]
```

we should get an error informing that:

1. you can't use unions with scalars (not supported by GraphQL)
2. you can only use union with strawberry types

jkimbo:
## Description

Fixes #571

Not sure whether it's worth creating specific Strawberry exceptions for this cases. What do you think @patrick91 ?

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* #571

## Checklist

### DIFF
--- a/tests/types/test_union.py
+++ b/tests/types/test_union.py
@@ -1,3 +1,4 @@
+from dataclasses import dataclass
 from typing import Generic, List, Optional, TypeVar, Union
 
 import pytest
@@ -138,3 +139,24 @@ def test_cannot_use_union_directly():
 
     with pytest.raises(ValueError, match=r"Cannot use union type directly"):
         Result()
+
+
+def test_error_with_empty_type_list():
+    with pytest.raises(Exception, match="No types passed to `union`"):
+        strawberry.union("Result", [])
+
+
+def test_error_with_scalar_types():
+    with pytest.raises(
+        TypeError, match="Scalar type `int` cannot be used in a GraphQL Union"
+    ):
+        strawberry.union("Result", (int,))
+
+
+def test_error_with_non_strawberry_type():
+    @dataclass
+    class A:
+        a: int
+
+    with pytest.raises(TypeError, match="Union type `A` is not a Strawberry type"):
+        strawberry.union("Result", (A,))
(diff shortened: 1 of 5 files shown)


botberry:

```markdown
Release type: patch

```

jkimbo:
Not sure why codecov thinks this change will reduce code coverage...

tests/types/test_union.py
```diff
@@ -138,3 +139,24 @@ class B:
 
     with pytest.raises(ValueError, match=r"Cannot use union type directly"):
         Result()
+
+
+def test_error_with_empty_type_list():
+    with pytest.raises(Exception, match="No types passed to `union`"):
```
BryceBeagle, line 145: TIL `match` argument. I will have to use this in the future.

BryceBeagle (changes requested):
Should we create custom exceptions (i.e. in `strawberry.exceptions`) for these?

jkimbo:
> Should we create custom exceptions (i.e. in `strawberry.exceptions`) for these?

I was wondering about that. Happy to do so though I can't see developers wanting to catch specific exceptions. Also I think the `len(types)` check should probably stay as a TypeError.

patrick91:
> Should we create custom exceptions (i.e. in `strawberry.exceptions`) for these?

Good question!

My idea around exception was that at some point we'd have a way to link to the docs for some of them[1], especially the ones that maybe don't make a lot of sense to a beginner :)

[1] Via a custom __repr__ or something :)

BryceBeagle:
> Also I think the len(types) check should probably stay as a TypeError

Technically speaking, our custom exception could just extend `TypeError`

jkimbo:
> My idea around exception was that at some point we'd have a way to link to the docs for some of them[1], especially the ones that maybe don't make a lot of sense to a beginner :)

I like this reasoning around when to add custom exceptions. Will update the PR with some.

patrick91 (approved):
Perfect, I was struggling understanding why we changed the test, but the I noticed the test was wrong 🤣 nice one!

PR #581, opened 2020-11-18
Safer way to get a mypy type

patrick91:
This PR uses a safer way to get named types. It should prevent a crash when mypy can't find a type.

Unfortunately it is quite hard to reproduce this, as with other mypy issues. I've tested on pollen's 
codebase and this seems to fix without causing other issues.


botberry:



---

Another small improvement for mypy, this should prevent mypy from crashing when it can't find a type

PR #582, opened 2020-11-18
Set origin_name to be the field origin name

patrick91:
We want origin_name to be the name of the field, not the resolver,
otherwise this would prevent us to the reuse a resolver with the
same name in a type.

Fixes use cases like this:

```python
def get_name(self) -> str:
    return "Name"

@strawberry.type
class Query:
    name: str = strawberry.field(resolver=get_name)
    name_2: str = strawberry.field(resolver=get_name)
```

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,13 @@
+Release type: patch
+
+Fix issue preventing reusing the same resolver for multiple fields, like here:
+
+```python
+def get_name(self) -> str:
+    return "Name"
+
+@strawberry.type
+class Query:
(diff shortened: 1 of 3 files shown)


botberry:



---

Fix issue preventing reusing the same resolver for multiple fields, like here:

```python
def get_name(self) -> str:
    return "Name"

@strawberry.type
class Query:
    name: str = strawberry.field(resolver=get_name)
    name_2: str = strawberry.field(resolver=get_name)
```

patrick91:
@BryceBeagle I'll merge this as I need to fix a bug :D but feel free to double check it 😊

PR #587, opened 2020-11-19
Fix issue when using union as type param for generics

patrick91:
Thanks @jordangarside for the bug report! 

This PR fixes an issue with using unions inside generic types, this is now
supported:

```python
@strawberry.type
class Dog:
    name: str

@strawberry.type
class Cat:
    name: str

@strawberry.type
class Connection(Generic[T]):
    nodes: List[T]

@strawberry.type
class Query:
    connection: Connection[Union[Dog, Cat]]
```

While previously it was failing trying to get the name of the Union

### DIFF
--- a/strawberry/types/generics.py
+++ b/strawberry/types/generics.py
@@ -1,34 +1,56 @@
 import builtins
 import dataclasses
-from typing import Dict, Iterable, Tuple, Type, cast
+from typing import Dict, Iterable, Tuple, Type, Union, cast
 
 from strawberry.union import StrawberryUnion, union
 from strawberry.utils.str_converters import capitalize_first
-from strawberry.utils.typing import is_type_var
+from strawberry.utils.typing import is_type_var, is_union
 
 from .types import FederationFieldParams, FieldDefinition, TypeDefinition
 
 
-def get_name_from_types(types: Iterable[Type]):
-    return "".join([capitalize_first(type.__name__) for type in types])
+def get_name_from_types(types: Iterable[Union[Type, StrawberryUnion]]):
+    names = []
+
+    for type_ in types:
+        if isinstance(type_, StrawberryUnion):
+            return type_.name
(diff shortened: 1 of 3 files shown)


botberry:



---

This release fixes an issue when usign unions inside generic types, this is now
supported:


```python
@strawberry.type
class Dog:
    name: str

@strawberry.type
class Cat:
    name: str

@strawberry.type
class Connection(Generic[T]):
    nodes: List[T]

@strawberry.type
class Query:
    connection: Connection[Union[Dog, Cat]]
```

jkimbo:
Do interfaces have the same issue?

patrick91:
> Do interfaces have the same issue?

Interfaces used inside generics?

strawberry/types/generics.py
```diff
@@ -1,35 +1,55 @@
...
-def get_name_from_types(types: Iterable[Type]):
-    return "".join([capitalize_first(type.__name__) for type in types])
+def get_name_from_types(types: Iterable[Union[Type, StrawberryUnion]]):
+    names = []
+
+    for type_ in types:
+        if isinstance(type_, StrawberryUnion):
+            return type_.name
+        else:
+            name = capitalize_first(type_.__name__)
```
patrick91, line 19: This is quite ugly, but I'm pretty sure that we'll improve this when the objectification proposed by @BryceBeagle gets done :)

jkimbo:
Yeah, like this:
```python
@strawberry.interface
class Pet:
	name: str

@strawberry.type
class Query:
    connection: Connection[Pet]
```

patrick91:
@jkimbo added a test, they work because we use the class name :)

jkimbo:
@patrick91 great! Usually if something breaks with unions it will also break for interfaces 😄

jkimbo (approved):

PR #588, opened 2020-11-19
create debug-server extra

### LINKED ISSUE #586: (dev server) support for Subscriptions requires `uvicorn[standard]`
To use `Subscription` websockets are required, to enable websockets you need to install `uvicorn[standard]`. Currently `pyproject.toml` / `poetry` is only installing `uvicorn`. Not having the "standard" extra results in a failure to upgrade the HTTP request to WebSocket, you can see this by increasing the uvicorn log_level in [`cli/commands/server.py`](https://github.com/strawberry-graphql/strawberry/blob/master/strawberry/cli/commands/server.py#L43) and attempting a Subscription.

I see two paths to improving this situation, I'm happy to submit a PR for either route - but I need advice on how to proceed 👍 

_Option 1: update dependencies in pyproject.toml_

Changing the [`uvicorn` dependency](https://github.com/strawberry-graphql/strawberry/blob/master/pyproject.toml#L27) to be `{version = ">=0.11.6,<0.13.0", extras = ["standard"]}` would result in websockets being supported. However as per [Uvicorn's QuickStart](https://www.uvicorn.org/#quickstart) that brings in Cython dependencies which may not be what folks want.

_Option 2: improve documentation to mention this requirement_

The safe route, at the risk of a bit more work for the user. It's likely that any "real" deployment would have folks needing to be a bit more opinionated about the way they run their ASGI sever.

ossareh:
Fixes #586 by introducing carving the uvicorn requirement off into an extra called "debug-server".

## Description

While poetry does not allow for aliases, the core "ask" in #586 cannot be solved, to that end I've updated the documentation to point out to folks new to the project that they need to provide their own websocket enabled server and that the debug-server can be made to do what they want.

A few other minor changes took place, I added some spaces to pyproject.toml (happy to revert if you want), and I included "asgiref" in the django extra since it is needed in the django tests ( I think the dev dependency was catching this ).

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation

* #586 

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
 
       - run: pip install poetry codecov
       - name: Install dependencies
-        run: poetry install --extras "django flask opentelemetry"
+        run: poetry install --extras "debug-server django flask pydantic opentelemetry"
       - name: Install dependencies
         run: poetry add django==${{ matrix.django }}
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
 
       - run: pip install poetry codecov
       - name: Install dependencies
-        run: poetry install --extras "debug-server django flask pydantic opentelemetry"
+        run: poetry install
       - name: Install dependencies
         run: poetry add django==${{ matrix.django }}
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,9 @@ flake8-eradicate = "^1.0.0"
 pytest-mypy-plugins = "^1.6"
 pytest-mock = "^3.3"
 pytest-django = {version = "^4.1"}
+asgiref = "^3.2"
 pytest-flask = {version = "^1.1.0"}
+flask = {version = "^1.1"}
 requests = "^2.25.0"
 pre-commit = "^2.8.2"
 pytest-benchmark = "^3.2.3"
@@ -62,6 +64,7 @@ flake8-isort = "^4.0.0"
 flake8-black = "^0.2.1"
 django = {version = ">=2,<4", optional = false}
 pydantic = {version = "^1.6.1", optional = false}
+uvicorn = ">=0.11.6,<0.13.0"
 
 [tool.poetry.extras]
 debug-server = ["uvicorn"]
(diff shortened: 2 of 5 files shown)


botberry:

```markdown
Release type: patch

```

.github/workflows/test.yml
```diff
@@ -21,7 +21,7 @@ jobs:
 
       - run: pip install poetry codecov
       - name: Install dependencies
-        run: poetry install --extras "django flask opentelemetry"
+        run: poetry install --extras "debug-server django flask pydantic opentelemetry"
```
patrick91, line 24: I think by default poetry installs dev dependencies, shall add the debug-server dependencies as dev deps too? so we don't have to specify it here? :)

patrick91 (approved):
This looks great, thanks so much! I only added a note about dev deps, let me know what you think!

.github/workflows/test.yml
ossareh: oh, that's nice! I'm coming from pipenv where you have to be explicit to ensure dev deps are installed. I'll trial the change and then send an update if it works out.

Typically should I rebase the commit into the right place or do you do squash merges when it is time?

patrick91: @ossareh awesome! no strong preference on rebasing vs squashing, but I think we leaning towards squashing the PRs /cc @BryceBeagle :)

pyproject.toml
```diff
@@ -51,7 +51,9 @@ flake8-eradicate = "^1.0.0"
 pytest-mypy-plugins = "^1.6"
 pytest-mock = "^3.3"
 pytest-django = {version = "^4.1"}
+asgiref = "^3.2"
 pytest-flask = {version = "^1.1.0"}
+flask = {version = "^1.1"}
```
patrick91, line 56: 🔥 thank you!

patrick91:
Very strong contribution contribution, thanks for that, will merge in a second!

Feel free to fill this for to request a sticker: https://forms.gle/7AQw5cA6kc5BKp2L9 😊

We also have a discord channel if you want to join: https://discord.gg/ZkRTEJQ

PR #592, opened 2020-11-21
Add interface docs

jkimbo:
## Description

Add documentation for interfaces.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation

* #328 

## Checklist

### DIFF
--- a/docs/02_types/interfaces.md
+++ b/docs/02_types/interfaces.md
@@ -14,7 +14,8 @@ member of that interface.
 
 For example, let's say a `Customer` (interface) can either be an `Individual`
 (object) or a `Company` (object). Here's what that might look like in the
-[GraphQL Schema Definition Language](https://graphql.org/learn/schema/#type-language) (SDL):
+[GraphQL Schema Definition Language](https://graphql.org/learn/schema/#type-language)
+(SDL):
 
 ```graphql
 interface Customer {
@@ -52,8 +53,9 @@ query {
 
 Whether the object is a `Company` or an `Individual`, it doesn’t matter – you
 still get their name. If you want some object-specific fields, you can query
-them with an [inline fragment](https://graphql.org/learn/queries/#inline-fragments),
-for example:
+them with an
+[inline fragment](https://graphql.org/learn/queries/#inline-fragments), for
+example:
 
 ```graphql
 query {
@@ -92,8 +94,8 @@ interface Customer {
 
 ## Implementing interfaces
 
-To define an object type that implements an interface, the type must inherit from
-the interface:
+To define an object type that implements an interface, the type must inherit
+from the interface:
 
 ```python
 import strawberry
@@ -107,7 +109,7 @@ class Company(Customer):
     # additional fields
 ```
 
-> ⚠️  If you add an object type which implements an interface, but that object
+> ⚠️ If you add an object type which implements an interface, but that object
 > type doesn’t appear in your schema as a field return type or a union member,
 > then you will need to add that object to the Schema defintion directly.
 
@@ -174,7 +176,6 @@ class Company(Customer):
     def name(self) -> str:
         return f"{self.name} Limited"
 ```
-
 
 ## Resolving an interface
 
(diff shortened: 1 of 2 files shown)


botberry:

```markdown
Release type: patch

```

docs/02_types/interfaces.md
```diff
@@ -5,4 +5,194 @@ path: /docs/types/interfaces
...
+        return f"{self.name} Limited"
+```
+
+
+## Resolving an interface
+
+When a field’s return type is an interface, GraphQL needs to know what specific
+object type to use for the return value. In the example above, each customer
+must be categorized as an `Individual` or `Company`. To do this you need to
+always return an instance of an object type:
```
jkimbo, line 183: @patrick91 currently the `resolve_type` method on interfaces is not exposed right?

docs/02_types/interfaces.md
```diff
@@ -5,4 +5,194 @@ path: /docs/types/interfaces
...
+  name: String!
+}
+
+type Query {
+  customers: [Customer!]!
+}
+```
+
+Notice that the `Customer` interface requires the `name: String!` field. Both
+`Company` and `Individual` implements that field so that they can implement the
```
BryceBeagle, line 40: `implements` -> `implement`

This sentence is kinda strange though. I wonder if something else could be said so we're not saying "_implement_ x so they can _implement_ y"

docs/02_types/interfaces.md
```diff
@@ -5,4 +5,194 @@ path: /docs/types/interfaces
...
+interface Customer {
+  name: String!
+}
+```
+
+> Note: Interface classes should never be instantiated directly.
+
+## Implementing interfaces
+
+To define object type that implement an interface the type must inherit from the
```
BryceBeagle, line 94: `define object type that implement` -> `define an object type that implements`

Also, add comment after `interface`

docs/02_types/interfaces.md
BryceBeagle, line 183: The example for this is kinda strange, as you're not returning "an instance", but instead a _list_ of instances, so I'm not exactly sure what you're trying to describe. Maybe add an example of what _not_ to do?

BryceBeagle (changes requested):
Looking really good 👍

docs/02_types/interfaces.md
jkimbo: Hmm yeah I'm trying to get across that you can't return a plain object or an instance of the interface because then GraphQL won't know what object type you're referring to. I agree it's confusing that in the example I'm returning a list but it's irrelevant if it's a list or not. I'll change the example to just return a single value so it's not as confusing.

docs/02_types/interfaces.md
jkimbo: Reworded the sentence to avoid the double `implement`

docs/02_types/interfaces.md
jkimbo: Done

docs/02_types/interfaces.md
jkimbo: @BryceBeagle updated the example. Does that help?

BryceBeagle: I think that's better. Should we also get rid of the unused `id` parameter?

jkimbo: Done

docs/02_types/interfaces.md
```diff
@@ -5,4 +5,191 @@ path: /docs/types/interfaces
...
+
+```python
+from typing import List
+import strawberry
+
+
+@strawberry.type
+class Query:
+    @strawberry.field
+    def bestCustomer(self) -> Customer:
```
patrick91, line 193:
```suggestion
    def best_customer(self) -> Customer:
```

docs/02_types/interfaces.md
```diff
@@ -5,4 +5,191 @@ path: /docs/types/interfaces
...
+query {
+  customers {
+    name
+  }
+}
+```
+
+Whether the object is a `Company` or an `Individual`, it doesn’t matter – you
+still get their name. If you want some object-specific fields, you can query
+them with an inline fragment, for example:
```
patrick91, line 55: Shall we add a link for docs on inline fragments? :)

patrick91 (approved):
This looks pretty good! Awesome stuff 😊

I've added two minor notes! ✨

BryceBeagle (approved):

PR #607, opened 2020-12-05
docs(subscriptions): add documentation about how subscriptions work

ossareh:
Documentation about how `@strawberry.subscription` works

## Description

For a project at work I had to fully understand how subscriptions work; with that new knowledge I've updated the documentation to reflect what I've learned.

I'd say I understand Async[Iterator|Generator] at approx 70-80% right now so in the "advanced" section I could really do with a second set of eyes on the code example, and the language surrounding it.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation

* #328 (tackles the subscription basics task)

## Checklist


botberry:

```markdown
Release type: patch

```

ossareh:
I'm choosing to not include a RELEASE.md because this is a doc only change, lmk if I should create one for this change.

patrick91:
Nice work! I'll take a better look soon! And I'll also ping @marcoacierno since he worked on subscriptions with Strawberry 😊

docs/01_general/04_subscriptions.md
```diff
@@ -5,4 +5,139 @@ path: /docs/subscriptions
...
+```python
+import asyncio
+
+import strawberry
+
+@strawberry.type
+class Subscription:
+    @strawberry.subscription
+    async def count(self, target: int = 100) -> int:
+        for i in range(0, target):
```
BryceBeagle, line 22:
```suggestion
        for i in range(target):
```

docs/01_general/04_subscriptions.md
```diff
@@ -5,4 +5,139 @@ path: /docs/subscriptions
 
 # Subscriptions
 
-Documentation coming soon
+In GraphQL you can use subscriptions to stream data from a server. To enable
+this with Strawberry your server must support ASGI and websockets.
+
+This is how you define a subscription capable resolver:
```
BryceBeagle, line 11:
```suggestion
This is how you define a subscription-capable resolver:
```

docs/01_general/04_subscriptions.md
```diff
@@ -5,4 +5,139 @@ path: /docs/subscriptions
...
+mutations the subscription can return any GraphQL type, not only scalars as
+demonstrated here.
+
+## Advanced Subscription Patterns
+
+Typically a GraphQL subscription is streaming something more interesting back.
+With that in mind your subscription function can return one of:
+
+- AsyncIterator, or
+- AsyncGenerator
```
BryceBeagle, line 56:
```suggestion
- `AsyncIterator`, or
- `AsyncGenerator`
```

docs/01_general/04_subscriptions.md
```diff
@@ -5,4 +5,139 @@ path: /docs/subscriptions
...
+        pass
+
+
+async def lines(stream: streams.StreamReader) -> AsyncIterator[str]:
+    """
+    lines reads all lines from the provided stream, decoding them as UTF-8
+    strings.
+    """
+    while True:
+        if b := await wait_for_call(stream.readline):
```
BryceBeagle, line 97: Strawberry is supposed to be compatible with python 3.7. Should the docs be using a 3.8 feature (`:=`)?

docs/01_general/04_subscriptions.md
```diff
@@ -5,4 +5,139 @@ path: /docs/subscriptions
...
+    """
+    return await asyncio.create_subprocess_exec(
+        "/bin/bash",
+        "-c",
+        f"for ((i = 0 ; i < {target} ; i++)); do echo $i; sleep 0.2; done",
+        stdout=subprocess.PIPE,
+    )
+
+
+async def tail(proc) -> AsyncGenerator[str, None]:
```
BryceBeagle, line 115:
```suggestion
async def tail(proc: subprocess.Proc) -> AsyncGenerator[str, None]:
```

BryceBeagle (changes requested):
Couple minor suggestions and a question. Overall this looks pretty good

BryceBeagle (approved):

docs/01_general/04_subscriptions.md
```diff
@@ -5,4 +5,140 @@ path: /docs/subscriptions
...
+    # this case the process can finish between the loop predicate and the call
+    # to read a line from stdout. This is a good example of why you need to
+    # be defensive by using asyncio.wait_for in wait_for_call().
+    while proc.returncode is None:
+        async for l in lines(proc.stdout):
+            yield l
+    else:
+        # read anything left on the pipe after the process has finished
+        async for l in lines(proc.stdout):
+            yield l
```
BryceBeagle, line 130: Just noticed this post-review. Is the `else` clause necessary? Is there a situation that the `while` body will exit early?

ossareh: In this case yes; since we're running a sub process it is reasonable that the process exits, set's the returncode and we've not yet read everything from the stdout pipe.

The "else" isn't strictly necessary and the call to `lines(proc.stdout)` would be un-indented one level. The `else` pattern is a hold over from my work where we use it to be explicit that there's another case to think about.

BryceBeagle (approved):

docs/01_general/04_subscriptions.md
```diff
@@ -5,4 +5,140 @@ path: /docs/subscriptions
...
+    """
+    while True:
+        b = await wait_for_call(stream.readline)
+        if b:
+            yield b.decode("UTF-8").rstrip()
+        else:
+            break
+
+
+async def exec_proc(target: int) -> subprocess.Process:
```
marcoacierno, line 104: I think that's a very cool example of what it's possible with subscriptions and so on, I just wonder if it's still complex and might make people afraid of trying subscriptions? 🤔 

I would merge like this and update with what people think in the future!

marcoacierno (approved):
Looks very good to me! :)

docs/01_general/04_subscriptions.md
jkimbo: I agree with this. It's a very impressive example but I worry that it's not very relatable to the use cases that most developers will be using GraphQL for (web development). Also: would it work on windows? 

As @marcoacierno said: I don't think it's a blocker. We should merge this (the documentation is excellent) but I think we should consider creating a different example in future. Maybe a live Twitter feed or a chat client.

ossareh: I hear you both, and respect the feedback. My thoughts on how I got here, and what options I didn't take:

I saw myself with two options:

 * build a chat server example [1]
 * use what I'm working on as the basis

I chose the latter because it was expedient for me to add as it's a distilled version of what I'm building at work. In that regard, it's nice to have something more complete than a toy example (the counter). After building the counter example I asked myself "would this have helped me?", the answer was a firm no. Mostly, honestly, because going from that example to something more fleshed out required learning about asyncio more completely.

When thinking about what I learned in the asyncio world this is what came up:

 * `asyncio.wait_for()`; to prevent some async task becoming jammed and keeping your Sub open when it is technically done
 * `lines()`; to demonstrate the `AsyncIterator` pattern
 * `exec_proc` is akin to a subscriber, the type of thing one may use when creating a chat system over graphql

It turns out that both my example and the chat server example need these patterns, but I wasn't certain that I could get a working minimal chat server up and running in the 2 hour window I gave myself to produce these docs.

RE: [1], and in general chat servers.

I saw this: https://discord.com/channels/689806334337482765/700765415764394066/779694329295011840 
by @jkimbo and it made me wonder if a chat server as a constant theme throughout the docs would be a good fit?

 * subscription for joining a room and receiving messages
 * mutations for sending a message to a room, setting a name, profile, etc
 * queries for things like room occupants, subject, chat participant profiles

It also allows for expansion into things as Strawberry fills out, i.e. uploading profile pic files.

docs/01_general/04_subscriptions.md
```diff
@@ -5,4 +5,140 @@ path: /docs/subscriptions
...
+
+@strawberry.type
+class Subscription:
+    @strawberry.subscription
+    async def count(self, target: int = 100) -> int:
+        for i in range(target):
+            yield i
+            await asyncio.sleep(0.5)
+
+schema = strawberry.Schema(subscription=Subscription)
```
jkimbo, line 26: I think you'll need to pass a `query` otherwise it will raise an exception.

ossareh:
Good catch!

docs/01_general/04_subscriptions.md
jkimbo: > After building the counter example I asked myself "would this have helped me?", the answer was a firm no.

If thats the case then I would say that it's probably not a good idea to have this example since it might just confuse people rather than help them.

> I saw this: discord.com/channels/689806334337482765/700765415764394066/779694329295011840
by @jkimbo and it made me wonder if a chat server as a constant theme throughout the docs would be a good fit?

I think building a chat server would be a great advanced tutorial to have in the docs as a pseudo real life example of how to use Strawberry. I'm not sure it's a good idea to make it the basis of all the docs though because it might imply that you have to have a real time application to use Strawberry. TBH I'm less sure now that it's necessary to have a constant theme throughout the docs because it will probably mean we have to come up with contrived examples to fit the theme. 

An alternative example: we could consume a realtime data feed and expose it through a subscription. Kraken has a public websocket that allows you subscribe to changes in a currency pair: https://support.kraken.com/hc/en-us/articles/360022327631-WebSocket-API-Public-Feeds-Python-Code-Examples

Here is a working example of it:

```python
from typing import AsyncGenerator
import json
from datetime import datetime

import strawberry
import websockets

@strawberry.type
class Query:
    @strawberry.field
    def hello() -> str:
        return "world"

@strawberry.type
class Ticker:
    best_ask_price: str
    best_bid_price: str
    close_price: str
    timestamp: datetime

@strawberry.type
class Subscription:
    @strawberry.subscription
    async def ticker(self, pair: str = "ETH/EUR") -> AsyncGenerator[Ticker, None]:
        async with websockets.connect('wss://ws.kraken.com') as websocket:
            await websocket.send(json.dumps({
				"event": "subscribe",
				"pair": [pair],
				"subscription": { "name": "ticker" }
			}))
            while True:
                data = await websocket.recv()
                data = json.loads(data)
                if isinstance(data, list):
                    yield Ticker(
                        best_ask_price=data[1]["a"][0],
                        best_bid_price=data[1]["b"][0],
                        close_price=data[1]["c"][0],
                        timestamp=datetime.now(),
                    )

schema = strawberry.Schema(query=Query, subscription=Subscription)
```

@ossareh do you think that shows enough to be useful for people?

ossareh: > If thats the case then I would say that it's probably not a good idea to have this example since it might just confuse people rather than help them.

I think I may have contributed some confusion here; to clear it up:

 * "counter example" is the example in the first portion of the doc. I consider it a "toy"
 * "process manager example", is the one we're discussing the virtues of.

The "counter example" would not have helped me understand how to integrate a subscription into my infrastructure, but it's helpful in terms of understanding the general principle (yield values) in a subscription. In that regard, I think the "counter example" should stay.

The "process manager example" is more helpful to me, in part because it's a copy of what I'm doing at work ^_^, but also because it touches on aspects of async python that you need to use to integrate subscriptions into a real world infra, i.e `Async[Generator|Iterator]`, `wait_for`, etc.

I believe it is outside of the realms of this documentation to teach asyncio, but a good example could also illustrate common pitfalls. For example, while building out the real solution at work, I was consistently holding open the subscription after the process had exited, this led me to `wait_for`, a refactor, and now a pattern where the server can end the subscription when there is no more data available. 

All that said, I'm happy to take direction on changes to make, things to delete, etc. I ultimately want this to be good documentation that is helpful to a broad range of folks - not "mikes pet process manager blog post".

-- 

RE: your example.

I think it's a good one, I like that it demonstrates a more involved problem space. That is, it's more than the "counter example" toy. What drew me to Strawberry is the `type` annotation and the automatic handling of serialization, etc. Currently that is a bit lost in your example because of needing to interact with the kraken websocket. This is why I moved the `exec_proc`, etc, stuff out of the `Subscription.run_command()` method.

Thus I think something like what follows works slightly better for me because it puts all of the non Strawberry stuff in it's own little unit allowing me to clearly see what it is that Strawberry does and what the bridge to my infra needs to do.

```python
async def kraken_feed(pair: str) -> AsyncIterator[<type for kraken data>]:
  # setup sub to kraken feed
  async with websockets.connect('wss://ws.kraken.com') as feed:
    # request data from kraken feed
    await feed.send(...)
    while True:
      # handle and yield feed data
      data = await feed.recv()
      data = json.loads(data)
      yield <type for kraken data...>

@strawberry.type
class Subscription:
    @strawberry.subscription
    async def ticker(self, pair: str = "ETH/EUR") -> AsyncGenerator[Ticker, None]:
        # at this point we should aim for the reader to say:
        # "OK, ignore crypto, this is how I integrate into my infra/product/etc"
        async for v in kraken_feed(pair):
          yield Ticker(
            best_ask_price=<from v>,
            best_bid_price=<from v>,
            close_price=<from v>,
            timestamp=datetime.now(),
        )            

schema = strawberry.Schema(query=Query, subscription=Subscription)
```

docs/01_general/04_subscriptions.md
ossareh: It seems the [mutations docs](https://github.com/strawberry-graphql/strawberry/blob/master/docs/01_general/03_mutations.md) have the same issue.

ossareh: @jkimbo thoughts on: https://github.com/strawberry-graphql/strawberry/pull/613 ?

docs/01_general/04_subscriptions.md
jkimbo: > I think I may have contributed some confusion here; to clear it up:

Ah sorry I did misunderstand you but it makes sense now.

I like the change you've done to my example and I think we should go with that as the example. What do you think @marcoacierno @BryceBeagle @patrick91 ?

patrick91: > I like the change you've done to my example and I think we should go with that as the example. What do you think @marcoacierno @BryceBeagle @patrick91 ?

Yes, I'd prefer to have a simpler example in the docs if possible. Also I think we should keep @ossareh's work somewhere. Maybe we can move it to a tutorial section of the website?

ossareh: 👍

will aim to move this forward tomorrow (before the Utd v City game....)

LouisAmon: I'm a web guy and I tried both the code from @ossareh and @jkimbo but none of these apply to my use-case

I have a MySQL DB and a React web page, I want to update my page content in real-time (in my case it's IoT events being sent)

The ticker with Kraken is cool and all but it's consuming an already functional websocket so it doesn't really prove anything, same goes for the counter which doesn't really help because it's fake data being generated on the fly so no link with any Mutation whatsoever

BTW the counter doesn't work at all on my GraphiQL, I just get:
```
"Your subscription data will appear here after server publication!"
```

Anyway... how would that work with an actual DB ? If my mutation creates say a `Message`, how would my new message get routed all the way to GraphiQL(or my React's Apollo Client) ? How do I know it will be the right message ?

Is the Pub/Sub logic directly implemented in Strawberry itself (like Apollo Server) or do I need to implement my own polling ?

I'm super confused 🙃

LouisAmon: Food for thought : wouldn't it be cool to demonstrate (and unit test) Subscriptions via SQLite and/or SQLAlchemy ?

This would relate nicely to a lot of Python web devs I think (Django is super similar anyway,)

patrick91: > Is the Pub/Sub logic directly implemented in Strawberry itself (like Apollo Server) or do I need to implement my own polling ?

No, at least not for now, we could provide something. Personally I've used https://asyncio-redis.readthedocs.io/en/latest/pages/examples.html#pubsub and https://pypi.org/project/aiopubsub/ in the past :)

Might be interesting to provide something with strawberry, that would lower the entry barrier for using subscriptions at least for experimentation, like apollo does: https://www.apollographql.com/docs/apollo-server/data/subscriptions/#pubsub-implementations

ossareh: Zoinks! I forgot about this 🤦🏻‍♂️

@patrick91 you had mentioned the notion of somewhere to put examples; could this be a case of benefiting from a strawberry-graphql/examples repo? The examples could be referenced by the documentation, and we could allow a bit more freedom there as to how best to demonstrate helpful approaches. If so, then I think the minimal "counter" case ([as defined here](https://github.com/strawberry-graphql/strawberry/pull/607#discussion_r539525567)) would probably be best here.

@LouisAmon

> "Your subscription data will appear here after server publication!"

I'll double check in case the example has bit rotted, but typically I've found this to be the case when the code is not written correctly and the result is not yielded to the ASGI server. Are you running under some sort of ASGI compatible server when you see this?

LouisAmon: Yup, I'm running the Strawberry CLI so ASGI/Starlette if I'm not mistaken !

ossareh: @LouisAmon did you follow the steps in the [readme about WebSockets](https://github.com/strawberry-graphql/strawberry/blob/master/README.md#websockets)?

LouisAmon: You mean having the debug-server and uvicorn installed ? I double checked and yes that seems good

But after looking at my chrome it seems I get error 400 during the WS handshake... odd

ossareh: @louisamon - yes, those extras are necessary for WebSockets to work. Especially the case with uvicorn: `pip install uvicorn` is not the same as `uvicorn[standard]` even though "standard" sounds like it should be. This tripped me up for a while, but the docs are clear the `standard` extra is necessary for WebSockets: https://www.uvicorn.org/#quickstart

> `$ pip install uvicorn[standard]`
> 
> This will install uvicorn with "Cython-based" dependencies (where possible) and other "optional extras".
> 
> In this context, "Cython-based" means the following:
> 
>  * the event loop uvloop will be installed and used if possible.
>  * the http protocol will be handled by httptools if possible.
> 
> Moreover, "optional extras" means that:
> 
> * the websocket protocol will be handled by websockets (should you want to use wsproto you'd need to install it manually) if possible.

patrick91: @ossareh 

> @patrick91 you had mentioned the notion of somewhere to put examples; could this be a case of benefiting from a strawberry-graphql/examples repo? The examples could be referenced by the documentation, and we could allow a bit more freedom there as to how best to demonstrate helpful approaches. If so, then I think the minimal "counter" case (as defined here) would probably be best here.

Yes, definitely, I'm thinking of writing a blog post about it as well, just didn't have time yet (and also I need to test it in production ;)).

By the way, we are having a call to discuss a roadmap tomorrow on our discord if you want to join :)

LouisAmon: Turns out you're right @ossareh  : I did have `uvicorn` but not `uvicorn[standard]`

It works now 👍  Thank you so much !

I wonder if it would be possible to include `uvicorn[standard]` in the dev dependencies. Maybe it should be part of the `debug-server` variant in the pyproject.toml ?

Still. I'm stuck because I have just a simple counter and I don't know how to manage my DB changes with this Subscription. Maybe my hosting service (AWS AppSync) will help me in that regard (I managed to deploy Strawberry on AWS Lambda + AWS AppSync, if anybody's interested 😉 )

ossareh: @LouisAmon, glad it's working! It's definitely not as straight forward as I would like! It turns out uvicorn is written to work in a wide variety of situations and Cython has dependencies which are not easily available / well tested on non x86_64 architectures. Or something to that end.

RE: `[debug-server]` including `uvicorn` - we looked at that in #586 and it's not possible because we cannot create aliases to pypi packages.

RE: subscription and db updates - the apollo graphql guide has details about this. There is a careful choreography around what query types to use. It's broadly like this:

 * change the system with mutations
 * subscribe to streams using subscription
 * or, subscribe to updates to an initial query using either polling ([pollInterval](https://www.apollographql.com/docs/react/data/queries/#polling)) or transition to a subscription ([subscribeToMore](https://www.apollographql.com/docs/react/data/subscriptions/#subscribing-to-updates-for-a-query))

Here's how I broadly think of it in terms of a chat system:

 * mutation to post a message
 * query for room name, topic, etc
 * subscription for message stream
 * query + subscription for room membership

So, in your case, I think your minimal app would need to have two arcs:

 * mutation for updating db
 * subscription if you want those updates streamed to you

With something managing the db updated -> subscription for you.

ossareh: @patrick91 ooh nice, I've a meeting which clashes with it so I can't make it - but thanks for the pointer :)

patrick91: > Still. I'm stuck because I have just a simple counter and I don't know how to manage my DB changes with this Subscription. Maybe my hosting service (AWS AppSync) will help me in that regard (I managed to deploy Strawberry on AWS Lambda + AWS AppSync, if anybody's interested 😉 )

That's cool! If you want we can chat on discord at some point, I can try to help figure out how to make it work :)

PR #614, opened 2020-12-09
Add execute context that can handle promises

jkimbo:
## Description

POC for a custom graphql-core `ExecutionContext` class that handles Promises (from https://github.com/syrusakbary/promise) so that you can use a sync dataloader. Couple of notes:

* This is a just a rough exploration of a potential solution. It needs a lot more testing and I've probably missed some things.
* There is no easy way to extend some of the `ExecutionContext` methods so I had to duplicate them and add custom logic (I also got rid of the code paths that handle async code). This will make maintenance a bit of a pain if those code paths change
* Regardless of whether we end up going down this route I think expanding the `execute` and `execute_sync` methods to allow passing of the `execution_context_class` would be beneficial

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,6 +65,7 @@ flake8-black = "^0.2.1"
 django = {version = ">=2,<4", optional = false}
 pydantic = {version = "^1.6.1", optional = false}
 uvicorn = ">=0.11.6,<0.13.0"
+promise = "^2.3"
 
 [tool.poetry.extras]
 debug-server = ["uvicorn"]
--- a/strawberry/schema/execute_context.py
+++ b/strawberry/schema/execute_context.py
@@ -1,0 +1,127 @@
+from typing import Any, Dict, List, Optional, Union
+
+from promise import Promise, is_thenable, promise_for_dict
+
+from graphql import (
+    ExecutionContext,
+    GraphQLError,
+    GraphQLField,
+    GraphQLFieldResolver,
+    GraphQLObjectType,
+    GraphQLOutputType,
+    GraphQLResolveInfo,
+)
+from graphql.execution.values import get_argument_values
+from graphql.language import FieldNode
+from graphql.pyutils import AwaitableOrValue, Path, Undefined
+
+
+class ExecutionContextWithPromise(ExecutionContext):
+    def build_response(self, data):
+        if is_thenable(data):
+            original_build_response = super().build_response
+
+            def on_rejected(error):
+                self.errors.append(error)
+                return None
+
+            def on_resolve(data):
+                return original_build_response(data)
+
+            promise = data.catch(on_rejected).then(on_resolve)
+            return promise
+        return super().build_response(data)
+
+    def resolve_field_value_or_error(
+        self,
+        field_def: GraphQLField,
+        field_nodes: List[FieldNode],
+        resolve_fn: GraphQLFieldResolver,
+        source: Any,
+        info: GraphQLResolveInfo,
+    ) -> Union[Exception, Any]:
+        """Resolve field to a value or an error.
+
+        Isolates the "ReturnOrAbrupt" behavior to not de-opt the resolve_field()
+        method. Returns the result of resolveFn or the abrupt-return Error object.
+
+        For internal use only.
+        """
+        try:
+            # Build a dictionary of arguments from the field.arguments AST, using the
+            # variables scope to fulfill any variable references.
+            args = get_argument_values(field_def, field_nodes[0], self.variable_values)
+
+            # Note that contrary to the JavaScript implementation, we pass the context
+            # value as part of the resolve info.
+            result = resolve_fn(source, info, **args)
+            if is_thenable(result):
+                return result
+
+            return result
+        except GraphQLError as error:
+            return error
+        except Exception as error:
+            return GraphQLError(str(error), original_error=error)
+
+    def complete_value_catching_error(
+        self,
+        return_type: GraphQLOutputType,
+        field_nodes: List[FieldNode],
+        info: GraphQLResolveInfo,
+        path: Path,
+        result: Any,
+    ) -> AwaitableOrValue[Any]:
+        """Complete a value while catching an error.
+        This is a small wrapper around completeValue which detects and logs errors in
+        the execution context.
+        """
+        completed: AwaitableOrValue[Any]
+        try:
+            if is_thenable(result):
+
+                def handle_error(error):
+                    self.handle_field_error(error, field_nodes, path, return_type)
+
+                completed = Promise.resolve(result).then(
+                    lambda resolved: self.complete_value(
+                        return_type, field_nodes, info, path, resolved
+                    ),
+                    handle_error,
+                )
+            else:
+                completed = self.complete_value(
+                    return_type, field_nodes, info, path, result
+                )
+            return completed
+        except Exception as error:
+            self.handle_field_error(error, field_nodes, path, return_type)
+            return None
+
+    def execute_fields(
+        self,
+        parent_type: GraphQLObjectType,
+        source_value: Any,
+        path: Optional[Path],
+        fields: Dict[str, List[FieldNode]],
+    ):
+        """Execute the given fields concurrently.
+
+        Implements the "Evaluating selection sets" section of the spec for "read" mode.
+        """
+        contains_promise = False
+        results = {}
+        for response_name, field_nodes in fields.items():
+            field_path = Path(path, response_name)
+            result = self.resolve_field(
+                parent_type, source_value, field_nodes, field_path
+            )
+            if result is not Undefined:
+                results[response_name] = result
+                if is_thenable(result):
+                    contains_promise = True
+
+        if contains_promise:
+            return promise_for_dict(results)
+
+        return results
--- a/tests/schema/test_sync_dataloader.py
+++ b/tests/schema/test_sync_dataloader.py
@@ -1,0 +1,33 @@
+from promise import Promise
+from promise.dataloader import DataLoader
+
+import strawberry
+
+
+def test_batches_correct():
+    load_calls = []
+
+    class TestDataLoader(DataLoader):
+        def batch_load_fn(self, keys):
+            load_calls.append(keys)
+            return Promise.resolve(keys)
+
+    @strawberry.type
+    class Query:
+        @strawberry.field
+        def get_id(self, info, id: str) -> str:
+            return info.context["dataloader"].load(id)
+
+    schema = strawberry.Schema(query=Query)
+    result = schema.execute_sync(
+        """
+        query {
+            id1: getId(id: "1")
+            id2: getId(id: "2")
+        }
+    """,
+        context_value={"dataloader": TestDataLoader()},
+    )
+    assert not result.errors
+    assert result.data == {"id1": "1", "id2": "2"}
+    assert load_calls == [["1", "2"]]
--- a/tests/schema/test_sync_dataloader.py
+++ b/tests/schema/test_sync_dataloader.py
@@ -2,6 +2,7 @@ from promise import Promise
 from promise.dataloader import DataLoader
 
 import strawberry
+from strawberry.schema.execute_context import ExecutionContextWithPromise
 
 
 def test_batches_correct():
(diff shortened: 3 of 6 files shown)


strawberry/schema/execute_context.py
```diff
@@ -0,0 +1,127 @@
...
+        """
+        try:
+            # Build a dictionary of arguments from the field.arguments AST, using the
+            # variables scope to fulfill any variable references.
+            args = get_argument_values(field_def, field_nodes[0], self.variable_values)
+
+            # Note that contrary to the JavaScript implementation, we pass the context
+            # value as part of the resolve info.
+            result = resolve_fn(source, info, **args)
+            if is_thenable(result):
```
Ambro17, line 58: What's the purpose of this if? Wouldn't the program flow be the same without it?

strawberry/schema/execute_context.py
```diff
@@ -0,0 +1,127 @@
...
+        """
+        completed: AwaitableOrValue[Any]
+        try:
+            if is_thenable(result):
+
+                def handle_error(error):
+                    self.handle_field_error(error, field_nodes, path, return_type)
+
+                completed = Promise.resolve(result).then(
+                    lambda resolved: self.complete_value(
```
Ambro17, line 87: Just for consistency, i would define both args as lambdas or both as 'normal' functions

tests/schema/test_sync_dataloader.py
```diff
@@ -0,0 +1,35 @@
...
+            id1: getId(id: "1")
+            id2: getId(id: "2")
+        }
+    """,
+        context_value={"dataloader": TestDataLoader()},
+        execution_context_class=ExecutionContextWithPromise,
+    )
+    assert not result.errors
+    assert result.data == {"id1": "1", "id2": "2"}
+    assert load_calls == [["1", "2"]]
```
Ambro17, line 35: Impressive 👏

strawberry/schema/execute_context.py
jkimbo: Oh yeah I don't need this anymore. In the original code there was an `if is_awaitable` branch as well which I removed.

strawberry/schema/execute_context.py
jkimbo: Yep agreed 👍

pyproject.toml
```diff
@@ -65,6 +65,7 @@ flake8-black = "^0.2.1"
 django = {version = ">=2,<4", optional = false}
 pydantic = {version = "^1.6.1", optional = false}
 uvicorn = ">=0.11.6,<0.13.0"
+promise = "^2.3"
```
patrick91, line 68: should we make this optional? 🤔 

also I found this the other day, wonder if it is a good alternative https://github.com/celery/vine (maybe they are swappable)

tests/schema/test_sync_dataloader.py
```diff
@@ -0,0 +1,143 @@
...
+
+    @strawberry.type
+    class Query:
+        @strawberry.field
+        def get_company(self, info, id: str) -> Company:
+            return (
+                info.context["company_dataloader"]
+                .load(id)
+                .then(lambda id: Company(id=id))
+            )
```
patrick91, line 111: would be possible to get a more complex test/example where we do something like

```
def resolver():
    a = loader.load("abc")
    return a.something
```

I guess we'd need to use .then and return a promise, I think it is useful to make sure that works and document it 😊

patrick91 (commented):
Awesome! Well done, I thought it was going to be more complicated to implement. It is a bit of a shame that we had to duplicate the code a bit, I wonder how we could make sure we keep in sync with graphql-core 🤔 

But otherwise it looks great 😊

pyproject.toml
jkimbo: Ooh vine looks good, I'll look into that. The promise library is not maintained so definitely in favour so switching to something else.

strawberry/schema/execute_context.py
```diff
@@ -0,0 +1,138 @@
...
+from graphql.language import FieldNode, OperationDefinitionNode
+from graphql.pyutils import (
+    AwaitableOrValue,
+    Path,
+    Undefined,
+    is_awaitable as default_is_awaitable,
+)
+
+
+def is_awaitable(value):
```
BryceBeagle, line 20:
```suggestion
def is_awaitable(value: object) -> bool:
```

strawberry/schema/execute_context.py
```diff
@@ -0,0 +1,138 @@
...
+        # Wrap execute in a Promise
+        original_execute_operation = super().execute_operation
+
+        def promise_executor(v):
+            return original_execute_operation(operation, root_value)
+
+        promise = Promise.resolve(None).then(promise_executor)
+        return promise
+
+    def build_response(self, data):
```
BryceBeagle, line 63: Can this be type annotated?

BryceBeagle: Same with the nested functions

strawberry/schema/execute_context.py
```diff
@@ -0,0 +1,138 @@
...
+            self.handle_field_error(error, field_nodes, path, return_type)
+            return None
+
+    def execute_fields(
+        self,
+        parent_type: GraphQLObjectType,
+        source_value: Any,
+        path: Optional[Path],
+        fields: Dict[str, List[FieldNode]],
+    ):
```
BryceBeagle, line 118: Can this be type annotated, potentially with the `S` TypeVar from above? If so, then `results` below could also use `S`?

tests/schema/test_sync_dataloader.py
```diff
@@ -0,0 +1,143 @@
...
+            return info.context["dataloader"].load(id)
+
+    schema = strawberry.Schema(query=Query)
+    result = schema.execute_sync(
+        """
+        query {
+            id1: getId(id: "1")
+            id2: getId(id: "2")
+        }
+    """,
```
BryceBeagle, line 29: Looks like the indentation got messed up here?

BryceBeagle: Appears to be an issue with a couple more `"""` strings below

jkimbo: It's how black formats it 🤷 I think it looks weird too.

BryceBeagle: Strange that black would format `"""` strings, seeing that the whitespace can be an important part of them :thinking:

jkimbo:
@patrick91 @BryceBeagle this is ready for review now

strawberry/promise/dataloader.py
```diff
@@ -0,0 +1,164 @@
...
+from strawberry.exceptions import WrongNumberOfResultsReturned
+from strawberry.promise import Promise, async_instance, default_scheduler
+
+
+T = TypeVar("T")
+K = TypeVar("K")
+
+
+# Private: cached resolved Promise instance
+cache = local()
```
BryceBeagle, line 15: If it's private, it should probably be `_cache`?

strawberry/promise/dataloader.py
```diff
@@ -0,0 +1,164 @@
...
+def dispatch_batch(loader: PromiseDataLoader, batch: Batch) -> None:
+    # If the batch has already been dispatched then bail out
+    if batch.dispatched is True:
+        return
+
+    batch.dispatched = True
+
+    keys = [task.key for task in batch.tasks]
+
+    # TODO: check if load_fn return an awaitable and it is a list
```
BryceBeagle, line 118: Was this TODO resolved?

strawberry/promise/dataloader.py
```diff
@@ -0,0 +1,164 @@
...
+        batch_promise = loader.load_fn(keys)
+    except Exception as e:
+        for task in batch.tasks:
+            task.reject(e)
+
+        return None
+
+    # Assert the expected response from load_fn
+    if not batch_promise or not isinstance(batch_promise, Promise):
+        # TODO
```
BryceBeagle, line 130: Was this TODO resolved?

strawberry/promise/promise_list.py
```diff
@@ -0,0 +1,165 @@
+# Copy of the Promise library (https://github.com/syrusakbary/promise) with some
```
BryceBeagle, line 1: Should this file (and the others from the promise lib) be marked excluded from the coverage suite?

BryceBeagle: Also, can we get a comment/mini readme to what "some modifications" means?

tests/promise/test_dataloader.py
```diff
@@ -0,0 +1,250 @@
...
+
+        assert load_calls == [["A", "B"], ["C"]]
+
+    do().get()
+
+
+# Represents Errors
+
+
+def test_resolves_to_error_to_indicate_failure():
```
BryceBeagle, line 137: Was this comment supposed to be attached to the below function, and the autoformatter separated it?

BryceBeagle: Even if so, I'm not too sure what it means

tests/promise/test_thread_safety.py
```diff
@@ -0,0 +1,238 @@
...
+                test_with_process_loop._force_stop = True
+                items.put(("ABORT",))
+                timeouts.append(format_exc())
+
+        items.put(("STOP",))
+
+    loop_thread = Thread(target=event_loop)
+    loop_thread.start()
+
+    worker_threads = [Thread(target=worker) for i in range(0, num_threads)]
```
BryceBeagle, line 229:
```suggestion
    worker_threads = [Thread(target=worker) for i in range(num_threads)]
```

BryceBeagle (changes requested):

strawberry/promise/dataloader.py
patrick91: I haven't fixed it in my implementation :)

patrick91: we can do it in another PR

strawberry/promise/promise_list.py
patrick91: I wonder, I guess in future we might change this code a lot (ie updating the code to only use py 3 things), maybe we can change say "originally based on promises"

strawberry/schema/execute_context.py
```diff
@@ -0,0 +1,149 @@
...
+        return_value = zip(value.keys(), resolved_values)
+        return dict(return_value)
+
+    return Promise.all(value.values()).then(handle_success)
+
+
+class ExecutionContextWithPromise(ExecutionContext):
+    """
+    Extend the default ExecutionContext with support of Promises.
+    Note: Only usable in a sync context
```
patrick91, line 52: shall we add a warning when running under async (like django does)?

patrick91 (approved):
Had a quick look and it looks pretty good, thanks for this! I

strawberry/promise/async_.py
```diff
@@ -0,0 +1,159 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+# Based on https://github.com/petkaantonov/bluebird/blob/master/src/promise.js
+from collections import deque
+from threading import local
+from typing import Any, Callable, Optional
+
+
+if False:
```
Ambro17, line 36: What's the purpose of this?

BryceBeagle: This comes from the upstream source of this file. I think it's supposed to be like `if TYPE_CHECKING` but I don't know why they didn't do that instead.

patrick91: maybe because they needed support for python 2 (and < 3.5) without having to install any other library :)

botberry:

```markdown
Release type: patch

```

jkimbo:
Apologies for the delay with this PR. Life events have meant I haven't had much time lately 👶 

I think the last thing that needs doing is to add support for lists.

patrick91:
> Apologies for the delay with this PR. Life events have meant I haven't had much time lately 👶
> 
> I think the last thing that needs doing is to add support for lists.

Awesome! Thanks for the update 😊 I was looking at fixing the tests but I got distracted by other things 😊

PR #615, opened 2020-12-10
Add support for unions in get_optional_annotation 

### LINKED ISSUE #602: strawberry.utils.typing.get_optional_annotation fails when provided an `Optional[Union]`
`strawberry.utils.typing.get_optional_annotation` fails when provided an `Optional[Union]`

```pycon
>>> from typing import Optional, Union
>>> from strawberry.utils.typing import get_optional_annotation

>>> get_optional_annotation(Optional[Union[int, str]])
<class 'int'>
```
This should return `Union[int, str]` instead

DeusFerrariis:
Modified `utils/typing.py::get_optional_annotation` when passed a union as demonstrated in issue #602.

## Description

`get_optional_annotation` when passed `Optional[Union[str, int]]` will now return Union[str, int]

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

Closes #602 

## Checklist

### DIFF
--- a/strawberry/utils/typing.py
+++ b/strawberry/utils/typing.py
@@ -42,7 +42,14 @@ def is_optional(annotation: Type) -> bool:
 
 def get_optional_annotation(annotation: Type) -> Type:
     types = annotation.__args__
-    non_none_types = [x for x in types if x != None.__class__]  # noqa:E711
+
+    non_none_types = tuple(x for x in types if x != None.__class__)  # noqa:E711
+
+    # if we have multiple non none types we want to return a copy of this
+    # type (normally a Union type).
+
+    if len(non_none_types) > 1:
+        return annotation.copy_with(non_none_types)
 
     return non_none_types[0]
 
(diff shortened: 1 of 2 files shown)


patrick91 (approved):
Nice! Thank you for this 😊 

Feel free to fill this form if you want a sticker: https://forms.gle/7AQw5cA6kc5BKp2L9 ✨

botberry:

```markdown
Release type: patch

```

BryceBeagle:
What's the advantage of throwing an exception with `Optional[Union]` types instead of just returning a `Union` type? I know the codebase currently doesn't do anything with `Optional[Union]`s, but I don't see why that would be a reason for the function to not support it.

DeusFerrariis:
> What's the advantage of throwing an exception with `Optional[Union]` types instead of just returning a `Union` type? I know the codebase currently doesn't do anything with `Optional[Union]`s, but I don't see why that would be a reason for the function to not support it.

Until a conclusion about the design around the function is made raising an explicit error is better than returning obscure and/or unexpected responses.

PR #622, opened 2020-12-17
Add docs for unions

jkimbo:
## Description

Add documentation for unions.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation

* #328

## Checklist

### DIFF
--- a/docs/02_types/interfaces.md
+++ b/docs/02_types/interfaces.md
@@ -186,12 +186,10 @@ class Company(Customer):
 When a field’s return type is an interface, GraphQL needs to know what specific
 object type to use for the return value. In the example above, each customer
 must be categorized as an `Individual` or `Company`. To do this you need to
-always return an instance of an object type:
+always return an instance of an object type from your resolver:
 
 ```python
-from typing import List
 import strawberry
-
 
 @strawberry.type
 class Query:
--- a/docs/02_types/interfaces.md
+++ b/docs/02_types/interfaces.md
@@ -168,7 +168,7 @@ class Customer:
 ```
 
 This resolve method will be called by objects who implement the interface.
-Object classes can override the implmentation by defining their own `name`
+Object classes can override the implementation by defining their own `name`
 field:
 
 ```python
--- a/docs/02_types/union.md
+++ b/docs/02_types/union.md
@@ -19,7 +19,7 @@ Whenever we return a `MediaItem` in our schema, we might get an `Audio`, a
 object types; you cannot create a union type out of interfaces, other unions or
 scalars.
 
-A good usecase for unions would be on a search field for example:
+A good use case for unions would be on a search field. For example:
 
 ```graphql
 searchMedia(term: "strawberry") {
--- a/docs/02_types/union.md
+++ b/docs/02_types/union.md
@@ -5,8 +5,10 @@ path: /docs/types/unions
 
 # Union types
 
-Union types are very similar to [interfaces](/docs/types/interfaces), but they don't have any common fields
-between the types. Here’s a union, expressed in
+Union types are similar to [interfaces](/docs/types/interfaces) however, while interfaces
+dictate fields that must be common to all implementations, unions do not. Unions
+just represent a selection of allowed types and make no requirements on those
+types. Here’s a union, expressed in
 [GraphQL Schema Definition Language](https://graphql.org/learn/schema/#type-language)
 (SDL):
 
@@ -36,8 +38,9 @@ searchMedia(term: "strawberry") {
 ```
 
 Here, the `searchMedia` field returns `[MediaItem!]!`, a list where each member
-is part of the `MediaItem` union. Since union members share no fields,
-selections are always made with [inline fragments](https://graphql.org/learn/queries/#inline-fragments).
+is part of the `MediaItem` union. So, for each member, we want to select different
+fields depending on which kind of object that member is. We can do that by using
+[inline fragments](https://graphql.org/learn/queries/#inline-fragments).
 
 ## Defining unions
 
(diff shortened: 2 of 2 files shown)


botberry:

```markdown
Release type: patch

```

docs/02_types/union.md
```diff
@@ -5,4 +5,133 @@ path: /docs/types/unions
...
+[GraphQL Schema Definition Language](https://graphql.org/learn/schema/#type-language)
+(SDL):
+
+```graphql
+union MediaItem = Audio | Video | Image
+```
+
+Whenever we return a `MediaItem` in our schema, we might get an `Audio`, a
+`Video` or an `Image`. Note that members of a union type need to be concrete
+object types; you cannot create a union type out of interfaces or other unions.
```
patrick91, line 19: scalars too, right?

docs/02_types/union.md
```diff
@@ -5,4 +5,133 @@ path: /docs/types/unions
...
+  }
+  ... on Image {
+    src
+  }
+}
+```
+
+Here, the `searchMedia` field returns `[MediaItem!]`, a list where each member
+is part of the `MediaItem` union. Since union members share no fields,
+selections are always made with [inline fragment](https://graphql.org/learn/queries/#inline-fragments).
```
patrick91, line 39:
```suggestion
selections are always made with [inline fragments](https://graphql.org/learn/queries/#inline-fragments).
```

docs/02_types/union.md
```diff
@@ -5,4 +5,133 @@ path: /docs/types/unions
...
+  ... on Video {
+    thumbnailUrl
+  }
+  ... on Image {
+    src
+  }
+}
+```
+
+Here, the `searchMedia` field returns `[MediaItem!]`, a list where each member
```
patrick91, line 37:
```suggestion
Here, the `searchMedia` field returns `[MediaItem!]!`, a list where each member
```

docs/02_types/union.md
```diff
@@ -5,4 +5,133 @@ path: /docs/types/unions
...
+is part of the `MediaItem` union. Since union members share no fields,
+selections are always made with [inline fragment](https://graphql.org/learn/queries/#inline-fragments).
+
+## Defining unions
+
+In Strawberry there are two ways to define a union:
+
+You can use the use the `Union` type from the `typing` module:
```
patrick91, line 45: shall we say that the name is auto generated based on the types names?

docs/02_types/union.md
```diff
@@ -5,4 +5,133 @@ path: /docs/types/unions
...
+```python
+from typing import Union
+import strawberry
+
+@strawberry.type
+class Query:
+    @strawberry.field
+    def latest_media(self) -> Union[Audio, Video, Image]:
+        return Video(
+            thumbnail_url="https://i.ytimg.com/vi/dQw4w9WgXcQ/hq720.jpg",
```
patrick91, line 135: 🤣

patrick91 (approved):
Looks great! Thank you 🙇

docs/02_types/union.md
jkimbo: Yep good point!

docs/02_types/union.md
```diff
@@ -5,4 +5,135 @@ path: /docs/types/unions
...
+}
+```
+
+Or if you need to specify a name or a description for a union you can use the
+`strawberry.union` function:
+
+```python+schema
+import strawberry
+
+MediaItem = strawberry.union("MediaItem", types=(Audio, Video, Image))
```
BryceBeagle, line 94: Does the string need to match the assigned symbol, similar to `TypeVar` and `namedtuple`?

docs/02_types/union.md
```diff
@@ -5,4 +5,135 @@ path: /docs/types/unions
...
+type Video {
+  thumbnailUrl: String!
+}
+
+type Image {
+  src: String!
+}
+```
+
+> Note: Union types should never be instantiated directly.
```
BryceBeagle, line 119: Probably give the alternative here?

> Union types should never be instantiated directly. Always use `strawberry.union`

BryceBeagle: Also, why is this the case? I feel like users should be able to make the objects manually. `strawberry.union` and `StrawberryUnion` share the same signature

BryceBeagle (changes requested):

docs/02_types/interfaces.md
```diff
@@ -186,13 +186,11 @@ class Company(Customer):
 When a field’s return type is an interface, GraphQL needs to know what specific
 object type to use for the return value. In the example above, each customer
 must be categorized as an `Individual` or `Company`. To do this you need to
-always return an instance of an object type:
+always return an instance of an object type from your resolver:
```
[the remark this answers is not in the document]
jkimbo: @BryceBeagle that is a good point: your example would work I think. Not sure how I would explain that though...

docs/02_types/union.md
jkimbo: No it doesn't. Do you think it's confusing that it does or is it more confusing it it doesn't match? Maybe inlining it in the Query definition would be better 🤔

docs/02_types/union.md
jkimbo: Oh by this I mean that you shouldn't do this: `MediaItem(...)`. This note might be more confusing that useful, I'll remove it.

jkimbo:
@BryceBeagle @patrick91 this is ready for another review

docs/02_types/interfaces.md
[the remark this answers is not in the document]
BryceBeagle: I guess it's probably fine to leave as is. Attempting to be more accurate will only make it more confusing for beginners

BryceBeagle (approved):

PR #625, opened 2020-12-25
Add more Pydantic types to the type conversion

marcoacierno:
## Description

EmailStr is a type of Pydantic to validate email addresses, I updated the type conversion code to convert it to the GraphQL String type.

I also added more types to the conversion (all to string):

- pydantic.AnyUrl
- pydantic.AnyHttpUrl
- pydantic.HttpUrl
- pydantic.PostgresDsn
- pydantic.RedisDsn

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -154,6 +154,33 @@ argon2 = ["argon2-cffi (>=16.1.0)"]
 bcrypt = ["bcrypt"]
 
 [[package]]
+name = "dnspython"
+version = "2.0.0"
+description = "DNS toolkit"
+category = "dev"
+optional = false
+python-versions = ">=3.6"
+
+[package.extras]
+dnssec = ["cryptography (>=2.6)"]
+doh = ["requests", "requests-toolbelt"]
+idna = ["idna (>=2.1)"]
+curio = ["curio (>=1.2)", "sniffio (>=1.1)"]
+trio = ["trio (>=0.14.0)", "sniffio (>=1.1)"]
+
+[[package]]
+name = "email-validator"
+version = "1.1.2"
+description = "A robust email syntax and deliverability validation library for Python 2.x/3.x."
+category = "dev"
+optional = false
+python-versions = "*"
+
+[package.dependencies]
+dnspython = ">=1.15.0"
+idna = ">=2.0.0"
+
+[[package]]
 name = "eradicate"
 version = "2.0.0"
 description = "Removes commented-out code."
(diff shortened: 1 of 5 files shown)


botberry:

```markdown
Release type: patch

```

pyproject.toml
```diff
@@ -64,6 +64,7 @@ flake8-isort = "^4.0.0"
 flake8-black = "^0.2.1"
 django = {version = ">=2,<4", optional = false}
 pydantic = {version = "^1.6.1", optional = false}
+email-validator = {version = "^1.1.2", optional = false}
```
marcoacierno, line 67: If someone is using pydantic with pydantic.networks.EmailStr they will need to install `email-validator` anyway otherwise pydantic will fail

So I think we should be fine including it only in the dev-dependencies for when we run the tests?

poetry.lock
```diff
@@ -71,7 +71,7 @@ python-versions = "*"
 
 [[package]]
 name = "certifi"
-version = "2020.11.8"
+version = "2020.12.5"
```
marcoacierno, line 74: I don't know why those changes happen to the lock file, any idea?

ulgens: I just created https://github.com/strawberry-graphql/strawberry/pull/627 . After that one is merged, this PR can be rebased from main branch.

ulgens:
@marcoacierno `poetry.lock` file is up to date now. If you can rebase your branch and relock the dependencies, lock file diff should get smaller.

ulgens (changes requested):
@marcoacierno poetry.lock file is up to date now. If you can rebase your branch and relock the dependencies, lock file diff should get smaller.

marcoacierno:
Have a look now

patrick91 (approved):

pyproject.toml
ulgens: @patrick91 👋

ulgens: I think it should be a non-dev dependency, or optional one, depending how we currently treat pydantic. Marking as a dev only sounds risky to me.

patrick91: this is fine :)

patrick91: @ulgens it is a dependency only used when using pydantic and the email field, in their doc they say you need to install email validator: https://pydantic-docs.helpmanual.io/usage/types/#pydantic-types

so for us it is ok to only have it in dev deps (because we use the field in tests), I don't think it is need to mark it as optional as Pydantic doesn't either

ulgens: Okay then.

ulgens (approved):

PR #662, opened 2021-01-14
More consistent/unified method of converting from Strawberry-based objects to graphql-core schema

BryceBeagle:
## Description

Currently, the code that converts from Strawberry objects/types to graphql-core objects is a bit... rough, presumably due to the continuous expansion and feature addition that has occurred during early development. This PR seeks to clean up, consolidate, and standardize that conversion. 

Having a consistent schema conversion process will allow us to more easily implement our own custom types support (esp. `StrawberryObjectType` and `StrawberryField`).

A few notes:
* ~~This is still in progress. First thing to note is that many tests do not pass -> especially those with Unions/Generics.~~
* ~~I am putting this out before completion so I can get feedback on what's going on.~~
* The main file to look at right now is `strawberry/schema/schema_converter.py`
* I am aiming to have migrated all the functions that exist `strawberry/schema` to this new paradigm. EDIT: I left the Scalar stuff behind as there's room for some more complicated logical changes, but in another PR.
* As of yet, there aren't any planned new features/breaking changes. This will most likely be a backend-only change not observable to most users. 
* ~~I have been committing with `--no-verify` while things are still very broken. None of these commits will pass the pre-commit checks.~~

I would like some feedback here:
* I'm not a fan of the monolithic design of `GraphQLCoreConverter`. I would like it to be split up into multiple classes/files.
* A goal would be to make `GraphQLCoreConverter` implement an abstract class. This would allow us to support/switch to another backend without having to rewrite the entire system.
* I am not sure how to best spread an abstract class implementation to more than a single class.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* Progress towards: #489 

## Checklist


strawberry/exceptions.py
```diff
@@ -62,7 +62,7 @@ class UnallowedReturnTypeForUnion(Exception):
     def __init__(
         self, field_name: str, result_type: str, allowed_types: Set[GraphQLObjectType]
     ):
-        formatted_allowed_types = [type_.name for type_ in allowed_types]
+        formatted_allowed_types = list(sorted(type_.name for type_ in allowed_types))
```
BryceBeagle, line 65: This should be one of the few functional changes. This makes the sorting of the set deterministic

strawberry/directive.py
```diff
@@ -13,6 +17,18 @@ class DirectiveDefinition:
     locations: List[DirectiveLocation]
     description: Optional[str] = None
 
+    @property
+    def arguments(self) -> List[ArgumentDefinition]:
+        annotations = self.resolver.__annotations__
+        annotations = dict(islice(annotations.items(), 1, None))
```
patrick91, line 23: is this to remove `self`?

strawberry/schema/schema_converter.py
```diff
@@ -0,0 +1,324 @@
...
+    def __init__(self):
+        self.type_map: Dict[str, ConcreteType] = {}
+
+    def get_graphql_type_field(self, field: FIELD_ARGUMENT_TYPE) -> GraphQLType:
+        # TODO: Completely replace with get_graphql_type
+
+        graphql_type: GraphQLType
+
+        if _is_list(field):
+            graphql_type = self.from_list(field)
```
patrick91, line 52: I like this `from_abc` method 👏

strawberry/schema/schema_converter.py
```diff
@@ -0,0 +1,324 @@
...
+        return GraphQLInputField(
+            type_=field_type, default_value=default_value, description=field.description
+        )
+
+    def from_input_object_type(self, object_type: Type) -> GraphQLInputObjectType:
+        type_definition = object_type._type_definition
+
+        return GraphQLInputObjectType(
+            name=type_definition.name,
+            # TODO: Does this need to be deferred?
```
patrick91, line 180: good question, I'd keep it deferred as there might be some places where people might have circular dependencies even for input types.

patrick91 (commented):
This looks great! I'm going to test it a bit and let you know if there's any issue 😊

patrick91 (approved):
Tested with our work projects and nothing broke 😊

strawberry/directive.py
BryceBeagle: Honestly, I'm not sure what it does. I just migrated it from here:
https://github.com/strawberry-graphql/strawberry/blob/9815872a78123cd80ae3fe2d9bf4e6b03262c402/strawberry/schema/types/directives.py#L13-L23

strawberry/schema/schema_converter.py
BryceBeagle: Hopefully they can all be made abstract, such that any backend we support could implement each of them

strawberry/directive.py
patrick91: ok! I think I’ll need to improve directives a bit in future :)

strawberry/schema/schema.py
```diff
@@ -36,27 +36,34 @@ def __init__(
...
+        query_type = self.schema_converter.from_object_type(query)
+        mutation_type = (
+            self.schema_converter.from_object_type(mutation) if mutation else None
+        )
         subscription_type = (
-            get_object_type(subscription, self.type_map) if subscription else None
+            self.schema_converter.from_object_type(subscription)
+            if subscription
+            else None
         )
```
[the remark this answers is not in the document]
BryceBeagle: Hmm it might make the code here visually cleaner, but I'm not sure it makes sense from the design of the schema converter. What would it really mean if we asked it to convert `None` to an GraphQL object? I'm not sure short circuiting a None-in-None-out situation will play nicely into the future continued abstraction of the converter.

botberry:

```markdown
Release type: patch

```

strawberry/schema/schema.py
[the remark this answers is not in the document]
patrick91: got it, I can live with that :D

strawberry/schema/schema.py
```diff
@@ -66,15 +73,15 @@ def __init__(
...
+        self.query = self.schema_converter.type_map[query_type.name]
 
     def get_type_by_name(
         self, name: str
     ) -> Optional[
         Union[TypeDefinition, ScalarDefinition, EnumDefinition, StrawberryUnion]
     ]:
-        if name in self.type_map:
-            return self.type_map[name].definition
+        if name in self.schema_converter.type_map:
```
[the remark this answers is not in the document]
BryceBeagle: It was mainly done to avoid passing the map around all the time. Do you think that the Schema and the schema converter should share a reference to the same type map?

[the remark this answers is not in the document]
patrick91: uhm, no, I think we can add a property to the schema if we need the type map in future, was just curious 😊

PR #677, opened 2021-01-22
Fix duplicated input types error

patrick91:


botberry:



---

Fix for a duplicated input types error.

PR #679, opened 2021-01-23
Add support for mypy 0.800

### LINKED ISSUE #678: Strawberry mypy plugin not working with new 0.800 mypy version
Due to the new `info: TypeInfo` arg in `DataclassAttribute` (see [here](https://github.com/python/mypy/pull/9380/files#diff-1fdad80f805993800524fbb9b684cc3d09879f54a0e51c71f2a447c7d46c1eb4R38)), `DataclassAttribute` instantiation fails [here](https://github.com/strawberry-graphql/strawberry/blob/master/strawberry/ext/mypy_plugin.py#L476).

I do not have any solution to offer yet, but I guess it will have to be addressed before bumping mypy.

Cheers

patrick91:
Closes #678

Add support for mypy 0.800 by doing a very naive try/except. We
could copy the updated code here, but I don't think it is worth 
doing so.

I also tried to use inspect to check if the class accepted info, 
but didn't manage to get it working (also mypy gets compiled, I 
wonder if that might cause issues with inspect).

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+This release adds support for MyPy==0.800
--- a/poetry.lock
+++ b/poetry.lock
@@ -369,7 +369,7 @@ zipp = ">=0.5"
 
 [package.extras]
 docs = ["sphinx", "jaraco.packaging (>=8.2)", "rst.linker (>=1.9)"]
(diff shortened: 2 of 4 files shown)


botberry:



---

This release adds support for MyPy==0.800

Issue #559, opened 2020-11-06
Improve mypy support for strawberry.field


patrick91:
I think it is time to make a full copy of the mypy dataclasses plugin: https://github.com/python/mypy/blob/a6c3e0740fb7ddf65a89c1742c9d10c5fe2ed604/mypy/plugins/dataclasses.py

Our plugin reuses it but it doesn't work with `strawberry.field`, since dataclasses don't know anything about strawberry.field.

Currently we get this error:

```python
import strawberry

def some_resolver() -> str:
    return ""

@strawberry.type
class Example:
    a: str
    b: str = strawberry.field(name="b")
```

```
demo_enum.py:9: error: Incompatible types in assignment (expression has type "StrawberryField", variable has type "str")
```

so we need to tell mypy that this is what we expect :)

MasterBiryer:
+1

peniqliotuv:
+1 - this is especially problematic when we have resolvers that explicitly invoke another resolver and expect the result from that resolver to be typed - leading to a lot of `# type: ignore` comments.

peniqliotuv:
@patrick91  @BryceBeagle Here's a case I added to `tests/mypy/test_fields.yml`:
```
- case: test_type_annotation_field
  main: |
    import strawberry

    @strawberry.type
    class User:
        @strawberry.field
        def age(self) -> int:
          return 42
    reveal_type(User().age())
  out: |
    main:8: note: Revealed type is 'builtins.int'
```
Output
```
~/code/strawberry  main !1 ❯ poetry run pytest tests/mypy/test_fields.yml                                                                                                                  Py 3.7.3 Go 1.10.7 10:17:14 AM
=================================================================================================== test session starts ===================================================================================================
platform darwin -- Python 3.7.3, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
benchmark: 3.2.3 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
django: settings: tests.django.django_settings (from ini)
rootdir: /Users/jerry.tsui/Code/strawberry, configfile: pyproject.toml
plugins: emoji-0.2.0, cov-2.11.1, mypy-plugins-1.6.1, asyncio-0.14.0, benchmark-3.2.3, django-4.1.0, mock-3.5.1, flask-1.2.0
collected 8 items

[...]

PR #685, opened 2021-01-25
Release 🍓 0.44.11

peniqliotuv:
I've noticed a new typing issue when upgrading from version 0.39 to 0.44.10 with regards to typing the programmatic invocation of functions decorated with strawberry.field() - I think it's related to https://github.com/strawberry-graphql/strawberry/commit/ceb555f8cfe6526bc9e7982d36e21bbe2b7b4a35?branch=ceb555f8cfe6526bc9e7982d36e21bbe2b7b4a35&diff=split

```
import strawberry

@strawberry.type
class X:
  @strawberry.field(name="bar")
    async def bar_resolver(self, info: Info) -> Bar:
        return Bar()

@strawberry.type
class Y
  @strawberry.field(name="foo")
    async def foo_resolver(self, info: Info) -> str:
        bar = await Something().bar_resolver(info)
        return bar.name
```

The problem is, the return type of `Something.bar_resolver` is revealed to be `Any`, whereas it should be typed to `Bar`.

@BryceBeagle and I talked a bit about this on discord and we think that the problem is that we need to make `StrawberryField` and `FieldDefinition` generic on `T` (the return type of the resolver), but due to my lack of familiarity with the codebase and limited amount of free time Im just gonna post this draft here if someone wants to continue working on this.


patrick91:
@peniqliotuv going to close this as it is documented in https://github.com/strawberry-graphql/strawberry/issues/559

PR #687, opened 2021-01-26
Allow to disable query validation

patrick91:
This PR adds a new parameter on schema to disable query validation. This is useful when you know that the queries sent by your clients are already validated.

You'd do this for improving performances a bit:

image

profiling (for sync code, same results for async):

validation on:

image

validation off:

image

benchmarking script -> https://gist.github.com/patrick91/30ea0131aba769c845c15d4d8c702ac1


botberry:



---

This release add the ability to disable query validation by setting
`validate_queries` to `False`

```python
import strawberry

@strawberry.type
class Query:
    @strawberry.field
    def hello(self) -> str:
        return "Hello"


schema = strawberry.Schema(Query, validate_queries=validate_queries)
```

patrick91:
I need to add tests and docs, but otherwise this is ready to review

RELEASE.md
```diff
@@ -0,0 +1,17 @@
...
+
+@strawberry.type
+class Query:
+    @strawberry.field
+    def hello(self) -> str:
+        return "Hello"
+
+
+schema = strawberry.Schema(Query, validate_queries=validate_queries)
+```
```
patrick91, line 17: I wonder if we should move this flag to schema.execute instead of enabling it globally.

It might be better for enabling/disabling validation for specific clients/queries since we usually instantiate the schema once 🤔

BryceBeagle:
What happens if validation is disabled and the server still receives an invalid query?

patrick91:
@BryceBeagle I've added a test here: https://github.com/strawberry-graphql/strawberry/pull/687/files#diff-2d9e49eeaff80ff54483f868c16c6ca48bdb2100cf86548fd28666eee94bda84R43-R67

I need to add a test for other cases (wrong variable types, wrong fields requested) and see if anything bad happens

patrick91:
@BryceBeagle added a couple more tests, only one is a bit surprising, I'll definitely mention it in the docs :)

RELEASE.md
patrick91: I've moved this option to the .execute function :)

docs/types/schema.md
```diff
@@ -0,0 +1,92 @@
...
+
+```python
+async Schema().execute(
+    query: str,
+    variable_values: Optional[Dict[str, Any]] = None,
+    context_value: Optional[Any] = None,
+    root_value: Optional[Any] = None,
+    operation_name: Optional[str] = None,
+    validate_queries: Optional[bool] = True,
+) -> ExecutionResult:
```
BryceBeagle, line 56: This arrow might confuse people into thinking it's real syntax. I was thinking this was a messed up `async def` at first. Maybe we should use something like
````
```pycon
>>> .execute(...)
ExecutionResult(...)
```
````
instead?

File: docs/types/schema.md [hunk 3]
```diff
@@ -0,0 +1,92 @@
...
+```
+
+Sync version of `Schema().execute`
+
+## Query validation
+
+When creating a schema you can decide to opt-out from validating the queries
```
BryceBeagle, line 86: Would it be better to talk about "disabling" the feature instead of "opting-out"?

BryceBeagle (changes requested):

docs/types/schema.md
patrick91: Yes, will try that :) I really dislike how it looks on the website at the moment, I need to find a better way to have docs for the API.

docs/types/schema.md
patrick91: Yes, will update ☺️

docs/types/schema.md
```diff
@@ -45,45 +45,39 @@ Creates a GraphQL schema
...
-    query: str,
-    variable_values: Optional[Dict[str, Any]] = None,
-    context_value: Optional[Any] = None,
-    root_value: Optional[Any] = None,
-    operation_name: Optional[str] = None,
-    validate_queries: Optional[bool] = True,
-) -> ExecutionResult:
-```
+| Parameter name   | Type                              | Description                                                                                            |
+| ---------------- | --------------------------------- | ------------------------------------------------------------------------------------------------------ |
```
BryceBeagle, line 55: Maybe it would be better to have a column for defaults instead of bundling it with Type?

BryceBeagle (changes requested):

docs/types/schema.md
patrick91: great idea!

docs/types/schema.md
```diff
@@ -0,0 +1,86 @@
...
+| query            | `str`                             | The document to be executed                                                                            |
+| variable_values  | `Optional[Dict[str, Any]] = None` | The variables for this operation                                                                       |
+| context_value    | `Optional[Any] = None`            | The value of the context that will be passed down to resolvers                                         |
+| root_value       | `Optional[Any] = None`            | The value for the root type that will passed down to root resolvers                                    |
+| operation_name   | `Optional[str] = None`            | The name of the operation you want to execute, useful when sending a document with multiple operations |
+| validate_queries | `Optional[bool] = True`           | This flag allows to disable query validation                                                           |
```
BryceBeagle, line 61: Why is this `Optional` if the default isn't `None`?

BryceBeagle (changes requested):

docs/types/schema.md
BryceBeagle: specifically the `validate_queries` field

patrick91: mistake :D

patrick91:
Docs look nice enough: 

image

BryceBeagle (approved):

PR #703, opened 2021-02-04
📚 🐞 Suggest installing with development server

Ambro17:
Just a documentation bug i noticed while setting up a new project.
As per the instructions on the docs, the command `strawberry server schema` command will fail, as uvicorn is not installed if we install the lib as currently suggested (`pip install strawberry-graphql`)

Note: I'm not sure if development server is the best choice of words, perhaps debug server might be better?

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -28,10 +28,10 @@ After that we need a new virtualenv:
 
     python -m venv virtualenv
 
-Activate the virtualenv and then install strawberry
+Activate the virtualenv and then install strawberry plus the development server. 
 
     source virtualenv/bin/activate
-    pip install strawberry-graphql
+    pip install strawberry-graphql[debug-server]
 
 ## Step 2: Define the schema
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -28,7 +28,7 @@ After that we need a new virtualenv:
 
     python -m venv virtualenv
 
-Activate the virtualenv and then install strawberry plus the development server. 
+Activate the virtualenv and then install strawberry plus the debug server.
 
     source virtualenv/bin/activate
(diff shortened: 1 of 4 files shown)


BryceBeagle:
I wonder if the docs should even be recommending to install with pip instead of Poetry :thinking: 
```bash
poetry install -E debug-server
```

Ambro17:
I wondered the same, but poetry itself can bring a lot of troubles to beginners (see main issues about the recommended installer installing with python2)
Based on that I imagined it was intentional and i think it's simpler to suggest the easiest, dependency free, way. At least until poetry is more beginner friendly

BryceBeagle:
That's a fair point. I'd imagine that anyone that's familiar with Poetry already would know how to translate an instruction to use Pip for installation into one to use Poetry.

docs/index.md
```diff
@@ -28,10 +28,10 @@ After that we need a new virtualenv:
 
     python -m venv virtualenv
 
-Activate the virtualenv and then install strawberry
+Activate the virtualenv and then install strawberry plus the development server.
```
patrick91, line 31:
```suggestion
Activate the virtualenv and then install strawberry plus the debug server. 
```

Let's call it debug server, since the optional install is also called debug server :)

patrick91 (approved):

botberry:



---

Suggest installing the debug server on the getting started docs, so examples can work without import errors of uvicorn

Ambro17:
Requested changes are ready! Don't know why so many builds are failing though, i haven't touch any code

patrick91:
not sure what happened, but a bump of opentelemetry-sdk fixed the issue 🤔  will double check this tomorrow :)

PR #707, opened 2021-02-08
Add docs for resolvers and expose Info type

### LINKED ISSUE #680: Document info parameter and export type from strawberry
Currently we can pass an info parameter to resolvers to get information about the GraphQL execution, it is exposed from graphql-core: https://github.com/graphql-python/graphql-core/blob/3359fe60d9940b174270426438a89d629b6cc6e2/src/graphql/type/definition.py#L547-L568

We should make our own version (a copy basically) and expose it from `strawberry.types`[1]

The reasoning behind this is that we are an abstraction layer on top of graphql-core so our users shouldn't need to know about it :)

[1] or a more suited place :)

patrick91:
This PR adds docs for resolvers and also creates a public `Info` type for the `info` parameter.

Currently it is a subset of the graphql-core so I didn't create a function that converts the original to our version, let me know what you think about that :)

Sorry for the big diff, I went and removed info where it wasn't needed. And added the type where needed :)

Closes #680

### DIFF
--- a/docs/types/resolvers.md
+++ b/docs/types/resolvers.md
@@ -30,15 +30,14 @@ type Query {
 }
 ```
 
-we have defined a `User` type and a `Query` type. To define how the data is
-returned from our server, we use `resolvers` and we attached them to fields.
+We have defined a `User` type and a `Query` type. Next, to define how the data
+is returned from our server, we will attach resolvers to our fields.
 
 ## Let's define a resolver
 
 Let's create a resolver and attach it to the `lastUser` field. A resolver is a
-python function that returns data. In Strawberry there are two ways of defining
-resolvers, the first one is to create a function and then pass it to the field
-definition, like this:
+Python function that returns data. In Strawberry there are two ways of defining
+resolvers; the first is to pass a function to the field definition, like this:
 
 ```python
 def get_last_user() -> Optional[User]:
@@ -49,7 +48,7 @@ class Query:
     last_user: Optional[User] = strawberry.field(resolver=get_last_user)
 ```
 
-When executing this following query Strawberry knows that it needs to call the
+Now when Strawberry executes the following query, it will call the
 `get_last_user` function to fetch the data for the `lastUser` field:
 
 ```graphql+response
@@ -87,15 +86,15 @@ very small resolvers.
 
 > _NOTE:_ the _self_ argument is a bit special here, when executing a GraphQL
 > query, in case of resolvers defined with a decorator, the _self_ argument
-> corresponds to the _root_ value that field, in this example the _root_ value
-> is the value `Query` type, which is usually `None`, you can change the _root_
-> value when calling the `execute` method on a `Schema`. More on root values
-> below
+> corresponds to the _root_ value that field. In this example the _root_ value
+> is the value `Query` type, which is usually `None`. You can change the _root_
+> value when calling the `execute` method on a `Schema`. More on _root_ values
+> below.
 
 ## Defining arguments
 
 Fields can also have arguments; in Strawberry the arguments for a field are
-defined on the resolver, as you would normally do in a python function. Let's
+defined on the resolver, as you would normally do in a Python function. Let's
 define a field on a Query that returns a user by ID:
 
 ```python+schema
@@ -127,9 +126,9 @@ type Query {
 ## Accessing parent's data
 
 It is quite common to want to be able to access the data from the parent in a
-resolver, for example let's say that we want to defined a `fullName` field on
-our user. We can defined a new field with a resolver that combines the first
-name and last name:
+resolver. For example let's say that we want to define a `fullName` field on
+our `User`. We can define a new field with a resolver that combines its first
+and last names:
 
 ```python+schema
 import strawberry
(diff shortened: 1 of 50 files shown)


botberry:



---

Expose execution info under `strawberry.types.Info`

botberry:



---

Expose execution info under `strawberry.types.Info`

botberry:



---

Expose execution info under `strawberry.types.Info`

docs/types/resolvers.md
```diff
@@ -4,4 +4,205 @@ title: Resolvers
...
+    lastName: String!
+    fullName: String!
+}
+```
+
+In the case of a decorated resolver you can use the _self_ parameter as you
+would do in a normal python class[^1].
+
+For resolvers defined as normal python functions you can use the special `root`
+parameter, when defined, Strawberry will pass to it the value of the parent:
```
BryceBeagle, line 158:
```suggestion
For resolvers defined as normal Python functions, you can use the special `root`
parameter, when defined, Strawberry will pass to it the value of the parent:
```

Also, might want to clarify what `defined` means. (i.e. needs to appear in arglist)

docs/types/resolvers.md
```diff
@@ -4,4 +4,205 @@ title: Resolvers
...
+class User:
+    first_name: str
+    last_name: str
+    full_name: str = strawberry.field(resolver=full_name)
+```
+
+## Accessing execution information
+
+Sometimes it is useful to access the information for the current execution
+context, to do so you can provide the `info` parameter to resolvers, like this:
```
BryceBeagle, line 176:
```suggestion
Sometimes it is useful to access the information for the current execution
context. To do so you can provide the `info` parameter to resolvers, like this:
```

docs/types/resolvers.md
```diff
@@ -4,4 +4,205 @@ title: Resolvers
...
+`Info[ContextType, RootValueType]`
+
+| Parameter name  | Type                       | Description                                                           |
+| --------------- | -------------------------- | --------------------------------------------------------------------- |
+| field_name      | `str`                      | The name of the current field                                         |
+| context         | `ContextType`              | The value of the context                                              |
+| root_value      | `RootValueType`            | The value for the root type                                           |
+| variable_values | `Optional[Dict[str, Any]]` | The variables for this operation                                      |
+| operation       | `OperationDefinitionNode`  | The ast for the current operation (public API might change in future) |
+| path            | `Path`                     | The path for the current field                                        |
```
BryceBeagle, line 203: I'm a bit confused on what this is the API for? Where do these parameters reside?

strawberry/resolvers.py
```diff
@@ -78,7 +80,8 @@ def get_result_for_field(
 
 
 def get_resolver(field: FieldDefinition) -> Callable:
-    def _check_permissions(source, info, **kwargs):
+    # TODO: convert info to our info
+    def _check_permissions(source, info: Info, **kwargs):
```
BryceBeagle, line 84: What does this TODO mean?

docs/types/resolvers.md
BryceBeagle: After looking at the other files, I see that `Info` is generic of types `ContextType` and `RootValueType`. I don't think this syntax here makes that clear

BryceBeagle (changes requested):
Great documentation! I left a few minor suggestions but otherwise looks good.

docs/types/resolvers.md
patrick91: do you have any suggestion, on to make it look nicer?

this is how it looks now: 
image

docs/types/resolvers.md
```diff
@@ -4,4 +4,205 @@ title: Resolvers
...
+@strawberry.type
+class User:
+    name: str
+
+
+@strawberry.type
+class Query:
+    @strawberry.field
+    def user(self, id: strawberry.ID) -> Optional[User]:
+        return User(name="Marco")
```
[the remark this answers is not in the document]
patrick91: I've added a comment for that, the other option is to add an id field to the user and return the passed id, but I'm not sure it makes a lot of sense

strawberry/resolvers.py
patrick91: the info that it is passed here comes from graphql-core, I was thinking we might need to convert it when and if our version starts differing or even now (it might be a bit weird to inspect the info parameter and see that it is a different type)

BryceBeagle: Let's open an issue and link it here :+1:

patrick91: done!

docs/types/resolvers.md
BryceBeagle: I think it might just be best to make it clear that Info is Generic (I don't think just the `[]`s will make that clear) and that those parameters are for it's `__init__` method

As far as the first is concerned maybe just use
```python
class Info(Generic[ContextType, RootValueType]): ...
```

patrick91: yes! similar to the python docs: 
image

image

BryceBeagle: A difference is that in the Python docs, the `__init__` argslist is in the parens, but here we have the base/generic classes. I think this is probably fine for now, though

BryceBeagle (approved):

PR #723, opened 2021-02-25
Enable .(dot) notation in the get_context of the django views.

ernestoarbitrio:
## Description
Currently the request coming from the `context` object within the `get_context`  in `views.py` in the `django` module, is a dict which can be accessed via standard python syntax. This PR adds the dot notation to the `request` object.
```python3
>>> context = {"request": "hi"}
>>> print(context.request)
hi
>>> print(context["request"]
hi
```
__Note: this works with nested dict too__

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [X] Enhancement/optimization
- [x] Documentation

## Checklist

- [X] My code follows the code style of this project.
- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [X] I have read the CONTRIBUTING document.
- [X] I have added tests to cover my changes.
- [X] I have tested the changes and verified that they work and don't break anything (as well as I can manage).

PR #724, opened 2021-02-25
Sanic Webserver Support

### LINKED ISSUE #541: Add support for sanic
Sanic seems to be quite popular for performance reasons, might be nice to add built-in support for it, should be similar to asgi :)

https://github.com/huge-success/sanic

shinybrar:
Sanic Webserver Support

## Description
Added very basic support for sanic web server based on the current implementation of flask support.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [x] New feature
- [ ] Documentation

Closes #541 .

## Checklist


BryceBeagle:
A lot of the changes here are formatting-only changes. While not inherently opposed to such changes (I certainly like good formatting), I wonder if we should move them to another PR?

shinybrar:
@BryceBeagle Sure, we can move the formatting changes to a different PR and only add the sanic webserver changes here. I will update the PR soon.

BryceBeagle:
Were all these changes done automatically by the pre-commit hooks? If so, I wonder why they weren't performed in prior commits/PRs :thinking: .

shinybrar:
@BryceBeagle There was a global config in my pre-commit for line length which caused a lot of formatting changes. All those have now been reverted.

docs/integrations/sanic.md
```diff
@@ -0,0 +1,113 @@
...
+
+The `GraphQLView` accepts two options at the moment:
+
+- schema: mandatory, the schema created by `strawberry.Schema`.
+- graphiql: optional, defaults to `True`, whether to enable the GraphiQL
+  interface.
+
+## Extending the view
+
+We allow to extend the base `GraphQLView`, by overriding the following methods:
```
BryceBeagle, line 33:
```suggestion
The base `GraphQLView` class can be extended by overriding the following methods:
```

tests/sanic/test_view.py
```diff
@@ -0,0 +1,117 @@
+from random import random
+
+import pytest
+
+import strawberry
+from sanic import Sanic
+from strawberry.sanic.views import GraphQLView as BaseGraphQLView
+from strawberry.types import ExecutionResult, Info
```
BryceBeagle, line 8: Was isort run here? If so, we may need to inspect our configuration, as the strawberry imports are supposed to be separated from the 3rd party ones.

tests/sanic/test_view.py
```diff
@@ -0,0 +1,117 @@
...
+    class Query:
+        hello: str = "strawberry"
+
+    schema = strawberry.Schema(query=Query)
+
+    class GraphQLView(BaseGraphQLView):
+        def get_root_value(self):
+            return Query()
+
+    app = Sanic("test-app-" + str(random()))
```
BryceBeagle, line 22: In the other tests, you hardcode the App name. Should this be made consistent, or did you need  the random name for a reason?

strawberry/sanic/graphiql.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def render_graphiql_page():
+    dir_path = abspath(join(dirname(__file__), ".."))
+    graphiql_html_file = f"{dir_path}/static/graphiql.html"
+
+    html_string = None
+
+    with open(graphiql_html_file, "r") as f:
+        html_string = f.read()
```
BryceBeagle, line 11: Is this .html file supposed to be committed? I'm not personally familiar with the GraphiQL system. At minimum, this may benefit from a bit of comments/docstring?

Also, I would use `Pathlib` instead of `os.path`. This project doesn't need to support old Python versions. [`Path.read_text()`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text) is really convenient.

docs/integrations/sanic.md
```diff
@@ -0,0 +1,113 @@
...
+
+- `get_context(self) -> Any`
+- `get_root_value(self) -> Any`
+- `process_result(self, result: ExecutionResult) -> GraphQLHTTPResponse`
+
+## get_context
+
+`get_context` allows to provide a custom context object that can be used in your
+resolver. You can return anything here, by default we return a dictionary with
+the request.
```
BryceBeagle, line 43:
```suggestion
By overriding `GraphQLView.get_context` you can provide a custom context object for
your resolvers. You can return anything here; by default GraphQLView returns a
dictionary with the request.
```

docs/integrations/sanic.md
```diff
@@ -0,0 +1,113 @@
...
+
+@strawberry.type
+class Query:
+    @strawberry.field
+    def example(self, info: Info) -> str:
+        return str(info.context["example"])
+```
+
+Here we are returning a custom context dictionary that contains only one item
+called "example".
```
BryceBeagle, line 59:
```suggestion
Here we are returning a custom context dictionary that contains only one item
called `"example"`.
```

docs/integrations/sanic.md
```diff
@@ -0,0 +1,113 @@
...
+
+Then we use the context in a resolver, the resolver will return "1" in this
+case.
```
BryceBeagle, line 62:
```suggestion
Then we can use the context in a resolver. In this case the resolver will return `1`.
```

BryceBeagle: I would also rearrange these code snippets to be a bit more interspersed with the descriptions

docs/integrations/sanic.md
```diff
@@ -0,0 +1,113 @@
...
+
+## get_root_value
+
+`get_root_value` allows to provide a custom root value for your schema, this is
+probably not used a lot but it might be useful in certain situations.
```
BryceBeagle, line 67:
```suggestion
By overriding `GraphQLView.get_root_value` you can provide a custom root value for your
schema. This is probably not used a lot but it might be useful in certain situations.
```

docs/integrations/sanic.md
```diff
@@ -0,0 +1,113 @@
...
+        return Query(name="Patrick")
+
+
+@strawberry.type
+class Query:
+    name: str
+```
+
+Here we are returning a Query where the name is "Patrick", so we when requesting
+the field name we'll return "Patrick" in this case.
```
BryceBeagle, line 83:
```suggestion
Here we configure a Query where requesting the `name` field will return `"Patrick"`
through the custom root value.
```

docs/integrations/sanic.md
```diff
@@ -0,0 +1,113 @@
...
+)
+```
+
+## Options
+
+The `GraphQLView` accepts two options at the moment:
+
+- schema: mandatory, the schema created by `strawberry.Schema`.
+- graphiql: optional, defaults to `True`, whether to enable the GraphiQL
+  interface.
```
BryceBeagle, line 29:
```suggestion
- `schema`: mandatory, the schema created by `strawberry.Schema`.
- `graphiql`: optional, defaults to `True`, whether to enable the GraphiQL
  interface.
```

docs/integrations/sanic.md
```diff
@@ -0,0 +1,113 @@
...
+```
+
+Here we are returning a Query where the name is "Patrick", so we when requesting
+the field name we'll return "Patrick" in this case.
+
+## process_result
+
+`process_result` allows to customize and/or process results before they are sent
+to the clients. This can be useful logging errors or hiding them (for example to
+hide internal exceptions).
```
BryceBeagle, line 89:
```suggestion
By overriding `GraphQLView.process_result` you can customize and/or process results
before they are sent to a client. This can be useful for logging errors, or even hiding
them (for example to hide internal exceptions).
```

tests/sanic/test_view.py
BryceBeagle: This applies to `strawberry/sanic/views.py` as well

strawberry/sanic/views.py
```diff
@@ -0,0 +1,79 @@
...
+from sanic.response import HTTPResponse, html
+from sanic.views import HTTPMethodView
+from strawberry.http import GraphQLHTTPResponse, process_result
+from strawberry.types import ExecutionResult
+
+from ..schema import BaseSchema
+from .graphiql import render_graphiql_page
+
+
+class GraphQLView(HTTPMethodView):
```
BryceBeagle, line 13: As this is a class where users are expected to create subclasses from, we should provide some docstring on what each function does.

BryceBeagle (changes requested):
This looks great! 

I added some requests for a couple minor changes, and wording improvements. Let me know if you have any questions/concerns :+1:.

docs/integrations/sanic.md
```diff
@@ -0,0 +1,113 @@
+---
+title: Sanic
+---
+
+# Sanic
+
+Strawberry comes with a basic Sanic integration. It provides a view that you can
+use to serve your GraphQL schema:
```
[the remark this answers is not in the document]
shinybrar: Done.

tests/sanic/test_view.py
shinybrar: Yes, pre-commit was run and it actually moves the `sanic` import after `import strawberry`.

tests/sanic/test_view.py
shinybrar: In this context, hardcoded app name is required so that we can run multiple iterations of the test in parallel using a single test rig.

strawberry/sanic/graphiql.py
shinybrar: The `html` file is actually a part of the strawberry project itself, and was not added in this PR. We are just using the established code practice in `strawberry/flask/graphiql.py` with some minor `sanic` changes.

strawberry/sanic/views.py
shinybrar: Agreed, added a doc-string

shinybrar:
@BryceBeagle We agree with all the documentation improvement changes, and implemented or commented on all the other code changes.

strawberry/sanic/views.py
BryceBeagle: On the methods too :)

strawberry/sanic/graphiql.py
BryceBeagle: Haha, I see: https://github.com/strawberry-graphql/strawberry/search?q=graphiql.html

There are already 4 different ways were accessing this file. I wonder if we should abstract it out (in another PR).

RELEASE.md
```diff
@@ -0,0 +1,3 @@
+Release type: patch
```
```suggestion
Release type: minor
```

😊

docs/README.md
```diff
@@ -45,6 +45,7 @@
 - [ASGI](./integrations/asgi.md)
 - [Django](./integrations/django.md)
 - [Flask](./integrations/flask.md)
+- [Sanic](./integrations/sanic.md)
```
patrick91, line 48: sweet!

strawberry/sanic/views.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def __init__(self, schema: BaseSchema, graphiql: bool = True):
+        self.graphiql = graphiql
+        self.schema = schema
+
+    def get_root_value(self):
+        return None
+
+    def get_context(self, request):
+        return {"request": request}
```
patrick91, line 41: can we return a class like we did here? https://github.com/strawberry-graphql/strawberry/pull/723

I think it's better to have a dataclass for this :)

strawberry/sanic/views.py
```diff
@@ -0,0 +1,96 @@
...
+        )
+
+        response_data = self.process_result(result)
+
+        return HTTPResponse(
+            json.dumps(response_data), status=200, content_type="application/json"
+        )
+
+    def should_display_graphiql(self, request):
+        if not self.graphiql or "raw" in request.args:
```
patrick91, line 89: when do you use `raw`?

tests/sanic/test_view.py
patrick91: this is so strange :D I'll see if we can improve it, for now let's leave it as it is :)

patrick91 (commented):
this looks *great*! I've added a few comments, nothing major though :)

Do you think we could add support for upload files with sanic? we could also do that in another PR if it is better :)

strawberry/sanic/views.py
```diff
@@ -0,0 +1,96 @@
...
+
+        data = request.json
+
+        try:
+            query = data["query"]
+            variables = data.get("variables")
+            operation_name = data.get("operationName")
+
+        except KeyError:
+            return HTTPResponse("No valid query was provided for the request", 400)
```
BryceBeagle, line 70: Can we add a test case on the `KeyError` here for coverage? It also seems as if the try/except should be around just the dict `[]` access and not the `.get` calls

BryceBeagle (changes requested):

strawberry/sanic/views.py
shinybrar: Agreed, the changes have been implemented in commit d1e0ced

strawberry/sanic/views.py
shinybrar: It was used for some dev tests, but it has been removed from production code.

strawberry/sanic/views.py
shinybrar: Implemented in  commit d1e0ced

shinybrar:
Hmm, I changed the following method call
```python 
schema.execute_sync()
````
to 
```
schema.execute()
```
to support sanic supported `async/awaitable` requests. But it leads to the following error in `mypy`

```python
strawberry/sanic/views.py:82:
error: Argument 1 to "process_result" of "GraphQLView" has incompatible type "strawberry.schema.base.ExecutionResult"; expected "strawberry.types.execution.ExecutionResult"
```

Essentially, its telling me, `schema.execute()` is returning `strawberry.schema.base.ExecutionResult` when I cannot even find it!!! Any help here would be much appreciated.

BryceBeagle:
It looks like Strawberry has two identical types named `ExecutionResult`:
https://github.com/strawberry-graphql/strawberry/blob/16dd965c06a574c257161864ee18385c719378b6/strawberry/schema/base.py#L15-L19 
and
https://github.com/strawberry-graphql/strawberry/blob/16dd965c06a574c257161864ee18385c719378b6/strawberry/types/execution.py#L15-L19

I'm guessing this is the cause of this problem. @patrick91  why do we have both types?

patrick91:
> It looks like Strawberry has two identical types named `ExecutionResult`:
> https://github.com/strawberry-graphql/strawberry/blob/16dd965c06a574c257161864ee18385c719378b6/strawberry/schema/base.py#L15-L19
> 
> 
> and
> https://github.com/strawberry-graphql/strawberry/blob/16dd965c06a574c257161864ee18385c719378b6/strawberry/types/execution.py#L15-L19
> 
> I'm guessing this is the cause of this problem. @patrick91 why do we have both types?

woops, this is my mistake, we can delete one of them :) I'd keep the one in types/execution.py :)

shinybrar:
> > It looks like Strawberry has two identical types named `ExecutionResult`:
> > https://github.com/strawberry-graphql/strawberry/blob/16dd965c06a574c257161864ee18385c719378b6/strawberry/schema/base.py#L15-L19
> > 
> > and
> > https://github.com/strawberry-graphql/strawberry/blob/16dd965c06a574c257161864ee18385c719378b6/strawberry/types/execution.py#L15-L19
> > 
> > I'm guessing this is the cause of this problem. @patrick91 why do we have both types?
> 
> woops, this is my mistake, we can delete one of them :) I'd keep the one in types/execution.py :)

Okay, I will try and do that in this PR, hopefully it works.

BryceBeagle:
> woops, this is my mistake, we can delete one of them :) I'd keep the one in types/execution.py :)

> Okay, I will try and do that in this PR, hopefully it works.


Make sure to add it to the changelog as this is technically a breaking change :+1:

shinybrar:
The excision of `from strawberry.schema.base.ExecutionResult` turned out to be easier than expected. I added the deprecation to the changelog as well.

The PR should be ready for merge now.

RELEASE.md
```diff
@@ -0,0 +1,16 @@
+Release type: minor
+
+# New Features
+Added support for sanic webserver.
+
+# Changelog
+Deprecated the following type annotation,
+```python
+from strawberry.schema.base import ExecutionResult # Deprecated
+```
```
BryceBeagle, line 10:
```suggestion
`ExecutionResult` was erroneously defined twice in the repository. The entry in `strawberry.schema.base` has been removed. If you were using it, switch to using
`strawberry.types.execution.ExecutionResult` instead.
```

BryceBeagle: It can't be deprecated if it's removed 🙂

BryceBeagle (changes requested):

RELEASE.md
shinybrar: Haha, fair

BryceBeagle (approved):

botberry:



---

# New Features
Added support for sanic webserver.

# Changelog
`ExecutionResult` was erroneously defined twice in the repository. The entry in `strawberry.schema.base` has been removed. If you were using it, switch to using
`strawberry.types.execution.ExecutionResult` instead.

It is now instead only supported under `strawberry.types`.

```python
from strawberry.types import ExecutionResult
```

patrick91 (approved):
@shinybrar this is amazing work, thanks for it and for handling all our comments!

I'll merge this now so it gets released in a few minutes 😊 

Oh, if you want a sticker feel free to fill this form https://forms.gle/9YxbHGXZGoymyzhSA :D

We also have a discord server if you want to chat: https://discord.com/invite/3uQ2PaY

RELEASE.md
```diff
@@ -0,0 +1,14 @@
...
+
+# Changelog
+`ExecutionResult` was erroneously defined twice in the repository. The entry in `strawberry.schema.base` has been removed. If you were using it, switch to using
+`strawberry.types.execution.ExecutionResult` instead.
+
+It is now instead only supported under `strawberry.types`.
+
+```python
+from strawberry.types import ExecutionResult
+```
```
patrick91, line 14:
```suggestion
`ExecutionResult` was erroneously defined twice in the repository. The entry in `strawberry.schema.base` has been removed. If you were using it, switch to using
`strawberry.types.ExecutionResult` instead:

```python
from strawberry.types import ExecutionResult
```
```

PR #728, opened 2021-02-28
Use custom json encoder to prevent encoding error

### LINKED ISSUE #714: UploadFile causes ASGI application's exception when it is in debug mode
Apparently when we use the ASGI application in debug mode, it cannot print the value of variables if they are not JSON serializable.

In my use case, when I tried to use the [file upload](https://strawberry.rocks/docs/features/file-upload) example in debug mode this issue ended up happening.

I believe it is because of this:

https://github.com/strawberry-graphql/strawberry/blob/de215370b247a417af8a8dd5fc382d71e305bcd7/strawberry/utils/debug.py#L26-L29

Perhaps converting variables to string might help.

patrick91:
## Description

This PR fixes an issue when using debug=True and uploading files.

We should probably change the prints to be logs, but we can do that in future :)

Closes #714

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/strawberry/utils/debug.py
+++ b/strawberry/utils/debug.py
@@ -1,6 +1,7 @@
 import datetime
 import json
-import typing
+from json import JSONEncoder
+from typing import Any, Dict
 
 from pygments import highlight, lexers
 from pygments.formatters import Terminal256Formatter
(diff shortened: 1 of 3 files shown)


botberry:



---

This release fixes a bug when using the debug server and upload a file

strawberry/utils/debug.py
```diff
@@ -1,13 +1,20 @@
 import datetime
 import json
 import typing
+from json import JSONEncoder
+from typing import Any
```
BryceBeagle, line 5: This file now uses both `Any` and `typing.Any`. Probably want to make that consistent

BryceBeagle (changes requested):

BryceBeagle (approved):

PR #755, opened 2021-03-12
Move get_http_response to asgi class to make it easier to override it

patrick91:
## Description

I need to override graphiql for an internal service but without this PR the changes is way too invasive.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation


botberry:

```markdown
Release type: patch

```

RELEASE.md
```diff
@@ -0,0 +1,3 @@
+Release type: patch
```
BryceBeagle, line 1: Minor? Do we expect anyone to have been using this function?

patrick91: no, I don't think anyone was using it :)

patrick91: I'm going to do minor just in case :D

BryceBeagle: Might want to make the change clear in the RELEASE.md too? (i.e. "moved x from y to z")

patrick91: will do :)

RELEASE.md
```diff
@@ -0,0 +1,3 @@
+Release type: minor
+
+This releases updates the ASGI class to make it easier to override `get_http_response`.
```
```suggestion
This releases updates the ASGI class to make it easier to override `get_http_response`.

`get_http_response` has been now removed from strawberry.asgi.http and been moved to be
a method on the ASGI class. 

A new `get_graphiql_response` method has been added to make it easier to provide a different GraphiQL interface.
```

BryceBeagle: Looks good to me 👍

BryceBeagle (approved):

PR #771, opened 2021-03-17
Combine sync and async resolvers into single function

la4de:
After this the resolver can be sync or async function and type is
detected at runtime and not at schema conversion time as before.

This is internal change and not affecting to API.

GraphQL library uses the same logic here: https://github.com/graphql-python/graphql-core/blob/main/src/graphql/execution/execute.py#L640


botberry:

```markdown
Release type: patch

```

patrick91 (approved):
This look great to me! Can you add a release file so it gets released too? 😊

PR #750, opened 2021-03-10
Add optional response to ASGI app context

mecampbellsoup:
Resolves https://github.com/strawberry-graphql/strawberry/issues/748.

## Description

HTTP requests to the strawberry GraphQL ASGI application server can now pre-set HTTP _response_ headers, useful for setting HTTP response headers e.g. `Set-Cookie: my-login-cookie-token-thing`, so that auth-related operations can be handled via GraphQL mutations.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Enhancement/optimization
- [ ] Documentation

## Issues Fixed or Closed by This PR

*

## Checklist

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the CONTRIBUTING document.
- [ ] I have added tests to cover my changes.
- [ ] I have tested the changes and verified that they work and don't break anything (as well as I can manage).

PR #773, opened 2021-03-18
Add temporal response in django's get context

patrick91:
Similar to #750 but for django 😊

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,14 @@
+Release type: minor
+
+This release updates get_context in the django integration to also receive a temporal response object that can be used to set headers, cookies and status code.
+
+
+```
+@strawberry.type
+class Query:
+    @strawberry.field
+    def abc(self, info: Info) -> str:
+        info.context.response.status_code = 418
+
+        return "ABC"
+```
--- a/strawberry/django/views.py
+++ b/strawberry/django/views.py
@@ -5,6 +5,7 @@ from typing import Any, Dict, Optional
 
 from django.core.exceptions import SuspiciousOperation
 from django.http import Http404, HttpRequest, HttpResponseNotAllowed, JsonResponse
+from django.http.response import HttpResponse
 from django.template import RequestContext, Template
 from django.template.exceptions import TemplateDoesNotExist
 from django.template.loader import render_to_string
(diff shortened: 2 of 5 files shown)


botberry:



---

This release updates get_context in the django integration to also receive a temporal response object that can be used to set headers, cookies and status code.


```
@strawberry.type
class Query:
    @strawberry.field
    def abc(self, info: Info) -> str:
        info.context.response.status_code = 418

        return "ABC"
```

RELEASE.md
```diff
@@ -0,0 +1,14 @@
+Release type: minor
```
alex-verve, line 1: Won't it be braking change for people who already inherit from  `GraphQLView` and expecting only one required argument in  `.get_context(request)` method?

strawberry/django/views.py
```diff
@@ -87,13 +88,31 @@ def _render_graphiql(self, request: HttpRequest, context=None):
...
+        self, response_data: GraphQLHTTPResponse, sub_response: HttpResponse
+    ) -> JsonResponse:
+        response = JsonResponse(response_data)
+        response._headers.update(sub_response._headers)
+
+        # unfortunately setting None as status code doesn't work
+        # so to make sure we only override status codes when they have been
+        # changed we use a status code that doesn't actually exist
+        # https://github.com/django/django/blob/03d0f12/django/http/response.py#L117-L124
+        if sub_response.status_code != 199:
```
alex-verve, line 101: Do you think if it's worth using some private constant for this integer?

alex-verve (approved):
🥇

RELEASE.md
patrick91: yes, it is breaking 😊 we are still pre 1.0 so it's fine to just bump to minor :)

strawberry/django/views.py
patrick91: that's a good suggestion, we could also:

1. have a subclass of httpresponse that allows to set status_code to none (or uses None as default)
2. set status_code as none after creating the instance

let me try 1

patrick91: @alex-verve 1 worked well 😊

AGASS007:
there's a new API for headers in 3.2

https://github.com/django/django/pull/13186

I guess we'll have to either differentiate based on django version or catch an `AttributeError`.

patrick91:
@AGASS007 thanks for pointing that out, I guess the current way will work for a while, right?

AGASS007:
hum, django 3.2 will be released in 2 weeks (rc1 is out today)

strawberry/django/views.py
```diff
@@ -23,6 +24,10 @@
 from .context import StrawberryDjangoContext
 
 
+class TemporalHttpResponse(HttpResponse):
```
alex-verve, line 27: I just realized it's better to inherit from `JsonResponse` here.
As `Content-Type` header will be overwritten  from json to `text/html` in line 99 below

patrick91: good spot! sending a PR now

blueyed:
Note that `status_code = None` makes Django's `HttpResponse.__repr__` crash: https://github.com/django/django/pull/15768 (just for info/reference, but maybe `__repr__` needs to (or even should?) be overridden/adopted here).

PR #777, opened 2021-03-19
Django 3.2 Support

AGASS007:
## Description

- added Django 3.2 in test matrix
- updated headers handling, accounting for new public headers API in django 3.2

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* As discussed here : https://github.com/strawberry-graphql/strawberry/pull/773

## Checklist

### DIFF
--- a/strawberry/django/views.py
+++ b/strawberry/django/views.py
@@ -99,7 +99,9 @@ class BaseView(View):
         self, response_data: GraphQLHTTPResponse, sub_response: HttpResponse
     ) -> JsonResponse:
         response = JsonResponse(response_data)
-        response._headers.update(sub_response._headers)
+
+        for name, value in sub_response.items():
+            response[name] = value
 
         if sub_response.status_code is not None:
             response.status_code = sub_response.status_code
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         python-version: ['3.7', '3.8', '3.9']
-        django: [3.1.4, 3.0.11, 2.2.17]
+        django: [3.2rc1, 3.1.7, 3.0.13, 2.2.19]
 
     steps:
       - uses: actions/checkout@v2
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+Django 3.2 support


botberry:

```markdown
Release type: patch

```

patrick91:
@AGASS007 thanks for doing this!

strawberry/django/views.py
```diff
@@ -99,7 +100,12 @@ def _create_response(
         self, response_data: GraphQLHTTPResponse, sub_response: HttpResponse
     ) -> JsonResponse:
         response = JsonResponse(response_data)
-        response._headers.update(sub_response._headers)
+
+        if django.VERSION >= (3, 2):
+            for name, value in sub_response.headers.items():
+                response.headers[name] = value
+        else:
+            response._headers.update(sub_response._headers)
```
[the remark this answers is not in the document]
patrick91: I'd move this to a function, what do you think?

patrick91 (commented):
this looks good to me, I added one comment :)

Shall we also start testing on django 3.2? https://github.com/strawberry-graphql/strawberry/blob/main/.github/workflows/test.yml#L13

strawberry/django/views.py
```diff
@@ -99,7 +100,12 @@ def _create_response(
         self, response_data: GraphQLHTTPResponse, sub_response: HttpResponse
     ) -> JsonResponse:
         response = JsonResponse(response_data)
-        response._headers.update(sub_response._headers)
+
+        if django.VERSION >= (3, 2):
+            for name, value in sub_response.headers.items():
+                response.headers[name] = value
```
jkimbo, line 106:
```suggestion
            for name, value in sub_response.headers.items():
                response[name] = value
```

We could always do this right? Regardless of the Django version.

AGASS007: Indeed!

See:
https://github.com/django/django/blob/3.1.7/django/http/response.py#L135
https://github.com/django/django/blob/3.2rc1/django/http/response.py#L72
https://github.com/django/django/blob/3.2rc1/django/http/response.py#L171

strawberry/django/views.py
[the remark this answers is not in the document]
AGASS007: went with @jkimbo's suggestion

AGASS007:
I added 3.2rc1 to the test matrix (didn't know where to look for at first).

AGASS007:
Mmmmmpf I went too fast...

strawberry/django/views.py
jkimbo: Sorry this won't work actually because `sub_response.headers` is not available pre 3.2. 

It also looks like the `.items` property (which is what I would expect to be able to use) has actually changed in 3.2 which I think is a breaking change. Not sure if thats ok for Django to do that...
https://github.com/django/django/blob/3.1.7/django/http/response.py#L152-L153
https://github.com/django/django/blob/3.2rc1/django/http/response.py#L186-L187

jkimbo: Anyway you're original suggestion is probably the best way of doing this now @AGASS007 . Sorry for the churn!

patrick91: > Sorry this won't work actually because `sub_response.headers` is not available pre 3.2.
> 
> It also looks like the `.items` property (which is what I would expect to be able to use) has actually changed in 3.2 which I think is a breaking change. Not sure if thats ok for Django to do that...
> [https://github.com/django/django/blob/3.1.7/django/http/response.py#L152-L153](https://github.com/django/django/blob/3.1.7/django/http/response.py?rgh-link-date=2021-03-19T11%3A47%3A18Z#L152-L153)
> [https://github.com/django/django/blob/3.2rc1/django/http/response.py#L186-L187](https://github.com/django/django/blob/3.2rc1/django/http/response.py?rgh-link-date=2021-03-19T11%3A47%3A18Z#L186-L187)

We could ask about this on the developer mailing list

AGASS007: Grrrrr `items()` on `HttpResponse` returns headers `items()`, but used to return `values()`

https://github.com/django/django/blob/3.1.7/django/http/response.py#L152
https://github.com/django/django/blob/3.2rc1/django/http/response.py#L186

AGASS007: Nevermind, for versions < 3.2, `values()` returns an iterable of `(key, value)` so the output of `HttpResponse.items()` is consistant across versions.

jkimbo (approved):
👍

patrick91 (approved):
Thanks for doing this @AGASS007 😊 Will merge now 😊

Fill this form if you want a sticker! https://forms.gle/dmnfQUPoY5gZbVT67

Issue #756, opened 2021-03-12
Crash on empty body


lijok:
If an empty body is sent to the server, it crashes with a json decode error on https://github.com/strawberry-graphql/strawberry/blob/master/strawberry/django/views.py#L43

should probably raise something akin to
https://github.com/strawberry-graphql/strawberry/blob/master/strawberry/django/views.py#L59

PR #781, opened 2021-03-21
[ Django ] Raise exception when JSON parsing fails within the parse_body method

kristofgilicze:
## Description

Raising exception when invalid, non parsable JSON provided to the Django integration.

Added a catch block for the parse_body method, catching the parsing exceptions and raised:
```python
SuspiciousOperation("Unable to parse request body as JSON")
```

## Types of Changes

- [ ] Core
- [X] Bugfix
- [ ] New feature
- [ ] Documentation

* #756 

## Checklist

### DIFF
--- a/tests/django/test_views.py
+++ b/tests/django/test_views.py
@@ -96,7 +96,7 @@ def test_fails_when_not_sending_query():
     with pytest.raises(SuspiciousOperation) as e:
         GraphQLView.as_view(schema=schema, graphiql=False)(request)
 
-        assert e.value.args == ("No GraphQL query found in the request",)
+    assert e.value.args == ("No GraphQL query found in the request",)
 
 
 def test_fails_when_request_body_has_invalid_json():
--- a/tests/django/test_views.py
+++ b/tests/django/test_views.py
@@ -99,6 +99,19 @@ def test_fails_when_not_sending_query():
         assert e.value.args == ("No GraphQL query found in the request",)
 
 
+def test_fails_when_request_body_has_invalid_json():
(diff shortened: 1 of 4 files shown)


tests/django/test_views.py
```diff
@@ -99,6 +99,19 @@ def test_fails_when_not_sending_query():
...
+    factory = RequestFactory()
+
+    request = factory.post(
+        "/graphql/", "definitely-not-json-string", content_type="application/json"
+    )
+
+    with pytest.raises(SuspiciousOperation) as e:
+        GraphQLView.as_view(schema=schema, graphiql=False)(request)
+
+    assert e.value.args == ("Unable to parse request body as JSON",)
```
kristofgilicze, line 112: Here with the tests I discovered the following behaviour, which I am fairly sure is not the intention.

### assert in the context ( as above on line 99 ):
The test **only checks whether the exception is raised or not**, the assertion of the exception message is ignored.

```python
with pytest.raises(SuspiciousOperation) as e:
    GraphQLView.as_view(schema=schema, graphiql=False)(request)
    assert e.value.args == ("blablabla",)  # test passes no matter what the string is
```


### If I move the assert outside the context, then it works as expected.

botberry:



---

Raise exception when un-serializable payload is provided to the Django view.

patrick91 (approved):
Nice one @kristofgilicze! I've pushed a fix for the tests, good spot! I'll merge this now 😊

Please fill this form if you want a sticker: https://forms.gle/dmnfQUPoY5gZbVT67 😊

PR #787, opened 2021-03-23
Include `field_nodes` in Strawberry info object.

coady:
## Description

The new Info wrapper omitted the `field_nodes`.

## Types of Changes

- [x] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+Include `field_nodes` in Strawberry info object.
--- a/strawberry/resolvers.py
+++ b/strawberry/resolvers.py
@@ -126,6 +126,7 @@ def strawberry_info_from_graphql(
 ) -> Info:
     return Info(
         field_name=info.field_name,
+        field_nodes=info.field_nodes,
         context=info.context,
         root_value=info.root_value,
         variable_values=info.variable_values,
--- a/strawberry/types/info.py
+++ b/strawberry/types/info.py
@@ -1,7 +1,8 @@
 import dataclasses
-from typing import Any, Dict, Generic, Optional, Type, TypeVar, Union
+from typing import Any, Dict, Generic, List, Optional, Type, TypeVar, Union
 
 from graphql import OperationDefinitionNode
+from graphql.language import FieldNode
 from graphql.pyutils.path import Path
 
 from strawberry.union import StrawberryUnion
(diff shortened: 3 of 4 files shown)


botberry:

```markdown
Release type: patch

```

patrick91:
@coady thanks for this, what do you use field nodes for? I'd like to see if we can abstract it from what graphql-core provides :)

coady:
> @coady thanks for this, what do you use field nodes for? I'd like to see if we can abstract it from what graphql-core provides :)

My project uses most of it, to determine [projected fields](https://github.com/coady/graphique/blob/e4f55a4df50318a6e2487984ff535b031fb18572/graphique/service.py#L119) for a data store.

But if the goal to move away from `graphql-core`, I think this will occur generally. The `GraphQLResolveInfo` object is considered part of the GraphQL spec. `OperationDefinitionNode` is also still `graphql-core`.

patrick91:
> > @coady thanks for this, what do you use field nodes for? I'd like to see if we can abstract it from what graphql-core provides :)
> 
> My project uses most of it, to determine [projected fields](https://github.com/coady/graphique/blob/e4f55a4df50318a6e2487984ff535b031fb18572/graphique/service.py#L119) for a data store.

Nice, do you think we can find an abstraction that makes it easier to use it for you?

> But if the goal to move away from `graphql-core`, I think this will occur generally. The `GraphQLResolveInfo` object is considered part of the GraphQL spec. 

Where did you see this?

coady:
> Nice, do you think we can find an abstraction that makes it easier to use it for you?

Sure, but it would need to replicate the recursive [Field](https://spec.graphql.org/June2018/#sec-Language.Fields) object.

> > But if the goal to move away from `graphql-core`, I think this will occur generally. The `GraphQLResolveInfo` object is considered part of the GraphQL spec.
> 
> Where did you see this?

I was referring to the [resolver docs](https://graphql.org/learn/execution/#root-fields-resolvers) and the [reference implementation](https://graphql.org/graphql-js/type/#graphqlobjecttype) linked there. I didn't mean to imply that `graphql-core` followed a struct exactly, just that the same data (including fields) should be there.

patrick91:
@coady I think if we can make this field_nodes better we should 😊 if this is breaking your project we can merge and improve it in future (I think we can even change the name `selected_fields` might make more sense).

What do you think?

coady:
> @coady I think if we can make this field_nodes better we should 😊 if this is breaking your project we can merge and improve it in future (I think we can even change the name `selected_fields` might make more sense).
> 
> What do you think?

Sounds great (both). Thanks.

patrick91:
@coady I'll merge this for now, hopefully I'll have some time in future to find an abstraction 😊

patrick91 (approved):

Issue #702, opened 2021-02-04
Add IDE support for strawberry.Private


lijok:
My apologies, not knowing the internals of pythons type hinting system, I can't articulate this issue well
Is there something that can be done to strawberry.Private to prevent this from happening?

![image](https://user-images.githubusercontent.com/28689084/106826807-90bb3600-667f-11eb-8e6a-a1e84e64d382.png)

BryceBeagle:
As discussed on Discord, `Private.__class_getitem__` is currently implemented like this:
```python
class Private:
    def __class_getitem(cls, item) -> Private:
        ...
```

PyCharm, unsurprisingly, does not understand that Strawberry change the field type behind the scenes. 

I mentioned that something along the lines of 
```python
class Private:
    def __class_getitem(cls, item: T) -> T:
        ...
```
is probably necessary to convince a type system of what's going on. This might involve lying to the type system and still returning a `Private` instance after annotating it as `T`, or actually changing the method to return `T`.

lijok:
Did another deep dive into this today, seems like this is not possible
https://stackoverflow.com/questions/48722835/custom-type-hint-annotation#comment84446327_48722879

BryceBeagle:
Thanks for looking into it some more. 

As a PyCharm devotee, I'm very interested in making all the typing work properly at the IDE level. Do you have any potential alternatives to the current implementation that would allow us to mark a field as to be excluded from the Schema?

lijok:
Can't think of anything that's as clean and intuitive as strawberry.Private
A less clean solution could be to add another field to `strawberry.type` i.e `private_fields`

```py
@strawberry.type(private_fields=["queryset"])
class Integration(Node):
    queryset: models.Integration
```

patrick91:
+1 for making it just work with PyCharm :)

The other option would be to have `field_name: str = strawberry.field(private=True)`

but let’s see if we can make it work with PyCharm. I’ll see if I can reach out someone from PyCharm 😊

jkimbo:
If we can make this work I think we should go with:

`field_name: str = strawberry.private()`

(No need to make it an argument to .field because it wouldn't make sense to pass any of the other arguments to it)

lijok:
> +1 for making it just work with PyCharm :)
> 
> The other option would be to have `field_name: str = strawberry.field(private=True)`
> 
> but let’s see if we can make it work with PyCharm. I’ll see if I can reach out someone from PyCharm 😊

[...]

PR #792, opened 2021-03-27
Add support for specifying private fields through a strawberry.type variable

lijok:
## Description
`strawberry.Private` is great for specifying private fields, however, does not play well with IDE's: https://github.com/strawberry-graphql/strawberry/issues/702

This PR adds supports for specifying which fields on a type should be made private, through a `private_fields` variable on `strawberry.type` decorator

Example:
```py
import strawberry

@strawberry.type(private_fields=["uid"])
class Post:
    uid: str
    title: str
    description: str
```

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation

* https://github.com/strawberry-graphql/strawberry/issues/702

## Checklist

### DIFF
--- a/strawberry/type.py
+++ b/strawberry/type.py
@@ -104,10 +104,9 @@ def _process_type(
     wrapped = _wrap_dataclass(cls)
 
     interfaces = _get_interfaces(wrapped)
+    fields = _get_fields(cls)
     if private_fields:
-        fields = [i for i in _get_fields(cls) if i.origin_name not in private_fields]
-    else:
-        fields = _get_fields(cls)
+        fields = [i for i in fields if i.origin_name not in private_fields]
 
     wrapped._type_definition = TypeDefinition(
         name=name,
--- a/strawberry/type.py
+++ b/strawberry/type.py
@@ -97,13 +97,17 @@ def _process_type(
     is_interface: bool = False,
     description: Optional[str] = None,
     federation: Optional[FederationTypeParams] = None,
+    private_fields: Optional[List[str]] = None,
 ):
     name = name or to_camel_case(cls.__name__)
 
     wrapped = _wrap_dataclass(cls)
 
     interfaces = _get_interfaces(wrapped)
-    fields = _get_fields(cls)
+    if private_fields:
+        fields = [i for i in _get_fields(cls) if i.origin_name not in private_fields]
+    else:
+        fields = _get_fields(cls)
 
     wrapped._type_definition = TypeDefinition(
         name=name,
@@ -137,6 +141,7 @@ def type(
     is_interface: bool = False,
     description: str = None,
     federation: Optional[FederationTypeParams] = None,
+    private_fields: Optional[List[str]] = None,
 ):
     """Annotates a class as a GraphQL type.
 
@@ -155,6 +160,7 @@ def type(
             is_interface=is_interface,
             description=description,
             federation=federation,
+            private_fields=private_fields,
         )
 
     if cls is None:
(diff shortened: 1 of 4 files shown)


botberry:

```markdown
Release type: patch

```

strawberry/type.py
```diff
@@ -97,13 +97,17 @@ def _process_type(
...
     name = name or to_camel_case(cls.__name__)
 
     wrapped = _wrap_dataclass(cls)
 
     interfaces = _get_interfaces(wrapped)
-    fields = _get_fields(cls)
+    if private_fields:
+        fields = [i for i in _get_fields(cls) if i.origin_name not in private_fields]
+    else:
+        fields = _get_fields(cls)
```
lijok, line 110: Would've written this as
```py
    fields = _get_fields(cls)
    if private_fields:
        fields = list(filter(lambda x: x.origin_name not in private_fields, fields))
```
but mypy was complaining about it
```sh
../../../../../../Users/user/strawberry/strawberry/type:109: error: Unsupported right operand type for in ("Optional[List[str]]") (diff)
```

tests/schema/test_private_field.py
```diff
@@ -23,6 +23,25 @@ class Query:
...
+        name: str
+        age: int
+
+    definition = Query._type_definition
+
+    assert definition.name == "Query"
+    assert len(definition.fields) == 1
+
+    assert definition.fields[0].name == "name"
+    assert definition.fields[0].type == str
```
BryceBeagle, line 38: I would add a check here that verifies that `age` is _not_ one of the fields on the `TypeDefinition`

strawberry/type.py
BryceBeagle:
```suggestion
    fields = _get_fields(cls)
    if private_fields:
        fields = [i for i in fields if i.origin_name not in private_fields]
```

BryceBeagle: Not sure why the `list(filter())` doesn't work, but I think this is a bit cleaner

BryceBeagle (changes requested):
I'm always a bit hesitant to keep tacking on more arguments to the function signatures for our types. I do feel like there's gotta be a cleaner way to get this done. 

I'll let @patrick91 give his opinions here, though

strawberry/type.py
BryceBeagle: Not sure why the list(filter()) doesn't work, but I think this is a bit cleaner

eimantas-gecas-axomic:
> I'm always a bit hesitant to keep tacking on more arguments to the function signatures for our types. I do feel like there's gotta be a cleaner way to get this done.
> 
> I'll let @patrick91 give his opinions here, though

I agree
It's not urgent on my end to get a solution for this in either, we can sleep on this for a while

lijok:
> I'm always a bit hesitant to keep tacking on more arguments to the function signatures for our types. I do feel like there's gotta be a cleaner way to get this done.
> 
> I'll let @patrick91 give his opinions here, though

I agree
It's not urgent to get a solution for this on my end, we can sleep on this

patrick91:
@lijok I think we should work on a plugin for PyCharm instead of changing our code, I don't have bandwidth for that at this time, but maybe in future :)

What do you think? Happy to close this PR for now?

lijok:
> @lijok I think we should work on a plugin for PyCharm instead of changing our code, I don't have bandwidth for that at this time, but maybe in future :)
> 
> What do you think? Happy to close this PR for now?

Yeah, that makes sense
Happy to close this

PR #798, opened 2021-03-30
Use schema to determine if we should camel case field names 

### LINKED ISSUE #542: Make camelCasing optional

### LINKED ISSUE #850: Introduce schema configurations
We discussed this on discord and a few issues here on GitHub, we want to be able to provide configuration to the schemas, for example:

- disabling/enabling camel casing
- changing how enums are generated
- having a "strict mode" for resolvers (ie throwing and error when a resolver doesn't have a type defined)

I think this might also be a good way to allow us to handle backwards incompatible changes, as we could put them under a flag here.

How could it look like? We could have a base class and subclass it (to get autocompletion):

```
class Config(StrawberryConfig):
   enum_values = StrawberryConfig.ENUM_MEMBER # or StrawberryConfig.ENUM_VALUE
   auto_camel_case = True

schema = strawberry.Schema(..., config=Config)
```

### LINKED ISSUE #1045: How to use snake_case in graphql schema for propeties too?
Is there a solution to use the `snake_case` as a **property name** inside a strawberry data class **and** the graphql schema?

My problem is, that my `snake_case` property name is always replaced by a `camelCase` one.

I searched inside the docs, but could only find the `name` setting for the class and field names.

Minimal example of what i'm trying:

```py
...
@strawberry.type
class SomeClass:
   creation_date: Optional[str]
```
graphql schema
```graphql
...
type SomeClass {
   creation_date: String
}

```

patrick91:
This PR introduces a new configuration parameter to `Schema`, right now it only supports disabling camel casing, like so: 

```python
@strawberry.type
class Query:
    example_field: str = "Example"

schema = strawberry.Schema(
    query=Query, config=StrawberryConfig(auto_camel_case=True)
)
```

I've decided to make it a dataclass to make it easier to instantiate directly. We can change it to a normal class if that's preferred :)

To make the auto_camel_case work I need to introduce a `GraphQLNameMixin` class that adds a `get_graphql_name(auto_camel_case: bool)` method. This mixin is used in both StrawberryField and StrawberryArgument.

The `get_graphql_name` method is used in arguments (when converting the argument) and in the schema converter when creating the schema. If there's a graphql_name passed explicitly (like in `strawberry.field(name="aName")` that will be used, otherwise we use the `python_name` and convert it to camel case if the option is enabled.

Maybe in future we will get rid of the mixin as the logic for converting arguments could be moved away from `arguments.py` but I didn't want to move that away in this PR.

TODO: docs, but it is ready to review

Closes #542
Closes #850
Closes #1045
Closes #1016
Closes #1013


strawberry/schema/schema_converter.py
```diff
@@ -164,7 +166,11 @@ def from_field(self, field: StrawberryField) -> GraphQLField:
         graphql_arguments = {}
         for argument in field.arguments:
             assert argument.name is not None
-            graphql_arguments[argument.name] = self.from_argument(argument)
+            argument_name = (
+                to_camel_case(argument.name) if self.auto_camel_case else argument.name
+            )
+
+            graphql_arguments[argument_name] = self.from_argument(argument)
```
patrick91, line 173: this should become a method: `self.get_graphql_name(argument.name)` or even `self.get_graphql_name(argument)` when we unify things :)

botberry:

```markdown
Release type: patch

```

strawberry/schema/schema.py
```diff
@@ -39,10 +40,12 @@ def __init__(
         types=(),
         extensions: Sequence[Type[Extension]] = (),
         execution_context_class: Optional[Type[GraphQLExecutionContext]] = None,
+        config: StrawberryConfig = None,
```
BryceBeagle, line 43:
```suggestion
        config: Optional[StrawberryConfig] = None,
```

strawberry/schema/schema_converter.py
```diff
@@ -312,12 +319,15 @@ def get_graphql_fields() -> Dict[str, GraphQLField]:
         return graphql_object_type
 
     def from_resolver(self, field: StrawberryField) -> Callable:
+        # TODO: convert things from graphql name to python name
```
BryceBeagle, line 322: Is this a TODO for a future PR?

strawberry/tools/create_type.py
```diff
@@ -33,14 +34,15 @@ def create_type(name: str, fields: List[StrawberryField]) -> Type:
...
-        dataclass_fields.append(
-            (
-                field.graphql_name,
-                field.type,
-                dataclasses_field(default=field),
-            )
-        )
+    for field in fields:
+        if not isinstance(field, StrawberryField):
+            raise TypeError("Field is not an instance of StrawberryField")
```
BryceBeagle, line 39: Do we want to include the field name/qualname in the error message?

strawberry/types/types.py
```diff
@@ -29,9 +29,9 @@ class TypeDefinition:
     _type_params: Dict[str, Type] = dataclasses.field(default_factory=dict, init=False)
 
     def get_field(self, name: str) -> Optional["StrawberryField"]:
-        return next(
-            (field for field in self.fields if field.graphql_name == name), None
-        )
+        # TODO: check if python name is ok for field name or if we should check
+        # the graphql name
+        return next((field for field in self.fields if field.python_name == name), None)
```
BryceBeagle, line 34: Is this a TODO for a different PR?

strawberry/utils/mixins.py
```diff
@@ -0,0 +1,19 @@
+from typing import Optional
+
+from strawberry.utils.str_converters import to_camel_case
+
+
+class GraphQLNameMixin:
+    python_name: str
+    graphql_name: Optional[str]
```
BryceBeagle, line 8: Couldn't a StrawberryField not have a `python_name` if defined using a lambda and an explicit `name` arg?

BryceBeagle (changes requested):

strawberry/field.py
```diff
@@ -2,22 +2,23 @@
...
 from .union import StrawberryUnion
-from .utils.str_converters import to_camel_case
 
 
 _RESOLVER_TYPE = Union[StrawberryResolver, Callable]
 
 
-class StrawberryField(dataclasses.Field):
+class StrawberryField(dataclasses.Field, GraphQLNameMixin):
+    python_name: str
```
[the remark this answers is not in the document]
patrick91: this is a left over I think, I did add it to give it a type :)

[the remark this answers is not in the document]
patrick91: ah, it is required for MyPy :( 

image

strawberry/schema/schema_converter.py
patrick91: this is stale :D removed!

strawberry/utils/mixins.py
patrick91: how would you define a field using a lambda?

strawberry/tools/create_type.py
patrick91: maybe! this code is duplicated by the way, bad rebase :D

strawberry/tools/create_type.py
```diff
@@ -33,14 +34,15 @@ def create_type(name: str, fields: List[StrawberryField]) -> Type:
...
+
+        namespace[field.python_name] = field
+        annotations[field.python_name] = field.type
+
+    namespace["__annotations__"] = annotations  # type: ignore
 
-    cls = make_dataclass(name, fields=dataclass_fields)
+    cls = types.new_class(name, (), {}, lambda ns: ns.update(namespace))
```
patrick91, line 42: oh, @jkimbo @BryceBeagle I've removed the make_dataclass creation as it was breaking the new code (calling dataclasses twice was removing our custom fields)

strawberry/types/types.py
patrick91: mmh, maybe not :) I think we use this in the printer, so it's better if I check this before merging

strawberry/arguments.py
```diff
@@ -54,8 +55,8 @@ def __init__(
         description: Optional[str] = None,
         default: object = UNSET,
     ) -> None:
-        self.python_name = python_name
-        self._graphql_name = graphql_name
+        self.python_name = python_name  # type: ignore
+        self.graphql_name = graphql_name
```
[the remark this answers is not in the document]
patrick91: the function signature has python name as optional, but in `GraphQLNameMixin` it is not optional

strawberry/utils/mixins.py
BryceBeagle: Oh right, that would be a resolver

strawberry/field.py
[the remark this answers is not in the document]
BryceBeagle: Why not just have a `._python_name` instance attribute that holds the non-optional str, and then the `python_name` property can be defined normally?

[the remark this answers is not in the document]
patrick91: I’d rather get rid of it being optional :) that should be happening with your PR, so I’d say we can leave as it is :)

strawberry/tools/create_type.py
jkimbo: But won't that mean that the create_type function no longer returns a dataclass?

Also I don't understand why this change means we're calling make_dataclass twice?

patrick91: before this change we are creating the dataclass twice:

1. cls = make_dataclass(name, fields=dataclass_fields)
2. return strawberry.type(cls)

`strawberry.type` runs `dataclasses.dataclass` on cls, effectively creating the dataclass again, that wouldn't be a be problem normally, but it was also overriding the values of the fields.

Doing a check in `strawberry.type` breaks inheritance unfortunately :)

hope this makes sense :)

we still return a dataclass since `strawberry.type` creates one for you :)

jkimbo: Ah yeah that makes sense. Good catch!

jkimbo (approved):
Lgtm

PR #801, opened 2021-04-01
Fix __eq__ and __repr__ when using resolver fields

patrick91:
## Description

Eq and repr where broken previously. We don't want to include fields with resolvers for eq and repr
as they are basically functions and doesn't make sense to compare functions :)

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

### DIFF
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -42,13 +42,16 @@ class StrawberryField(dataclasses.Field):
     ):
         federation = federation or FederationFieldParams()
 
+        # basic fields are fields with no provided resolver
+        is_basic_field = not base_resolver
+
         super().__init__(  # type: ignore
             default=dataclasses.MISSING,
             default_factory=dataclasses.MISSING,
             init=base_resolver is None,
-            repr=True,
+            repr=is_basic_field,
+            compare=is_basic_field,
             hash=None,
-            compare=True,
             metadata=None,
         )
 
@@ -90,6 +93,9 @@ class StrawberryField(dataclasses.Field):
 
         self.base_resolver = resolver
         self.type = resolver.type
+
+        self.repr = False
+        self.compare = False
 
         return self
 
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -48,7 +48,7 @@ class StrawberryField(dataclasses.Field):
         super().__init__(  # type: ignore
             default=dataclasses.MISSING,
             default_factory=dataclasses.MISSING,
-            init=base_resolver is None,
+            init=is_basic_field,
             repr=is_basic_field,
             compare=is_basic_field,
             hash=None,
@@ -94,9 +94,6 @@ class StrawberryField(dataclasses.Field):
         self.base_resolver = resolver
         self.type = resolver.type
 
-        self.repr = False
-        self.compare = False
-
         return self
 
     @property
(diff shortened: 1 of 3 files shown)


botberry:



---

This relases fixes an issue with the generated `__eq__` and `__repr__` methods when defining
fields with resolvers.

This now works properly:

```python
@strawberry.type
class Query:
    a: int

    @strawberry.field
    def name(self) -> str:
        return "A"

assert Query(1) == Query(1)
assert Query(1) != Query(2)
```

botberry:



---

This relases fixes an issue with the generated `__eq__` and `__repr__` methods when defining
fields with resolvers.

This now works properly:

```python
@strawberry.type
class Query:
    a: int

    @strawberry.field
    def name(self) -> str:
        return "A"

assert Query(1) == Query(1)
assert Query(1) != Query(2)
```

botberry:



---

This relases fixes an issue with the generated `__eq__` and `__repr__` methods when defining
fields with resolvers.

This now works properly:

```python
@strawberry.type
class Query:
    a: int

    @strawberry.field
    def name(self) -> str:
        return "A"

assert Query(1) == Query(1)
assert Query(1) != Query(2)
```

jkimbo (approved):

strawberry/field.py
```diff
@@ -91,6 +94,9 @@ def __call__(self, resolver: _RESOLVER_TYPE) -> "StrawberryField":
         self.base_resolver = resolver
         self.type = resolver.type
 
+        self.repr = False
+        self.compare = False
+
```
BryceBeagle, line 99: This is done in `@base_resolver.setter`. Is that not working?

BryceBeagle: Here: https://github.com/strawberry-graphql/strawberry/blob/f14305b6b2918424f2ac70519a61617c7136f113/strawberry/field.py#L130-L132

patrick91: uh, I missed it, but no it is not working, let me check why

patrick91: oh we were missing `compare`, updated!

strawberry/field.py
```diff
@@ -127,8 +130,9 @@ def base_resolver(self, resolver: StrawberryResolver) -> None:
         self._base_resolver = resolver
         self.origin = resolver.wrapped_func
 
-        # Don't add field to __init__ or __repr__ once it has a resolver
+        # Don't add field to __init__, __repr__ and __eq__ once it has a resolver
         self.init = False
+        self.compare = False
```
BryceBeagle, line 135: Should we not be comparing fields with resolvers? If I have a type with only fields with resolvers, should instances _always_ be equal?

patrick91: I'd say yes, how would you compare resolvers?

BryceBeagle: Is executing them unreasonable? I guess only if they have no arguments

patrick91: @BryceBeagle I'll merge this for now, I think it makes sense to compare just values :)

patrick91: > Is executing them unreasonable? I guess only if they have no arguments

oh sorry, I missed your comment. I'd say so, also there might be side effects too :)

PR #807, opened 2021-04-03
Fixed django multipart/form-data upload

evasseure:
## Description

This PR aims to fix an issue with django multipart/form-data uploads.

The current code only check if the content type is equal to `multipart/form-data`, but doesn't take into consideration that each item in a multipart message is separated by a boundary marker. For exemple webkit based browsers put "WebKitFormBoundary" in the name of that boundary.

So for webkit based browsers the content-type in the header would look like that: 
`Content-Type: multipart/form-data; boundary=----WebKitFormBoundaryXQjBNYe84AdfPItt`

This PR simply changes the way we check the content type for `multipart/form-data`.

## Types of Changes

- [ ] Core
- [X] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,3 @@
-Release type: minor
+Release type: patch
 
 Fixed issue with django multipart/form-data uploads
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: minor
+
+Fixed issue with django multipart/form-data uploads
--- a/strawberry/django/views.py
+++ b/strawberry/django/views.py
@@ -40,7 +40,7 @@ class BaseView(View):
         self.graphiql = graphiql
 
     def parse_body(self, request) -> Dict[str, Any]:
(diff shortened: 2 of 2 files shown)


RELEASE.md
```diff
@@ -0,0 +1,3 @@
+Release type: minor
```
patrick91, line 1: I'd say this is a patch version 😊
```suggestion
Release type: patch
```

patrick91 (approved):
Awesome, didn't know about `Content-Type: multipart/form-data; boundary=----WebKitFormBoundaryXQjBNYe84AdfPItt`

The change looks solid! I'll push a commit do this also for flask and ASGI :)

Feel free to fill this for to get a sticker! https://forms.gle/dmnfQUPoY5gZbVT67

patrick91:
oh looks like only the django view had this problem, perfect then, will update the release note and merge 😊

botberry:



---

Fixed issue with django multipart/form-data uploads

PR #808, opened 2021-04-03
Add cli option to specify src directory

nat-n:
In my project python packages live under the src directory. I want to be able to run the strawberry serve command from the root directory. The uvicorn cli [has a `--app-dir` option](https://github.com/encode/uvicorn/blob/master/uvicorn/main.py#L322) to support this use case. The present change adds the same mechanism to the strawberry CLI, also in the form of an `--app-dir` option.

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,11 @@
+Release type: minor
+
+Add --app-dir CLI option to specify where to find the schema module to load when using
+the debug server.
+
+For example if you have a _schema_ module in a _my_app_ package under ./src, then you
+can run the debug server with it using:
+
+```bash
+strawberry server --app-dir src my_app.schema
+```
--- a/strawberry/cli/commands/server.py
+++ b/strawberry/cli/commands/server.py
@@ -19,8 +19,11 @@ from strawberry.asgi import GraphQL
     default=".",
     type=str,
     show_default=True,
-    help="Look for the module in the specified directory, by adding this to the "
-    "PYTHONPATH. Defaults to the current working directory.",
+    help=(
+        "Look for the module in the specified directory, by adding this to the "
+        "PYTHONPATH. Defaults to the current working directory. "
+        "Works the same as `--app-dir` in uvicorn."
+    ),
 )
 def server(module, host, port, app_dir):
     sys.path.insert(0, app_dir)
--- a/strawberry/cli/commands/server.py
+++ b/strawberry/cli/commands/server.py
@@ -1,5 +1,4 @@
 import importlib
-import os
 import sys
 
 import click
@@ -15,8 +14,16 @@ from strawberry.asgi import GraphQL
 @click.argument("module", type=str)
 @click.option("-h", "--host", default="0.0.0.0", type=str)
 @click.option("-p", "--port", default=8000, type=int)
-def server(module, host, port):
-    sys.path.append(os.getcwd())
+@click.option(
+    "--app-dir",
+    default=".",
+    type=str,
+    show_default=True,
+    help="Look for the module in the specified directory, by adding this to the "
+    "PYTHONPATH. Defaults to the current working directory.",
+)
+def server(module, host, port, app_dir):
+    sys.path.insert(0, app_dir)
 
     reloader = hupper.start_reloader("strawberry.cli.run", verbose=False)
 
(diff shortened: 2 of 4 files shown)


botberry:

```markdown
Release type: patch

```

patrick91:
@nat-n looks good to me, should we follow uvicorn and name the parameter `--app-dir` as well?

nat-n:
@patrick91 Thanks.

I was thinking the semantics are a slightly different. Uvicorn targets an ASGI _app_, whereas strawberry server targets a _schema_. Though consistency is also valuable; I don't feel strongly about it.

I'll prepare the rest of the PR.

RELEASE.md
```diff
@@ -0,0 +1,3 @@
+Release type: minor
+
+Add --app-dir CLI option to specify where to find the schema module to load
```
patrick91, line 3: Let's add an example of using this here 😊
```suggestion
Add --app-dir CLI option to specify where to find the schema module to load when using the debug server
```

strawberry/cli/commands/server.py
```diff
@@ -15,8 +14,16 @@
 @click.argument("module", type=str)
 @click.option("-h", "--host", default="0.0.0.0", type=str)
 @click.option("-p", "--port", default=8000, type=int)
-def server(module, host, port):
-    sys.path.append(os.getcwd())
+@click.option(
+    "--app-dir",
```
patrick91, line 18: @BryceBeagle what do you think of `--app-dir` vs `--src-dir`?

patrick91 (commented):
Nice work so far, added a minor comment and asked @BryceBeagle for more feedback around the argument name 😊

strawberry/cli/commands/server.py
BryceBeagle: I think `--src-dir` is more reflective of what the option actually does, but consistency is nice. I don't have a strong opinion on which is used.

Regardless, I think the `help` argument to the `@click.option` should provide a little note mentioning that this is the same functionality as the `--app-dir` flag that the uvicorn CLI provides.

nat-n: @bryanculver The help message is copied almost verbatim from the uvicorn CLI. Are you suggesting to make it refer to the uvicorn CLI for context? It's already kinda verbose.

Screenshot 2021-04-05 at 18 24 15

nat-n:
Updated the RELEASE.md with an example.

strawberry/cli/commands/server.py
BryceBeagle: I assume you meant to tag me instead :stuck_out_tongue:. I was just suggesting adding a quick blurb that says something like "This is the same functionality as the `--app-dir` option that Uvicorn provides". I don't think it would be too verbose :slightly_smiling_face:

RELEASE.md
```diff
@@ -0,0 +1,11 @@
...
+
+Add --app-dir CLI option to specify where to find the schema module to load when using
+the debug server.
+
+For example if you have a _schema_ module in a _my_app_ package under ./src, then you
+can run the debug server with it using:
+
+```bash
+strawberry server --app-dir src my_app.schema
+```
```
patrick91, line 11: 👍

strawberry/cli/commands/server.py
```diff
@@ -15,8 +14,16 @@
...
+    default=".",
+    type=str,
+    show_default=True,
+    help="Look for the module in the specified directory, by adding this to the "
+    "PYTHONPATH. Defaults to the current working directory.",
```
patrick91, line 23:
```suggestion
    help=(
        "Look for the module in the specified directory, by adding this to the "
        "PYTHONPATH. Defaults to the current working directory. "
        "Works the same as `--app-dir` in uvicorn."
    ),
```

@BryceBeagle does this sound good?

patrick91 (commented):
Nice work! Thanks for updating the PR with our suggestions. I think we can merge this after updating the help value :)

strawberry/cli/commands/server.py
BryceBeagle: Looks perfect :+1:

patrick91 (approved):
@nat-n almost forgot, feel free to fill this form to get a sticker https://forms.gle/dmnfQUPoY5gZbVT67

And congrats on your first PR and release 😊

nat-n:
@patrick91 Thanks! I like the way you've set things up here :)

Issue #788, opened 2021-03-23
Field level data validators and pre/post processors


la4de:
It would be very useful to have possibility to define field level validators especially for those fields which do not have resolvers. My proposal is to add validators parameter to field to `strawberry.field` and validator decorator. This idea came to my mind when I was designing API for `strawberry-graphql-django` package (https://github.com/strawberry-graphql/strawberry-graphql-django/issues/10). Soon we realized that this could be core feature of `strawberry` library.

Validators could be used for following purposes
* field data validation
* data pre processing (input types)
* data post processing (output types)

I'd propose following syntax for validators
```python
def permission_validator(value, info):
    if not info.context.request.user.has_permission():
        raise Exception('Permission denied')
    return value

@strawberry.type
class User:
    name: str = strawberry.field(validators=[permission_validator])
```

I'd also propose decorator syntax, which is quite similar with existing field resolver syntax. Strawberry field would provide validator decorator which could be then use to tie the field together with validator function.
```python
@strawberry.input
class UserInput:
    name: str = strawberry.field()

    @name.validator
    def name_validator(value, info):
        if 'bad' in value:
            raise ValueError('name contains bad word')
      return value.title()
```

BryceBeagle:
I still think that maybe it's possible do add this functionality without first-classing support for validation, through other techniques such as function decorators (this gets a bit tricky with `input` types w/o resolvers, so maybe not?).

However, if we do want to first-class validation, I wonder if we should take a similar approach that Python's `@property` decorator does the `getter` and `setter` functions have the same symbol name:
```python
####### Property

class SomeClass:
    @property                     # implicitly my_field.getter
    def my_field() -> bool:
        return True

    @my_field.setter
    def my_field(value: bool):
        print("Set {value=}"

####### Strawberry

@strawberry.type
class User:
    @strawberry.field             # implicitly name.resolver
    def name() -> str:
        return "First Last"

    @name.validator
    def name(value):
        if 'bad' in value:
            raise ValueError('name contains bad word')
        return value.title()
```

Here, the same function symbol name is used for both the `resolver` and `validator` functions for the `name` field.

la4de:
I added 3rd option. I think I personally would like implement both 2nd and 3rd. Any opinions?

BryceBeagle:
Unfortunately, I don't think you can use decorators around assignments/expressions, only function/class definitions.

la4de:
[...]

PR #809, opened 2021-04-03
WIP: field data processors and validators

la4de:
The purpose of this pull request is to demonstrate the functionality of validators. This new feature has been proposed and discussed in #788.

## Description

New `validators` list is added to `StrawberryField` structure which contains list of callable validators. User defined validator function takes value as an parameter and returns validated value. Validator function is called after resolver functions with output types and before resolver with input types. See discussion in #788.

Both output and input types are supported.

```python
def permission_validator(value, info):
    if not info.context.request.user.has_permission():
        raise Exception('Permission denied')
    return value

@strawberry.type
class User:
    # option 1, define validator as a field parameter
    name: str = strawberry.field(validators=[permission_validator])
    age: int = strawberry.field()

    # option 2, define validator with field decorator    
    @age.validator
    def age(value, info):
        if not info.context.request.user.has_permission():
            raise Exception('Permission denied')
        return value
```

## Types of Changes

- [x] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* #788

## Checklist

### DIFF
--- a/tests/types/test_validators.py
+++ b/tests/types/test_validators.py
@@ -1,0 +1,70 @@
+import strawberry
+
+
+def upper_validator(value, info):
+    return value.upper()
+
+
+def test_validator():
+    @strawberry.type
+    class Type:
+        string: str = strawberry.field(validators=[upper_validator])
+
+    @strawberry.type
+    class Query:
+        @strawberry.field
+        def type() -> Type:
+            return Type("hello")
+
+    schema = strawberry.Schema(Query)
+
+    result = schema.execute_sync("{ type { string } }")
+
+    assert not result.errors
+    assert result.data["type"]["string"] == "HELLO"
+
+
+def test_validator_decorator():
+    @strawberry.type
+    class Type:
+        string: str = strawberry.field()
+
+        @string.validator
+        def validator(value, info):
+            return value.title()
+
+    @strawberry.type
+    class Query:
+        @strawberry.field
+        def type() -> Type:
+            return Type("hello")
+
+    schema = strawberry.Schema(Query)
+
+    result = schema.execute_sync("{ type { string } }")
+
+    assert not result.errors
+    assert result.data["type"]["string"] == "Hello"
+
+
+def test_input_validation():
+    @strawberry.input
+    class Input:
+        string: str = strawberry.field(validators=[upper_validator])
+
+    @strawberry.type
+    class Query:
+        s: str
+
+    @strawberry.type
+    class Mutation:
+        @strawberry.mutation
+        def mutation(self, input: Input) -> str:
+            return input.string
+
+    schema = strawberry.Schema(query=Query, mutation=Mutation)
+
+    result = schema.execute_sync('mutation { mutation(input: { string: "hello" }) }')
+
+    assert not result.errors
+    assert result.data["mutation"] == "HELLO"
--- a/tests/types/test_validators.py
+++ b/tests/types/test_validators.py
@@ -2,6 +2,8 @@ import strawberry
 
 
 def upper_validator(value, info):
+    if info.context and info.context in value:
+        raise ValueError("Invalid value")
     return value.upper()
 
 
@@ -22,6 +24,11 @@ def test_validator():
 
     assert not result.errors
     assert result.data["type"]["string"] == "HELLO"
+
+    # error case
+    result = schema.execute_sync("{ type { string } }", context_value="hello")
+
+    assert result.errors[0].message == "Invalid value"
 
 
 def test_validator_decorator():
@@ -68,3 +75,10 @@ def test_input_validation():
 
     assert not result.errors
     assert result.data["mutation"] == "HELLO"
+
+    # error case
+    result = schema.execute_sync(
+        'mutation { mutation(input: { string: "hello" }) }', context_value="hello"
+    )
+
+    assert result.errors[0].message == "Invalid value"
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -209,7 +209,10 @@ class StrawberryField(dataclasses.Field):
 
         if self.base_resolver:
             args, kwargs = self._get_arguments(kwargs, source=source, info=info)
-            validators.validate_arguments(kwargs, info)
+
+            error = validators.validate_arguments(kwargs, info)
+            if error:
+                return error
 
             return self.base_resolver(*args, **kwargs)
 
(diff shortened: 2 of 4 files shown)


tests/types/test_validators.py
```diff
@@ -0,0 +1,70 @@
+import strawberry
+
+
+def upper_validator(value, info):
+    return value.upper()
```
patrick91, line 5: I haven't seen the implementation yet, but I'm a bit confused on the terms, we are passing validators, but we are not doing any validation here 🤔

la4de: I'll add some data validation.

As discussed in #788, this same method can be used for data pre/post processing and validation. Could we have more generic name for this? Or do we want to separate these so that there would be separate validator and processor? Any opinions?

la4de:
Hello team,

Are you happy with chosen approach? I'm happy to contribute and to start working on final implementation but before that I'd like to hear your feedback. Is this the way to go?

patrick91:
I have a few comments, is this going to (also) replace our permission_classes? What about decorators, we wanted to support them too, see #473?

Also, do you think we can make it easier to support this pattern for handling errors with this approach: https://sachee.medium.com/200-ok-error-handling-in-graphql-7ec869aec9bc ?

la4de:
> I have a few comments, is this going to (also) replace our permission_classes? What about decorators, we wanted to support them too, see #473?

Validators can be used for permission validation as well. This can replace permission_classes if we decide so.

Decorators are nice but they can be used only with those fields which have resolver function. Validators can be used with both, fields with and without resolvers. I'd support both, decorators and validators. Decorators are more flexible, but as said they cannot be always used. @jkimbo, any comments?


> Also, do you think we can make it easier to support this pattern for handling errors with this approach: https://sachee.medium.com/200-ok-error-handling-in-graphql-7ec869aec9bc ?

This is very good question. I haven't thought about that earlier. :)

I think we could interpret exception as errors. Alternative results could be provided by validators. For example:

```python
@strawberry.type
class User:
    groups: List[Group] = strawberry.field()

    @groups.validator
    def groups(value, info):
        if not info.context.user.is_active:
            # inactive user, return error
            raise Exception('Invalid User')
        if not info.context.user.has_permission()
            # no permissions, return alternative response
            return []
        # validation passed, return original value
        return value
```

jkimbo:
I've been thinking about the proposal a lot and I think there is some really interesting ideas here but I'm not sure if it's the right approach. I also feel like the feature has expanded in scope quite a bit and I'd like to focus just on the validation part first. Some thoughts (numbered for easy replying):

1. I find it strange that the validators can modify the return value of the field. If we look at the example from `attrs`: the return value of the validator functions is ignored https://www.attrs.org/en/stable/examples.html#validators That makes more sense to me.

2. I think validators make the most sense for input types since they deal with user input. Object types are constructed by the developer so I would imagine that adding validation there is mostly unnecessary since you have full control over the values already before they get passed to the object type.

3. As @patrick91 points out this API doesn't allow you to model errors as part of the return type very easily. For example take this mutation which can either return a success state or an error:

```python
@strawberry.input
class UserInput:
	email: str = strawberry.field(validators=[valid_email])

@strawberry.type
class InvalidEmail:
	message: str

@strawberry.type
class Success:
	user: User

Response = strawberry.union("Response", [InvalidEmail, Success])

@strawberry.mutation
def create_user(user_input: UserInput) -> Response:
	# validation has already happened by this point
```

If the input contains an invalid email the error will be raised before the mutation function body gets called and so there is no opportunity to return the `InvalidEmail` type. 

<details>
<summary>Aside</summary>

A way to support this would be to add an explicit `validate` function to the input types so that you could run the validation in the resolver body. (or an `is_valid` function like in Django forms)

</details>

4. The above points then lead me to the question: are validators part of the GraphQL request/response or are they just concerned with the `strawberry.type/input` construction? I can't think of a use case where a validator would need access to the GraphQL context (other than permissions but I think that is better done elsewhere) and so I would lean to just supporting validation when creating the strawberry types. That would leave decorators to deal with any logic that does require the GraphQL context (e.g. permissions). This is closer to what Pydantic and attrs do and I think would be a valuable feature for strawberry.

<details>
<summary>Aside</summary>

It would actually be possible to support this form of validation in user land using the `__post_init__` method:

```python
@strawberry.input
class UserInput:
	email: str

	def __post_init__(self):
		if not valid_email(self.email):
			raise ValidationError("Invalid email")
```

I think there is still value in adding first class support for validators though FYI.
</details>

---

Sorry for such a long response @la4de but this proposal has been very thought provoking for me! Looking forward to hearing what you think. Also happy to jump on a call in Discord if that helps.

patrick91:
> 3. As @patrick91 points out this API doesn't allow you to model errors as part of the return type very easily. For example take this mutation which can either return a success state or an error [...]

Oh, by the way, if we find a way to manage this and validation we could use the same approach for Pydantic (https://strawberry.rocks/docs/integrations/pydantic#input-types). 

At the moment we do exactly what @jkimbo proposed:

```python
@strawberry.mutation
def create_user(user_input: UserInput) -> Response:
	try:
		user_input.to_pydantic()
	except ValidationError as error:
		return InvalidEmail(message="Invalid email address")
```

which isn't that bad to be honest, but I wonder if we could simplify things a bit (if that's actually useful)

la4de:
```python
@strawberry_django.type(models.Group)
class Group:
    pass

@strawberry_django.type(models.User, fields=['groups'])
class User:
    groups: List[Group] = strawberry_django.field()

    @groups.validator
    def groups(value, info):
```

la4de:
> 1. I find it strange that the validators can modify the return value of the field. If we look at the example from `attrs`: the return value of the validator functions is ignored https://www.attrs.org/en/stable/examples.html#validators That makes more sense to me.

I'm still preparing better example but meanwhile I want give another example: https://pydantic-docs.helpmanual.io/usage/validators/

Pydantic validators return value, which can be used for data cleaning and normalization.

jkimbo:
> > 1. I find it strange that the validators can modify the return value of the field. If we look at the example from `attrs`: the return value of the validator functions is ignored https://www.attrs.org/en/stable/examples.html#validators That makes more sense to me.
> 
> 
> 
> I'm still preparing better sample code but meanwhile I want give another example: https://pydantic-docs.helpmanual.io/usage/validators/
> 
> 
> 
> Pydantic validators return value, which can be used for data cleaning and normalization.

@la4de that is a fair point. I still find it a bit odd because the naming doesn't really match the functionality. I guess this is why Django forms have the `clean_*` methods rather than calling them validators.

patrick91:
This is also a similar approach here: https://docs.nestjs.com/graphql/field-middleware

patrick91:
Closing as we have extensions now 😊

PR #814, opened 2021-04-05
Add support for a default and default_factory with strawberry.field

### LINKED ISSUE #569: Parity with `dataclasses.field`
While starting to use Strawberry 🍓  and admittedly being a beginner, I noticed I had to make a choice : I have to choose between Strawberry features on my fields via `strawberry.field` (e.g. adding a description) and using native `dataclasses.field` features.

Some of the native `dataclasses.field` attributes that I do use on my dataclasses are `default_factory` and `metadata` (which allows me to pass an arbitrary dictionary of extra settings to my field)

Is there a reason why `strawberry.field` doesn't directly extend `dataclasses.field` ? Wouldn't that be more pythonic ? 😺 

Practically speaking:
```python
import faker

# https://faker.readthedocs.io/en/master/
fake = faker.Faker()

@strawberry.type
class User:
    first_name: str = strawberry.field(description="First name",  default_factory=fake.first_name, metadata={"foo": "bar"})
```

This isn't currently possible because `@strawberry.field` doesn't accept `default_factory` and `metadata`.
Would've been cool though, especially in the context of being able to generate fake data on the fly

jkimbo:
## Description

Add `default` and `default_factory` arguments to `strawberry.field` to bring it inline with `dataclasses.field`.

Note: I haven't added the `metadata` attribute because I'm not sure there is a compelling reason to add it.

# Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* Closes #569

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,10 @@
+Release type: minor
+
+Add support for `default` and `default_factory` arguments in `strawberry.field`
+
+```python
+@strawberry.type
+class Droid:
+    name: str = strawberry.field(default="R2D2")
+    aka: List[str] = strawberry.field(default_factory=["Artoo"])
+```
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -6,5 +6,5 @@ Add support for `default` and `default_factory` arguments in `strawberry.field`
 @strawberry.type
 class Droid:
     name: str = strawberry.field(default="R2D2")
-    aka: List[str] = strawberry.field(default_factory=["Artoo"])
+    aka: List[str] = strawberry.field(default_factory=lambda: ["Artoo"])
 ```
--- a/strawberry/exceptions.py
+++ b/strawberry/exceptions.py
@@ -153,3 +153,13 @@ class FieldWithResolverAndDefaultValueError(Exception):
         )
 
         super().__init__(message)
+
+
+class FieldWithResolverAndDefaultFactoryError(Exception):
+    def __init__(self, field_name: str, type_name: str):
+        message = (
+            f'Field "{field_name}" on type "{type_name}" cannot define a '
+            "default_factory and a resolver."
+        )
+
+        super().__init__(message)
--- a/strawberry/exceptions.py
+++ b/strawberry/exceptions.py
@@ -143,3 +143,13 @@ class WrongNumberOfResultsReturned(Exception):
         )
 
         super().__init__(message)
+
+
+class FieldWithResolverAndDefaultValueError(Exception):
+    def __init__(self, field_name: str, type_name: str):
+        message = (
+            f'Field "{field_name}" on type "{type_name}" cannot define a default '
+            "value and a resolver."
+        )
+
+        super().__init__(message)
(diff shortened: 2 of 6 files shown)


botberry:



---

Add support for `default` and `default_factory` arguments in `strawberry.field`

```python
@strawberry.type
class Droid:
    name: str = strawberry.field(default="R2D2")
    aka: List[str] = strawberry.field(default_factory=["Artoo"])
```

botberry:



---

Add support for `default` and `default_factory` arguments in `strawberry.field`

```python
@strawberry.type
class Droid:
    name: str = strawberry.field(default="R2D2")
    aka: List[str] = strawberry.field(default_factory=["Artoo"])
```

botberry:



---

Add support for `default` and `default_factory` arguments in `strawberry.field`

```python
@strawberry.type
class Droid:
    name: str = strawberry.field(default="R2D2")
    aka: List[str] = strawberry.field(default_factory=["Artoo"])
```

RELEASE.md
```diff
@@ -0,0 +1,10 @@
+Release type: minor
+
+Add support for `default` and `default_factory` arguments in `strawberry.field`
+
+```python
+@strawberry.type
+class Droid:
+    name: str = strawberry.field(default="R2D2")
+    aka: List[str] = strawberry.field(default_factory=["Artoo"])
```
patrick91, line 9: should this be? 
```suggestion
    aka: List[str] = strawberry.field(default_factory=lambda: ["Artoo"])
```

strawberry/exceptions.py
```diff
@@ -143,3 +143,23 @@ def __init__(self, expected: int, received: int):
...
+
+
+class FieldWithResolverAndDefaultFactoryError(Exception):
+    def __init__(self, field_name: str, type_name: str):
+        message = (
+            f'Field "{field_name}" on type "{type_name}" cannot define a '
+            "default_factory and a resolver."
+        )
+
+        super().__init__(message)
```
patrick91, line 165: Nice!

strawberry/field.py
```diff
@@ -39,6 +39,7 @@ def __init__(
         base_resolver: Optional[StrawberryResolver] = None,
         permission_classes: List[Type[BasePermission]] = (),  # type: ignore
         default_value: Any = undefined,
+        default_factory: Union[Callable, object] = undefined,
```
patrick91, line 42: can we use `UNSET` instead of undefined? 😊

patrick91 (commented):
Looks good! I was going to suggest checking how the defaults values were printed in the schema, but they are only printed for arguments (and we support that already). So all good from my side once resolved those two minor comments 😊

RELEASE.md
jkimbo: Yes it should be!

strawberry/field.py
jkimbo: Sure. Should I change `default_value` as well?

patrick91: yes, then we can remove this: https://github.com/strawberry-graphql/strawberry/blob/44dd42d4f442eb77078c8def079b22ed4800c583/strawberry/types/type_resolver.py#L31-L32

but if it starts looking like a big change we can defer it to another PR :)

patrick91 (approved):
💯

patrick91:
@jkimbo will fix the pre-commit issue (just send a PR to update our config too)

PR #817, opened 2021-04-05
Use identity comparisons for `dataclasses.MISSING`

BryceBeagle:
## Description

(Post-merge, nit-picky code review for #814)

`dataclasses.MISSING` is a singleton objects and should be compared against using identity operators instead of equality operators.

I don't think a release is necessary for this specific change.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist


botberry:

```markdown
Release type: patch

```

strawberry/field.py
```diff
@@ -48,9 +48,11 @@ def __init__(
         is_basic_field = not base_resolver
 
         super().__init__(  # type: ignore
-            default=(default_value if default_value != UNSET else dataclasses.MISSING),
+            default=(
+                default_value if default_value is not UNSET else dataclasses.MISSING
+            ),
             default_factory=(
-                default_factory if default_factory != UNSET else dataclasses.MISSING
+                default_factory if default_factory is not UNSET else dataclasses.MISSING
```
patrick91, line 55: we have a function called `is_unset` we should use that 😊

strawberry/types/type_resolver.py
```diff
@@ -381,7 +381,10 @@ class if one is not set by either using an explicit strawberry.field(name=...) o
                 raise PrivateStrawberryFieldError(field.python_name, cls.__name__)
 
             # Check that default is not set if a resolver is defined
-            if field.default != dataclasses.MISSING and field.base_resolver is not None:
+            if (
+                field.default is not dataclasses.MISSING
```
patrick91, line 385: 👍

strawberry/field.py
BryceBeagle: `is_unset` is in `strawberry.arguments` currently. If we want to expand its usage, where should we put it?

BryceBeagle: Actually, `UNSET` is in there too. I think both should be moved somewhere

jkimbo:
Tldr

strawberry/field.py
patrick91: if we move them it will be a breaking change (unless we re-export them). maybe we can put them in `strawberry.utils.unset`?

BryceBeagle: > it will be a breaking change
That's true... I wonder how many people are directly importing `UNSET`/`is_unset` though. 

> maybe we can put them in `strawberry.utils.unset`
I'm not sure `utils` is the correct place for `UNSET`. Maybe `is_unset`, though

jkimbo: +1 to moving them. Felt weird importing them from arguments into .field

patrick91: > That's true... I wonder how many people are directly importing UNSET/is_unset though.

we use it at work and on pycon 😊

> I'm not sure utils is the correct place for UNSET. Maybe is_unset, though

where would you put UNSET `strawberry.datastructures`[1]?

[1] https://github.com/encode/starlette/blob/master/starlette/datastructures.py

BryceBeagle: I guess a question that might help here is, do we have any _other_ "values" floating around the library? I was thinking it could go on `StrawberryType` when that's made, but decided against that because it's not the _type_, but the _value_. Is there room/want for a `StrawberryValue`? In theory this could be used to wrap around _any_ default value provided to a field/argument, but I'm not sure how helpful that would be.

patrick91: I don't know if we will have more values or singleton. I'd guess no, but you never know 😊

Not sure about `StrawberryValue`, can't see how it could be useful 🤔

jkimbo: I think it should be exported at the top level: `strawberry.UNSET`

Since it has to be used by arguments and input types it's not really an implementation detail.

patrick91: we'd still need to find a good place for it though 😊

jkimbo: Oh sure but that's less important if ultimately it gets imported from the top level. For what's it's worth I like the idea of a data structures file.

jkimbo: BTW this change was done as part of https://github.com/strawberry-graphql/strawberry/pull/916

BryceBeagle:
With #916, this PR only needs to touch `dataclasses.MISSING` now. This diff is very minimal

patrick91 (approved):

PR #824, opened 2021-04-06
Improve how we generate types from Pydantic

patrick91:
## Description

This PR improves how we generate types from Pydantic, it starts by using the StrawberryField (h/t @BryceBeagle) and then changes where we call @dataclass when using the decorator, we do this because @dataclass replaces StrawberryField with dataclasses.Field which breaks things unfortunately.
This is fine to do since we already create a dataclass when creating a type from Pydantic :)

I also tried to add a check with `dataclasses.isdataclass` but that was breaking interfaces and inheritance :(

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,5 @@
+Release type: patch
+
+This release fixes the pydantic integration where you couldn't
+convert objects to pydantic instance when they didn't have a
+default value.
--- a/strawberry/experimental/pydantic/type.py
+++ b/strawberry/experimental/pydantic/type.py
@@ -5,14 +5,13 @@ from typing import Any, Dict, List, Optional, Type
 from pydantic import BaseModel
 from pydantic.fields import ModelField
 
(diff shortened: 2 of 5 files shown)


botberry:



---

This release fixes the pydantic integration where you couldn't
convert objects to pydantic instance when they didn't have a
default value.

botberry:



---

This release fixes the pydantic integration where you couldn't
convert objects to pydantic instance when they didn't have a
default value.

botberry:



---

This release fixes the pydantic integration where you couldn't
convert objects to pydantic instance when they didn't have a
default value.

BryceBeagle:
How does this work when a field is a resolver?

patrick91:
> How does this work when a field is a resolver?

what do you mean?

marcoacierno (approved):

PR #832, opened 2021-04-08
Initial implementation of StrawberryEnum

### LINKED ISSUE #370: Use @strawberry.enum value instead of name
I was a bit surprised that the following code snippet prints `{'userType': 'BLAH'}` instead of `{'userType': 'c'}`
```python
from enum import Enum

import graphql
import strawberry


@strawberry.type
class User:
    @strawberry.enum
    class UserType(Enum):
        A = "a"
        B = "b"
        BLAH = "c"

    name: str
    user_type: UserType


@strawberry.type
class Query:
    @strawberry.field
    def user(self) -> User:
        return User(
            name="Bob",
            user_type=User.UserType.BLAH
        )


query = "{ user { userType } }"

result = graphql.graphql_sync(schema, query)

user = result.data['user']
print(user)
```

Is there a way to have the response use the Enum's value instead of its name? If the name is supposed to be used, is there any reason to define values instead of using `Enum.auto()`?

patrick91:
## Description

This PR implements StrawberryEnum, and adds supports to choose whether using names or values for enum generation

## Types of Changes

- [x] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* Closes #370

## TODO

- [ ] Allow to choose values or names #370


strawberry/arguments.py
```diff
@@ -173,6 +174,9 @@ def convert_argument(value: Any, argument: StrawberryArgument) -> Any:
 
     # Convert Enum fields to instances using the value. This is safe
     # because graphql-core has already validated the input.
+    if isinstance(argument_type, StrawberryEnum):
+        return argument_type(value)  # type: ignore
+
     if isinstance(argument_type, enum.EnumMeta):
         return argument_type(value)  # type: ignore
```
patrick91, line 181: I think we can delete this

strawberry/enum.py
```diff
@@ -11,16 +11,27 @@ class EnumValue:
...
 
-@dataclasses.dataclass
-class EnumDefinition:
+@dataclasses.dataclass(frozen=True)
+class StrawberryEnum:
+    enum: EnumMeta
+    # TODO: graphql_name, python_name
     name: str
-    values: List[EnumValue]
+    values: List[EnumValue] = dataclasses.field(hash=False)
```
patrick91, line 19: this can become a property

strawberry/enum.py
```diff
@@ -11,16 +11,27 @@ class EnumValue:
...
     description: Optional[str]
 
+    def __call__(self, *args: Any, **kwds: Any) -> Any:
+        return self.enum(*args, **kwds)
+
+    def __getattr__(self, attr):
+        if hasattr(self.enum, attr):
+            return self.enum[attr]
+
+        return super().__getattribute__(attr)
```
patrick91, line 29: I need a sanity check on this :D

strawberry/exceptions.py
```diff
@@ -92,6 +92,7 @@ def __init__(self, field_name: str, annotation):
 
 class UnsupportedTypeError(Exception):
     def __init__(self, annotation):
+        breakpoint()
```
patrick91, line 95: 👀

strawberry/enum.py
```diff
@@ -11,16 +11,27 @@ class EnumValue:
...
+    # TODO: graphql_name, python_name
     name: str
-    values: List[EnumValue]
+    values: List[EnumValue] = dataclasses.field(hash=False)
     description: Optional[str]
 
+    def __call__(self, *args: Any, **kwds: Any) -> Any:
+        return self.enum(*args, **kwds)
+
+    def __getattr__(self, attr):
```
patrick91, line 33: get item is needed too

strawberry/schema/schema_converter.py
```diff
@@ -110,23 +110,20 @@ def from_argument(self, argument: StrawberryArgument) -> GraphQLArgument:
...
-    def from_enum(self, enum: EnumDefinition) -> GraphQLEnumType:
-
-        assert enum.name is not None
-
+    def from_enum(self, enum: StrawberryEnum) -> GraphQLEnumType:
         # Don't reevaluate known types
-        if enum.name in self.type_map:
-            graphql_enum = self.type_map[enum.name].implementation
+        if enum.graphql_name in self.type_map:
+            graphql_enum = self.type_map[enum.graphql_name].implementation
```
patrick91, line 116: I need to check if types are unique globally or if they are unique by type (ie I can have a scalar called User and an Enum called User too)

strawberry/enum.py
BryceBeagle: I think this is correct (`object` doesn't define `__getattr__` by default, only `__getattribute__`), but I would type annotate this function a bit. This should always take a `str` and return an `Enum`, right?

strawberry/enum.py
```diff
@@ -11,45 +12,69 @@ class EnumValue:
...
+        if hasattr(self.enum, attr):
+            return self.enum[attr]
+
+        return super().__getattribute__(attr)
+
+    @property
+    def values(self) -> List[EnumValue]:
+        return [
+            EnumValue(item.name, item.value) for item in cast(Iterable[Enum], self.enum)
+        ]
```
BryceBeagle, line 41: Do we need `EnumValue` anymore? Or can the schema converter directly grab the values?

strawberry/enum.py
```diff
@@ -11,45 +12,69 @@ class EnumValue:
...
         description=description,
     )
 
-    return cls
-
 
 def enum(
     _cls: EnumMeta = None, *, name=None, description=None
-) -> Union[EnumMeta, Callable[[EnumMeta], EnumMeta]]:
+) -> Union[StrawberryEnum, Callable[[EnumMeta], StrawberryEnum]]:
```
BryceBeagle, line 70: Why does this take a `Callable`?

BryceBeagle: Why would you pass a `StrawberryEnum` to this function?

BryceBeagle (changes requested):
Will we be adding support for `auto()` and using the Enum name as the value?

strawberry/enum.py
patrick91: yes! the return type depends on what attribute we are requesting though

strawberry/enum.py
patrick91: this is the return type :)

patrick91:
Going to close this, might reimplement it in future 😊

Spartanforever1:
It would have been very cool to be able to pick between names or values for enum generation :pensive: :point_right::point_left:

Issue #839, opened 2021-04-12
Add default_resolver for strawberry.type


la4de:
I'd propose to add `default_resolver` for `strawberry.type`.

```python
def default_resolver():
    return Car(model='Ferrari')

@strawberry.type(default_resolver=default_resolver)
class Car:
    model: str
```

Following schema would use default resolver which would be overridable by field.
```python
@strawberry.type
class Query:
    car: Car
    @strawberry.field
    def another_car() -> Car:
        return Car(models='BMW')
```

With that schema, following query
```graphql
query {
  car { model }
  anotherCar { model }
}
```
would return 
```js
{
  data: {
    car: { model: "Ferrari" },
    anotherCar: { model: "BMW" },
}
```

That could be used by django integration for auto resolution of related objects as an example.

It would be nice if new resolver would have access to default resolver and could "extend" `default_resolver` something like this.
```python
@strawberry.type
class Query:
    @strawberry.field
    def yet_another_car(info) -> Car:
        # I'm not sure if info is right place for it, but you got the idea
        car = info.default_resolver()
        car.model = car.model.upper()
        return car
```

la4de:
@patrick91, any comments? I created draft pull request for this: #842.

This would help us in django integration. Default resolver can be used to resolve relation fields of django types.

patrick91:
I haven't seen the need of this, but if it can help the development of strawberry django then I'm for it.  

I'm only worried about the name, `default_resolver` sounds like we are change the default resolver for this type's fields, what do you think?

Apollo federation has a resolve reference function, what if used a similar approach?
https://www.apollographql.com/docs/federation/api/apollo-federation/#__resolvereference

la4de:
> I haven't seen the need of this, but if it can help the development of strawberry django then I'm for it.

With this the developer could just write type like this
```python
@strawberry_django.type(models.User)
class User:
    name: auto
    groups: List[Group]

@strawberry.type
class Query:
    user: User
```
and then just query all groups of requested user without need to worry about resolver. Default resolver would handle filters, relations and everything =)

> I'm only worried about the name, `default_resolver` sounds like we are change the default resolver for this type's fields, what do you think?

Good point, it may mislead the reader. 


> Apollo federation has a resolve reference function, what if used a similar approach?
> https://www.apollographql.com/docs/federation/api/apollo-federation/#__resolvereference

Relation resolving is just one example. This should be generic in my opinion.

patrick91:
[...]

PR #842, opened 2021-04-12
WIP: add default resolver for type

la4de:
This pull request adds new `default_resolver` parameter for `strawberry_type`, which is used to resolve the type if `base_resolver` is not given.

See more detailed description and examples are in the ticket #839.

### DIFF
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -72,6 +72,10 @@ class StrawberryField(dataclasses.Field):
         self._base_resolver: Optional[StrawberryResolver] = None
         if base_resolver is not None:
             self.base_resolver = base_resolver
+        elif hasattr(self.type, "_type_definition"):
+            default_resolver = self.type._type_definition.default_resolver
+            if default_resolver is not None:
+                self.base_resolver = StrawberryResolver(default_resolver)
 
         self.default_value = default_value
 
--- a/strawberry/type.py
+++ b/strawberry/type.py
@@ -1,6 +1,6 @@
 import dataclasses
 from functools import partial
-from typing import List, Optional, Type, cast
+from typing import Callable, List, Optional, Type, cast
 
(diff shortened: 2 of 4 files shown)


BryceBeagle:
I think the implementation should probably be touching these lines here: 
https://github.com/strawberry-graphql/strawberry/blob/27b375ff770dd0394c2b5722e5b301dd6c87b1f2/strawberry/field.py#L208-L213

The `getattr` is the de-facto current default resolver. I think it should be ripped out into an instance of `StrawberryResolver` and stored somewhere.

la4de:
> I think the implementation should probably be touching these lines here:
> https://github.com/strawberry-graphql/strawberry/blob/27b375ff770dd0394c2b5722e5b301dd6c87b1f2/strawberry/field.py#L208-L213
> 
> The `getattr` is the de-facto current default resolver. I think it should be ripped out into an instance of `StrawberryResolver` and stored somewhere.

Hmm. Should be then modify this too?
https://github.com/strawberry-graphql/strawberry/blob/043d9131b7d040ee38921c255de9c93b8adce96a/strawberry/field.py#L107-L112

How about the API and concept in general? Do you think if this would be useful?

BryceBeagle:
> Hmm. Should we then modify this too?

Yeah, by the looks of it, line `110` there should return the arguments for the default resolver. It's currently fine because the default resolver has no arguments, but if we allow changing it in the future, that will need to be updated too.

la4de:
> > Hmm. Should we then modify this too?
> 
> Yeah, by the looks of it, line `110` there should return the arguments for the default resolver. It's currently fine because the default resolver has no arguments, but if we allow changing it in the future, that will need to be updated too.

Default resolver may have arguments. For example arguments for filtering m2m relations.

la4de:
I closed ticket so I close this pull request too. See my last comment in ticket.

PR #846, opened 2021-04-12
Fix return type value from info field

la4de:
This fixes the bug where return type was some times resolved incorrectly.

### DIFF
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -275,12 +275,18 @@ class StrawberryField(dataclasses.Field):
         return _resolver
 
     def _get_return_type(self):
+        # using type ignore to make mypy happy,
+        # this codepath will change in future anyway, so this is ok
         if self.is_list:
-            type_ = List[self.child._get_return_type()]
+            assert self.child
+
+            type_ = List[self.child._get_return_type()]  # type: ignore
         else:
             type_ = self.type
+
         if self.is_optional:
-            type_ = Optional[type_]
+            type_ = Optional[type_]  # type: ignore
+
         return type_
 
 
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -250,7 +250,7 @@ class StrawberryField(dataclasses.Field):
                 context=info.context,
                 root_value=info.root_value,
                 variable_values=info.variable_values,
-                return_type=self.type,
+                return_type=self._get_return_type(),
                 operation=info.operation,
                 path=info.path,
             )
@@ -273,6 +273,15 @@ class StrawberryField(dataclasses.Field):
 
         _resolver._is_default = not self.base_resolver  # type: ignore
         return _resolver
+
+    def _get_return_type(self):
+        if self.is_list:
+            type_ = List[self.child._get_return_type()]
+        else:
+            type_ = self.type
+        if self.is_optional:
+            type_ = Optional[type_]
+        return type_
 
 
 def field(
--- a/tests/schema/test_info.py
+++ b/tests/schema/test_info.py
@@ -1,3 +1,7 @@
+from typing import List, Optional
+
+import pytest
+
 import strawberry
 from strawberry.types import Info
 
@@ -62,3 +66,44 @@ def test_info_has_the_correct_shape():
         "variableValues": "{}",
         "returnType": "<class 'tests.schema.test_info.test_info_has_the_correct_shape.<locals>.Result'>",  # noqa
     }
+
+
+@pytest.mark.parametrize(
+    "return_type,return_value",
+    [
+        (str, "text"),
+        (List[str], ["text"]),
+        (Optional[List[int]], None),
+    ],
+)
+def test_return_type_from_resolver(return_type, return_value):
+    @strawberry.type
+    class Query:
+        @strawberry.field
+        def field(info) -> return_type:
+            assert info.return_type == return_type  # type: ignore
+            return return_value
+
+    schema = strawberry.Schema(query=Query)
+
+    result = schema.execute_sync("{ field }")
+
+    assert not result.errors
+    assert result.data["field"] == return_value
+
+
+def test_return_type_from_field():
+    def resolver(info):
+        assert info.return_type == int
+        return 0
+
+    @strawberry.type
+    class Query:
+        field: int = strawberry.field(resolver=resolver)
+
+    schema = strawberry.Schema(query=Query)
+
+    result = schema.execute_sync("{ field }")
+
+    assert not result.errors
+    assert result.data["field"] == 0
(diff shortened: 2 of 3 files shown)


botberry:



---

This release fixes the return type value from info argument of resolver.

botberry:



---

This release fixes the return type value from info argument of resolver.

botberry:



---

This release fixes the return type value from info argument of resolver.

tests/schema/test_info.py
```diff
@@ -62,3 +64,38 @@ def hello(self, info: Info[str, str]) -> Result:
...
+
+
+def test_return_type():
+    def resolver_factory(return_type, return_value):
+        def resolver(info) -> return_type:
+            # check return type
+            assert info.return_type == return_type
+            return return_value
+
+        return resolver
```
patrick91, line 76: could we use pytest parametrization for this? 😊

la4de: Nice idea, but I'm not sure if it'd be easy task. I added different type of resolvers too. some return types are inherited from resolver, some from field

patrick91: we can split the test in two 😊

la4de: that's even better idea =)

strawberry/field.py
```diff
@@ -250,7 +255,7 @@ def _strawberry_info_from_graphql(info: GraphQLResolveInfo) -> Info:
                 context=info.context,
                 root_value=info.root_value,
                 variable_values=info.variable_values,
-                return_type=self.type,
+                return_type=self._return_type,
```
BryceBeagle, line 258: Can we just use 
```suggestion
                return_type=self.type or self.base_resolver.type,
```
and not store the extra variable?

la4de:
I have a question. Is this valid case to leave return type from resolver function and define it only in field? I'd expect that this would return type `int` and not `None`.

```python
        def resolver(info):
            assert info.return_type == expected
            return return_value

        @strawberry.type
        class Query:
            field: int = strawberry.field(resolver=resolver)
 ```

**Comment by patrick91** [comment]
> I have a question. Is this valid case to leave return type from resolver function and define it only in field? I'd expect that this would return type `int` and not `None`.
> 
> ```python
>         def resolver(info):
>             assert info.return_type == expected
>             return return_value
> 
>         @strawberry.type
>         class Query:
>             field: int = strawberry.field(resolver=resolver)
> ```

I think we briefly discussed this, but never decided if we should enforce types of resolver defined with `resolver=resolver`

Internally we could change the return type (at least on our StrawberryResolver object)

**Comment by BryceBeagle** [comment]
Yeah that should probably do one of two things:
1. Use the field type because the resolver is not annotated
2. Raise an exception because the resolver's (implicit `None`) return annotation does not match the field's type

My vote would be for #2

**Comment by patrick91** [comment]
> Yeah that should probably do one of two things:
> 
> 1. Use the field type because the resolver is not annotated
> 2. Raise an exception because the resolver's (implicit `None`) return annotation does not match the field's type
> 
> My vote would be for #2

#2 would probably be a big breaking change, I think we should only throw if there's a type that doesn't match, at least for now :)

**Comment by BryceBeagle** [comment]
> #⁠2 would probably be a big breaking change

I guess this could be another idea for the "things configurable on the Schema" bucket

**Comment by patrick91** [comment]
> > #⁠2 would probably be a big breaking change
> 
> I guess this could be another idea for the "things configurable on the Schema" bucket

gotta love feature flags :D

**Comment by la4de** [comment]
Many tests start to fail, if I access `self.base_resolver.type` in `_strawberry_info_from_graphql()`. In other other words, following change
```diff
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -244,6 +244,7 @@ class StrawberryField(dataclasses.Field):
             return result

         def _strawberry_info_from_graphql(info: GraphQLResolveInfo) -> Info:
+            self.base_resolver.type
             return Info(
                 field_name=info.field_name,
                 field_nodes=info.field_nodes,
```

causes this
```
> pytest
...
=== 83 failed, 354 passed, 2 xfailed in 40.93s ===
```

la4de (commented):
> Many tests start to fail, if I access `self.base_resolver.type` in `_strawberry_info_from_graphql()`. In other other words, following change
> 
> ```diff
> --- a/strawberry/field.py
> +++ b/strawberry/field.py
> @@ -244,6 +244,7 @@ class StrawberryField(dataclasses.Field):
>              return result
> 
>          def _strawberry_info_from_graphql(info: GraphQLResolveInfo) -> Info:
> +            self.base_resolver.type
>              return Info(
>                  field_name=info.field_name,
>                  field_nodes=info.field_nodes,
> ```
> 
> causes this
> 
> ```
> > pytest
> ...
> === 83 failed, 354 passed, 2 xfailed in 40.93s ===
> ```

I finally found the reason. `base_resolver` is sometimes `None` and `None` does not have `type` attribute.

It would be very nice to get full back traces from errors in unit tests. Now GraphQL exception capturing making all this invisible :)

tests/schema/test_info.py
la4de: Done

strawberry/field.py
la4de: After playing with the code and debugging multiple weird issues I chose another approach. Please take a look :)

la4de:
@patrick91 , @BryceBeagle, I ended up implementing getter for return type. At least functional tests are happy now. Unfortunately mypy wasn't so happy. I'd need some help from you now to resolve that.

I'm running out of ideas how to solve typing issues. Could you please take a look, what's wrong there. Please also review the code. Thanks.

strawberry/field.py
```diff
@@ -274,6 +274,15 @@ async def await_result(result):
         _resolver._is_default = not self.base_resolver  # type: ignore
         return _resolver
 
+    def _get_return_type(self):
+        if self.is_list:
+            type_ = List[self.child._get_return_type()]
```
patrick91, line 279: worth nothing that this in future will change to StrawberryList 😊

patrick91 (approved):
Looks good to me! Let's double check with @BryceBeagle before merging 😊

strawberry/field.py
BryceBeagle: Or potentially `StrawberryType` with `.is_list == True`. I'm still waffling on the implementation...

BryceBeagle (approved):
This looks to me like a good enough fix until `StrawberryType` is implemented

strawberry/field.py
patrick91: ok! no rush, I think this will be only used internally for now 😊

PR #861, opened 2021-04-17
Fix how we check if a type is generic

patrick91:
## Description

This PR fixes how we check if a type is generic, which was preventing us from being able to subclass a generic type while passing a type to it.

```python
T = typing.TypeVar("T")

@strawberry.interface
class Node(typing.Generic[T]):
    id: strawberry.ID

    def _resolve(self) -> typing.Optional[T]:
        return None

@strawberry.type
class Book(Node[str]):
    name: str

@strawberry.type
class Query:
    @strawberry.field
    def books(self) -> typing.List[Book]:
        return list()
```

**NOTE**: we should probably also fix this use case:

```python
T = typing.TypeVar("T")

@strawberry.interface
class Node(typing.Generic[T]):
    id: strawberry.ID
    name: T

@strawberry.type
class Book(Node[str]):
    ...
```

which is still unsupported, but I can be done in another PR

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,25 @@
+Release type: patch
+
+This release fixes a bug that prevented from extending a generic type when
+passing a type, like here:
+
+```python
+T = typing.TypeVar("T")
+
+@strawberry.interface
+class Node(typing.Generic[T]):
+    id: strawberry.ID
+
+    def _resolve(self) -> typing.Optional[T]:
+        return None
+
+@strawberry.type
+class Book(Node[str]):
+    name: str
+
+@strawberry.type
+class Query:
+    @strawberry.field
+    def books(self) -> typing.List[Book]:
+        return list()
+```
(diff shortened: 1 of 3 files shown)


botberry:



---

This release fixes a bug that prevented from extending a generic type when
passing a type, like here:

```python
T = typing.TypeVar("T")

@strawberry.interface
class Node(typing.Generic[T]):
    id: strawberry.ID

    def _resolve(self) -> typing.Optional[T]:
        return None

@strawberry.type
class Book(Node[str]):
    name: str

@strawberry.type
class Query:
    @strawberry.field
    def books(self) -> typing.List[Book]:
        return list()
```

sarahhenkens:
🎉  Thanks for the quick fix!

PR #862, opened 2021-04-19
types/info: Add fragments field to Info

staticfox:
## Description
My project currently uses fragments in order to break up data, though I was unable to find a way to look up those fragment definitions from within my resolvers. This patch adds the vanilla `.fragments` field to the custom Info field provided by Strawberry in order to allow for full document parsing from within the resolver. Please feel free to let me know if there's some other way I can achieve this in case I missed it. I also only added minimal tests since the functionality itself just exposes a vanilla python-graphql field, though I can add more detailed tests if desired.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation

## Checklist

### DIFF
--- a/docs/types/resolvers.md
+++ b/docs/types/resolvers.md
@@ -192,14 +192,15 @@ Info objects contain information for the current execution context:
 
 `class Info(Generic[ContextType, RootValueType])`
 
-| Parameter name  | Type                      | Description                                                           |
-| --------------- | ------------------------- | --------------------------------------------------------------------- |
-| field_name      | `str`                     | The name of the current field                                         |
-| context         | `ContextType`             | The value of the context                                              |
-| root_value      | `RootValueType`           | The value for the root type                                           |
-| variable_values | `Dict[str, Any]`          | The variables for this operation                                      |
-| operation       | `OperationDefinitionNode` | The ast for the current operation (public API might change in future) |
-| path            | `Path`                    | The path for the current field                                        |
+| Parameter name  | Type                                | Description                                                           |
+| --------------- | ----------------------------------- | --------------------------------------------------------------------- |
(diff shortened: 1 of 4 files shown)


botberry:

```markdown
Release type: patch

```

patrick91 (commented):
Looks good to me! Before merging though, how do you use fragments? Ideally we should make an abstraction for them instead of relying on the ast :)

staticfox:
Maybe I should have been less vague, but my fragment usage was meant more in terms of using them normally when creating queries on the client and not so much on the server side. I purely need access to them within the resolver in order to re-build a local map of the document within the resolver itself. Part of my project maps types to to sqlalchemy models, hence why knowing what fields the client wants will make the mapping process easier, e.g. how far I have to traverse through tables, what relationships I can skip, etc. An example use case I made looks something like this: https://gist.github.com/staticfox/2717c6940578010ff55e68dfb37c1d16

Note that I didn't include the AST builder code right now since it is currently in the "it's finally works, I can comment out all the breakpoints, print statements, and asserts now!" phase. 😅 

But yeah, let me know your thoughts on this. I don't really know what I could do in terms of abstraction for physical fragments, since they're being used in the same way as regular field nodes.

BryceBeagle:
@patrick91, going through old, open PRs. What's the status of this one? Did you self-assign so that you would remember to get back to it?

patrick91:
> @patrick91, going through old, open PRs. What's the status of this one? Did you self-assign so that you would remember to get back to it?

yes!

patrick91:
@staticfox I'll close this as it is a bit old and there's a few conflicting files 😊

Feel free to reopen or make a new one if you still want to work on it!

PR #873, opened 2021-04-25
Fix for allowing `source` as argument name

patrick91:
## Description

Us using `source` as argument name didn't allow `source` to be used as an argument name. 
This PR fixes it, but I'm not sure I like the solution, what do you think @BryceBeagle?

/cc @sarahhenkens

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+This release fixes an issue that prevented using `source` as name of an argument
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -172,7 +172,7 @@ class StrawberryField(dataclasses.Field):
         return None
 
     def _get_arguments(
-        self, kwargs: Dict[str, Any], source: Any, info: Any
+        self, kwargs: Dict[str, Any], source_: Any, info: Any
     ) -> Tuple[List[Any], Dict[str, Any]]:
         assert self.base_resolver is not None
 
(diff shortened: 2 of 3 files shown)


botberry:



---

This release fixes an issue that prevented using `source` as name of an argument

BryceBeagle (commented):
I think this still leaves `source_` as a reserved kwarg. Is there not a way to completely get around this problem?

patrick91:
> I think this still leaves `source_` as a reserved kwarg. Is there not a way to completely get around this problem?

We could do this: https://www.python.org/dev/peps/pep-0570/ (I've tested it, works), but it's only for python 3.8+ so not ready yet :D

do you have any other ideas in mind? I can change `source_` to `_source`, that's more appropriate

BryceBeagle:
So I looked at the source that these changes affect, but I'm still a bit confused on _how_ the underlying function's argument names leak to the field's argument list? 

Could we use `__source`? Does name mangling for work for function arguments?

patrick91:
> So I looked at the source that these changes affect, but I'm still a bit confused on _how_ the underlying function's argument names leak to the field's argument list?

This is because we always create a resolver for our users, meaning they don't get to choose the name of the `source/root` param. In pure graphql core you'd have something like this:

```
resolver=lambda obj, **kwargs: none
```

meaning that you could use source a field param as it is not used in the function signature. This case would break if you had an a obj param :) hope this clarifies the issue more
 
> Could we use `__source`? Does name mangling for work for function arguments?

The test seem to pass, not sure it does any mangling though and I don't really like it :D

PR #874, opened 2021-04-25
Pythonic wrapper for graphql core `FieldNode`.

coady:
## Description

Follow-up to #787, and may also be reusable in #862. Instead of directly relying on graphql-core node objects, creates strawberry types to be used on `info.field`.

Uses dataclasses, and creates a flatter more Pythonic representation of the node objects provided by GraphQL-core.

Note `field_nodes` is a list, but appears to always be a list of 1 - namely the field being resolved.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+Translate graphql-core `FieldNode` into a strawberry type on the `info` object.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,7 @@
 Release type: patch
 
-Translate graphql-core `FieldNode` into a strawberry type on the `info` object.
+This releases adds `selected_fields` on the `info` objects and it 
+allows to introspect the fields that have been selected in a GraphQL
+operation.
+
+This can become useful to run optimisation based on the queried fields.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,6 +1,6 @@
 Release type: patch
 
-This releases adds `selected_fields` on the `info` objects and it 
+This releases adds `selected_fields` on the `info` objects and it
 allows to introspect the fields that have been selected in a GraphQL
 operation.
 
--- a/strawberry/types/info.py
+++ b/strawberry/types/info.py
@@ -7,6 +7,8 @@ from graphql.pyutils.path import Path
 
 from strawberry.type import StrawberryType
 
+from .nodes import SelectedField
+
 
 ContextType = TypeVar("ContextType")
 RootValueType = TypeVar("RootValueType")
@@ -15,7 +17,8 @@ RootValueType = TypeVar("RootValueType")
 @dataclasses.dataclass
 class Info(Generic[ContextType, RootValueType]):
     field_name: str
-    field_nodes: List[FieldNode]
+    field_nodes: List[FieldNode]  # deprecated
+    selected_fields: List[SelectedField]
     context: ContextType
     root_value: RootValueType
     variable_values: Dict[str, Any]
--- a/strawberry/types/nodes.py
+++ b/strawberry/types/nodes.py
@@ -1,0 +1,106 @@
+"""
+Abstraction layer for graphql-core field nodes.
+
+Call `SelectedField` on a graphql `FieldNode`, such as in `info.field_nodes`.
+
+If a node has only one useful value, it's value is inlined.
+
+If a list of nodes have unique names, it's transformed into a mapping.
+Note Python dicts maintain ordering (for all supported versions).
+"""
+
+import dataclasses
+from typing import Any, Dict, Iterable, List, NewType, Optional, Union
+
+from graphql.language import (  # type: ignore
+    ArgumentNode,
+    DirectiveNode,
+    FieldNode,
+    FragmentSpreadNode,
+    InlineFragmentNode,
+    Node,
+    ValueNode,
+)
+
+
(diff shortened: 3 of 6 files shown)


botberry:



---

Translate graphql-core `FieldNode` into a strawberry type on the `info` object.

BryceBeagle:
Could/should the `Field` class be renamed something a bit more descriptive to avoid confusion with the more common field type (the one `@strawberry.field` returns (`StrawberryField`))

patrick91:
@coady that's cool, nice work! What do you think of renaming `Field` to `SelectedField` or `SelectionField`? That should be more descriptive for what it actually is :)

coady:
Renamed to `SelectionField`.

RELEASE.md
```diff
@@ -0,0 +1,3 @@
+Release type: patch
+
+Translate graphql-core `FieldNode` into a strawberry type on the `info` object.
```
````suggestion

This release adds full support for getting information about the field(s) that are being requested. The information is exposed as `field` inside the `info` parameter on resolvers.

```python
@strawberry.type
class Query:
    @strawberry.field
    def hello(self, info: Info[str, str]) -> int:
        return len(info.field.selections)
```

````

File: strawberry/types/nodes.py [hunk 2]
```diff
@@ -0,0 +1,106 @@
...
+
+@dataclasses.dataclass
+class SelectionField(FragmentSpread):
+    """Wrapper for a FieldNode."""
+
+    alias: Optional[str]
+    arguments: Arguments
+    selections: List[Selection]
+
+    def __init__(self, node: FieldNode):
```
patrick91, line 100: I think we can change this (in another) PR to something like `from_graphql_core` what do you think @BryceBeagle?

patrick91 (approved):
Looks good to me, sorry it took me a lot of time to review this!

I've added a couple of minor comments 😊

coady:
Fixed merge conflicts and kept `field_nodes` as deprecated.

strawberry/types/nodes.py
BryceBeagle: I think that's a good idea. We should open an issue to do so after this PR is merged

BryceBeagle (approved):

patrick91:
I'll test this soon and merge! Thanks @coady, this is great 😊

strawberry/types/info.py
```diff
@@ -15,7 +17,8 @@
 @dataclasses.dataclass
 class Info(Generic[ContextType, RootValueType]):
     field_name: str
-    field_nodes: List[FieldNode]
+    field_nodes: List[FieldNode]  # deprecated
+    field: SelectionField
```
patrick91, line 21: what do you think of renaming this to `current_field` or `selected_field`? 

not sure just `field` is a good name :) what do you think?

strawberry/types/nodes.py
```diff
@@ -0,0 +1,106 @@
...
+    def __init__(self, node: InlineFragmentNode):
+        self.type_condition = node.type_condition.name.value
+        self.selections = list(
+            map(selection, getattr(node.selection_set, "selections", []))
+        )
+        self.directives = directives(node.directives)
+
+
+@dataclasses.dataclass
+class SelectionField(FragmentSpread):
```
patrick91, line 93: Should this be `FieldSelection`? or `SelectedField`?

tests/schema/test_info.py
```diff
@@ -56,7 +58,19 @@ def hello(self, info: Info[str, str]) -> Result:
...
+        "rootEqual",
+        "operation",
+        "contextEqual",
+        "variableValues",
+        "returnType",
+        "fieldName",
+    }
+    assert field == {"name": "hello", "directives": {}, "alias": None, "arguments": {}}
     assert info == {
         # TODO: abstract this (in future)
```
patrick91, line 75: I think we don't need this comment anymore :)

patrick91 (commented):
Looks great, I did a small tests and seems solid! I've suggested a couple of name changes, let me know what you think, other than that this should be ready to merge :)

strawberry/types/info.py
coady: Oh, I found a problem. It is possible to have multiple field nodes: if a client splits an identical field. So this will have to be a list, just like the original `field_nodes`.

As for naming, for the other nodes I basically just stripped the suffix `Node`, but that just leaves us with `Field` again. I guess I don't have a preference.

patrick91: `selected_fields` might be a good name then. Do we get multiple selection when doing a query like this:

```
{
  field {
    name
    name
  }
}
```

?

coady: Yes, exactly.

Renamed to `selected_fields: List[SelectedField]`.

patrick91: I'd have expected it to merge them :D interesting!

what about field aliases?

```
{
  field {
    name
    some: name
  }
}
```

coady: Aliases are resolved separately.

But apparently duplicate fields are allowed as long as they're identical (w.r.t. inputs), and then resolved once with multiple `field_nodes`. I don't know why anyone would need that feature, but it's there.

patrick91:
@coady sorry I messed up while trying to rebase this PR :D

image

can you rebase? not sure how I managed to push some code and now I get permission error 😅

coady:
Rebased. Looks like I don't have permissions to reopen.

Issue #437, opened 2020-08-28
aiohttp integration


hyzyla:
It would be nice to see [aiohttp](https://docs.aiohttp.org/en/stable/) integration with your library.

patrick91:
@hyzyla doesn't aiohttp support asgi? (never used it)

hyzyla:
No, there is not resolved closed issue about ASGI support for aiohttp https://github.com/aio-libs/aiohttp/issues/2902

patrick91:
Closing this as it was done in #895

PR #895, opened 2021-05-01
Add a basic AIOHTTP integration

DoctorJohn:
This PR adds a basic AIOHTTP integration, tests and documentation.

## Description

As mentioned in #437 AIOHTTP does not support ASGI and thus needs a separate integration. This PR adds a basic aiohttp integration with query, mutation and file upload support (I will possibly attempt to add subscription support in a separate PR). This PR also includes tests and documentation based on the Sanic integration tests and docs.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation

* #437 

## Checklist

### DIFF
--- a/strawberry/aiohttp/views.py
+++ b/strawberry/aiohttp/views.py
@@ -1,6 +1,7 @@
 import json
 from io import BytesIO
 from pathlib import Path
+from typing import Any, Dict
 
 from aiohttp import web
 from strawberry.file_uploads.data import replace_placeholders_with_files
@@ -82,17 +83,17 @@ class GraphQLView:
 
     async def parse_multipart_body(self, request: web.Request) -> dict:
         reader = await request.multipart()
-        operations = {}
-        files_map = {}
-        files = {}
+        operations: Dict[str, Any] = {}
+        files_map: Dict[str, Any] = {}
+        files: Dict[str, Any] = {}
         try:
             async for field in reader:
                 if field.name == "operations":
-                    operations = await field.json()
+                    operations = (await field.json()) or {}
                 elif field.name == "map":
-                    files_map = await field.json()
+                    files_map = (await field.json()) or {}
                 elif field.filename:
-                    files[field.name] = BytesIO(await field.read(decode=False))
+                    files[field.filename] = BytesIO(await field.read(decode=False))
         except ValueError:
             raise web.HTTPBadRequest(reason="Unable to parse the multipart body")
         try:
--- a/strawberry/aiohttp/views.py
+++ b/strawberry/aiohttp/views.py
@@ -93,7 +93,9 @@ class GraphQLView:
                 elif field.name == "map":
                     files_map = (await field.json()) or {}
                 elif field.filename:
-                    files[field.filename] = BytesIO(await field.read(decode=False))
+                    assert field.name
+
+                    files[field.name] = BytesIO(await field.read(decode=False))
         except ValueError:
             raise web.HTTPBadRequest(reason="Unable to parse the multipart body")
         try:
--- a/strawberry/aiohttp/views.py
+++ b/strawberry/aiohttp/views.py
@@ -32,12 +32,12 @@ class GraphQLView:
     ) -> GraphQLHTTPResponse:
         return process_result(result)
 
-    async def get(self, request: web.Request) -> web.Response:
+    async def get(self, request: web.Request) -> web.StreamResponse:
         if self.should_render_graphiql(request):
             return self.render_graphiql()
         return web.HTTPNotFound()
 
-    async def post(self, request: web.Request) -> web.Response:
+    async def post(self, request: web.Request) -> web.StreamResponse:
         operation_context = await self.get_execution_context(request)
         context = await self.get_context(request)
         root_value = await self.get_root_value(request)
@@ -94,7 +94,7 @@ class GraphQLView:
             raise web.HTTPBadRequest(reason="Unable to parse the multipart body")
         return replace_placeholders_with_files(operations, files_map, files)
 
-    def render_graphiql(self) -> web.Response:
+    def render_graphiql(self) -> web.StreamResponse:
         html_string = self.graphiql_html_file_path.read_text()
         html_string = html_string.replace("{{ SUBSCRIPTION_ENABLED }}", "false")
         return web.Response(text=html_string, content_type="text/html")
--- a/strawberry/aiohttp/views.py
+++ b/strawberry/aiohttp/views.py
@@ -24,8 +24,8 @@ class GraphQLView:
     async def get_root_value(self, request: web.Request) -> object:
         return None
 
-    async def get_context(self, request: web.Request) -> object:
-        return {"request": request}
+    async def get_context(self, request: web.Request, response: web.Response) -> object:
+        return {"request": request, "response": response}
 
     async def process_result(
         self, request: web.Request, result: ExecutionResult
@@ -39,7 +39,8 @@ class GraphQLView:
 
     async def post(self, request: web.Request) -> web.StreamResponse:
         operation_context = await self.get_execution_context(request)
-        context = await self.get_context(request)
+        response = web.Response()
+        context = await self.get_context(request, response)
         root_value = await self.get_root_value(request)
 
         result = await self.schema.execute(
@@ -51,7 +52,9 @@ class GraphQLView:
         )
 
         response_data = await self.process_result(request, result)
-        return web.json_response(response_data)
+        response.text = json.dumps(response_data)
+        response.content_type = "application/json"
+        return response
 
     async def get_execution_context(self, request: web.Request) -> ExecutionContext:
         data = await self.parse_body(request)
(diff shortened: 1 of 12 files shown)


strawberry/aiohttp/views.py
```diff
@@ -0,0 +1,106 @@
...
+
+class GraphQLView(aiohttp.web.View):
+    graphiql = True
+    schema: Optional[BaseSchema] = None
+
+    async def get_root_value(self) -> Any:
+        return None
+
+    async def get_context(self) -> Any:
+        return {"request": self.request}
```
BryceBeagle, line 21: Nitpick: Make these return `object` so we don't lose all type-hinting

strawberry/aiohttp/views.py
```diff
@@ -0,0 +1,106 @@
...
+            data = await self.parse_body()
+        except json.JSONDecodeError:
+            raise aiohttp.web.HTTPBadRequest(
+                reason="Unable to parse request body as JSON"
+            )
+
+        try:
+            query = data["query"]
+            variables = data.get("variables")
+            operation_name = data.get("operationName")
```
BryceBeagle, line 58: Nitpick: Can you move these out of the `try/except` as they cannot raise a `KeyError`?

strawberry/aiohttp/views.py
```diff
@@ -0,0 +1,106 @@
...
+                elif field.name == "map":
+                    files_map = await field.json()
+                elif field.filename:
+                    files[field.name] = BytesIO(await field.read(decode=False))
+            return replace_placeholders_with_files(operations, files_map, files)
+        return await self.request.json()
+
+    def render_graphiql(self) -> aiohttp.web.Response:
+        with open(self.graphiql_html_file_path, "r") as f:
+            html_string = f.read()
```
BryceBeagle, line 88:
```suggestion
        html_string = self.graphiql_html_file_path.read_text()
```

strawberry/aiohttp/views.py
```diff
@@ -0,0 +1,106 @@
...
+    def should_render_graphiql(self) -> bool:
+        if not self.graphiql:
+            return False
+        return "text/html" in self.request.headers.get("Accept", "")
+
+    @classmethod
+    def as_view(cls, schema: BaseSchema, graphiql=True) -> Type["GraphQLView"]:
+        cls.schema = schema
+        cls.graphiql = graphiql
+        return cls
```
BryceBeagle, line 106: This modifies _and_ returns the global object. Is that intentional?

BryceBeagle (changes requested):

botberry:

```markdown
Release type: patch

```

strawberry/aiohttp/views.py
DoctorJohn: It somewhat is. `aiohttp.web.Application.router.add_view` expects a subclass of `aiohttp.web.View` as parameter. The `as_view` method is what I came up with to make the subclass configurable and more consistent with the other integrations.

The class based view appeared to be the easiest for the user to configure routing for. I could explore two alternatives to class based views which would not depend on my somewhat hacky configuration approche.

BryceBeagle: Would it not be unreasonable for auser to subclass the `aiohttp` `GraphQLView` object to override those two values?

DoctorJohn: I thought about that. Creating a subclass would be required since the abviously `schema` is not optional. Requiring the user to create a subclass in order to configure it felt incovenient to me. It's an option though.

I just realized that I could probably achive something like the following by not using a class based view and utilizing a `__call__` method:

```python
app = aiohttp.web.Application()
app.add_routes([web.route('*', '/graphql', GraphQLView(schema=schema, graphiql=False))])
```

I'll give this approche a try.

strawberry/aiohttp/views.py
```diff
@@ -0,0 +1,104 @@
...
+        self.graphiql = graphiql
+
+    async def __call__(self, request: web.Request) -> web.StreamResponse:
+        if request.method == "GET":
+            return await self.get(request)
+        if request.method == "POST":
+            return await self.post(request)
+        raise web.HTTPMethodNotAllowed(request.method, ["GET", "POST"])
+
+    async def get_root_value(self, request: web.Request) -> object:
```
patrick91, line 24: why object instead of Any? 😊

strawberry/aiohttp/views.py
```diff
@@ -0,0 +1,104 @@
...
+        return None
+
+    async def get_context(self, request: web.Request) -> object:
```
patrick91, line 27: do you think we should also add a `response` object here? We do that in django and asgi to allow setting cookies, not sure how that could be done in aiohttp :)

patrick91 (commented):
Awesome one! Looks good to me, I've added a minor comment, hopefully it makes sense 😊

strawberry/aiohttp/views.py
DoctorJohn: @bryanculver noted in https://github.com/strawberry-graphql/strawberry/pull/895#discussion_r624566099 the following:

> Nitpick: Make these return `object` so we don't lose all type-hinting

strawberry/aiohttp/views.py
DoctorJohn: I'll check whether this would be feasible :)

strawberry/aiohttp/views.py
patrick91: @BryceBeagle is `object` better than `Any`?

strawberry/aiohttp/views.py
DoctorJohn: I manage to implement the solution mentioned in the comment before. Feel free to take a look :)

strawberry/aiohttp/views.py
DoctorJohn: After some research it appears uncommon to pre-create an empty response object and later add data and content-type headers. It could be done though. Is setting cookies from `get_context` a common use case? To me it feels like setting cookies would be something that is done after schema execution. Similar to the `process_result` method a `process_response` method could be introduced.

patrick91: so, what I've done and seen others do is to set cookies from a resolver, via `info.context.response.set_cookies`, not sure if there's a better way, we got the inspiration from fastapi :)

DoctorJohn: Ah, I didn't think about setting cookies from resolvers. Good point. I'll try to implement it.

strawberry/aiohttp/views.py
BryceBeagle: Yes. `Any` tells the type checker to just _ignore_ anything that makes use of this value. `object` specifies that the type is so generic that it can be basically anything, _but_ it's still a Python object. This makes it so the type checker insists only methods/attributes/etc that are common to _all_ Python objects are used.

strawberry/aiohttp/views.py
DoctorJohn: `get_context` now takes an extra response argument. The solution works suprisingly well.

I also added some extra tests and a `RELEASE.md` file.

BryceBeagle (commented):
LGTM other than the merge conflict :+1:

DoctorJohn:
> LGTM other than the merge conflict +1

I just rebased my branch on main.

patrick91 (approved):
Nice work! I'll merge this in a bit, not sure why mypy is failing 🤔 

By the way we have a discord server if you want to join 😊 https://strawberry.rocks/discord

BryceBeagle:
Tests fail, but I'm not sure if they're related to your changes...

strawberry/aiohttp/views.py
```diff
@@ -82,17 +83,17 @@ async def parse_body(self, request: web.Request) -> dict:
...
             async for field in reader:
                 if field.name == "operations":
-                    operations = await field.json()
+                    operations = (await field.json()) or {}
                 elif field.name == "map":
-                    files_map = await field.json()
+                    files_map = (await field.json()) or {}
                 elif field.filename:
-                    files[field.name] = BytesIO(await field.read(decode=False))
+                    files[field.filename] = BytesIO(await field.read(decode=False))
```
DoctorJohn, line 96: Should be `files[field.name]`

patrick91:
@DoctorJohn there might be an issue with python 3.7, unrelated to this PR, maybe it was due to the merge of the latest typing_extensions. Will try to fix this tomorrow!

Error is 
```
E     ../../../../../../Users/patrick/Documents/github/strawberry-graphql/strawberry/.venv/lib/python3.7/site-packages/pydantic/typing:131: error: Module 'typing' has no attribute '_GenericAlias' (diff)
```

But it works when importing normally: 

```
❯ python
Python 3.7.10 (default, May  3 2021, 01:28:14)
[Clang 12.0.5 (clang-1205.0.22.9)] on darwin
>>> from typing import _GenericAlias
```

Also this error doesn't happen when running mypy normally. So maybe something is wrong with the test plugin

patrick91:
Not sure why we get this now and only when running the tests, but the error seems to be trigger by typeshed not having some private classes defined: https://github.com/python/typeshed/issues/4694

I've ignored the error in pydantic, so we are good I'd say :)

Issue #889, opened 2021-04-29
subscription AsyncGenerator raises error during client disconnected finalization process


wuyuanyi135:
Hello, I found a very weird problem (Python 3.8): 
When run the following code and open the GraphiQL interface (http://127.0.0.1/graphql), then create a query `subscription { test }` and refresh, the console will give: cannot reuse already awaited coroutine and task (detailed trace back attached later)

This happens when the Queue is blocked for a while without new put into the queue, i.e. if you  refresh before 2.0 is yield, it may not give you this error.

## Server code
```py
import asyncio
from strawberry.asgi import GraphQL
from starlette.applications import Starlette
import strawberry

@strawberry.type
class Query:
    @strawberry.field
    def hello(self) -> str:
        return "World"

@strawberry.type
class Subscription:
    @strawberry.subscription
    async def test(self) -> float:
        q = asyncio.Queue()
        q.put_nowait(1.0)
        q.put_nowait(2.0)
        async def agen():
            try:
                while True:
                    val = await q.get()
                    yield val
                    await asyncio.sleep(1.0)
            except GeneratorExit:
                print("bye")
            except BaseException as e:
                print(e)
        agen_obj = agen()
        return agen_obj


schema = strawberry.Schema(Query, subscription=Subscription)

graphql = GraphQL(schema)

app = Starlette(debug=True)
app.mount("/graphql", graphql)
```

## GraphQL query
```gql
subscription {
  test
}
```

Similarly, if we do not return AsyncGenerator but use the function as an AsyncGenerator, the same problem is still there:"
```py
@strawberry.type
class Subscription:
    @strawberry.subscription
    async def test(self) -> float:
        q = asyncio.Queue()
        q.put_nowait(1.0)

        try:
            while True:
                val = await q.get()
                yield val
                await asyncio.sleep(1.0)
        except GeneratorExit:
            print("bye")
        except BaseException as e:
            print(e)
```

## Trace back
After enabling `PYTHONASYNCIODEBUG=1` I could see the detailed traceback now.
```
Task was destroyed but it is pending!
source_traceback: Object created at (most recent call last):
  File "C:\Program Files\JetBrains\PyCharm 2019.3.1\plugins\python\helpers\pydev\pydevd.py", line 2173, in <module>
    main()
  File "C:\Program Files\JetBrains\PyCharm 2019.3.1\plugins\python\helpers\pydev\pydevd.py", line 2164, in main
[...]

PR #897, opened 2021-05-01
Fix inconsistent subscription generator clean up when the client disconnects

wuyuanyi135:
## Description

1. catch the `asyncio.CancelledError` raised when the client disconnects. Otherwise, the task will exit and leave un-retrieved exception error.
2. The checks on L202 seems unable to catch the disconnected state and insist to send the completion to the disconnected socket, leading to an error printed in the console. I wonder whether we could completely remove the `if` check and just use the try-except?
3. Unit-test pending

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

#889 

## Checklist

### DIFF
--- a/strawberry/asgi/__init__.py
+++ b/strawberry/asgi/__init__.py
@@ -1,12 +1,13 @@
 import asyncio
 import json
 import typing
-from websockets.exceptions import ConnectionClosedError
+
 from starlette import status
 from starlette.requests import Request
 from starlette.responses import HTMLResponse, JSONResponse, PlainTextResponse, Response
 from starlette.types import Receive, Scope, Send
 from starlette.websockets import WebSocket, WebSocketDisconnect, WebSocketState
+from websockets.exceptions import ConnectionClosedError
 
 from graphql import ExecutionResult as GraphQLExecutionResult, GraphQLError
 from graphql.error import format_error as format_graphql_error
@@ -184,6 +185,14 @@ class GraphQL:
                     ]
 
                 await self._send_message(websocket, GQL_DATA, payload, operation_id)
+        except asyncio.CancelledError:
+            # CancelError will be received when the client disconnects
+            # and the `handle_async_results` task gets cancelled
+
+            # Suppress the error and prevent sending any message, since the
+            # client was disconnected now
+            return
+
         except Exception as error:
             if not isinstance(error, GraphQLError):
                 error = GraphQLError(str(error), original_error=error)
@@ -194,10 +203,6 @@ class GraphQL:
                 {"data": None, "errors": [format_graphql_error(error)]},
                 operation_id,
             )
-        except asyncio.CancelledError:
-            # CancelError will be received when the client disconnects and the handle_async_results task being cancelled
-            # Suppress the error and prevent contining to the sending completion message stage
-            return
 
         if (
             websocket.client_state != WebSocketState.DISCONNECTED
--- a/strawberry/asgi/__init__.py
+++ b/strawberry/asgi/__init__.py
@@ -1,7 +1,7 @@
 import asyncio
 import json
 import typing
-
+from websockets.exceptions import ConnectionClosedError
 from starlette import status
 from starlette.requests import Request
 from starlette.responses import HTMLResponse, JSONResponse, PlainTextResponse, Response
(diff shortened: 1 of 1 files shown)


strawberry/asgi/__init__.py
```diff
@@ -194,12 +194,19 @@ async def handle_async_results(
                 {"data": None, "errors": [format_graphql_error(error)]},
                 operation_id,
             )
+        except asyncio.CancelledError:
+            # CancelError will be received when the client disconnects and the handle_async_results task being cancelled
+            # Suppress the error and prevent contining to the sending completion message stage
+            return
```
BryceBeagle, line 200: Can you make sure you've [installed the `pre-commit` hooks](https://github.com/strawberry-graphql/strawberry#pre-commit)? There's some minor formatting issues here, and the imports haven't been sorted.

BryceBeagle (changes requested):

botberry:

```markdown
Release type: patch

```

strawberry/asgi/__init__.py
wuyuanyi135: Sorry, I probably forgot that. I will first settle the PR on `graphql-core` then get back here.

wuyuanyi135:
We will wait for the upstream to decide whether the CancelledError should interrupt the MapAsyncIterator with StopAsyncIteration (normal exit) or some other exception. If this will be deemed as a normal exit we will consider put a return after the `async for`.

strawberry/asgi/__init__.py
patrick91: @wuyuanyi135 does that mean we need to wait for the updated release of `graphql-core` to merge this PR?

wuyuanyi135: @patrick91 The fix has been merged in `graphql-core`. We do need the updated release here.

strawberry/asgi/__init__.py
```diff
@@ -194,12 +194,19 @@ async def handle_async_results(
                 {"data": None, "errors": [format_graphql_error(error)]},
                 operation_id,
             )
+        except asyncio.CancelledError:
+            # CancelError will be received when the client disconnects and the handle_async_results task being cancelled
+            # Suppress the error and prevent contining to the sending completion message stage
```
marcoacierno, line 199:
```suggestion
            # Suppress the error and prevent continuing to the sending completion message stage
```
?

strawberry/asgi/__init__.py
```diff
@@ -194,12 +194,19 @@ async def handle_async_results(
...
 
         if (
             websocket.client_state != WebSocketState.DISCONNECTED
             and websocket.application_state != WebSocketState.DISCONNECTED
         ):
-            await self._send_message(websocket, GQL_COMPLETE, None, operation_id)
+            try:
+                await self._send_message(websocket, GQL_COMPLETE, None, operation_id)
+            except ConnectionClosedError:
+                pass
```
marcoacierno, line 209: I think we should catch ConnectionClosedError on L198 - L196 as well? 
There is a `except Exception as error` which might capture this error as well and try to send it as error to the client

strawberry/asgi/__init__.py
```diff
@@ -194,12 +194,19 @@ async def handle_async_results(
                 {"data": None, "errors": [format_graphql_error(error)]},
                 operation_id,
             )
+        except asyncio.CancelledError:
```
marcoacierno, line 197: I think this should come before `except Exception as error:`?

here https://docs.python.org/3.8/library/asyncio-exceptions.html it says that it's using BaseException on 3.8 as base so it will be captured by `except Exception` above?

strawberry/asgi/__init__.py
DoctorJohn: @wuyuanyi135 Strawberry was upgraded to graphql-core 3.1.5 (24080d57817dabd1e571f5463a82e4eeaf0b2dde) which includes your upstream changes.

strawberry/asgi/__init__.py
DoctorJohn: Just out of curiosity: where could `websockets.ConnectionClosedError` be raised?

I noticed `self._send_message` calls `starlette.WebSocket.send_json` calls `starlette.WebSocket.send`, but these methods appear to only raise `RuntimeError`s. (Finally `starlette.WebSocket.send` calls the `send` method passed to `strawberry.asgi.GraphQL.__call__`, thats where I lost track).

wuyuanyi135:
@patrick91 Sorry for the late response. I wonder what needs to be worked on in this PR now?

patrick91:
> @patrick91 Sorry for the late response. I wonder what needs to be worked on in this PR now?

I wanted to add some tests, but didn’t manage to do so, are you able to write one? 😌

wuyuanyi135:
I will work on this one tomorrow.

Issue #589, opened 2020-11-19
Return extensions results from our view


patrick91:
Our `process_result` default function at the moment doesn't return `extensions`, we should return them when set :)

Also might be good to change the typing for  `GraphQLHTTPResponse` to allow to add more keys when needed :)

PR #903, opened 2021-05-03
add optional extensions to GraphQLHTTPResponse

lijok:
## Types of Changes

- [x] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* https://github.com/strawberry-graphql/strawberry/issues/589

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,5 @@
 Release type: minor
 
-Add extensions to GraphQLHTTPResponse.
+This release adds an extensions field to the `GraphQLHTTPResponse` type and also exposes it in the view's response.
+
+This field gets populated by Strawberry extensions: https://strawberry.rocks/docs/guides/extensions#get-results
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: minor
+
+Add extensions to GraphQLHTTPResponse.
--- a/strawberry/http.py
+++ b/strawberry/http.py
@@ -10,6 +10,7 @@ from strawberry.types import ExecutionResult
 class GraphQLHTTPResponse(TypedDict, total=False):
     data: Optional[Dict[str, Any]]
     errors: Optional[List[Any]]
+    extensions: Optional[Dict[str, Any]]
 
 
 def process_result(result: ExecutionResult) -> GraphQLHTTPResponse:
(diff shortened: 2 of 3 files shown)


lijok:
Not sure of the best place to add tests for this?

botberry:

```markdown
Release type: patch

```

patrick91:
@lijok looks good to me! I think we should add test for all our integrations, what do you think? :)

lijok:
> @lijok looks good to me! I think we should add test for all our integrations, what do you think? :)

This is what I was wondering, where best would it be to add tests for these?
New ones or extend existing ones?

patrick91:
> > @lijok looks good to me! I think we should add test for all our integrations, what do you think? :)
> 
> This is what I was wondering, where best would it be to add tests for these?
> New ones or extend existing ones?

new ones 😊

lijok:
@patrick91 this should be ready now

strawberry/http.py
```diff
@@ -10,12 +10,15 @@
 class GraphQLHTTPResponse(TypedDict, total=False):
     data: Optional[Dict[str, Any]]
     errors: Optional[List[Any]]
+    extensions: Optional[Dict[str, Any]]
```
BryceBeagle, line 13: This is marked `Optional` but can never be `None`. Should we be setting `data["extensions"] = None` if there aren't any? (this would apply to `data` and `errors` too.)

BryceBeagle (changes requested):

RELEASE.md
```diff
@@ -0,0 +1,3 @@
+Release type: minor
+
+Add extensions to GraphQLHTTPResponse.
```
```suggestion
This release adds an extensions field to GraphQLHTTPResponse and exposes it in the view's response.
```

strawberry/http.py
patrick91: `data` and `errors` already work like that, for extensions we always return a dict: https://github.com/strawberry-graphql/strawberry/blob/98b4e6728251bfba3fa59b35a65f348f273b2e3b/strawberry/extensions/runner.py#L50-L56

I think we can keep them as optional, I'm not sure if we will change the dict to None at somepoint (for example if there's no extension). So I'd say this is fine :)

We can always update this in another PR :)

patrick91 (approved):
Looks good to me 😊 I think we can merge it!

Feel free to request a sticker here: https://forms.gle/dmnfQUPoY5gZbVT67 ✨

DoctorJohn:
@patrick91 I think #589 can be closed now.

Issue #872, opened 2021-04-25
Add support for implicit UNSET as a type annotation


sarahhenkens:
Currently, we need to set the default value to `strawberry.arguments.UNSET` to support optional arguments. This object has the `Any` annotation which hides itself from type checkers.

Example:
```python
id: Optional[strawberry.ID] = UNSET
```
MyPy will consider `id` to be a `Union[ID, None]` while in reality, it's a `Union[ID, None, _Unset]`. This results in potential bugs in code where the type-checker would catch it.

For example, the following snippet will raise an error without a type checker catching it:
```php
def foo(id: Optional[str] = UNSET) -> None:
  if id is not None:
    print(id.lower()) # Will attempt to call `lower()` on `UNSET`
```

## Proposal

My proposal is to introduce a new type alias as a first-class citizen in the Strawberry type system:

```python
T = TypeVar("T")
UnsetOrOptional = Union[_Unset, Optional[T]]
```

And its usage in resolvers:
```python
@strawberry.field
def foo(id: UnsetOrOptional[strawberry.ID]) -> None:
  pass
```

- The default assignment of the arguments remain used for actual default values in GraphQL
- The body of our resolver has the current triple Union to enforce it to check for unset as well
- The upstream logic will automatically assign `UNSET` to its value if it was not included in the request

### Just optional

This allows us to also provide the option for resolvers to define a simpler default behavior when we only care about `strawberry.ID` vs `None`:

```python
@strawberry.field
def foo(id: Optional[strawberry.ID]) -> None:
  pass
```

- No default assignment needed if you don't want it to show up in the generated GraphQL schema
- The upstream logic will automatically assign `None` to its value if it was not included in the request

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/strawberry-graphql) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/strawberry-graphql/strawberry/issues/872">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/872/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/872/pledge.svg">
</picture>
</a>

jkimbo:
[...]

PR #916, opened 2021-05-06
Clean up StrawberryField and StrawberryArgument interface

jkimbo:
## Description

* Change the signature of StrawberryField to make it easier to instantiate directly (defaulting arguments to None)
* Rename `default_value` to `default` in StrawberryField
* Rename `default_value` to `default` in StrawberryArgument
* Remove usages of `undefined` in favour of `UNSET` (note: I haven't actually removed `undefined` but I think we can do that now)

These changes are pretty minor but they are the ground work for allowing custom fields. I was just going to do all the changes necessary but #906 changes some of the code I was going to touch so I'll write up my proposal for it first before doing the code changes.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/strawberry/arguments.py
+++ b/strawberry/arguments.py
@@ -13,6 +13,21 @@ from .union import StrawberryUnion
 from .utils.str_converters import to_camel_case
 
 
+class _Unset:
+    def __str__(self):
+        return ""
+
+    def __bool__(self):
+        return False
+
+
+UNSET: Any = _Unset()
+
+
+def is_unset(value: Any) -> bool:
+    return type(value) is _Unset
+
+
 class StrawberryArgumentAnnotation:
     description: Optional[str]
 
@@ -35,7 +50,7 @@ class StrawberryArgument:
         is_list: bool = False,
         is_union: bool = False,
         description: Optional[str] = None,
-        default_value: Any = undefined,
+        default: Any = UNSET,
     ) -> None:
         self.python_name = python_name
         self._graphql_name = graphql_name
@@ -48,7 +63,7 @@ class StrawberryArgument:
         self.is_list = is_list
         self.is_union = is_union
         self.description = description
-        self.default_value = default_value
+        self.default = default
 
     @property
     def graphql_name(self) -> Optional[str]:
@@ -63,7 +78,7 @@ class StrawberryArgument:
         cls,
         python_name: str,
         annotation: Type[Annotated],  # type: ignore
-        default_value: Any,
+        default: Any,
         origin: Any,
     ) -> StrawberryArgument:
         annotated_args = get_args(annotation)
@@ -94,7 +109,7 @@ class StrawberryArgument:
             python_name=python_name,
             # TODO: fetch from StrawberryArgumentAnnotation
             graphql_name=None,
-            default_value=default_value,
+            default=default,
         )
 
 
@@ -110,7 +125,7 @@ def get_arguments_from_annotations(
     for name, annotation in annotations.items():
         default_value = parameters[name].default
         default_value = (
-            undefined
+            UNSET
             if default_value is inspect.Parameter.empty or is_unset(default_value)
             else default_value
         )
@@ -119,7 +134,7 @@ def get_arguments_from_annotations(
             argument = StrawberryArgument.from_annotated(
                 python_name=name,
                 annotation=annotation,
-                default_value=default_value,
+                default=default_value,
                 origin=origin,
             )
         else:
@@ -127,7 +142,7 @@ def get_arguments_from_annotations(
                 type_=annotation,
                 python_name=name,
                 graphql_name=None,
-                default_value=default_value,
+                default=default_value,
                 description=None,
                 origin=origin,
             )
@@ -137,21 +152,6 @@ def get_arguments_from_annotations(
         arguments.append(argument)
 
     return arguments
-
-
-class _Unset:
-    def __str__(self):
-        return ""
-
-    def __bool__(self):
-        return False
-
-
-UNSET: Any = _Unset()
-
-
-def is_unset(value: Any) -> bool:
-    return type(value) is _Unset
 
 
 def convert_argument(value: Any, argument: StrawberryArgument) -> Any:
--- a/strawberry/arguments.py
+++ b/strawberry/arguments.py
@@ -50,7 +50,7 @@ class StrawberryArgument:
         is_list: bool = False,
         is_union: bool = False,
         description: Optional[str] = None,
-        default: Any = UNSET,
+        default: object = UNSET,
     ) -> None:
         self.python_name = python_name
         self._graphql_name = graphql_name
@@ -78,7 +78,7 @@ class StrawberryArgument:
         cls,
         python_name: str,
         annotation: Type[Annotated],  # type: ignore
-        default: Any,
+        default: object,
         origin: Any,
     ) -> StrawberryArgument:
         annotated_args = get_args(annotation)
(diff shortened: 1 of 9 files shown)


botberry:

```markdown
Release type: patch

```

botberry:

```markdown
Release type: patch

```

botberry:

```markdown
Release type: patch

```

strawberry/arguments.py
```diff
@@ -13,6 +13,21 @@
 from .utils.str_converters import to_camel_case
 
 
+class _Unset:
+    def __str__(self):
+        return ""
+
+    def __bool__(self):
+        return False
```
BryceBeagle, line 21: Advantage of adding truthiness?

strawberry/arguments.py
```diff
@@ -35,7 +50,7 @@ def __init__(
         is_list: bool = False,
         is_union: bool = False,
         description: Optional[str] = None,
-        default_value: Any = undefined,
+        default: Any = UNSET,
```
BryceBeagle, line 53:
```suggestion
        default: object = UNSET,
```

strawberry/arguments.py
```diff
@@ -63,7 +78,7 @@ def from_annotated(
         cls,
         python_name: str,
         annotation: Type[Annotated],  # type: ignore
-        default_value: Any,
+        default: Any,
```
BryceBeagle, line 81:
```suggestion
        default: object,
```

strawberry/arguments.py
```diff
@@ -119,15 +134,15 @@ def get_arguments_from_annotations(
...
+                default=default_value,
                 origin=origin,
             )
         else:
             argument = StrawberryArgument(
                 type_=annotation,
                 python_name=name,
                 graphql_name=None,
-                default_value=default_value,
+                default=default_value,
```
BryceBeagle, line 145: Can we rename the `default_value`s to `default` as well?

strawberry/field.py
```diff
@@ -38,7 +38,7 @@ def __init__(
         description: Optional[str] = None,
         base_resolver: Optional[StrawberryResolver] = None,
         permission_classes: List[Type[BasePermission]] = (),  # type: ignore
-        default_value: Any = UNSET,
+        default: Any = UNSET,
```
BryceBeagle, line 41:
```suggestion
        default: object = UNSET,
```

strawberry/field.py
```diff
@@ -48,7 +48,7 @@ def __init__(
         is_basic_field = not base_resolver
 
         super().__init__(  # type: ignore
-            default=(default_value if default_value != UNSET else dataclasses.MISSING),
+            default=(default if default != UNSET else dataclasses.MISSING),
```
BryceBeagle, line 51:
```suggestion
            default=(default if default is not UNSET else dataclasses.MISSING),
```

strawberry/field.py
```diff
@@ -48,7 +48,7 @@ def __init__(
...
             default_factory=(
                 default_factory if default_factory != UNSET else dataclasses.MISSING
```
BryceBeagle, line 53:
```suggestion
                default_factory if default_factory is not UNSET else dataclasses.MISSING
```

BryceBeagle (changes requested):

strawberry/field.py
```diff
@@ -73,7 +73,11 @@ def __init__(
         if base_resolver is not None:
             self.base_resolver = base_resolver
 
-        self.default_value = default_value
+        # Note: StrawberryField.default is the same as
+        # StrawberryField.default_value except that `.default` uses
+        # `dataclasses.MISSING` to represent an "undefined" value and
+        # `.default_value` uses `UNSET`
+        self.default_value = default
```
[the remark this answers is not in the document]
jkimbo: That's right. There are a couple of cases where it's compared to UNSET and I didn't want to make this change bigger than it has to. I could change it to compare against dataclasses.MISSING instead? What do you think?

strawberry/arguments.py
jkimbo: I just moved this class around. Not sure what the point of this is. @patrick91 do you know?

strawberry/arguments.py
jkimbo: Yep will do

strawberry/arguments.py
patrick91: it's to make sure you can use it in `if` statements:

```
def something(a: int = UNSET):
    if a:
        return 

    ...
```

(note this check is silly but it's to demonstrate the idea :) )

BryceBeagle: That seems like it could lead to bugs. What if `a` is `0`?

strawberry/field.py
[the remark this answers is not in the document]
jkimbo: > We should probably leave an explicit comment there referring to the super class 👍

@BryceBeagle not sure I follow. What kind of comment do you mean?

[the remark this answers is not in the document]
jkimbo: Oh I think we've misunderstood each other. This is not to make `dataclasses` work, it's to make the rest of the Strawberry code work. For example we check against `default_value` here (not `default`): https://github.com/strawberry-graphql/strawberry/blob/b36b30c18aa5a3e131972830305ab8f912339ec2/strawberry/schema/schema_converter.py#L179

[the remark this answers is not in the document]
jkimbo: > So `.default` is only used by `dataclasses` at this point right?

Yep thats right. I think we should just make things consistent and check against `.default` (and use `dataclasses.MISSING` as the indicator that it's not set). That way we don't have 2 attributes that are basically the same thing but use different values to represent a missing value.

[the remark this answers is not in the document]
BryceBeagle: Something I've mentioned to @patrick91 is for us to find a way to make `StrawberryField` stop inheriting from `dataclasses.Field`.

The only reason it does currently is for our interal `dataclasses.dataclass` wrapping of the type to conveniently pick up the fields from the annotations. But it causes a lot of complexity and headaches.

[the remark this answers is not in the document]
jkimbo: > Something I've mentioned to @patrick91 is for us to find a way to make `StrawberryField` stop inheriting from `dataclasses.Field`.

Oh I see, I didn't realise that was the idea. Don't have a strong opinion on replacing dataclasses or not but I'd rather not do it in this PR 😄

jkimbo:
@patrick91 @BryceBeagle I've added some release notes. I think this is good to go now.

strawberry/field.py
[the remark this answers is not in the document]
BryceBeagle: > `StrawberryField.default` is an implementation detail related to the `dataclasses.Field` base class and should not be used. Use `StrawberryField.default_value` instead.

I'd like this comment somewhere, but I'm not sure of the best place for it

strawberry/arguments.py
patrick91: sorry I missed this comment! that's a good point 😊 

but then that means you **always** need to check if it is unset before doing anything with it, right? I guess that's similar to using None, so I think we can remove it :)

strawberry/field.py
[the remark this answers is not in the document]
patrick91: can we make a property for default? 🤔

patrick91 (approved):
Looks good to me! There's a couple of comments, but we can merge the PR and deal with them later if you prefer @jkimbo 😊

strawberry/field.py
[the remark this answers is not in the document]
BryceBeagle: What purpose would the property serve?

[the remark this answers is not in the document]
patrick91: to add the comment :D and, maybe, to trigger a warning?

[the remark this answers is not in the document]
BryceBeagle: I think we could just do something like

```def __init__(self):
    default: Any  # no value
    """The comment explaining it"
```

I'm not sure we could throw a warning without having a mechanism to determine if a user is using the value, or if `dataclasses` is internally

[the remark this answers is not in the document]
BryceBeagle: Discussed on discord: leave as-is

strawberry/arguments.py
jkimbo: BTW UNSET will be changing significantly with this change anyway: https://github.com/strawberry-graphql/strawberry/issues/872

PR #936, opened 2021-05-12
Only run release check when a PR is opened or updated

jkimbo:
## Description

To prevent multiple release check comments.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/.github/workflows/release-check.yml
+++ b/.github/workflows/release-check.yml
@@ -2,7 +2,7 @@ name: Release file check
 
 on:
   pull_request:
-    types: [opened, synchronize, reopened, ready_for_review, review_requested]
+    types: [opened, synchronize, reopened]
     branches:
     - main
 
--- a/.github/workflows/release-check.yml
+++ b/.github/workflows/release-check.yml
@@ -2,7 +2,7 @@ name: Release file check
 
 on:
   pull_request:
-    types: [opened, synchronize, reopened]
+    types: [synchronize, reopened, ready_for_review]
     branches:
     - main
 


botberry:

```markdown
Release type: patch

```

BryceBeagle:
IIRC, those extra states were added so that the RELEASE file check would get triggered when a PR transitioned from Draft to Ready for Review. Potentially removing `opened` would be better?

.github/workflows/release-check.yml
```diff
@@ -2,7 +2,7 @@ name: Release file check
 
 on:
   pull_request:
-    types: [opened, synchronize, reopened, ready_for_review, review_requested]
+    types: [opened, synchronize, reopened]
```
patrick91, line 5: If we remove this then we don’t get the comment when the PR goes from draft to ready, what if debounced events on the bot instead? 🤔

patrick91:
> IIRC, those extra states were added so that the RELEASE file check would get triggered when a PR transitioned from Draft to Ready for Review. Potentially removing `opened` would be better?

Oh, maybe that’s a better solution than mine ☺️

jkimbo:
That makes sense. Will change.

jkimbo:
Updated @patrick91 @BryceBeagle . I think we can still remove the "review_requested" event because the "ready_for_review" event should cover it.

patrick91 (approved):
Looks good to me 😊 let's see how it works!

PR #937, opened 2021-05-12
Fix federation field requires and provides

marcoacierno:
## Description

Right now using `@strawberry.federation.field(requires=["email"])` does nothing, email is not sent to resolve_reference, after checking the schema I found out that the `requires` was printed as an array instead of how graphql wants it: https://www.apollographql.com/docs/federation/federation-spec/#requires

```graphql
# extended from the Users service
extend type User @key(fields: "id") {
  id: ID! @external
  email: String @external
  reviews: [Review] @requires(fields: "email")
}
```

this field allows people to do

```python
@strawberry.federation.field(requires=["size", "weight"])
@strawberry.federation.field(requires=["size"])
```

to specify what fields they want

`@provides` has the same bug, so I fixed it as well

## Types of Changes

- [x] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/tests/federation/test_printer.py
+++ b/tests/federation/test_printer.py
@@ -137,3 +137,67 @@ def test_fields_requires_are_printed_correctly():
     assert schema.as_str() == textwrap.dedent(expected).strip()
 
     del Review
+
+
+def test_field_provides_are_printed_correctly():
+    global Review
+
+    @strawberry.federation.type
+    class User:
+        username: str
+
+    @strawberry.federation.type(keys=["upc"], extend=True)
+    class Product:
+        upc: str = strawberry.federation.field(external=True)
+        name: str = strawberry.federation.field(external=True)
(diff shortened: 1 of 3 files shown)


botberry:

```markdown
Release type: patch

```

tests/federation/test_printer.py
```diff
@@ -66,3 +66,74 @@ def top_products(self, first: int) -> List[Product]:
     assert schema.as_str() == textwrap.dedent(expected).strip()
 
     del Review
+
+
+def test_fields_requires_are_printed_correctly():
+    global Review
+
```
marcoacierno, line 73: This is the only test I can think of, I am not sure how to test it better without having a federation server and so on?

BryceBeagle (approved):

patrick91 (approved):

PR #945, opened 2021-05-17
Prevent the debug server from setting up its own event loop

### LINKED ISSUE #940: Can't get DataLoader to work
Hello! I'm trying examples from this page https://strawberry.rocks/docs/guides/dataloaders.
Running the following code on Python 3.8:
```python
import strawberry
from strawberry.dataloader import DataLoader
from typing import List


@strawberry.type
class User:
    id: strawberry.ID


async def load_users(keys) -> List[User]:
    return [User(id=key) for key in keys]

loader = DataLoader(load_fn=load_users)


@strawberry.type
class Query:
    @strawberry.field
    async def get_user(self, id: strawberry.ID) -> User:
        return await loader.load(id)


```
I get the following error message:
```
Task <Task pending name='Task-8' coro=<ExecutionContext.resolve_field.<locals>.await_result() 
running at /Users/-/Documents/src/dataservice-poc/virtualenv/lib/python3.8/site-packages/graphql/execution/execute.py:625> 
cb=[gather.<locals>._done_callback() at /usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/tasks.py:758]> 
got Future <Future pending> attached to a different loop
```

When I try my own code (which is pretty much the same, but the loader is real - it reads data from the db) I get this: "RuntimeError: await wasn't used with future".

I'm stuck, don't really know where to look. I thought Strawberry is supposed to manage async processing, but looks like it doesn't work that way. Any help would be greatly appreciated.

DoctorJohn:
## Description

This PR is a possible fix for #940. It configures `uvicorn` to not create its own event loop. This way the `strawberry.dataloader.DataLoader.loop` would be the same as the loop used by `uvicorn`.

While I would probably not configure `uvicorn` the same way in production code, I think it's the simplest (maybe the only) way to make dataloaders work with the debug server command.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* Fixes #940 

## Checklist

### DIFF
--- a/strawberry/cli/commands/server.py
+++ b/strawberry/cli/commands/server.py
@@ -57,4 +57,4 @@ def server(schema, host, port, app_dir):
         app.add_websocket_route(path, graphql_app)
 
     print(f"Running strawberry on http://{host}:{port}/ 🍓")
-    uvicorn.run(app, host=host, port=port, log_level="error")
+    uvicorn.run(app, loop="none", host=host, port=port, log_level="error")
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,4 @@
+Release type: patch
+
+This releases fixes an issue with the debug server that prevented the
+usage of dataloaders, see: https://github.com/strawberry-graphql/strawberry/issues/940


strawberry/cli/commands/server.py
```diff
@@ -57,4 +57,4 @@ def server(schema, host, port, app_dir):
         app.add_websocket_route(path, graphql_app)
 
     print(f"Running strawberry on http://{host}:{port}/ 🍓")
-    uvicorn.run(app, host=host, port=port, log_level="error")
+    uvicorn.run(app, loop="none", host=host, port=port, log_level="error")
```
patrick91, line 60: Is this documented in the uvicorn docs? I was only able to find it here: https://github.com/encode/uvicorn/blob/bf1c64e2c141971c546671c7dc91b8ccf0afeb7d/uvicorn/config.py#L55-L60

I'll test this with the dataloader example and see if it works :)

Thank you!

DoctorJohn: It appears to be undocumented on their website which matches the uvicorn cli help command. Maybe it was intended for programmatically usage only. However, according to [this PR](https://github.com/encode/uvicorn/pull/455) they added it specifically to allow the setup of custom event loops.

patrick91 (approved):
Just tested and it seems to work! Good spot!

botberry:



---

This releases fixes an issue with the debug server that prevented the
usage of dataloaders, see: https://github.com/strawberry-graphql/strawberry/issues/940

Issue #849, opened 2021-04-13
Internals restructure


patrick91:
@BryceBeagle and I had an impromptu chat last Friday about restructuring the internals. Our current structure is a bit messy, and moving forwards we want to 

<details>
<summary>Current structure</summary>

<img src="https://user-images.githubusercontent.com/667029/114539704-a7b47080-9c4c-11eb-8714-2787c6f47eb7.png">
</details>

<details>
<summary>Proposed structure</summary>

<img src="https://user-images.githubusercontent.com/667029/114540597-a172c400-9c4d-11eb-9068-50b45e691656.jpg">
</details>

we also have miro board with the structure, which should make it easier to read and navigate: https://miro.com/app/board/o9J_lKpif3E=/?moveToWidget=3074457357166650259&cot=14

This also takes into account @BryceBeagle's proposed re-architecture of the types and in general feels much more organised as a structure.

**Note** that this is not going to touch public api and imports. The idea is that public imports should be short and memorable

@BryceBeagle feel free to edit this and put more details, I wrote it quickly during a break 😊

scratchmex:
So far the structure seems very nice. I only have one question, why is the core module inside types? Should not be swapped? `strawberry.core.types`?

BryceBeagle:
IIRC, these were supposed to be the "core" types of strawberry. The other things in the `strawberry.types` directory aren't things that go on the schema and were to be just related to typing (e.g. `unset` and `private`). I don't remember why there's both a `strawberry.types.scalar` and a `strawberry.types.core.StrawberryScalar`, though. Maybe @patrick91 does?

PR #949, opened 2021-05-18
strawberry-graphql-django package integration

la4de:
This PR integrates `strawberry-graphql-django` package into strawberry core so that it is possible to use django package directly through `strawberry.django` namespace.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,10 @@
+release type: patch
+
+This release integrates the `strawberry-graphql-django` package into Strawberry
+core so that it's possible to use the Django extension package directly via the
+`strawberry.django` namespace.
+
+You still need to install `strawberry-graphql-django` if you want to use the
+extended Django support.
+
+See: https://github.com/strawberry-graphql/strawberry-graphql-django
--- a/strawberry/django/__init__.py
+++ b/strawberry/django/__init__.py
@@ -1,0 +1,18 @@
+try:
+    # import modules and objects from external strawberry-graphql-django
+    # package so that it can be used through strawberry.django namespace
+    from strawberry_django import *  # noqa: F401, F403
+except ModuleNotFoundError:
+    import importlib
+
+    def __getattr__(name):
+        # try to import submodule and raise exception only if it does not exist
+        import_symbol = f"{__name__}.{name}"
+        try:
+            return importlib.import_module(import_symbol)
+        except ModuleNotFoundError:
+            raise AttributeError(
+                f"Attempted import of {import_symbol} failed. Make sure to install the"
+                "'strawberry-graphql-django' package to use the Strawberry Django "
+                "extension API."
+            )


strawberry/django/__init__.py
```diff
@@ -0,0 +1,6 @@
+try:
+    # import modules and objects from external strawberry-graphql-django
+    # package so that it can be used through strawberry.django namespace
+    from strawberry_django import *  # noqa: F401, F403
+except ModuleNotFoundError:
+    pass
```
BryceBeagle, line 6: Do we want to add a custom [module `__getattr__`](https://www.python.org/dev/peps/pep-0562/) that tells the user to install the extension package if they try using it without installing it?

la4de: I thought about it and I'd like to do that.

The challenge is that we have few modules already under django subfolder and we should probably move them into somewhere else if would like to do that.

Let's take an example.
```python
# this should raise an exception and nice tip
strawberry.django.type

# this and few other submodules should still work
strawberry.django.views.GraphQLView
```

BryceBeagle: https://github.com/strawberry-graphql/strawberry/issues/849 should hopefully help with that

botberry:



---

This release integrates `strawberry-graphql-django` package into strawberry core so that it is possible to use django package directly through `strawberry.django` namespace.

strawberry/django/__init__.py
la4de: > Do we want to add a custom module __getattr__ that tells the user to install the extension package if they try using it without installing it?

I tried module level `__getattr__` and it seems to work pretty well. 

Please let me know if this is feasible solution or if you think I need to change anything else.

BryceBeagle: Can you still use `strawberry.django.views.GraphQLView` if the extension isn't installed?

la4de: Yes, you can use GraphQLView in the way we have it in our documentation.
https://strawberry.rocks/docs/integrations/django

```python
```

but unfortunately following does not work anymore because it tries to access `views` through `__getattr__`
```python
from strawberry.django import views
```

do you think if that's a problem? I can remove the last commit if you think so.

BryceBeagle: Can we just add a bit of logic to `__getattr__` to try the import, and if that fails only then throw an exception?

la4de: Probably yes, Let me try :)

BryceBeagle: https://stackoverflow.com/a/49412969/8134178

Oh this may not even be a problem

la4de: All following import variants are working with latest change
```python
from strawberry.django import views
import strawberry.django.views
```

strawberry/django/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+
+    def __getattr__(name):
+        # try to import submodules and raise exception only if it does not exist
+        try:
+            return importlib.import_module(f"{__name__}.{name}")
+        except ModuleNotFoundError:
+            raise ModuleNotFoundError(
+                "Install 'strawberry-graphql-django' package to use"
+                " Strawberry Django API"
+            )
```
BryceBeagle, line 16:
```suggestion
    def __getattr__(name):
        # try to import submodules and raise exception only if it does not exist
        import_symbol = f"{__name__}.{name}"
        try:
            return importlib.import_module(import_symbol)
        except ModuleNotFoundError:
            raise ModuleNotFoundError(
                f"Attempted import of {import_symbol} failed. Make sure to install the"
                "'strawberry-graphql-django' package to use the Strawberry Django "
                "extension API."
            )
```

RELEASE.md
```diff
@@ -0,0 +1,3 @@
+release type: patch
+
+This release integrates `strawberry-graphql-django` package into strawberry core so that it is possible to use django package directly through `strawberry.django` namespace.
```
```suggestion
Integrate the `strawberry-graphql-django` package into Strawberry core so that it's 
possible to use the Django extension package directly through the `strawberry.django`
namespace.
```

BryceBeagle (approved):
Couple minor wording change suggestions, then it LGTM :+1:

BryceBeagle (changes requested):
Wait, tests fail

strawberry/django/__init__.py
la4de: Updated as suggested.

strawberry/django/__init__.py
la4de: I pushed suggested changes to the branch. Test are passing now after changing exception from `ModuleNotFoundError`to `AttributeError`.

la4de: Haha, README.md check did not pass. Need to fix that too :)

BryceBeagle: pre-commit check still fails because of an errant trailing space :slightly_smiling_face:

la4de: > pre-commit check still fails because of an errant trailing space 🙂

fixed :)

BryceBeagle (approved):

strawberry/django/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+        # try to import submodule and raise exception only if it does not exist
+        import_symbol = f"{__name__}.{name}"
+        try:
+            return importlib.import_module(import_symbol)
+        except ModuleNotFoundError:
+            raise AttributeError(
+                f"Attempted import of {import_symbol} failed. Make sure to install the"
+                "'strawberry-graphql-django' package to use the Strawberry Django "
+                "extension API."
+            )
```
patrick91, line 18: nice!

RELEASE.md
```diff
@@ -0,0 +1,5 @@
+release type: patch
+
+Integrate the `strawberry-graphql-django` package into Strawberry core so that it's
+possible to use the Django extension package directly through the `strawberry.django`
+namespace.
```
patrick91, line 5:
```suggestion
This release tntegrates the `strawberry-graphql-django` package into Strawberry core so that it's possible to use the Django extension package directly via the `strawberry.django`
namespace.

You still need to install `strawberry-graphql-django` if you want to use the extended django support.

See: https://github.com/strawberry-graphql/strawberry-graphql-django
```

patrick91 (approved):
Looks good to me! Only suggested a change to the release notes :)

RELEASE.md
la4de: RELEASE.md updated

la4de:
Release note updated. Is it good time to push the merge button? :)

patrick91:
Yes! Will do now! ☺️

PR #950, opened 2021-05-19
Start permissions documentation

taybenlor:
Adds basic examples from permissions tests, shows how errors will be returned and explains how this could be combined with `get_context` to obtain details from the `Request` object.

## Description

This change adds documentation for the current state of the `BasePermission` class and its behaviour in the system. I am not particularly familiar with Strawberry, but I needed this documentation myself, so I thought a good learning process would be to write it.

To help readers understand the documentation I've explained how the `has_permission` method will return an error, and included details about how you might implement the `has_permission` method. Including suggesting how to add details to the `Info.context` object.

I haven't checked that my approach works, and I don't know if this is the suggested way to do permission. However I figured that just writing something would be a good start and would allow for iteration.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation

Not sure, seems like documentation is in meta-issues

## Checklist

Hi, I haven't done any of this! I've just submitted this through the github editor. Would love pointers if there is anything I'm missing. I couldn't see anything in CONTRIBUTING about documentation.

### DIFF
--- a/docs/guides/permissions.md
+++ b/docs/guides/permissions.md
@@ -25,10 +25,12 @@ class Query:
     user: str = strawberry.field(permission_classes=[IsAuthenticated])
 ```
 
-If the `has_permission` method returns `False` or a falsy value then an error
-will be raised using the `message` class attribute. See
-[Dealing with Errors](/docs/guides/errors) for more information on how errors
-are handled.
+Your `has_permission` method should do the work to check if this request
+has permission to the field. If the `has_permission` method returns `True` or
+a truthy value then the field access will go ahead. If the `has_permission`
+method returns `False` or a falsy value then an error will be raised using
+the `message` class attribute. See [Dealing with Errors](/docs/guides/errors)
+for more information on how errors are handled.
 
 ```json
 {
@@ -40,6 +42,7 @@ are handled.
   ]
 }
 ```
+
 
 ## Accessing user information
 
--- a/docs/guides/permissions.md
+++ b/docs/guides/permissions.md
@@ -35,41 +35,29 @@ If the `has_permission` method fails then an error will be raised using the `mes
 }
 ```
 
-In a simple use case you could obtain all the information you need to execute the `has_permission` check from the `source` and `info` objects.
-
-For more complex cases it might need more information from the Request. This can be provided by extending the `GraphQL` class to provide more context. The `Permission` class can then use that context from the `Info` object.
+In most use cases you can obtain all the information you need to execute the `has_permission` check from the `source` and `info` objects. For example the `info` object contains a `context` dictionary, which includes the Request object. The Request object can be used to retrieve headers, query parameters, or cookies that can be used for auth like so:
 
 ```python
-from strawberry.asgi import GraphQL
+from myauth import authenticate_header, authenticate_query_param
 
 from starlette.requests import Request
 from starlette.websockets import WebSocket
-
-
-class MyGraphQL(GraphQL):
-    async def get_context(self, request: typing.Union[Request, WebSocket]) -> Any:
-        return {
-            "auth": request.headers["Authorization"] or request.query_params["auth"]
-        }
-```
-
-In this example the `MyGraphQL` class extends the Strawberry `GraphQL` class to provide a `get_context` helper. This helper is used to fill the `Info` object. You should be aware that `Request` and `WebSocket` objects have different attributes and methods, and you may need to handle retrieving authorization details from them differently.
-
-*For more information on using context see the [Data Loaders](/docs/guides/dataloaders) docs.*
-
-The `Info` object can then be accessed in a `Permission` class like:
-
-```python
-from myauth import authenticate
 
 class IsAuthenticated(BasePermission):
     message = "User is not authenticated"
 
     def has_permission(self, source: Any, info: Info, **kwargs) -> bool:
-        return authenticate(info.context["auth"])
+        request: Union[Request, Websocket] = info.context["request"]
+        if "Authorization" in request.headers:
+          return authenticate_header(request)
+        if "auth" in request.query_params:
+          return authenticate_query_params(request)
+        return False
 ```
 
-Here we are taking the `auth` key out of the `info.context` dictionary and passing it to an `authenticate` method we have implemented somewhere else in our codebase.
+Here we retrieve the `request` object from the `context` provided by `info`. This object will be either a `Request` or `Websocket` instance from `starlette` (see: [Request docs](https://www.starlette.io/requests/) and [Websocket docs](https://www.starlette.io/websockets/)).
+
+In the next step we take either the `Authorization` header or the `auth` query parameter out of the `request` object, depending on which is available. We then pass those on to some authenticate methods we've implemented ourselves.
 
 Beyond providing hooks, Authentication is not currently Strawberry's responsibility. You should provide your own helpers to figure out if a request has the permissions you expect.
 
--- a/docs/guides/permissions.md
+++ b/docs/guides/permissions.md
@@ -6,7 +6,7 @@ title: Permissions
 
 Permissions can be managed using `Permission` classes. A `Permission` class
 extends `BasePermission` and has a `has_permission` method. It can be added to a
-field using the `permission_classes` keyword argument. A simple example looks
+field using the `permission_classes` keyword argument. A basic example looks
 like this:
 
 ```python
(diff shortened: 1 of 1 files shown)


docs/guides/permissions.md
```diff
@@ -4,4 +4,73 @@ title: Permissions
 
 # Permissions
 
-Documentation coming soon
+Permissions can be managed using `Permission` classes. A `Permission` class extends `BasePermission` and has a `has_permission` method. It can be hooked up to a field using the `permission_classes` keyword argument. A simple example looks like this:
```
jkimbo, line 7:
```suggestion
Permissions can be managed using `Permission` classes. A `Permission` class extends `BasePermission` and has a `has_permission` method. It can be added to a field using the `permission_classes` keyword argument. A simple example looks like this:
```

docs/guides/permissions.md
```diff
@@ -4,4 +4,73 @@ title: Permissions
...
+
+    def has_permission(self, source: Any, info: Info, **kwargs) -> bool:
+        return False
+
+@strawberry.type
+class Query:
+    user: str = strawberry.field(permission_classes=[IsAuthenticated])
+```
+
+If the `has_permission` method fails then an error will be raised using the `message` class attribute. See [Dealing with Errors](/docs/guides/errors) for more information.
```
jkimbo, line 25:
```suggestion
If the `has_permission` method fails then an error will be raised using the `message` class attribute. See [Dealing with Errors](/docs/guides/errors) for more information on how errors are handled.
```

docs/guides/permissions.md
```diff
@@ -4,4 +4,73 @@ title: Permissions
...
+    {
+      "message": "User is not authenticated"
+    }
+  ]
+}
+```
+
+In a simple use case you could obtain all the information you need to execute the `has_permission` check from the `source` and `info` objects.
+
+For more complex cases it might need more information from the Request. This can be provided by extending the `GraphQL` class to provide more context. The `Permission` class can then use that context from the `Info` object.
```
jkimbo, line 40: Most (maybe all?) of the views already pass the request as part of the context object. Wouldn't it be better to extract what you need in the permission class rather than modifying the context?

jkimbo (changes requested):
Thanks for the contribution! I've added a couple of comments. Also I'm not sure the part explaining how to modify the context is necessary here.

docs/guides/permissions.md
taybenlor: I didn't know that! I'll update the documentation.

In this case I'm basing the documentation off a solution I used to place a `user` object into the `context` (using the `Authorization` header of the request).

taybenlor:
Thanks! I understand not wanting to introduce using the `context` object in this particular guide, but it was something I found useful while implementing. I think if not here, then there should be a specific guide on using it.

taybenlor:
After trying to understand the properties available on the `context` object I think it would be very useful to document it.

taybenlor:
I've finished up changes now and ready for re-review.

docs/guides/permissions.md
```diff
@@ -4,4 +4,61 @@ title: Permissions
...
+
+class IsAuthenticated(BasePermission):
+    message = "User is not authenticated"
+
+    def has_permission(self, source: Any, info: Info, **kwargs) -> bool:
+        request: Union[Request, Websocket] = info.context["request"]
+        if "Authorization" in request.headers:
+          return authenticate_header(request)
+        if "auth" in request.query_params:
+          return authenticate_query_params(request)
```
jkimbo, line 54:
```suggestion
        if "Authorization" in request.headers:
            return authenticate_header(request)
        if "auth" in request.query_params:
            return authenticate_query_params(request)
```

docs/guides/permissions.md
```diff
@@ -4,4 +4,61 @@ title: Permissions
...
+  "data": null,
+  "errors": [
+    {
+      "message": "User is not authenticated"
+    }
+  ]
+}
+```
+
+In most use cases you can obtain all the information you need to execute the `has_permission` check from the `source` and `info` objects. For example the `info` object contains a `context` dictionary, which includes the Request object. The Request object can be used to retrieve headers, query parameters, or cookies that can be used for auth like so:
```
jkimbo, line 38:
```suggestion

## Accessing user information

Accessing the current user information to implement your permission checks depends on the web framework you are using. Most frameworks will have a Request object where you can either access the current user directly or access headers/cookies/query parameters to authenticate the user. All the Strawberry integrations will provide this Request object in the `info.context` object that is accessible in every resolver and in the `has_permission` function. You can find more details about a specific framework integration under the "Integrations" heading in the navigation. 

In this example we are using `starlette` which uses the [ASGI](/docs/integrations/asgi) integration:
```

jkimbo (changes requested):
I agree that documenting what is available on the context object is a good idea but I think that should go in the specific integration pages because they will differ slightly between them. Like the example you've provided for starlette as well. I've added some suggestions on how to reword it to make it clear that the specifics on how you implement the auth checks are going to differ between integrations. I don't want to imply that Strawberry always uses starlette for example.

botberry:

```markdown
Release type: patch

```

docs/guides/permissions.md
taybenlor: Ahhh of course. I hadn't realised they would be different.

taybenlor:
Fantastic, thanks for the suggestions. I'm happy with where that is now.

docs/guides/permissions.md
```diff
@@ -4,4 +4,66 @@ title: Permissions
...
+
+    def has_permission(self, source: Any, info: Info, **kwargs) -> bool:
+        return False
+
+@strawberry.type
+class Query:
+    user: str = strawberry.field(permission_classes=[IsAuthenticated])
+```
+
+If the `has_permission` method fails then an error will be raised using the `message` class attribute. See [Dealing with Errors](/docs/guides/errors) for more information on how errors are handled.
```
BryceBeagle, line 25: Might want to be explicit on what `fails` means. Returns false? Throws an exception?

docs/guides/permissions.md
```diff
@@ -4,4 +4,66 @@ title: Permissions
...
+      "message": "User is not authenticated"
+    }
+  ]
+}
+```
+
+
+## Accessing user information
+
+Accessing the current user information to implement your permission checks depends on the web framework you are using. Most frameworks will have a Request object where you can either access the current user directly or access headers/cookies/query parameters to authenticate the user. All the Strawberry integrations will provide this Request object in the `info.context` object that is accessible in every resolver and in the `has_permission` function. You can find more details about a specific framework integration under the "Integrations" heading in the navigation.
```
BryceBeagle, line 41:
```suggestion
Accessing the current user information to implement your permission checks depends on the web framework you are using. Most frameworks will have a `Request` object where you can either access the current user directly or access headers/cookies/query parameters to authenticate the user. All the Strawberry integrations provide this Request object in the `info.context` object that is accessible in every resolver and in the `has_permission` function. You can find more details about a specific framework integration under the "Integrations" heading in the navigation. 
```

BryceBeagle: > You can find more details about a specific framework integration under the "Integrations" heading in the navigation

Can we make this a link?

BryceBeagle (changes requested):
Can we add some line wrapping to each long line to match the style of the other docs in the repo?

patrick91:
This is amazing! Thanks a lot for taking time to write this, let me know if @BryceBeagle and @jkimbo comments make sense 😊

docs/guides/permissions.md
jkimbo: @BryceBeagle not at the moment. It's just a heading in the sidebar and not an actual page.

docs/guides/permissions.md
taybenlor: Have cleared this up as `False` or falsy.

taybenlor:
Thanks @patrick91!

@BryceBeagle have reworded and reformatted per review.

2malh:
I think the guide is missing the info that `has_permission` must
- a) contain some logic and 
- b) return a value other than False to properly execute without raising message.

To include Django as an example for `IsAuthenticated`
```py
def has_permission(self, source: Any, info: Info, **kwargs) -> bool:
        return info.context.request.user.is_authenticated
```

taybenlor:
Thanks @2malh - have fixed this.

@BryceBeagle and @jkimbo your feedback has also been fixed.

jkimbo (approved):
Thanks @taybenlor

taybenlor:
Cheers!

PR #966, opened 2021-05-23
Fix aiohttp subscription max recursion depth error

DoctorJohn:
## Description

This PR fixes a max recursion depth error in the AIOHTTP subscription implementation.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* None

## Checklist

### DIFF
--- a/strawberry/aiohttp/views.py
+++ b/strawberry/aiohttp/views.py
@@ -166,9 +166,9 @@ class WebSocketHandler(BaseGraphQLView, ABC):
         await self.cleanup_operation(operation_id)
 
     async def handle_keep_alive(self, ws: web.WebSocketResponse) -> None:
-        await asyncio.sleep(self.keep_alive_interval)
-        await ws.send_json({"type": GQL_CONNECTION_KEEP_ALIVE})
-        await self.handle_keep_alive(ws)
+        while True:
+            await asyncio.sleep(self.keep_alive_interval)
+            await ws.send_json({"type": GQL_CONNECTION_KEEP_ALIVE})
 
     async def handle_async_results(
         self,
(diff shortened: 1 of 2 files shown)


botberry:



---

This release fixes a max recursion depth error in the AIOHTTP subscription
implementation.

strawberry/aiohttp/views.py
```diff
@@ -166,9 +166,9 @@ async def handle_stop(self, message: dict) -> None:
         await self.cleanup_operation(operation_id)
 
     async def handle_keep_alive(self, ws: web.WebSocketResponse) -> None:
-        await asyncio.sleep(self.keep_alive_interval)
-        await ws.send_json({"type": GQL_CONNECTION_KEEP_ALIVE})
-        await self.handle_keep_alive(ws)
+        while True:
+            await asyncio.sleep(self.keep_alive_interval)
+            await ws.send_json({"type": GQL_CONNECTION_KEEP_ALIVE})
```
patrick91, line 171: oh, so the recursion was happening when the code was left running for a long time, right?

patrick91: should we do the same here? https://github.com/strawberry-graphql/strawberry/blob/aa3ab0e51a25a7b2b0c664820187c82d2783e27b/strawberry/asgi/__init__.py#L70-L79

DoctorJohn: > oh, so the recursion was happening when the code was left running for a long time, right?

exactly

> should we do the same here? 

I don't think the ASGI implementation has the same problem since it's using `asycio.create_task` which is non blocking as far as I know.

(However, I planned to refactor the ASGI subscription implementation in a separate PR to bring both implementations closer together and increase ASGI test coverage.)

patrick91 (approved):
👍 nice work!

strawberry/aiohttp/views.py
patrick91: gotcha, thanks for explaining!

PR #980, opened 2021-05-31
Add e2e tests for federation

marcoacierno:
## Description

Run federation tests on a real infrastructure to make sure it works

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/.github/workflows/e2e-federation-tests.yml
+++ b/.github/workflows/e2e-federation-tests.yml
@@ -1,0 +1,37 @@
+name: e2e Federation tests
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - 'e2e-federation-tests/**/*'
+      - 'strawberry/federation.py'
+      - 'strawberry/printer.py'
+      - 'tests/federation/**/*'
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.7"
+      - name: Install poetry
+        run: pip install poetry
+      - name: Start products service
+        uses: strawberry-graphql/e2e-federation/.github/start-services@main
+        with:
+          service: products
+      - name: Start orders service
+        uses: strawberry-graphql/e2e-federation/.github/start-services@main
+        with:
+          service: orders
+      - name: Start gateway service
+        uses: strawberry-graphql/e2e-federation/.github/start-services@main
+        with:
+          service: gateway
+      - name: Install project dependencies
+        run: poetry install
+      - name: Run e2e federation tests
+        run: poetry run pytest e2e-federation-tests/
--- a/.github/workflows/e2e-federation-tests.yml
+++ b/.github/workflows/e2e-federation-tests.yml
@@ -4,19 +4,25 @@ on:
   pull_request:
     branches: [main]
     paths:
-      - 'e2e-federation-tests/**/*'
-      - 'strawberry/federation.py'
-      - 'strawberry/printer.py'
-      - 'tests/federation/**/*'
+      - "e2e-federation-tests/**/*"
+      - "strawberry/**"
+      - "tests/federation/**/*"
+      - "pyproject.toml"
+      - "poetry.lock"
 
 jobs:
   test:
+    name: Federation e2e on ${{ matrix.python-version }}
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.7', '3.8', '3.9', '3.10']
     steps:
       - uses: actions/checkout@master
       - uses: actions/setup-python@v2
         with:
-          python-version: "3.7"
+          python-version: ${{ matrix.python-version }}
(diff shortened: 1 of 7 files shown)


e2e-federation-tests/infrastructure/orders/api/types.py
```diff
@@ -0,0 +1,35 @@
...
+        elif self.id == "2":
+            return [Order(id=ID("2"), customer_name="Patrick")]
+        elif self.id == "500":
+            return [Order(id=ID("3"), customer_name="Ethan Winters")]
+        else:
+            return []
+
+    @classmethod
+    async def resolve_reference(cls, id: ID, name: Optional[str] = None):
+        return cls(id=id, name=name)  # type: ignore
```
marcoacierno, line 35: `type: ignore` because mypy fails because `name` is Optional in `resolve_reference` but it can't be in the Product fields list because it needs to be the same type as on GraphQL

The reason `name` is Optional and by default `= None` is that the Gateway will pass the `name` only if the client asked for the `code` field (i.e the field that `@requires` it)

botberry:

```markdown
Release type: patch

```

e2e-federation-tests/infrastructure/orders/api/types.py
BryceBeagle: Can this comment be added w/in the code?

marcoacierno: sure, will do it (this code is now in the other repo)

marcoacierno: Have a look here: https://github.com/strawberry-graphql/e2e-federation/pull/1/files

BryceBeagle:
Can you explain how these tests work (and potentially add some brief doc readme into the source)? I'm not sure I fully understand how the other repo comes into play with this GitHub workflow?

marcoacierno:
> Can you explain how these tests work (and potentially add some brief doc readme into the source)? I'm not sure I fully understand how the other repo comes into play with this GitHub workflow?

Yeah, so:

There is the github action ".github/workflows/e2e-federation-tests.yml" that is run when a PR changes a file that is inside one of those folders/files:

```
      - 'e2e-federation-tests/**/*'
      - 'strawberry/federation.py'
      - 'strawberry/printer.py'
      - 'tests/federation/**/*'
```

The github action clones the Strawberry repos in the action and then we run a custom action we defined in the **e2e-federation** repo:

```
uses: strawberry-graphql/e2e-federation/.github/start-services@main
```

You can see the code of the action here: https://github.com/strawberry-graphql/e2e-federation/blob/main/.github/start-services/action.yml

When you do `uses: X` github action clones the repo you specify there in a folder `blabla/_actions/your-action` meaning that now we have the Strawberry source code at (/home/runner/work/strawberry) and the other repo in `blabla/_actions/your-action`

Docs on `uses`: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsuses

e2e tests need all the servers running. 
The code of each server inside the e2e-federation repo

So when we did `uses: X` Github cloned our action together with the rest of our git repo, meaning that we have the code of the services.

Now we need to run the servers so we can run our tests:

1. We first need to install the dependencies of each project (https://github.com/strawberry-graphql/e2e-federation/blob/main/.github/start-services/action.yml#L10). Only the gateway uses Typescript, so we use yarn there otherwise we run poetry install.
2. Once the dependencies are installed we can just run the services like we would do locally: https://github.com/strawberry-graphql/e2e-federation/blob/main/.github/start-services/action.yml#L24 With the difference that we do `&` at the end of the command so they are moved in background and do not block the github action

When we install the dependencies we install the Strawberry from the PR where is it running, see the pyproject here: https://github.com/strawberry-graphql/e2e-federation/blob/main/orders/pyproject.toml#L9

(Note we need an absolute fixed path due to a bug of Poetry we cannot do `poetry add [github workspace]` dynamically)

Each service starts on hard-coded ports (7000, 7100, 7200)

Once the services are up and running we install Strawberry dependencies and run pytest on e2e-federation-tests that contain tests that run on the real infrastructure

@BryceBeagle please let me know if it's clear now? I probably can write something better for the readme later

BryceBeagle:
It's fairly clear now, thanks.

Is there a reason that we don't always run these tests, instead of just for some path changes? Do they take a long time?

marcoacierno:
>Is there a reason that we don't always run these tests, instead of just for some path changes? Do they take a long time?

it doesn't take much time for now, and I don't think they will ever take lots of time to run as I doubt we will do anything more complex than what is done right now.

The reason I decided to run them only when editing federation related fields is that it seemed kind of a waste to keep always ran them when nothing changed

(but I am also happy to change it to always run them :))

DoctorJohn:
Could a change to the strawberry core break federation support? If so, it might be a good idea to always run the tests.

patrick91:
@marcoacierno closing this as we have the Apollo Federation test action now 😊

PR #991, opened 2021-06-02
Allow creating pre-release of PRs

marcoacierno:
## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/.github/workflows/pre-release-pr.yml
+++ b/.github/workflows/pre-release-pr.yml
@@ -1,0 +1,83 @@
+name: Release test pre-release version
+
+on:
+  repository_dispatch:
+    types: [pre-release-command]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.client_payload.pull_request.head.sha }}
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+      - name: Install deps
+        run: |
+          python -m pip install pip --upgrade
+          pip install poetry
+          pip install githubrelease
+          pip install autopub
+          pip install httpx
+      - name: Check if release exists
+        id: check_release
+        run: |
+          set +e
+          echo ::set-output name=release::$(autopub check)
+      - name: Metadata
+        id: metadata
+        if: steps.check_release.outputs.release == ''
+        run: echo "::set-output name=commit::$(git rev-parse HEAD)"
+      - name: Find Release Comment
+        uses: peter-evans/find-comment@v1
+        id: find_comment
+        if: steps.check_release.outputs.release == ''
+        with:
+          token: ${{ secrets.BOT_TOKEN }}
+          issue-number: ${{ github.event.client_payload.github.payload.issue.number }}
+          comment-author: botberry
+          body-includes: "# Pre-release"
+      - name: Create or update comment
+        uses: peter-evans/create-or-update-comment@v1
+        if: steps.check_release.outputs.release == ''
+        with:
+          token: ${{ secrets.BOT_TOKEN }}
+          comment-id: ${{ steps.find_comment.outputs.comment-id }}
+          issue-number: ${{ github.event.client_payload.github.payload.issue.number }}
+          body: |
+            # Pre-release
+            :wave:
+            Releasing commit [${{ steps.metadata.outputs.commit }}] to PyPi as pre-release! :package:
+          edit-mode: replace
+      - name: Publish pre-release version
+        if: steps.check_release.outputs.release == ''
+        id: release
+        env:
+          GITHUB_TOKEN: ${{ secrets.BOT_TOKEN }}
+        run: |
+          autopub prepare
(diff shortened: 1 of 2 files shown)


botberry:

```markdown
Release type: patch

```

.github/workflows/pre-release-pr.yml
```diff
@@ -0,0 +1,83 @@
...
+          python -m pip install pip --upgrade
+          pip install poetry
+          pip install githubrelease
+          pip install autopub
+          pip install httpx
+      - name: Check if release exists
+        id: check_release
+        run: |
+          set +e
+          echo ::set-output name=release::$(autopub check)
```
BryceBeagle, line 29: What's the `set +e` for?

.github/workflows/pre-release-pr.yml
```diff
@@ -0,0 +1,83 @@
...
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.client_payload.pull_request.head.sha }}
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
```
BryceBeagle, line 17: Do we care about using a matrix for different python versions?

.github/workflows/pre-release-pr.yml
```diff
@@ -0,0 +1,83 @@
+name: Release test pre-release version
+
+on:
+  repository_dispatch:
+    types: [pre-release-command]
```
BryceBeagle, line 5: Could we not use a `workflow_dispatch` event w/o the bot?

.github/workflows/pre-release-pr.yml
```diff
@@ -0,0 +1,83 @@
...
+      - name: Publish pre-release version
+        if: steps.check_release.outputs.release == ''
+        id: release
+        env:
+          GITHUB_TOKEN: ${{ secrets.BOT_TOKEN }}
+        run: |
+          autopub prepare
+          poetry version $(poetry version -s)-dev.$(date '+%s')
+          poetry build
+          poetry publish --username __token__ --password ${{ secrets.PYPI_TOKEN }}
```
BryceBeagle, line 64: Could we use the `POETRY_PYPI_TOKEN_PYPI` env var instead? I'm always hesitant to provide secrets in the command line. I know GitHub Actions attempts to obfuscate secretes, but I think it would still be better to use the env var

.github/workflows/pre-release-pr.yml
```diff
@@ -0,0 +1,83 @@
...
+          pip install httpx
+      - name: Check if release exists
+        id: check_release
+        run: |
+          set +e
+          echo ::set-output name=release::$(autopub check)
+      - name: Metadata
+        id: metadata
+        if: steps.check_release.outputs.release == ''
+        run: echo "::set-output name=commit::$(git rev-parse HEAD)"
```
BryceBeagle, line 33: Nitpick: your use of quotes around your `::set-output` echos seems inconsistent

BryceBeagle (changes requested):

.github/workflows/pre-release-pr.yml
marcoacierno: What would be the point? 🤔 seeing if the library builds with different python versions? I think we already have the unit-tests for that

.github/workflows/pre-release-pr.yml
marcoacierno: I just used what the library suggested, do you know why we would prefer workflow_dispatch over repository_dispatch? It has been a while since I used the former so I am not 100% sure why it exists

.github/workflows/pre-release-pr.yml
marcoacierno: Sure, I will also change it in the the other workflow if we are worried about leaking it

.github/workflows/pre-release-pr.yml
BryceBeagle: If that's what the library suggests, then we may as well just stick to it. My understanding is that `workflow_dispatch` is for github->github events, while `repository_dispatch` is for external->github events.

marcoacierno: I will read about this a bit and see what makes more sense

.github/workflows/pre-release-pr.yml
patrick91: we should use latest :D

.github/workflows/pre-release-pr.yml
marcoacierno: It changes how the shell behaves when a command exits with a status code different from zero

I actually might not need it in this case, as we just use `autopub check`, I will have a look at that as well

.github/workflows/pre-release-pr.yml
marcoacierno: I guess this is what happens when you get used to having some tool fix those stuff for joy :D 😂

PR #994, opened 2021-06-03
Import django into strawberry namespace

la4de:
This fixes django symbol resolution in some IDE which was originally reported here:
* https://github.com/strawberry-graphql/strawberry-graphql-django/issues/35
* https://github.com/strawberry-graphql/strawberry-graphql-django/issues/36

### DIFF
--- a/strawberry/django/__init__.py
+++ b/strawberry/django/__init__.py
@@ -1,7 +1,7 @@
 try:
     # import modules and objects from external strawberry-graphql-django
     # package so that it can be used through strawberry.django namespace
-    from strawberry_django import *  # noqa: F401, F403
+    from strawberry_django import *  # type: ignore # noqa: F401, F403
 except ModuleNotFoundError:
     import importlib
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+release type: patch
+
+This release fixes the IDE integration where package `strawberry.django` could not be find by some editors like vscode.
--- a/strawberry/__init__.py
+++ b/strawberry/__init__.py
@@ -1,4 +1,4 @@
-from . import experimental, federation
+from . import django, experimental, federation
 from .arguments import argument
 from .custom_scalar import scalar
 from .directive import directive
@@ -23,6 +23,7 @@ __all__ = [
     "Schema",
     "argument",
     "directive",
+    "django",
     "enum",
     "federation",
     "field",


botberry:



---

This release fixes the IDE integration where package `strawberry.django` could not be find by some editors like vscode.

la4de:
Tests are passing locally but failing on server. Needs more investigation.

la4de:
I finally were able to reproduce the problem locally and fix failures.

@patrick91, would you have time to take a look?

strawberry/django/__init__.py
```diff
@@ -1,7 +1,7 @@
 try:
     # import modules and objects from external strawberry-graphql-django
     # package so that it can be used through strawberry.django namespace
-    from strawberry_django import *  # noqa: F401, F403
+    from strawberry_django import *  # type: ignore # noqa: F401, F403
```
patrick91, line 4: why do we need the type ignore? 😊

patrick91 (approved):
Looks good to me! Just added one question, we can merge after "resolving" that one :D

strawberry/django/__init__.py
la4de: mypy was unhappy because strawberry_django package was not available.

I think we have two options
* install strawberry-graphql-django package
* ignore type and make mypy happy in that way

what's your opinion?

patrick91: considering we are going to merge the packages and the issue will go away by doing that, I'd go with this solution 😊

la4de: I'm getting this error. "fix" was to ignore it

```
/Users/l/git/graphql/strawberry/django/tests/mypy/test_dataloaders.yml:10:
E   pytest_mypy_plugins.utils.TypecheckAssertionError: Invalid output:
E   Expected:
E     main:8: note: Revealed type is 'strawberry.dataloader.DataLoader[Any, Any]' (diff)
E     main:13: note: Revealed type is 'Any'         (diff)
E   Actual:
E     ../../../../../../Users/l/git/graphql/strawberry/django/strawberry/django/__init__:4: error: Cannot find implementation or library stub for module named 'strawberry_django' (diff)
E     ../../../../../../Users/l/git/graphql/strawberry/django/strawberry/django/__init__:4: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports (diff)
E     main:8: note: Revealed type is 'strawberry.dataloader.DataLoader[Any, Any]' (diff)
E     main:13: note: Revealed type is 'Any'         (diff)
E
E   Alignment of first line difference:
E     E: main:8: note: Revealed type is 'strawberry.dataloader.DataLoader[Any, An...
E     A: ../../../../../../Users/l/git/graphql/strawberry/django/strawberr...
```

patrick91: I see, fine to ignore it for now :)

la4de:
> Looks good to me! Just added one question, we can merge after "resolving" that one :D

It was fast. Thanks =)

PR #1016, opened 2021-06-14
 Introduce schema configurations [fix #850]

### LINKED ISSUE #850: Introduce schema configurations

peach-lasagna:
## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

#850 

## Checklist

### DIFF
--- a/strawberry/config.py
+++ b/strawberry/config.py
@@ -1,4 +1,4 @@
-from enum import IntEnum
+# from enum import IntEnum
 
 # class EnumGenerate(IntEnum):
 #     MEMBER = 0
--- a/strawberry/config.py
+++ b/strawberry/config.py
@@ -1,0 +1,10 @@
+from enum import IntEnum
+
+class EnumGenerate(IntEnum):
+    MEMBER = 0
+    VALUE  = 1
+
+class StrawberryConfig:
+    auto_camelcase: bool = True
+    resolvers_strict_mode: bool = True
+    enum_values = EnumGenerate.MEMBER # or VALUE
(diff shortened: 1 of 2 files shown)


strawberry/config.py
```diff
@@ -0,0 +1,10 @@
+from enum import IntEnum
+
+class EnumGenerate(IntEnum):
+    MEMBER = 0
+    VALUE  = 1
+
+class StrawberryConfig:
+    auto_camelcase: bool = True
+    resolvers_strict_mode: bool = True
+    enum_values = EnumGenerate.MEMBER # or VALUE
```
patrick91, line 10: what do you think of only supporting auto_camel_case for now? So we can have this new configuration option that also provides a new future?

So I'd have this for now:

```
@dataclass
class StrawberryConfig:
    auto_camelcase: bool = True
```

Then we can expand in future?

peach-lasagna: Yes, ok

peach-lasagna:
@patrick91

Issue #1017, opened 2021-06-14
0.65.4 – possible breaking change – "partially initialized module 'strawberry' has no attribute 'type'" (circular import?)


2malh:
Hey, upgrading to 0.65.4 throws me the following error:
```py
Exception in thread django-main-thread:
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/threading.py", line 954, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.9/threading.py", line 892, in run
    self._target(*self._args, **self._kwargs)
  File "/venv/lib/python3.9/site-packages/django/utils/autoreload.py", line 64, in wrapper
    fn(*args, **kwargs)
  File "/venv/lib/python3.9/site-packages/django/core/management/commands/runserver.py", line 110, in inner_run
    autoreload.raise_last_exception()
  File "/venv/lib/python3.9/site-packages/django/utils/autoreload.py", line 87, in raise_last_exception
    raise _exception[1]
  File "/venv/lib/python3.9/site-packages/django/core/management/__init__.py", line 375, in execute
    autoreload.check_errors(django.setup)()
  File "/venv/lib/python3.9/site-packages/django/utils/autoreload.py", line 64, in wrapper
    fn(*args, **kwargs)
  File "/venv/lib/python3.9/site-packages/django/__init__.py", line 24, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/venv/lib/python3.9/site-packages/django/apps/registry.py", line 91, in populate
    app_config = AppConfig.create(entry)
  File "/venv/lib/python3.9/site-packages/django/apps/config.py", line 224, in create
    import_module(entry)
  File "/usr/local/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 972, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
[...]

PR #1018, opened 2021-06-14
Only import django when type checking

patrick91:
#994 broke imports when using strawberry and strawberry django. 

To replicate the issue:

```python
import strawberry
```

with both strawberry-graphql and strawberry-graphql-django installed.

This fix should work as only import django when typechecking :)

### DIFF
--- a/strawberry/__init__.py
+++ b/strawberry/__init__.py
@@ -1,4 +1,6 @@
-from . import django, experimental, federation
+from typing import TYPE_CHECKING
+
+from . import experimental, federation
 from .arguments import argument
 from .custom_scalar import scalar
 from .directive import directive
@@ -12,6 +14,10 @@ from .scalars import ID
 from .schema import Schema
 from .type import input, interface, type
 from .union import union
+
+
+if TYPE_CHECKING:
+    from . import django
 
 
 __all__ = [
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+This release fixes an import error when using strawberry with strawberry django.


patrick91:
Should fix https://github.com/strawberry-graphql/strawberry/issues/1017

botberry:



---

This release fixes an import error when using strawberry with strawberry django.

strawberry/__init__.py
```diff
@@ -14,6 +16,10 @@
 from .union import union
 
 
+if TYPE_CHECKING:
+    from . import django
+
+
 __all__ = [
```
patrick91, line 23: not sure if should change the all though (as it now includes django)

BryceBeagle: Well the reason we have `__all__` is _for_ type checking

patrick91: Does it do anything at run time though?

BryceBeagle: Afaik, the only runtime effect it has is on `from strawberry import *`

BryceBeagle: But we could just dynamically add to `__all__`:
```python
if TYPE_CHECKING:
    __all__.append("django")
```

patrick91: Yeah, I was thinking that, but looks ugly :D and I’m not 100% sure type checker would support it, as they don’t run python. 

If it only affects star imports we are good I’d say :)

la4de:
I tried this approach. This fixes circular import, but not help with IDE symbol resolution problem which was the purpose of #994. I propose to revert my change and try another solution. I created PR #1019

Issue #960, opened 2021-05-21
Add utilities for preventing malicious queries


patrick91:
Some ideas here: https://www.apollographql.com/blog/graphql/security/securing-your-graphql-api-from-malicious-queries/

We can definitely implement some of those :)

jkimbo:
Because graphql-core is a direct port of graphql-js I think https://github.com/stems/graphql-depth-limit can be directly ported to Python.

jkimbo:
@MeRuslan points out a another useful library to implement cost analysis: https://github.com/pa-bru/graphql-cost-analysis (ref: https://github.com/strawberry-graphql/strawberry/issues/902#issuecomment-873305453)

Note that graphql-js has decided not to support adding directives through code definitions: https://github.com/graphql/graphql-js/issues/1343

jkimbo:
Another simple thing would be to add an extension that prevents introspection queries. That way you can add/enable it on production only to prevent people from introspecting your schema.

Speedy1991:
Maybe this is also a thing to note:
https://www.apollographql.com/docs/react/api/link/persisted-queries/

Obviousy only useful for closed API's - it makes a gql scheme more "static" with the precalculated hashes

jkimbo:
> Maybe this is also a thing to note:
> https://www.apollographql.com/docs/react/api/link/persisted-queries/

Persisted queries is the ultimate way to protect against malicious queries because it means you can’t execute a query that isn’t trusted. It’s very powerful but requires some effort with tooling to make it work well.

We should provide an extension that lets you implement persisted queries like envelop has: https://www.envelop.dev/plugins/use-persisted-operations

erikwrede:
For reference from strawberry discord on cost calculation (will refine later) 

Roadmap could be the following
## M1: basic cost calc 
- field costs
- defaults for
     - scalar field
     - list of scalars
     - other type (:1 relationship)
     - list of types (:m relationship)
- cost per list item
   - based on limit argument(projected cost exceeds the max cost)
- fitting design of schema directives (so we can do the eval in rust later and frontend clients can write codegens for that 😉 or apollo router plugin)
## M2 advanced cost calc
- projected vs actual cost (limit of list vs actual list length)
- rate limiting (max x cost tokens per second per user via an extension)
## M3 full integration
- dry run query to find out projected cost (no resolvers called) (see shopify api)
- cost headers in all web servers
e.g. 
```
x-actual-query-cost
x-tokens-remaining
[...]

PR #1021, opened 2021-06-15
Add depth limit validator

jkimbo:
## Description

Adds a python port of https://github.com/stems/graphql-depth-limit

Also allow passing custom validators to `execute_sync` and `execute`.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* #960

## Checklist

### DIFF
--- a/docs/guides/tools.md
+++ b/docs/guides/tools.md
@@ -47,19 +47,29 @@ type Query {
 
 ### `depth_limit_validator`
 
-Limit the complexity of queries by their depth.
+Limit the complexity of queries by their depth to protect against malicious
+queries.
 
 ```python
+from graphql import ValidationRule
+
 class DepthLimitOptions(TypedDict):
-    ignore: Union[str, re.Pattern, Callable[[str], bool]]
+    ignore: List[Union[str, re.Pattern, Callable[[str], bool]]]
 
 def depth_limit_validator(
-  max_depth: int,  # The maximum allowed depth for any operation in a GraphQL document
-  options: DepthLimitOptions = None,
-  callback: Callable[Dict[str, int]] = None  # Called each time validation runs.  Receives an Object which is a map of the depths for each operation
+    max_depth: int,
+    options: Optional[DepthLimitOptions] = None,
+    callback: Optional[Callable[Dict[str, int]]] = None
 ) -> ValidationRule:
     ...
 ```
+
+| Parameter name   | Type                       | Default | Description                                                                                            |
+| ---------------- | -------------------------- | ------- | ------------------------------------------------------------------------------------------------------ |
+| max_depth        | `int`                      | N/A     | The maximum allowed depth for any operation in a GraphQL document                                      |
+| options   | `Optional[DepthLimitOptions]` | `None` |                                                                                                             |
+| options.ignore   | `List[Union[str, re.Pattern, Callable[[str], bool]]]` | `None` | Stops recursive depth checking based on a field name. Either a string or regexp to match the name, or a function that reaturns a boolean. |
+| callback         | `Optional[Callable[[Dict[str, int]], None]]` | `None` | Called each time validation runs.  Receives an Object which is a map of the depths for each operation |
 
 Example:
 
@@ -69,10 +79,15 @@ from strawberry.schema import default_validation_rules
 from strawberry.tools import depth_limit_validator
 
 
+# Add the depth limit validator to the list of default validation rules
+validation_rules = (
+  default_validation_rules + [depth_limit_validator(3)]
+)
+
 # assuming you already have a schema
 result = schema.execute_sync(
-  """
-    query {
+    """
+    query MyQuery {
       user {
         pets {
           owner {
@@ -83,11 +98,10 @@ result = schema.execute_sync(
         }
       }
     }
-  """,
-  validation_rules=(
-    default_validation_rules
-    + [depth_limit_validator(2)]
-    )
+    """,
+    validation_rules=validation_rules,
+  )
 )
-assert result.errors
+assert len(result.errors) == 1
+assert result.errors[0].message == "'MyQuery' exceeds maximum operation depth of 3"
 ```
--- a/docs/guides/tools.md
+++ b/docs/guides/tools.md
@@ -53,23 +53,19 @@ queries.
 ```python
 from graphql import ValidationRule
 
-class DepthLimitOptions(TypedDict):
-    ignore: List[Union[str, re.Pattern, Callable[[str], bool]]]
-
 def depth_limit_validator(
     max_depth: int,
-    options: Optional[DepthLimitOptions] = None,
+    ignore: Optional[List[Union[str, re.Pattern, Callable[[str], bool]]]] = None,
     callback: Optional[Callable[Dict[str, int]]] = None
 ) -> ValidationRule:
     ...
 ```
 
-| Parameter name | Type                                                  | Default | Description                                                                                                                               |
-| -------------- | ----------------------------------------------------- | ------- | ----------------------------------------------------------------------------------------------------------------------------------------- |
-| max_depth      | `int`                                                 | N/A     | The maximum allowed depth for any operation in a GraphQL document                                                                         |
-| options        | `Optional[DepthLimitOptions]`                         | `None`  |                                                                                                                                           |
-| options.ignore | `List[Union[str, re.Pattern, Callable[[str], bool]]]` | `None`  | Stops recursive depth checking based on a field name. Either a string or regexp to match the name, or a function that reaturns a boolean. |
-| callback       | `Optional[Callable[[Dict[str, int]], None]]`          | `None`  | Called each time validation runs. Receives an Object which is a map of the depths for each operation                                      |
+| Parameter name | Type                                                            | Default | Description                                                                                                                               |
+| -------------- | --------------------------------------------------------------- | ------- | ----------------------------------------------------------------------------------------------------------------------------------------- |
+| max_depth      | `int`                                                           | N/A     | The maximum allowed depth for any operation in a GraphQL document                                                                         |
+| ignore         | `Optional[List[Union[str, re.Pattern, Callable[[str], bool]]]]` | `None`  | Stops recursive depth checking based on a field name. Either a string or regexp to match the name, or a function that reaturns a boolean. |
+| callback       | `Optional[Callable[[Dict[str, int]], None]]`                    | `None`  | Called each time validation runs. Receives an Object which is a map of the depths for each operation                                      |
 
 Example:
 
--- a/docs/guides/tools.md
+++ b/docs/guides/tools.md
@@ -42,3 +42,52 @@ type Query {
   myName: String!
 }
 ```
+
+---
+
+### `depth_limit_validator`
+
+Limit the complexity of queries by their depth.
+
+```python
+class DepthLimitOptions(TypedDict):
+    ignore: Union[str, re.Pattern, Callable[[str], bool]]
+
+def depth_limit_validator(
+  max_depth: int,  # The maximum allowed depth for any operation in a GraphQL document
+  options: DepthLimitOptions = None,
+  callback: Callable[Dict[str, int]] = None  # Called each time validation runs.  Receives an Object which is a map of the depths for each operation
+) -> ValidationRule:
+    ...
+```
+
+Example:
+
+```python
+import strawberry
+from strawberry.schema import default_validation_rules
+from strawberry.tools import depth_limit_validator
+
+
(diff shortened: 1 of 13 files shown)


botberry:

```markdown
Release type: patch

```

strawberry/tools/depth_limit_validator.py
```diff
@@ -0,0 +1,171 @@
...
+    FieldNode,
+    FragmentDefinitionNode,
+    FragmentSpreadNode,
+    InlineFragmentNode,
+    OperationDefinitionNode,
+)
+from graphql.validation import ValidationContext, ValidationRule
+
+
+def depth_limit_validator(max_depth: int, options=None, callback=None):
```
DoctorJohn, line 16: Should we add type hints for `options` and `callback`?

The type of `callback` is already part of the documentation. To properly type `options` we would probably have to use a TypedDict.

strawberry/tools/depth_limit_validator.py
```diff
@@ -0,0 +1,171 @@
...
+            ] = definition
+
+    return operations
+
+
+introspection_regex = re.compile(r"^__")
+
+
+def determine_depth(
+    node, fragments, depth_so_far, max_depth, context, operation_name, options
```
DoctorJohn, line 87: Maybe we want to add type hints for these as well? Not sure if necessary, since it's an internal API.

strawberry/tools/depth_limit_validator.py
```diff
@@ -0,0 +1,171 @@
...
+            GraphQLError(
+                f"'{operation_name}' exceeds maximum operation depth of {max_depth}",
+                [node],
+            )
+        )
+        return depth_so_far
+
+    if isinstance(node, FieldNode):
+        # by default, ignore the introspection fields which begin with double underscores
+        should_ignore = introspection_regex.match(node.name.value) or see_if_ignored(
```
DoctorJohn, line 100: We could use `node.name.value.startswith("__")` instead of the regex, which may be slightly faster and maybe easier to read. (I know a regex is used because its a port :) )

DoctorJohn: There is also this little documented helper function which could be reused if we want to https://github.com/strawberry-graphql/strawberry/blob/e6b005ce98df53f923c477fc11ab69c0a8f58736/strawberry/extensions/tracing/utils.py#L8

DoctorJohn (commented):
Nice work, I'm very excited for this. I added a couple of small comments, just things that came to my mind :)

strawberry/tools/depth_limit_validator.py
jkimbo: Yes thats a good idea.

strawberry/tools/depth_limit_validator.py
jkimbo: Ah good spot, will change.

jkimbo:
Thanks for the review @DoctorJohn , I'll get those changes in as soon as I can.

docs/guides/tools.md
```diff
@@ -42,3 +42,66 @@ type Query {
...
+
+---
+
+### `depth_limit_validator`
+
+Create a validator to limit the complexity of queries by their depth to protect against malicious
+queries.
+
+```python
+from graphql import ValidationRule
```
patrick91, line 54: can we make our own class for this? 🤔 

to make sure we are abstracted from GraphQL core

docs/guides/tools.md
```diff
@@ -42,3 +42,66 @@ type Query {
...
+
+class DepthLimitOptions(TypedDict):
```
patrick91, line 56: we wanted to use a class for configuration, should we use the same approach here?

docs/guides/tools.md
```diff
@@ -42,3 +42,66 @@ type Query {
...
+    ignore: List[Union[str, re.Pattern, Callable[[str], bool]]]
+
+def depth_limit_validator(
+    max_depth: int,
+    options: Optional[DepthLimitOptions] = None,
+    callback: Optional[Callable[Dict[str, int]]] = None
+) -> ValidationRule:
```
patrick91, line 63: the way we use this below makes me think we could use a class instead of a function

jkimbo: @patrick91 how can we use a function here? This is essentially a factory method to create a `ValidationRule` class.

docs/guides/tools.md
jkimbo: I'm not sure there would be much benefit to that because it would have to be a subclass of the `ValidationRule` class anyway for this check to work: https://github.com/graphql-python/graphql-core/blob/03c2b06e4636ed8a89c7a4c7e56d244fd9d00bde/src/graphql/validation/validate.py#L51-L56

docs/guides/tools.md
jkimbo: Actually, since there is only 1 option, I'm just going to inline it.

strawberry/tools/depth_limit_validator.py
```diff
@@ -0,0 +1,186 @@
...
+                    max_depth,
+                    context,
+                    operation_name,
+                    ignore,
+                ),
+                node.selection_set.selections,
+            )
+        )
+    else:
+        raise Exception(f"uh oh! depth crawler cannot handle: {node.kind}")
```
DoctorJohn, line 165: I guess this would be hard to test since this function currently appears to handle all kinds of nodes.

```suggestion
        raise Exception(f"Depth crawler cannot handle: {node.kind}")  # pragma: no cover
```

DoctorJohn (approved):
LGTM 👍

docs/guides/tools.md
patrick91: uhm ok :) we could convert it as we do we the schema converter, but that's fine for now :)

strawberry/tools/depth_limit_validator.py
```diff
@@ -0,0 +1,186 @@
...
+            GraphQLError(
+                f"'{operation_name}' exceeds maximum operation depth of {max_depth}",
+                [node],
+            )
+        )
+        return depth_so_far
+
+    if isinstance(node, FieldNode):
+        # by default, ignore the introspection fields which begin with double underscores
+        should_ignore = is_instrospection_key(node.name.value) or see_if_ignored(
```
patrick91, line 114:
```suggestion
        should_ignore = is_instrospection_key(node.name.value) or is_ignored(
```

strawberry/tools/depth_limit_validator.py
```diff
@@ -0,0 +1,186 @@
...
+                    ignore,
+                ),
+                node.selection_set.selections,
+            )
+        )
+    else:
+        raise Exception(f"uh oh! depth crawler cannot handle: {node.kind}")
+
+
+def see_if_ignored(node: FieldNode, ignore: Optional[List[IgnoreType]] = None) -> bool:
```
patrick91, line 168:
```suggestion
def is_ignored(node: FieldNode, ignore: Optional[List[IgnoreType]] = None) -> bool:
```

I'd say this reads a tiny bit better 😊

patrick91 (approved):
💯

strawberry/tools/depth_limit_validator.py
jkimbo: Yep agreed

docs/types/schema.md
```diff
@@ -53,14 +53,15 @@ Executes a GraphQL operation against a schema (async)
...
+| query            | `str`                                  | N/A     | The document to be executed                                                                            |
+| variable_values  | `Optional[Dict[str, Any]]`             | `None`  | The variables for this operation                                                                       |
+| context_value    | `Optional[Any]`                        | `None`  | The value of the context that will be passed down to resolvers                                         |
+| root_value       | `Optional[Any]`                        | `None`  | The value for the root type that will passed down to root resolvers                                    |
+| operation_name   | `Optional[str]`                        | `None`  | The name of the operation you want to execute, useful when sending a document with multiple operations |
+| validate_queries | `bool`                                 | `True`  | This flag allows to disable query validation                                                           |
```
BryceBeagle, line 63:
```suggestion
| validate_queries | `bool`                                 | `True`  | This flag enables/disables query validation                                                           |
```

strawberry/tools/depth_limit_validator.py
```diff
@@ -0,0 +1,186 @@
...
+    Creates a validator for the GraphQL query depth
+
+    This is a Python port of https://github.com/stems/graphql-depth-limit
+
+    - max_depth - The maximum allowed depth for any operation in a GraphQL document.
+    - ignore - Stops recursive depth checking based on a field name.
+        Either a string or regexp to match the name, or a function that returns
+        a boolean.
+    - callback - Called each time validation runs. Receives an Object which is a
+    map of the depths for each operation.
```
BryceBeagle, line 37:
```suggestion
    - callback - Called each time validation runs. Receives an Object which is a
        map of the depths for each operation.
```

strawberry/tools/depth_limit_validator.py
```diff
@@ -0,0 +1,186 @@
...
+            for name in queries:
+                query_depths[name] = determine_depth(
+                    queries[name],
+                    fragments,
+                    0,
+                    max_depth,
+                    validation_context,
+                    name,
+                    ignore,
+                )
```
BryceBeagle, line 58:
```suggestion
            for name in queries:
                query_depths[name] = determine_depth(
                    node=queries[name],
                    fragments=fragments,
                    depth_so_far=0,
                    max_depth=max_depth,
                    context=validation_context,
                    operation_name=name,
                    ignore=ignore,
                )
```

strawberry/tools/depth_limit_validator.py
```diff
@@ -0,0 +1,186 @@
...
+def get_queries_and_mutations(
+    definitions: List[DefinitionNode],
+) -> Dict[str, OperationDefinitionNode]:
+    operations = {}
+
+    for definition in definitions:
+        if isinstance(definition, OperationDefinitionNode):
+            operations[
+                definition.name.value if definition.name else "anonymous"
+            ] = definition
```
BryceBeagle, line 89:
```suggestion
            operation = definition.name.value if definition.name else "anonymous"
            operations[operation] = definition
```

strawberry/tools/depth_limit_validator.py
```diff
@@ -0,0 +1,186 @@
...
+            if field_name == rule:
+                return True
+        elif isinstance(rule, re.Pattern):
+            if rule.match(field_name):
+                return True
+        elif callable(rule):
+            if rule(field_name):
+                return True
+        else:
+            raise Exception(f"Invalid ignore option: {rule}")
```
BryceBeagle, line 184: Should this be a `ValueError`?

BryceBeagle (changes requested):

strawberry/tools/depth_limit_validator.py
```diff
@@ -0,0 +1,171 @@
...
+            depth_so_far,
+            max_depth,
+            context,
+            operation_name,
+            options,
+        )
+    elif (
+        isinstance(node, InlineFragmentNode)
+        or isinstance(node, FragmentDefinitionNode)
+        or isinstance(node, OperationDefinitionNode)
```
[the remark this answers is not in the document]
jkimbo: TIL

strawberry/tools/depth_limit_validator.py
```diff
@@ -0,0 +1,183 @@
...
+
+def depth_limit_validator(
+    max_depth: int,
+    ignore: Optional[List[IgnoreType]] = None,
+    callback: Callable[[Dict[str, int]], None] = None,
+):
+    """
+    Creates a validator for the GraphQL query depth
+
+    This is a Python port of https://github.com/stems/graphql-depth-limit
```
BryceBeagle, line 30: If this is a port, do we need to add the upstream license for this?

BryceBeagle (changes requested):

strawberry/tools/depth_limit_validator.py
BryceBeagle: Potentially the port should even be placed in its own repo?

DoctorJohn: Good call mentioning the upstream license. For reference: graphql-core is a port of graphql.js. Both are licensed under the MIT license (just like strawberry and the original graphql-depth-limit code). graphql-core retained graphql.js' license and copyright notice.

jkimbo: Yes good point @BryceBeagle , I'll add the copyright notice. I don't think it needs to be in its own repo though since it's a pretty small amount of code.

PR #1044, opened 2021-07-04
Move where we create the resolvers for our fields

patrick91:
## Description

This PR moves where we create our custom resolver for fields. Our custom logic resolver
is pretty much tied to GraphQL core and the schema. So it makes sense to move it there, 
it will also help with the auto_camel_case changes, as we need to know if we have auto
camel casing on when fetching the arguments.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

### DIFF
--- a/strawberry/schema/schema_converter.py
+++ b/strawberry/schema/schema_converter.py
@@ -314,7 +314,7 @@ class GraphQLCoreConverter:
     def from_resolver(self, field: StrawberryField) -> Callable:
         def _get_arguments(
             source: Any,
-            info: Any,
+            info: Info,
             kwargs: Dict[str, Any],
         ) -> Tuple[List[Any], Dict[str, Any]]:
             if field.base_resolver is None:
--- a/strawberry/schema/schema_converter.py
+++ b/strawberry/schema/schema_converter.py
@@ -1,4 +1,5 @@
 from enum import Enum
+from inspect import isasyncgen, iscoroutine
 from typing import Any, Callable, Dict, Type, cast
 
 from graphql import (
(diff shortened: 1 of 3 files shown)


botberry:

```markdown
Release type: patch

```

strawberry/schema/schema_converter.py
```diff
@@ -308,8 +311,91 @@ def get_graphql_fields() -> Dict[str, GraphQLField]:
...
+        if field.base_resolver.has_self_arg:
+            args.append(source)
+
+        if field.base_resolver.has_root_arg:
+            kwargs["root"] = source
+
+        if field.base_resolver.has_info_arg:
+            kwargs["info"] = info
+
+        return args, kwargs
```
BryceBeagle, line 343: Can this methods be embedded in `from_resolver` with the other functions? It doesn't need any state on the class and is only consumed by that one method.

BryceBeagle: I know that `from_resolver` is getting quite long, but I still think that would be cleaner

BryceBeagle (changes requested):

strawberry/schema/schema_converter.py
patrick91: yes, can do :)

strawberry/schema/schema_converter.py
```diff
@@ -309,7 +312,88 @@ def get_graphql_fields() -> Dict[str, GraphQLField]:
         return graphql_object_type
 
     def from_resolver(self, field: StrawberryField) -> Callable:
-        return field.get_wrapped_resolver()
+        def _get_arguments(
+            source: Any,
+            info: Any,
```
BryceBeagle, line 317: Do `source` and `info` have known types now?

BryceBeagle (approved):

strawberry/schema/schema_converter.py
patrick91: info yes, source depends on the resolver, so I think any is good :)

PR #1046, opened 2021-07-06
Removed csrf_exempt from django views

Speedy1991:
I think it's not a good idea to disable security features by default. If one wants an csrf exempt, you can wrap your view in urls with this decorator.

E.g.:

```
from django.views.decorators.csrf import csrf_exempt

urlpatterns = [
    path("graphql/", csrf_exempt(GraphQLView.as_view(), name='graphql'))
]
```

I didn't expect this exemp and the server could have been vulnerable to CSRF attacks by default.

## Description

Removed the decorators

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

- [x] My change requires a change to the documentation. (maybe - not sure?)
- [x] I have tested the changes and verified that they work and don't break anything (as well as I can manage) (tested in my own project with overwritten dispatch method)

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+Remove the (default and likely unexpected) csrf_exempt from django
--- a/strawberry/django/views.py
+++ b/strawberry/django/views.py
@@ -12,7 +12,6 @@ from django.template.exceptions import TemplateDoesNotExist
 from django.template.loader import render_to_string
 from django.template.response import TemplateResponse
 from django.utils.decorators import classonlymethod, method_decorator
-from django.views.decorators.csrf import csrf_exempt
 from django.views.generic import View
 
 import strawberry
@@ -144,7 +143,6 @@ class GraphQLView(BaseView):
     ) -> GraphQLHTTPResponse:
         return process_result(result)
 
-    @method_decorator(csrf_exempt)
     def dispatch(self, request, *args, **kwargs):
         if not self.is_request_allowed(request):
             return HttpResponseNotAllowed(
@@ -184,7 +182,6 @@ class AsyncGraphQLView(BaseView):
         view._is_coroutine = asyncio.coroutines._is_coroutine
         return view
 
-    @method_decorator(csrf_exempt)
     async def dispatch(self, request, *args, **kwargs):
         if not self.is_request_allowed(request):
             return HttpResponseNotAllowed(
--- a/strawberry/django/views.py
+++ b/strawberry/django/views.py
@@ -13,6 +13,7 @@ from django.template.exceptions import TemplateDoesNotExist
 from django.template.loader import render_to_string
 from django.template.response import TemplateResponse
 from django.utils.decorators import classonlymethod, method_decorator
+from django.views.decorators.csrf import csrf_exempt
 from django.views.generic import View
 
 import strawberry
--- a/strawberry/django/views.py
+++ b/strawberry/django/views.py
@@ -12,7 +12,7 @@ from django.template import RequestContext, Template
 from django.template.exceptions import TemplateDoesNotExist
 from django.template.loader import render_to_string
 from django.template.response import TemplateResponse
-from django.utils.decorators import classonlymethod
+from django.utils.decorators import classonlymethod, method_decorator
 from django.views.generic import View
 
 import strawberry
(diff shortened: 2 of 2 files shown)


botberry:

```markdown
Release type: patch

```

Speedy1991:
Anything missing to get this merged?

friendly ping @patrick91

patrick91:
> Anything missing to get this merged?
> 
> friendly ping @patrick91

I'll do a review this week, I want to check that:

1. the graphiql interface still works by default
2. we have a friend error


I think point 2 is quite important because we don't users to spend a lot of time debugging issue with CSRF tokens, especially if they are beginners. We might need to write docs for that. What do you think?

jkimbo:
I don't think this is the right approach for a couple of reasons:

1. It's quite a disruptive change for current users because it will likely break current applications. Also it probably won't show up in tests because they usually don't test the http layer.

2. I think that defaulting to exempting the view from csrf checks is actually a good default for Strawberry because it matches how most people use graphql. I don't have any solid evidence for this but I would assume that most people use graphql in a single page application or mobile app that is not hosted by Django and so getting the csrf token is not trivial. 

---

I think a better solution would be to make csrf protection an option on the GraphQLView. It could be a required option so that people are forced to choose whether to opt in or out. That way it would be obvious at app startup rather than later at runtime. 

What do you think @Speedy1991 @patrick91 ?

patrick91:
Uh, an option could be a better solution! 

I agree with all your points :)

Speedy1991:
I'm still not 100% convinced. Because "it will likely break current applications" we will keep a potential security issue? At least we could add a warning about disabled CSRF

Anyway, I like the suggested "options" if it is a required arg/kwarg :)

jkimbo:
@Speedy1991 yep I agree, a required option is the way forward IMO. Could you update this PR?

BryceBeagle:
Could the option be not required, but default to the more secure approach?

jkimbo:
> Could the option be not required, but default to the more secure approach?

I think we should require the option to be set first to avoid accidentally breaking things (it would break my application for example). We can then default to requiring csrf protection in the future.

g-as:
My two cents to your point 2. @jkimbo 

It makes sense to disable CSRF checks for a _public_ API where third parties consume it. But for a _private_ API used by a single client (SPA), the cost of fetching the CSRF token (and keep the safety on) is low IMHO.

And +1 on your latest proposal!

Speedy1991:
> ```python
> urlpatterns = [
>     path("graphql/", GraphQLView.as_view(schema=schema, csrf_protection=True)),
> ]
> ```

I dislike the idea to add an arg to `as_view` when `path("graphql/", csrf_exempt(GraphQLView.as_view(), name='graphql'))` does the same (like "do not reinvent the wheel")

And url capsulation is more the "django way" IMO

Maybe we can put it in the v1 release as breaking change?

However, I'm glad if this gets fixed in any way

Speedy1991:
Rebased to current master

To prevent breaking changes i kept the classes "as is" but added a warning about the disabled csrf stuff

paltman:
A simple work around is just to add the following to your project:

```python


class ProtectedGraphQLView(GraphQLView):

    def dispatch(self, request, *args, **kwargs):
        return super().dispatch(request, *args, **kwargs)
```

Then use `ProtectedGraphQLView` in your `urls.py`.  This seems to override the `@method_decorator` on the `GraphQLView` base class.

patrick91:
@bellini666 do you have opinions on this PR? 😊

bellini666:
> @bellini666 do you have opinions on this PR? blush

I mostly agree with @jkimbo's points. I also don't have any evidence of that, but I have a strong feeling that most people using strawberry and/or DRF are not using csrf by default.

I'm not totally against this, but I would say that based on the fact that the minority of users would use CSRF for graphql APIs, and when someone wants to use it for such they probably know that they need to write some extra code in the client to get that and send it in a different way (e.g. in a header), it is probably better to have an example aimed at them at the docs with [paltman's suggestion](https://github.com/strawberry-graphql/strawberry/pull/1046#issuecomment-1086757715) to enable it.

patrick91:
@bellini666 makes sense, I'll close this then!

Arfey:
@bellini666 

if u have API which use cookie for authentication and some query can mutate user data then csrf is critical security issue. When I use `django.middleware.csrf.CsrfViewMiddleware` I expect that all my views are protected.

>I have a strong feeling that most people using strawberry and/or DRF are not using csrf by default.

I have a strong feeling that many people don't know they have this problem in the project and will be surprised 😅

@jkimbo good suggestion with `csrf_protection` parameter

PR #1047, opened 2021-07-07
Always pass `info` to Field.get-result

la4de:
This pull request tries to fix regression caused by #1044.

This needs still more testing and changes to strawberry-django package as well. See https://github.com/strawberry-graphql/strawberry-graphql-django/tree/wip/fix-strawberry-regression.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,5 @@
+Release type: patch
+
+This release fixes a regression that broke strawberry-graphql-django.
+
+`Field.get_results` now always receives the `info` argument.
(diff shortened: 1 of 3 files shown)


la4de:
@patrick91 , do these changes make sense? I'd like to hear your feedback before I continue?

patrick91:
> @patrick91 , do these changes make sense? I'd like to hear your feedback before I continue?

Will check this tonight!

botberry:

```markdown
Release type: patch

```

patrick91 (approved):
Yup, this makes sense to me!

PR #1066, opened 2021-07-18
Add option to graphiql.html to use newer graphql-ws subsctiption library.

Object905:
## Description
Excerpts from subscriptions-transport-ws:
[README.md](https://github.com/apollographql/subscriptions-transport-ws#graphql-ws) The subscriptions-transport-ws library is not being actively maintained. It is recommended that you use the graphql-ws library instead.
[CHANGELOG.md](https://github.com/apollographql/subscriptions-transport-ws/blob/master/CHANGELOG.md#v0919-2021-06-08) Apollo Server 3 will remove its superficial integration with this package entirely. We encourage users of Apollo Server 2 to disable the integration with this unmaintained package, and consider packages such as graphql-ws to power GraphQL subscriptions.

## Breaking changes
`strawberry.asgi.utils.get_graphiql_html`, `strawberry.flask.graphiql.render_graphiql_page`, `strawberry.sanic.graphiql.render_graphiql_page` are removed in favor of `strawberry.utils.graphiql.get_graphiql_html`.

### Migration to new [protocol](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md).
Should we migrate existing implementations, dropping old one altogether? Or change names for old protocol, replacing it with new one? Then how selection of protocol will be handled?

#### Off-topic Question
I have an implementation of the old protocol for django [channels](https://github.com/django/channels).
In my free time I will work on new one. When I'm done should I make pull request here, adding optional dependency on `channels` or in [strawberry-graphql-django](https://github.com/strawberry-graphql/strawberry-graphql-django) repo?

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

## Checklist

- [x] I have added tests to cover my changes. (Already covered)


DoctorJohn:
> Should we migrate existing implementations, dropping old one altogether? Or change names for old protocol, replacing it with new one? Then how selection of protocol will be handled?

I already planned on adding support for the newer protocol. Since both protocols ([old](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md), [new](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md)) are not compatible and there isn't really an official standard, I planned on keeping support for both protocols for backwards compatibility.

Clients can choose which protocol they want to use by specifying a websocket subprotocol when initialising the connection. (Most client libraries do this automatically). Developers of strawberry based servers can decide which protocols they want to support, probably by passing a tuple of subprotocols to a integrations view (that's how things like the [keep alive interval](https://github.com/strawberry-graphql/strawberry/blob/5b601be026a329ed5e04f48b2291b311efa79520/strawberry/asgi/__init__.py#L53) is currently configured).

> I have an implementation of the old protocol for django channels.
In my free time I will work on new one. When I'm done should I make pull request here, adding optional dependency on channels or in strawberry-graphql-django repo?

Currently two of the strawberry integrations support subscriptions: ASGI and AIOHTTP. Django is ASGI compatible, meaning currently one would use the ASGI integration to add subscriptions support to your Django application. I prefer this approche to a separate channels specific implementation. That said, the channels based implementation you're working on can probably ported to work with our vanilla ASGI approche.

Object905:
@DoctorJohn 
Yes asgi implementation already works with django asgi, but using starlette stack, which might mean more unwanted dependencies for users that already use channels, not only for graphql (which is my case).
Also django `runserver` command is using wsgi stack, which means you can't use it with subsciptions. You have to use uvicorn for dev server.
Django channels patch `runserver` command both in wsgi and asgi mode, so websockets "just work".
Patching `runserver` in `strawberry-graphql-django`, for example would be a bad idea, since that would break channels, whitenoise and other libs that override runserver.

DoctorJohn:
Fair point @Object905. I thought there was a way to somehow integrate the strawberry ASGI app with django channels. But if there isn't, a dedicated django channels integration or just a channels based subscriptions implementation extending the existing django integration could be a great addition to strawberry! The code would probably go directly into strawberry since thats were the other subscription code lives (and because there are plans to merge strawberry-django into strawberry). Channels would become an optional dependency of strawberry, just like starlette.

I would recommend opening a new issue so further details can be discussed separate from this PR.

Object905:
@DoctorJohn. As I said earlier django can be adapted to use asgi, all asgi apps are extensible, because they are just callables.

But it has two downsides that I see.
1. You can't use `runserver` command, you have to use uvicorn to run dev server, which is somewhat fine, but I found that in some cases it can't handle file reload as well as vanila runserver, and may crash on certain exceptions.
2. You have to depend on `starlette`. 

I've refactored protocol handling, to work on abstraction over `BaseWebSocket` and `BaseProtocol` what handler now does is 
 selects protocol by `sec-websocket-protocol` header and provides it with websocket.

Haven't got to implementing new protocol or integrating this new abstraction into any current implementation. Will open separate PR once I do.

strawberry/aiohttp/views.py
```diff
@@ -307,8 +308,10 @@ async def parse_multipart_body(self, request: web.Request) -> dict:
             raise web.HTTPBadRequest(reason="File(s) missing in form data")
 
     def render_graphiql(self) -> web.StreamResponse:
-        html_string = self.graphiql_html_file_path.read_text()
-        html_string = html_string.replace("{{ SUBSCRIPTION_ENABLED }}", "true")
+        html_string = get_graphiql_html(
+            subscription_enabled=True,
+            graphiql_html_file_path=self.graphiql_html_file_path,
```
DoctorJohn, line 313: aiohttp can use the same default graphiql file as the other integrations. The `HTTPHandler.graphiql_html_file_path` property can then be removed and the `graphiql_html_file_path` argument as well :)

strawberry/django/views.py
```diff
@@ -78,21 +77,17 @@ def _render_graphiql(self, request: HttpRequest, context=None):
...
-        except TemplateDoesNotExist:
             template = Template(
-                open(
-                    os.path.join(
-                        os.path.dirname(os.path.abspath(strawberry.__file__)),
-                        "static/graphiql.html",
-                    ),
-                    "r",
-                ).read()
+                render_to_string("graphql/graphiql.html", context=context)
```
DoctorJohn, line 87: I think it would be good for consistency if we would let your function handle interpolation here as well. As far as I can tell we do not actually require a custom context nor the posibility to override the graphiql template using the django template system. cc @patrick91 

(If we were to leverage your functions interpolation here as well, we could remove its `raw` parameter which appears to be only used here.)

DoctorJohn (changes requested):
Thanks for this PR @Object905 ! Changes like these make it easier to maintain different integrations :)

jkimbo:
@Object905 we were having some issues with JSDelivr so we changed the CDN to unkpg instead: https://github.com/strawberry-graphql/strawberry/pull/1096

Could you update this PR pointing to unpkg instead?

patrick91:
@Object905 I've merged master into this PR, but I'm getting this error when testing it: 

```
Error in opening handshake
  File "/Users/patrick/Library/Caches/pypoetry/virtualenvs/strawberry-graphql-SRZ1TVPA-py3.9/lib/python3.9/site-packages/websockets/server.py", line 130, in handler
    path = await self.handshake(
  File "/Users/patrick/Library/Caches/pypoetry/virtualenvs/strawberry-graphql-SRZ1TVPA-py3.9/lib/python3.9/site-packages/websockets/server.py", line 532, in handshake
    early_response = await early_response_awaitable
  File "/Users/patrick/Library/Caches/pypoetry/virtualenvs/strawberry-graphql-SRZ1TVPA-py3.9/lib/python3.9/site-packages/uvicorn/protocols/websockets/websockets_impl.py", line 136, in process_request
    websockets.legacy.handshake.check_request(headers)
AttributeError: module 'websockets' has no attribute 'legacy'
```

is it expected?

botberry:



---

`strawberry.asgi.utils.get_graphiql_html`, `strawberry.flask.graphiql.render_graphiql_page`, `strawberry.sanic.graphiql.render_graphiql_page` are removed in favor of `strawberry.utils.graphiql.get_graphiql_html`.

`graphiql.html` template is now using latest version of `graphiql`. Added `SUBSCRIPTION_GRAPHQL_WS` template variable into `graphiql.html` that specifies whether it should use newer [graphql-ws](https://github.com/enisdenjo/graphql-ws) instead of unmaintained [subscriptions-transport-ws](https://github.com/apollographql/subscriptions-transport-ws). Implementations of new [protocol](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md) are yet to come.


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Zverev Konstantin for the PR 👏

```

patrick91:
closing this as we updated GraphiQL to v2 a while back 😊

PR #1071, opened 2021-07-19
Change how we map Pydantic fields to types

### LINKED ISSUE #1072: Compatibility with pydantic 1.4
I'm trying to use strawberry in a project that has pydantic pinned at 1.4. I chatted with @patrick91 on discord about this, and he thought it would be reasonable to achieve compatibility with this version.

Pydantic appears to only be used in the [strawberry.experimental](https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/experimental/__init__.py) module, which only gets loaded if pydantic is present. One way to solve this for me in particular would be to lazily load strawberry.experimental/pydantic, such that when an older version of pydantic is present, one can still import other packages in strawberry.

Thank you!

patrick91:
This PR changes how we map Pydantic fields to types to allow for older version of Pydantic which might not have some field like pydantic.ConstrainedSet in 1.4.

I've also allowed any Pydantic version before 2.0.

Not sure I 100% like this solution. Happy to hear feedback :)

Closes #1072


botberry:



---

This release changes how we map Pydantic fields to types
to allow using older version of Pydantic.

jkimbo (approved):
Lgtm

PR #1073, opened 2021-07-20
dependencies: require typing-extensions 3.10.0

tobiaswiese:
Strawberry uses TypeGuard from typing-extensions, which is only
available starting with version 3.10.0

Signed-off-by: Tobias Wiese <tobias@tobiaswiese.com>


patrick91:
@tobiaswiese I completely missed this! I did change how we require TypeGuards in this PR: https://github.com/strawberry-graphql/strawberry/pull/1074

so we don't need to update typing_extensions, does that work for you? :)

tobiaswiese:
As it is no longer a hard dependency on 3.10.0 anymore that should be enough

PR #1074, opened 2021-07-20
Don't import TypeGuard if TYPE_CHECKING is false

patrick91:
## Description

Small patch to prevent issues when using older version of `typing_extensions`.

I don't think we should force people to upgrade as we only use TypeGuards for
type checking.

## Types of Changes

- [x] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,5 @@
+Release type: patch
+
+
+This release removes a `TypeGuard` import to prevent errors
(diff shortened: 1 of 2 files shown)


botberry:



---

This release removes a `TypeGuard` import to prevent errors
when using older versions of `typing_extensions`.

marcoacierno (approved):

PR #1076, opened 2021-07-21
Fix field inheritance with new `StrawberryAnnotation`s

### LINKED ISSUE #1075: Type refactoring has a regression with inheritance and explicit fields.
```python
@strawberry.input
class A:
    a: str = strawberry.field(default='', desc='')

@strawberry.input
class B(A):
    b: Optional[str] = strawberry.field(default=None, desc='')

@strawberry.type
class Query:
    @strawberry.field
    def field(self, arg: B) -> str:
        return ''

result = schema.execute_sync('{ field(arg: {}) }')
assert not result.errors
```
raises `TypeError: B fields cannot be resolved. unhashable type: 'StrawberryAnnotation'`.

`StrawberryAnnotation` has a custom `__eq__` without a `__hash__`, causing a set lookup to fail. However adding a suitable `__hash__` just lead to the next `TypeError`.

`StrawberryOptional` likely has the same problem.

coady:
## Description

The type resolver ensures that fields' type annotations are StrawberryAnnotations, but a typo was checking the wrong field. This resulted in StrawberryAnnotations getting recursively wrapped and causing errors downstream.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* Fixes #1075.

## Checklist

### DIFF
--- a/strawberry/types/type_resolver.py
+++ b/strawberry/types/type_resolver.py
@@ -103,7 +103,7 @@ def _get_fields(cls: Type) -> List[StrawberryField]:
             field.origin = field.origin or cls
 
             # Make sure types are StrawberryAnnotations
-            if not isinstance(field.type, StrawberryAnnotation):
+            if not isinstance(field.type_annotation, StrawberryAnnotation):
                 module = sys.modules[field.origin.__module__]
                 field.type_annotation = StrawberryAnnotation(
                     annotation=field.type_annotation, namespace=module.__dict__
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+Fix for regression when defining inherited types with explicit fields.
--- a/tests/objects/test_inheritance.py
+++ b/tests/objects/test_inheritance.py
@@ -1,0 +1,15 @@
+from typing import Optional
+
+import strawberry
+
+
+def test_inherited_fields():
+    @strawberry.type
+    class A:
+        a: str = strawberry.field(default="")
+
+    @strawberry.type
+    class B(A):
+        b: Optional[str] = strawberry.field(default=None)
+
+    assert strawberry.Schema(query=B)


botberry:

```markdown
Release type: patch

```

strawberry/annotation.py
```diff
@@ -43,6 +43,9 @@ def __init__(
     ):
         self.annotation = annotation
         self.namespace = namespace
+        if isinstance(annotation, StrawberryAnnotation):
+            self.annotation = annotation.annotation
+            self.namespace = annotation.namespace
```
patrick91, line 48: I wonder if we should prevent passing a StrawberryAnnotation to StrawberryAnnotation itself, we could reuse the existing one, what do you think @BryceBeagle ?

BryceBeagle: I'll take a closer look tomorrow, but this looks to be a bandaid solution instead of addressing the underlying problem. I don't think inheritance should lead to a situation where `StrawberryAnnotation` is constructed recursively. I think we should investigate that.

coady: Updated with a fix for the root cause; just a typo.

strawberry/types/type_resolver.py
```diff
@@ -103,7 +103,7 @@ class if one is not set by either using an explicit strawberry.field(name=...) o
             field.origin = field.origin or cls
 
             # Make sure types are StrawberryAnnotations
-            if not isinstance(field.type, StrawberryAnnotation):
```
patrick91, line 106: well spotted!

patrick91 (approved):
This looks good to me! Let's see what @BryceBeagle thinks and then we can ship it!

strawberry/annotation.py
BryceBeagle: This looks good to me just looking at the diff. Good spot. I'll run the changes in a bit to double check this is what we want :slightly_smiling_face:

BryceBeagle (changes requested):
Hey @coady, thanks for this change. I think the change is what we need to fix this bug, but I think the test should be moved to a different place. The file (and directory) you put this in is concerned with directly resolving the `StrawberryAnnotation`s, agnostic of how the rest of the Strawberry library uses/creates them.

If it could be moved to `tests/objects/test_inheritance.py` (a new file), I think that would be best.

BryceBeagle (approved):
If you could just add a release file, I think this is good to ship :+1:

PR #1077, opened 2021-07-21
Change how we determine if a path is strawberry decorator in MyPy

patrick91:
## Description

When you have `follow_imports = skip` in mypy the path inside the plugin is not the actual path, as show in this example:

image

I've changed the plugin to support this kind of paths by check if the path ends with our decorator name; it's pretty unlikely that we'd get someone with a different `strawberry.type` function :D

## Types of Changes

- [x] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,4 +1,4 @@
-Release type patch:
+Release type: patch
 
 This releases improves the MyPy plugin to be more forgiving of
 settings like `follow_imports = skip` which would break the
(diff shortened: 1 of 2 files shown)


botberry:

```markdown
Release type: patch

```

patrick91:
Going to merge this so I can update strawberry at work 😊 we can change the approach if needed to in future :)

PR #1078, opened 2021-07-21
Fix more mypy full name checks

patrick91:
Continuation of #1077

I didn't realise we change more paths. This PR fixes all of them 😊

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,7 @@
+Release type: patch
+
+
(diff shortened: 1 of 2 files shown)


botberry:



---

This releases improves the MyPy plugin to be more forgiving of
settings like follow_imports = skip which would break the type checking.

This is a continuation of the previous release and fixes for type checking issues.

PR #1091, opened 2021-07-28
Add workflow to send tweets when new release is published

patrick91:
This PR updates our release check workflow to also include a release card preview.

I'll add the automated tweet in another PR.

Note that I've split the release-check action in two, one for checking the release and one for sending the comment. In other PRs I'll move the comment action to a new repo (or the bot repo).


botberry:

```markdown
Release type: patch

```

botberry:
**Hello**
🌏

!

botberry:
**Hello**
🌏

!

botberry:
**Hello**
🌏

!

.github/bot-action/action.yml
```diff
@@ -0,0 +1,15 @@
...
+    required: true
+  change_type:
+    required: true
+  changelog:
+    required: true
+  release_card_url:
+    required: false
+runs:
+  using: 'docker'
+  image: 'Dockerfile'
```
patrick91, line 15: My plan is to move this action to the strawberry-bot repo and allow multiple "commands", for now it only supports adding a release comment so that's why almost everything is optional.

When moving it to the new repo will make sure to add validation and to support multiple commands :)

.github/release-check-action/check.py
```diff
@@ -44,28 +43,19 @@
...
 
-response = httpx.post(
-    API_URL,
-    json={"query": mutation, "variables": {"input": mutation_input}},
-    timeout=120,
-)
-response.raise_for_status()
+if release_info:
+    changelog = release_info["changelog"]
+    changelog = changelog.replace("`", r"\`")
```
patrick91, line 52: this was needed for generating the preview, might not be needed anymore, will check soon

patrick91: yes, it is needed: 
image

patrick91:
This is how the comment looks like: 

image

.github/workflows/release-check.yml
```diff
@@ -10,9 +10,60 @@ jobs:
...
+          --header "Authorization: Client-ID $IMGUR_CLIENT_ID" \
+          --form "image=@screenshot.png;filename=screenshot.png;type=image/png" \
+          | jq  --raw-output '.data.link')
+      env:
+        IMGUR_CLIENT_ID: ${{ secrets.IMGUR_CLIENT_ID }}
+
+  send-comment:
+    runs-on: ubuntu-latest
+    needs: [release-file-check, generate-preview]
+    if: always() && github.event.pull_request.draft == false
```
marcoacierno, line 58: Why the always()?

patrick91: I always want to send the comment, even if any of the previous job fail (the release file check fails if the file is missing or wrong) :)

.github/release-check-action/check.py
marcoacierno: Does this ^ mean I could write shell code in the changelog and have it run on the action? E.g can I write code that sends the secrets to me via curl?

patrick91: not with the escape, I hope so at least :D I've got another idea to improve this, will test soon

wb21268-bit (approved):

Issue #920, opened 2021-05-07
Proposal: Custom fields


jkimbo:
## Objective

Allow developers to create custom fields to encapsulate common resolver logic, allow modification of field arguments and return types.

## Motivation

At the moment the only way to define a field is to use the `strawberry.field` function. There is no ability to modify arguments or types programmatically so defining common patterns in a schema means duplicating logic throughout the code base.

@BryceBeagle work on refactoring StrawberryField into a proper class has now provided us with a stable foundation on which to provide an API for developers to extend the default StrawberryField class to define their own custom fields.

## Benefit

Custom fields will allow developers to create fields to abstract away common patterns in their schema and make it more maintainable. Strawberry can also provide common custom fields to help with things like pagination, permissions and validation.

## API design

```python
from strawberry import StrawberryField

class MyCustomField(StrawberryField):
    def get_type(self) -> Type:
		type_ = super().get_type()
        # Modify type
		return type_

    def get_arguments(self) -> Dict[str, Type]:
		arguments = super().get_arguments()
        # Modify arguments
		return arguments

    def get_result(self, source: Any, info: Info, arguments: Dict[str, Any]) -> Union[Awaitable[Any], Any]:
		result = super().get_result(source, info, arguments)
        # Modify result
		return result
```

#### Simple upper case field

```python
class UpperCaseField(StrawberryField):
	def get_type(self) -> Type:
		type_ = super().get_type()
		# Make sure that this field is only used on str fields
		assert type_ is str
		return type_

	def get_result(
		self, source: Any, info: Any, arguments: Dict[str, Any]
	) -> Union[Awaitable[Any], Any]:
		result = super().get_result(arguments, source, info)
		return cast(str, result).upper()


@strawberry.type
class Query:
	name: str = UpperCaseField(default="Jonathan")

	@UpperCaseField()
	def alt_name() -> str:
		return "jkimbo"

schema = strawberry.Schema(query=Query)

result = schema.execute_sync("{ name, altName }", root_value=Query())

assert result.data == {"name": "JONATHAN", "altName": "JKIMBO"}
```

#### Auth required field

```python
class AuthenticationRequiredField(StrawberryField):
	def get_type(self) -> Type:
		type_ = super().get_type()
		# Make sure the return type is optional
		return Optional[type_]

	def get_result(
		self, source: Any, info: Any, arguments: Dict[str, Any]
	) -> Union[Awaitable[Any], Any]:
		# Check authentication
[...]

PR #1100, opened 2021-08-02
[WIP] Custom fields

jkimbo:
## Description

Refactor `StrawberryField` to allow for custom fields (ref: https://github.com/strawberry-graphql/strawberry/issues/920)

Putting this out there for some early feedback to make sure I'm going in the right direction.

Things to note:

* I've removed the `.type` property override from StrawberryField to let dataclasses set it as it normally would. I've then added `.resolved_type` property that does the conversion to the Strawberry type.
* This does mean that setting the type for a field means you create the field and then assign the type. It's a bit clunky but should only happen internally so I don't think it's a big deal
* `.resolved_type` should potentially be a cached property since it's doing a fair amount of work. What do you think @BryceBeagle ?

## TODO

* [x] Fix tests
* [ ] Add documentation
* [ ] Add a proper pagination custom field to test out API
* [x] Allow stacking of custom fields

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -212,7 +212,7 @@ class StrawberryField(dataclasses.Field, GraphQLNameMixin):
         #       removed.
         _ = resolver.arguments
 
-    def get_type(self) -> Optional[Union[StrawberryType, object]]:
+    def get_type(self) -> Union[StrawberryType, object]:
         if self.base_resolver is not None:
             # Handle unannotated functions (such as lambdas)
             if self.base_resolver.type is not None:
@@ -238,15 +238,15 @@ class StrawberryField(dataclasses.Field, GraphQLNameMixin):
     # TODO: add this to arguments (and/or move it to StrawberryType)
     @property
     def type_params(self) -> List[TypeVar]:
-        if hasattr(self.resolved_type, "_type_definition"):
+        if hasattr(self.type, "_type_definition"):
             parameters = getattr(self.type, "__parameters__", None)
 
             return list(parameters) if parameters else []
 
         # TODO: Consider making leaf types always StrawberryTypes, maybe a
         #       StrawberryBaseType or something
-        if isinstance(self.resolved_type, StrawberryType):
-            return self.resolved_type.type_params
+        if isinstance(self.type, StrawberryType):
+            return self.type.type_params
         return []
 
     def copy_with(
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -131,7 +131,7 @@ class StrawberryField(dataclasses.Field, GraphQLNameMixin):
 
     @property
     def arguments(self) -> List[StrawberryArgument]:
-        arguments_as_types = self.get_arguments()
+        arguments_as_types = self.get_field_arguments()
 
         if not self.base_resolver:
             return []
@@ -149,7 +149,7 @@ class StrawberryField(dataclasses.Field, GraphQLNameMixin):
 
         return arguments
 
-    def get_arguments(self) -> Dict[str, object]:
+    def get_field_arguments(self) -> Dict[str, object]:
         """
         Hook to modify the GraphQL arguments for the field
         """
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -121,6 +121,25 @@ class StrawberryField(dataclasses.Field, GraphQLNameMixin):
     def __call__(self, resolver: _RESOLVER_TYPE) -> "StrawberryField":
         """Add a resolver to the field"""
 
+        # If we are stacking fields on top of each other
+        if isinstance(resolver, StrawberryField):
+            base_field = resolver
+            # Create a new field type that merges the 2 types together
+            NewFieldType = type(
+                f"{self.__class__.__name__}{resolver.__class__.__name__}",
+                (self.__class__, resolver.__class__),
+                {},
+            )
+            new_field = NewFieldType(
+                graphql_name=base_field.graphql_name,
+                description=base_field.description,
+                permission_classes=base_field.permission_classes,
+                deprecation_reason=base_field.deprecation_reason,
+                base_resolver=base_field.base_resolver,
+            )
+
+            return new_field
+
         # Allow for StrawberryResolvers or bare functions to be provided
         if not isinstance(resolver, StrawberryResolver):
             resolver = StrawberryResolver(resolver)
@@ -408,7 +427,6 @@ def field(
 
     it can be used both as decorator and as a normal function.
     """
-
     field_ = StrawberryField(
         python_name=None,
         graphql_name=name,
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -221,7 +221,7 @@ class StrawberryField(dataclasses.Field, GraphQLNameMixin):
         namespace = module.__dict__
         return namespace
 
-    def get_type(self) -> Union[object, str]:
+    def get_return_type(self) -> Union[object, str]:
         """
         Hook to allow custom fields to modify the return type of a field. This
         hook should only be called once at schema creation time.
@@ -243,7 +243,7 @@ class StrawberryField(dataclasses.Field, GraphQLNameMixin):
         # We should have an origin set by now
         assert self.origin
 
-        type_ = self.get_type()
+        type_ = self.get_return_type()
 
         type_annotation = StrawberryAnnotation(
             annotation=type_, namespace=self.origin_namespace
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -30,6 +30,25 @@ from .types.types import FederationFieldParams, TypeDefinition
 _RESOLVER_TYPE = Union[StrawberryResolver, Callable]
 
 
+@dataclasses.dataclass
+class BareStrawberryArgument:
+    """
+    Container for a StrawberryArgument where we don't know the argument name
+    yet. This lets us create arguments and then bind them to a name later.
+    """
+
+    type_annotation: StrawberryAnnotation
+    default: Any
+
+    def as_argument(self, arg_name):
+        return StrawberryArgument(
+            python_name=arg_name,
+            graphql_name=None,
+            type_annotation=self.type_annotation,
+            default=self.default,
+        )
+
+
 class StrawberryField(dataclasses.Field, GraphQLNameMixin):
     python_name: str
 
@@ -116,13 +135,13 @@ class StrawberryField(dataclasses.Field, GraphQLNameMixin):
         # Convert the arguments to StrawberryArgument types
         arguments = []
         for arg_name, annotation in arguments_as_annotations.items():
-            if isinstance(annotation, StrawberryArgument):
+            if isinstance(annotation, BareStrawberryArgument):
                 argument = annotation
             else:
                 default = self.base_resolver.get_argument_default(arg_name)
-                argument = self.create_argument(arg_name, annotation, default)
+                argument = self.create_argument(annotation, default)
 
-            arguments.append(argument)
+            arguments.append(argument.as_argument(arg_name))
 
         return arguments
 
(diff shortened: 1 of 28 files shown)


strawberry/field.py
```diff
@@ -27,14 +28,32 @@
...
 
+@dataclasses.dataclass
+class BareStrawberryArgument:
+    """
+    Container for a StrawberryArgument where we don't know the argument name
+    yet. This lets us create arguments and then bind them to a name later.
+    """
+
+    type_annotation: StrawberryAnnotation
+    default: Any
```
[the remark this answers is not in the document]
BryceBeagle: Also, I'm wondering if it would be a smart idea to have a more general-purpose type that pairs an annotation with a default value. This could be used on both fields _and_ arguments

strawberry/field.py
```diff
@@ -105,11 +122,51 @@ def __call__(self, resolver: _RESOLVER_TYPE) -> "StrawberryField":
...
+        # Convert the arguments to StrawberryArgument types
+        arguments = []
+        for arg_name, annotation in arguments_as_annotations.items():
+            if isinstance(annotation, BareStrawberryArgument):
+                argument = annotation
+            else:
+                default = self.base_resolver.get_argument_default(arg_name)
+                argument = self.create_argument(annotation, default)
+
+            arguments.append(argument.as_argument(arg_name))
```
BryceBeagle, line 139: It might be a bit confusing to call the mapping values here `annotation`s. `StrawberryField.get_arguments` returns `Dict[str, object]` not `Dict[str, StrawberryAnnotation]`.

strawberry/field.py
```diff
@@ -105,11 +122,51 @@ def __call__(self, resolver: _RESOLVER_TYPE) -> "StrawberryField":
...
+
+        return arguments
+
+    def get_arguments(self) -> Dict[str, object]:
```
BryceBeagle, line 143: Should this be a `_private` method?

strawberry/field.py
```diff
@@ -105,11 +122,51 @@ def __call__(self, resolver: _RESOLVER_TYPE) -> "StrawberryField":
...
+            return {}
+
         return self.base_resolver.arguments
 
+    def create_argument(
+        self, type_annotation: object, default: Any = UNSET
+    ) -> BareStrawberryArgument:
+        """
+        Helper function to create StrawberryArgument
+        """
```
BryceBeagle, line 157:
```suggestion
        """Helper function to create StrawberryArgument"""
```
(nitpicky suggestion for a draft PR, sorry 😅)

strawberry/field.py
```diff
@@ -152,32 +209,28 @@ def base_resolver(self, resolver: StrawberryResolver) -> None:
...
-    @type.setter
-    def type(self, type_: Any) -> None:
-        self.type_annotation = type_
+    def get_type(self) -> Union[StrawberryType, object]:
+        if self.base_resolver is not None:
+            # Handle unannotated functions (such as lambdas)
+            if self.base_resolver.type is not None:
+                return self.base_resolver.type
+
+        assert self.type is not None
```
BryceBeagle, line 218: At minimum, this should probably throw a more descriptive error

strawberry/field.py
```diff
@@ -234,17 +287,22 @@ def copy_with(
...
         Calls the resolver defined for the StrawberryField. If the field doesn't have a
         resolver defined we default to using getattr on `source`.
         """
 
         if self.base_resolver:
-            return self.base_resolver(*args, **kwargs)
+            args = []
+            if self.base_resolver.has_self_arg:
+                args = [source]
+            return self.base_resolver(*args, **arguments)
```
BryceBeagle, line 305: Do you know how this works with async resolvers?

tests/schema/test_custom_fields.py
```diff
@@ -0,0 +1,177 @@
...
+
+        def get_arguments(self) -> Dict[str, object]:
+            arguments = super().get_arguments()
+            arguments["first"] = Optional[int]
+
+            return arguments
+
+    @strawberry.type
+    class Query:
+        @PageNumberPagination()
```
BryceBeagle, line 55: Does this need to be instantiated?

strawberry/field.py
```diff
@@ -234,17 +287,22 @@ def copy_with(
             default_factory=self.default_factory,  # type: ignore[misc]
             deprecation_reason=self.deprecation_reason,
         )
+        field.type = new_type
+        return field
 
     def get_result(
-        self, source: Any, info: Info, args: List[Any], kwargs: Dict[str, Any]
+        self, source: Any, info: Info, arguments: Dict[str, Any]
     ) -> Union[Awaitable[Any], Any]:
```
BryceBeagle, line 295: Should we give this function a better name, esp. if users are expected to override it?

BryceBeagle (commented):
I think this a really good start. I think that inheriting from the base `StrawberryField` is probably the best library design for our users and this looks pretty good.

> I've removed the .type property override from StrawberryField to let dataclasses set it as it normally would. I've then added .resolved_type property that does the conversion to the Strawberry type.

I want the final user-facing API to just have a property called `.type`, but obviously dataclasses gets in the way. Is is possible to make this change without having renaming the property? Or should we rename it back later if we ever get rid of the `dataclasses.Field` super class?

> .resolved_type should potentially be a cached property since it's doing a fair amount of work. What do you think @BryceBeagle ?

I think that would be wise. I don't think the typing information on arguments/fields/resolvers should ever dynamically change after first access.

strawberry/field.py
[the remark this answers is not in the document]
jkimbo: The typing on `StrawberryArgument` implies that setting both `python_name` and `graphql_name` to `None` is not possible: https://github.com/strawberry-graphql/strawberry/blob/ba9de7c905d790ff8d416340d26097c4de067200/strawberry/arguments.py#L45-L46

> Also, I'm wondering if it would be a smart idea to have a more general-purpose type that pairs an annotation with a default value. This could be used on both fields and arguments

I like this idea. It could contain the argument description as well (which I've just realised I don't have here).

strawberry/types/fields/resolver.py
```diff
@@ -39,12 +39,16 @@ def __call__(self, *args, **kwargs) -> T:
         return self.wrapped_func(*args, **kwargs)
 
     @cached_property
-    def arguments(self) -> List[StrawberryArgument]:
+    def function_parameters(self) -> Mapping[str, inspect.Parameter]:
+        return inspect.signature(self.wrapped_func).parameters
+
+    @cached_property
+    def arguments(self) -> Mapping[str, Type]:
```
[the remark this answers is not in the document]
jkimbo: Good point, will change.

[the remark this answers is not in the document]
jkimbo: Also what should the second argument here be? Is `Type` correct or should it just be `object`?

strawberry/field.py
jkimbo: No, you should be able to override this in a custom field.

strawberry/field.py
jkimbo: Yep agreed. I copied this from the previous function

strawberry/field.py
jkimbo: It should be fine? I haven't changed it's behaviour from before, just moved the optional source arg logic here.

tests/schema/test_custom_fields.py
jkimbo: It's just simpler, otherwise we would have to do the awkward dance of checking which way the decorator is called.

strawberry/field.py
jkimbo: Any suggestions? I don't think `get_result` is a bad name.

jkimbo:
> I want the final user-facing API to just have a property called .type, but obviously dataclasses gets in the way. Is is possible to make this change without having renaming the property? Or should we rename it back later if we ever get rid of the dataclasses.Field super class?

I can have a go at making `.type` work (probably by storing the actual type annotation on a private attribute like `._type` or something. I'm in two minds about it though because would you expect `.type` to return the type annotation or the resolved Strawberry type? For example:

```python
@strawberry.type
class MyType:
	my_field: Optional[str]

field = MyField._type_definition.fields[0]

# should field.type == Optional[str]
# or StrawberryOptional
```

jkimbo:
@BryceBeagle so I tried to make `.type` property return the resolved StrawberryType but it really doesn't work because it's accessed from all over the place and it might get accessed before the namespace is ready to resolve any forward references. So the structure I've gone with is this:

```python
class StrawberryField: # also StrawberryResolver and StrawberryArgument
	@property
	def type(self) -> Optional[Union[object, str]]:
		# This is the "raw" type annotation without any processing
		# Note: for StrawberryField this can optionally come from the resolver

	@property
	def resolved_type(self) -> Union[StrawberryType, type]:
		# This is the resolved type + conversion to StrawberryType
		# This property should only be accessed at schema creation time because it tries to resolve any forward refs
```

Does that make sense?

strawberry/field.py
```diff
@@ -152,32 +212,57 @@ def base_resolver(self, resolver: StrawberryResolver) -> None:
...
+        namespace = module.__dict__
+        return namespace
+
+    def get_type(self) -> Union[object, str]:
+        """
+        Hook to allow custom fields to modify the return type of a field. This
+        hook should only be called once at schema creation time.
+
+        Note: the return value of this function is always the resolved type if
+        it's a forward reference. This is so it's easier to manipulate.
```
patrick91, line 227: this note is a bit confusing 😊

strawberry/schema/schema_converter.py
```diff
@@ -126,12 +128,14 @@ def from_directive(self, directive: DirectiveDefinition) -> GraphQLDirective:
     def from_field(self, field: StrawberryField) -> GraphQLField:
         field_type: GraphQLType
 
-        if isinstance(field.type, StrawberryOptional):
-            field_type = self.from_optional(field.type)
+        resolved_type = field.resolved_type
```
patrick91, line 131: do we need this variable seems redundant :D

strawberry/schema/schema_converter.py
```diff
@@ -294,13 +300,13 @@ def get_graphql_fields() -> Dict[str, GraphQLField]:
...
     def from_resolver(
-        self, field: StrawberryField
+        self, field: StrawberryField, return_type: Union[StrawberryType, type]
     ) -> Callable:  # TODO: Take StrawberryResolver
         def _get_arguments(
             source: Any,
             info: Info,
             kwargs: Dict[str, Any],
-        ) -> Tuple[List[Any], Dict[str, Any]]:
+        ) -> Dict[str, Any]:
```
patrick91, line 309: 👍

patrick91 (commented):
looks good so far, nice one!

strawberry/field.py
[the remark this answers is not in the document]
BryceBeagle: I'm not sold on the idea of of having the description here, as it's unrelated to the _value_ of the data

strawberry/types/fields/resolver.py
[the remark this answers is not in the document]
BryceBeagle: Oh I meant to comment on it. I think it should be `object` for now. I think in the long term, the arguments' typing can get fancy with Generics, though.

Also, when you do use it, I think the best practice is to use `type` instead of `Type` if you're not using the argument to `Type`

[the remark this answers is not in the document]
BryceBeagle: Oh wait, I think I was looking at this wrong. I think this should be `Dict[str, StrawberryType]`?

BryceBeagle:
> @BryceBeagle so I tried to make .type property return the resolved StrawberryType but it really doesn't work because it's accessed from all over the place and it might get accessed before the namespace is ready to resolve any forward references

Unfortunate, but not entirely unexpected. I think the route you've taken here is probably the best. 

One comment though:
> # This property should only be accessed at schema creation time because it tries to resolve any forward refs

This isn't _entirely_ accurate. It's OK to access this value before schema generation (as long as all the objects have been fully built), and also _after_ schema generation

strawberry/schema/schema_converter.py
jkimbo: It's just so I don't have to access it multiple times

jkimbo:
@patrick91 @BryceBeagle I think this ready for another review.

PR #1116, opened 2021-08-06
Use object_type methods to correctly construct strawberry fields

### LINKED ISSUE #1109: Type resolver not called when extending from pydantic model
I have 2 types, Category and Articles, and I want the Category type to return its related Articles(I have tried with other simpler types too), but when defining a resolver in the Category type its not being called:

```python
def articles_helper(root):  
    return []      #  <-whatever I return here, it always shows null
   
   
    model=Category,
    fields=[
        'title',
        'description',
        'content',
        'template',
        'slug',
        'date_published',
        'date_updated',
    ])
class CategoryType(Category):
    id: int
    articles: Optional[List[ArticleType]] = strawberry.field(resolver=articles_helper)  # <- resolver not called
```    

Query type:
```python
@strawberry.type()
class Query:
    category: Optional[CategoryType] = field(resolver=get_category)  # this resolver is OK
```

Always returns `null`:
```
Query:

{
  category(categoryId: 1) {
    id
    title
    articles {
      title
    }
  }
}

Response:

{
  "data": {
    "category": {
      "id": 1,
      "title": "test",
      "articles": null      <- always
    }
  }
}
```

**I works when I don't extend from a pydantic model**

jkimbo:
## Description

Fix for #1109

Note: This PR also changes how we avoid the issue where a field without a default value is defined after a field with a default value. The fix I've implemented is to sort the fields. This will change the order in which they are shown in the schema but I don't think it's a big issue. 

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* Fixes #1109

## Checklist

### DIFF
--- a/strawberry/experimental/pydantic/object_type.py
+++ b/strawberry/experimental/pydantic/object_type.py
@@ -12,7 +12,8 @@ from strawberry.experimental.pydantic.conversion import (
 )
 from strawberry.experimental.pydantic.fields import get_basic_type
 from strawberry.field import StrawberryField
-from strawberry.object_type import _process_type
+from strawberry.object_type import _process_type, _wrap_dataclass
+from strawberry.types.type_resolver import _get_fields
 from strawberry.types.types import FederationTypeParams, TypeDefinition
 
 from .exceptions import MissingFieldsListError, UnregisteredTypeException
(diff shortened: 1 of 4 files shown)


patrick91:
> This will change the order in which they are shown in the schema but I don't think it's a big issue.

Preserving the schema order is useful sometimes, ie apollo federation uses that to check types are the same (not sure it makes a lot of sense to be honest though)

botberry:

```markdown
Release type: patch

```

jkimbo:
> > This will change the order in which they are shown in the schema but I don't think it's a big issue.
> 
> Preserving the schema order is useful sometimes, ie apollo federation uses that to check types are the same (not sure it makes a lot of sense to be honest though)

Yeah it's not ideal but to be honest the best way of resolving it is to replace dataclasses with something that doesn't require fields without defaults to be defined before fields with.

strawberry/experimental/pydantic/object_type.py
```diff
@@ -91,34 +92,36 @@ def wrap(cls):
...
+                    field,
                 )
-                for name, type_ in cls_annotations.items()
+                for field in extra_fields
             )
         )
 
+        # Sort fields so that fields with missing defaults go first
+        # because dataclasses require that fields with no defaults are defined
+        # first
```
patrick91, line 111: should we add a test for this case?

jkimbo: The changes I made to `test_basic` already cover this I think (the order changed)

patrick91: oh man, I just realised this is just for Pydantic :D

PR #1117, opened 2021-08-06
Use action from docker to make it faster

patrick91:
Continuation of #1091, this makes the action a bit faster as it uses the image from docker hub instead of building it all the time :)


botberry:



---

This is a test 🍓🍓

PR #1125, opened 2021-08-08
Add support for async permission classes

### LINKED ISSUE #1085: Allow async `has_permission` methods for `permission_classes` on fields
I am trying to use sqlalchemy async ORM everywhere but there is not currently support for async `has_permission` methods on permission classes. As the actual resolver can be async, I can simply check/raise at the beginning of each resolver, but it is a bit clumsy to do something in the resolver when there is a mechanism there for it. A brief look makes me think this will quite likely be a lot of work, and with a simple workaround, there may not be a pressing need for it.

DoctorJohn:
## Description

This PR adds support for async permissions. I made sure sync and async permission classes and resolvers can be combined and that permissions are always checked before the protected resolver was called.

This ended up being really messy (just like when we had to write async wrappers to handle enum conversion), so I decided to have separate sync and async resolve functions in the schema coverter. Looking at graphql-core code that's how it's supposed to be. Before we would put extra async functionality (enum conversion and now permission checks) into an async wrapper that graphql-core would call to finalize async values returned from resolvers, which worked but was hard to extend and maintain.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

* Fixes #1085

## Checklist

### DIFF
--- a/docs/guides/permissions.md
+++ b/docs/guides/permissions.md
@@ -10,6 +10,7 @@ field using the `permission_classes` keyword argument. A basic example looks
 like this:
 
 ```python
+import typing
 import strawberry
 from strawberry.permission import BasePermission
 from strawberry.types import Info
@@ -17,7 +18,8 @@ from strawberry.types import Info
 class IsAuthenticated(BasePermission):
     message = "User is not authenticated"
 
-    def has_permission(self, source: Any, info: Info, **kwargs) -> bool:
+    # This method can also be async!
+    def has_permission(self, source: typing.Any, info: Info, **kwargs) -> bool:
         return False
 
 @strawberry.type
@@ -26,11 +28,16 @@ class Query:
 ```
 
 Your `has_permission` method should do the work to check if this request
-has permission to the field. If the `has_permission` method returns `True` or
-a truthy value then the field access will go ahead. If the `has_permission`
-method returns `False` or a falsy value then an error will be raised using
-the `message` class attribute. See [Dealing with Errors](/docs/guides/errors)
-for more information on how errors are handled.
+has permission to the field.
+
+If the `has_permission` method returns a truthy value or an awaitable resolving to a
+truthy value then the field access will go ahead.
+
+If the `has_permission` method returns a falsy value or an awaitable resolving to a
+falsy value then an error will be raised using the `message` class attribute.
+
+Take a look at our [Dealing with Errors Guide](/docs/guides/errors) for more information
+on how errors are handled.
 
 ```json
 {
@@ -46,32 +53,41 @@ for more information on how errors are handled.
 ## Accessing user information
 
 Accessing the current user information to implement your permission checks
-depends on the web framework you are using. Most frameworks will have a
-`Request` object where you can either access the current user directly or access
-headers/cookies/query parameters to authenticate the user. All the Strawberry
-integrations provide this Request object in the `info.context` object that is
-accessible in every resolver and in the `has_permission` function. You can find
-more details about a specific framework integration under the "Integrations"
-heading in the navigation.
+depends on the web framework you are using.
+
+Most frameworks will have a `Request` object where you can either access the current
+user directly or access headers/cookies/query parameters to authenticate the user.
+
+All the Strawberry integrations provide this Request object in the `info.context` object
+that is accessible in every resolver and in the `has_permission` function.
+
+You can find more details about a specific framework integration under the
+"Integrations" heading in the navigation.
 
 In this example we are using `starlette` which uses the
 [ASGI](/docs/integrations/asgi) integration:
 
 ```python
+import typing
 from myauth import authenticate_header, authenticate_query_param
 
 from starlette.requests import Request
 from starlette.websockets import WebSocket
+from strawberry.permission import BasePermission
+from strawberry.types import Info
 
 class IsAuthenticated(BasePermission):
     message = "User is not authenticated"
 
-    def has_permission(self, source: Any, info: Info, **kwargs) -> bool:
-        request: Union[Request, Websocket] = info.context["request"]
+    def has_permission(self, source: typing.Any, info: Info, **kwargs) -> bool:
+        request: typing.Union[Request, WebSocket] = info.context["request"]
+
         if "Authorization" in request.headers:
             return authenticate_header(request)
+
         if "auth" in request.query_params:
             return authenticate_query_params(request)
+
         return False
 ```
 
--- a/docs/guides/permissions.md
+++ b/docs/guides/permissions.md
@@ -27,14 +27,11 @@ class Query:
     user: str = strawberry.field(permission_classes=[IsAuthenticated])
 ```
 
-Your `has_permission` method should do the work to check if this request
-has permission to access the field.
+Your `has_permission` method should check if this request has permission to access the
+field. Note that the `has_permission` method can also be asynchronous.
 
-If the `has_permission` method returns a truthy value or an awaitable resolving to a
-truthy value then the field access will go ahead.
-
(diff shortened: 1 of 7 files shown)


strawberry/field.py
```diff
@@ -246,6 +247,13 @@ def get_result(
         if self.base_resolver:
             return self.base_resolver(*args, **kwargs)
 
+        if inspect.isawaitable(source):
```
jkimbo, line 250: Is this necessary? When would this be asynchronous?

strawberry/schema/schema_converter.py
```diff
@@ -336,6 +336,32 @@ def _check_permissions(source: Any, info: Info, kwargs: Dict[str, Any]):
...
+                else:
+                    has_permission = permission.has_permission(  # type: ignore
+                        source, info, **kwargs
+                    )
+
+                if not has_permission:
+                    message = getattr(permission, "message", None)
+                    raise PermissionError(message)
+
+        def _has_async_permission_classes() -> bool:
```
jkimbo, line 359: It would be good to hoist this to the field definition so that it only runs once. Having it here means that it will run every time the resolver is called.

strawberry/schema/schema_converter.py
```diff
@@ -352,33 +378,35 @@ def _resolver(_source: Any, info: GraphQLResolveInfo, **kwargs):
...
+                source=_source, info=strawberry_info, kwargs=kwargs
+            )
 
-                return await_result(result)
+            return field.get_result(
+                _source, info=strawberry_info, args=field_args, kwargs=field_kwargs
+            )
 
-            return result
+        has_async_base_resolver = inspect.iscoroutinefunction(field.base_resolver)
```
jkimbo, line 401: Hoisting this the field definition would also be good.

jkimbo (changes requested):
Change looks sensible to me. Added some comments.

strawberry/field.py
DoctorJohn: In a nutshell: Our tests pass without it but the following schema would produce an error when queried using the debug server

```python
import strawberry


class CanView(strawberry.BasePermission):
    async def has_permission(self, source, info, **kwargs):
        return True


@strawberry.type
class User:
    name: str


@strawberry.type
class Query:
    @strawberry.field(permission_classes=[CanView])
    async def user(self, name: str) -> User:
        return User(name=name)


```

I'm currently further investigating this issue...

DoctorJohn: Found the issue, in some cases unresolved results were returned from the async resolver function I added. Didn't noticed this was an issue because I missinterpreted some of the graphql-core code. I updated the test to cover this case and fixed it.

strawberry/schema/schema_converter.py
DoctorJohn: I'm undecided between calculating whether there are async permission classes in the constructor vs. having a cached property. The calculated value depends on the StrawberryField.permission_classes property which technically is mutable at the moment. I went with having a cached_property for now but I'm open for suggestions.

(we could make permission classes immutable by adding a property, or a setter that recalculates whether there are async permission classes).

strawberry/schema/schema_converter.py
DoctorJohn: Inspecting the base_resolver turned out to be unnecessary.

botberry:



---

This release adds support for asynchronous permission classes. The only difference to
their synchronous counterpart is that the `has_permission` method is asynchronous.

```python

    async def has_permission(self, source, info, **kwargs):
        return True
```


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Jonathan Ehwald for the PR 👏

```

strawberry/schema/schema_converter.py
```diff
@@ -336,6 +336,26 @@ def _check_permissions(source: Any, info: Info, kwargs: Dict[str, Any]):
...
+                has_permission: bool
+
+                if inspect.iscoroutinefunction(permission.has_permission):
+                    has_permission = await permission.has_permission(  # type: ignore
+                        source, info, **kwargs
+                    )
+                else:
+                    has_permission = permission.has_permission(  # type: ignore
+                        source, info, **kwargs
+                    )
```
jkimbo, line 353: The `await_me_maybe` function would make this code a bit cleaner (https://simonwillison.net/2020/Sep/2/await-me-maybe/):

```suggestion
                
                has_permission = await await_me_maybe(
                	permission.has_permission(source, info, **kwargs)
                )
```

strawberry/schema/schema_converter.py
```diff
@@ -348,37 +368,38 @@ def _strawberry_info_from_graphql(info: GraphQLResolveInfo) -> Info:
...
+            return _get_result(_source, strawberry_info, **kwargs)
 
-            if iscoroutine(result):  # pragma: no cover
+        async def _async_resolver(_source: Any, info: GraphQLResolveInfo, **kwargs):
+            strawberry_info = _strawberry_info_from_graphql(info)
+            await _check_permissions_async(_source, strawberry_info, kwargs)
 
-                async def await_result(result):
-                    return await result
+            result = _get_result(_source, strawberry_info, **kwargs)
```
jkimbo, line 390: The `await_me_maybe` function could be used here as well.

strawberry/schema/schema_converter.py
```diff
@@ -348,37 +368,38 @@ def _strawberry_info_from_graphql(info: GraphQLResolveInfo) -> Info:
...
 
-                return await_result(result)
+            if inspect.isawaitable(result):
+                return await result
 
             return result
 
-        _resolver._is_default = not field.base_resolver  # type: ignore
-        return _resolver
+        if field.has_async_permission_classes:
```
jkimbo, line 397:
```suggestion
        if field.has_async_permission_classes or (field.base_resolver is not None and field.base_resolver.is_async):
```

docs/guides/permissions.md
```diff
@@ -26,11 +28,16 @@ class Query:
 ```
 
 Your `has_permission` method should do the work to check if this request
-has permission to the field. If the `has_permission` method returns `True` or
-a truthy value then the field access will go ahead. If the `has_permission`
-method returns `False` or a falsy value then an error will be raised using
-the `message` class attribute. See [Dealing with Errors](/docs/guides/errors)
-for more information on how errors are handled.
+has permission to the field.
```
jkimbo, line 31:
```suggestion
has permission to access the field.
```

(I know this was the original wording)

docs/guides/permissions.md
```diff
@@ -26,11 +28,16 @@ class Query:
...
+
+If the `has_permission` method returns a truthy value or an awaitable resolving to a
+truthy value then the field access will go ahead.
```
jkimbo, line 34: I don't think you need to spell this out, it's enough explaining that the `has_permission` method can be an async method.

docs/guides/permissions.md
```diff
@@ -26,11 +28,16 @@ class Query:
...
+
+If the `has_permission` method returns a falsy value or an awaitable resolving to a
+falsy value then an error will be raised using the `message` class attribute.
```
jkimbo, line 37: Same comment as above.

strawberry/schema/schema_converter.py
jkimbo: Might be nice to have an `is_async` property on the field that encompasses all this logic.

jkimbo (changes requested):

strawberry/schema/schema_converter.py
DoctorJohn: Excellent idea, didn't saw base_resolver had a similar property already.

jkimbo (approved):
Very nice 👍

PR #1126, opened 2021-08-08
Add workflow to send tweet when release is published

patrick91:
This PR updates our workflows to send a tweet when a new release is published.

I've changed a bit how our actions work:

- The bot action now also sends the tweet, so we can have a preview in the PR
- There's a new action to get info about a PR even when triggered on the main branch on merge (we can reuse to invite people to the organisation)
- The action that reads tweets also has a default tweet value, and supports variables in the tweet message ($contributor, $version and $release_url)
- The release action doesn't fail, failure needs to be trigger manually, this to make the action more reusable
- I've removed the inclusive organisation workflow, it wasn't working
- The release workflow is more complex now, it reads the tweet, generates the preview and after release it sends a tweet


botberry:



---

This is a test


---

PR #1136, opened 2021-08-09
Use pull_request_target

patrick91:


patrick91:
Seems to work, as tested here: https://github.com/strawberry-graphql/strawberry/pull/1138

Issue #1061, opened 2021-07-13
Overriding parsers/serializers of existing scalars


Apakottur:
Is there a way to override parsers/serializers of existing scalars like `Date`, `DateTime` or `ID`?

For example, if I want to use the following function
```python
def my_cool_parser(value: Any) -> datetime.datetime:
    if ...:
       return dateutil.parser.isoparse(value)
    else:
       ...
```

Instead of `dateutil.parser.isoparse` for parsing datetimes, it would be awesome to have something along:

```python
from strawberry.??? import scalar_mapping
scalar_mapping[datetime.datetime].parse_value = my_cool_parser
```

------

I didn't find a clean way of doing this so I ended up overriding `strawberry.schema.types.scalar.DEFAULT_SCALAR_REGISTRY` before the schema is being generated, which seems to work.

jkimbo:
I think this is related to #1028 - ideally scalars would be scoped to a schema and they could be overridden.

jkimbo:
@Apakottur you should now be able to override the built in scalar implementations in [v0.75.0](https://github.com/strawberry-graphql/strawberry/releases/tag/0.75.0)

Let me know if you have any issues with it.

PR #1147, opened 2021-08-13
Bring scalar definitions into schema and allow overriding

### LINKED ISSUE #1028: Rewrite storage scalars
if I create two Schemas, there may be scalars from both schemas.
https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/custom_scalar.py#L21

jkimbo:
## Description

This PR removes the global scalar registry in favour of a registry per schema. It also adds a new function to the Schema `get_scalar` that allows you to override the built in scalars with a custom implementation.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* Fixes #1028 and #1061 

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,39 @@
+Release type: minor
+
+This release improves how we deal with custom scalars. Instead of being global
+they are now scoped to the schema. This allows you to have multiple schemas in
+the same project with differnet scalars.
+
+Also you can now override the built in scalars with your own custom
+implementation. Out of the box Strawberry provides you with custom scalars for
+common Python types like `datetime` and `Decimal`. If you require a custom
+implementation of one of these built in scalars you can how subclass the
+`Schema` class and override the `get_scalar` function to return your own custom
+implementation:
+
+```python
+from datetime import datetime, timezone
+import strawberry
+
+EpocDateTime = strawberry.scalar(
+    datetime,
+    serialize=lambda value: int(value.timestamp()),
+    parse_value=lambda value: datetime.fromtimestamp(int(value), timezone.utc),
+)
+
+class MySchema(strawberry.Schema):
+    def get_scalar(self, scalar):
+        if scalar == datetime:
+            return EpocDateTime
+        return super().get_scalar(scalar)
+
+@strawberry.type
+class Query:
+    @strawberry.field
+    def current_time(self) -> datetime:
+        return datetime.now()
+
+schema = MySchema(Query)
+result = schema.execute_sync("{ currentTime }")
+assert result.data == {"currentTime": 1628683200}
+```
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,7 +2,7 @@ Release type: minor
 
 This release improves how we deal with custom scalars. Instead of being global
 they are now scoped to the schema. This allows you to have multiple schemas in
-the same project with differnet scalars.
+the same project with different scalars.
 
 Also you can now override the built in scalars with your own custom
 implementation. Out of the box Strawberry provides you with custom scalars for
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -7,9 +7,8 @@ the same project with differnet scalars.
 Also you can now override the built in scalars with your own custom
 implementation. Out of the box Strawberry provides you with custom scalars for
 common Python types like `datetime` and `Decimal`. If you require a custom
-implementation of one of these built in scalars you can how subclass the
-`Schema` class and override the `get_scalar` function to return your own custom
-implementation:
+implementation of one of these built in scalars you can now pass a map of
+overrides to your schema:
 
 ```python
 from datetime import datetime, timezone
@@ -21,19 +20,18 @@ EpocDateTime = strawberry.scalar(
     parse_value=lambda value: datetime.fromtimestamp(int(value), timezone.utc),
 )
 
-class MySchema(strawberry.Schema):
-    def get_scalar(self, scalar):
-        if scalar == datetime:
-            return EpocDateTime
-        return super().get_scalar(scalar)
-
 @strawberry.type
 class Query:
     @strawberry.field
     def current_time(self) -> datetime:
         return datetime.now()
 
-schema = MySchema(Query)
+schema = strawberry.Schema(
+  Query,
+  scalar_overrides={
+    datetime: EpocDateTime,
+  }
+)
 result = schema.execute_sync("{ currentTime }")
 assert result.data == {"currentTime": 1628683200}
 ```
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -14,7 +14,7 @@ overrides to your schema:
 from datetime import datetime, timezone
 import strawberry
 
-EpocDateTime = strawberry.scalar(
+EpochDateTime = strawberry.scalar(
     datetime,
     serialize=lambda value: int(value.timestamp()),
     parse_value=lambda value: datetime.fromtimestamp(int(value), timezone.utc),
@@ -29,7 +29,7 @@ class Query:
 schema = strawberry.Schema(
   Query,
   scalar_overrides={
-    datetime: EpocDateTime,
+    datetime: EpochDateTime,
   }
 )
 result = schema.execute_sync("{ currentTime }")
--- a/TWEET.md
+++ b/TWEET.md
@@ -1,0 +1,3 @@
+This release improves how we deal with custom scalars. Instead of being
+global they are now scoped to the current schema. You can also override
+the built in scalars with your own custom implementation.
--- a/TWEET.md
+++ b/TWEET.md
@@ -1,4 +1,4 @@
-🆕 Release $version is out! Thanks to @jkimbo for this great new feature! 
+🆕 Release $version is out! Thanks to @jkimbo for this great new feature!
 We are now able to override the default scalars, so you don't have to stick
 with the defaults provided by us 😊
 
--- a/TWEET.md
+++ b/TWEET.md
@@ -1,3 +1,11 @@
+🆕 Release $version is out! Thanks to @jkimbo for this great new feature! 
+We are now able to override the default scalars, so you don't have to stick
+with the defaults provided by us 😊
+
+Get it here 👉 $release_url
+
+---
+
 This release improves how we deal with custom scalars. Instead of being
 global they are now scoped to the current schema. You can also override
 the built in scalars with your own custom implementation.
--- a/docs/types/scalars.md
+++ b/docs/types/scalars.md
@@ -160,3 +160,38 @@ query ExampleDataQuery {
   }
 }
 ```
+
+## Overriding built in scalars
+
+To override the behaviour of the built in scalars you can create a custom Schema
+and implement a custom `get_scalar` function.
+
+Here is a full example of replacing the built in `DateTime` scalar with one that
+serializes all datetimes as unix timestamps:
+
+```python
+from datetime import datetime, timezone
+import strawberry
+
+EpocDateTime = strawberry.scalar(
+    datetime,
+    serialize=lambda value: int(value.timestamp()),
+    parse_value=lambda value: datetime.fromtimestamp(int(value), timezone.utc),
+)
+
+class MySchema(strawberry.Schema):
+    def get_scalar(self, scalar):
+        if scalar == datetime:
+            return EpocDateTime
+        return super().get_scalar(scalar)
+
+@strawberry.type
+class Query:
+    @strawberry.field
+    def current_time(self) -> datetime:
+        return datetime.now()
+
+schema = MySchema(Query)
+result = schema.execute_sync("{ currentTime }")
+assert result.data == {"currentTime": 1628683200}
+```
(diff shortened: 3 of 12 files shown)


botberry:



---

This release improves how we deal with custom scalars. Instead of being global
they are now scoped to the schema. This allows you to have multiple schemas in
the same project with differnet scalars.

Also you can now override the built in scalars with your own custom
implementation. Out of the box Strawberry provides you with custom scalars for
common Python types like `datetime` and `Decimal`. If you require a custom
implementation of one of these built in scalars you can how subclass the
`Schema` class and override the `get_scalar` function to return your own custom
implementation:

```python
from datetime import datetime, timezone
import strawberry

EpocDateTime = strawberry.scalar(
    datetime,
    serialize=lambda value: int(value.timestamp()),
    parse_value=lambda value: datetime.fromtimestamp(int(value), timezone.utc),
)

class MySchema(strawberry.Schema):
    def get_scalar(self, scalar):
        if scalar == datetime:
            return EpocDateTime
        return super().get_scalar(scalar)

@strawberry.type
class Query:
    @strawberry.field
    def current_time(self) -> datetime:
        return datetime.now()

schema = MySchema(Query)
result = schema.execute_sync("{ currentTime }")
assert result.data == {"currentTime": 1628683200}
```


---




Here's the tweet text: 
```

```

tests/schema/test_custom_scalar.py
```diff
@@ -85,10 +81,3 @@ def my_str(self, arg: MyStr) -> MyStr:
 
     assert not result.errors
     assert result.data["myStr"] == "valueSuffix"
-
-
-def test_error_when_registering_duplicate_scalar():
```
patrick91, line 90: what happens now if we create a scalar with the same name twice?

strawberry/schema/schema_converter.py
```diff
@@ -64,9 +68,11 @@ def serialize(self, output_value: Any) -> str:
 class GraphQLCoreConverter:
     # TODO: Make abstract
 
-    def __init__(self, config: StrawberryConfig):
+    def __init__(self, schema: "Schema"):
         self.type_map: Dict[str, ConcreteType] = {}
-        self.config = config
+        self.config = schema.config
+        self.schema = schema
```
patrick91, line 74: do you think we can find a way to decouple the converter from the schema?
now we have a circular dependency between schema and schema converter, not sure this is ideal /cc @BryceBeagle

patrick91 (commented):
This looks very useful! I've added a couple of comments 😊

RELEASE.md
```diff
@@ -0,0 +1,39 @@
+Release type: minor
+
+This release improves how we deal with custom scalars. Instead of being global
+they are now scoped to the schema. This allows you to have multiple schemas in
+the same project with differnet scalars.
```
BryceBeagle, line 5:
```suggestion
This release improves how we deal with custom scalars. Instead of being global
they are now scoped to the schema. This allows you to have multiple schemas in
the same project with different scalars.
```

RELEASE.md
```diff
@@ -0,0 +1,39 @@
...
+    datetime,
+    serialize=lambda value: int(value.timestamp()),
+    parse_value=lambda value: datetime.fromtimestamp(int(value), timezone.utc),
+)
+
+class MySchema(strawberry.Schema):
+    def get_scalar(self, scalar):
+        if scalar == datetime:
+            return EpocDateTime
+        return super().get_scalar(scalar)
```
BryceBeagle, line 28: This seems a bit redundant as `datetime` is the first argument to `strawberry.scalar`. I wonder if there's a different mechanism we could use to have a schema "use" a scalar?

strawberry/schema/schema.py
```diff
@@ -216,3 +218,12 @@ def introspect(self) -> Dict[str, Any]:
...
         return introspection.data
+
+    def get_scalar(
+        self, scalar: Type
+    ) -> Union[ScalarWrapper, ScalarDefinition, GraphQLScalarType]:
+        if scalar in DEFAULT_SCALAR_REGISTRY:
+            return DEFAULT_SCALAR_REGISTRY[scalar]
+
+        scalar_definition = scalar._scalar_definition
+        return scalar_definition
```
BryceBeagle, line 229: How do we ensure that `get_scalar` is only called on objects with `._scalar_definition`?

BryceBeagle (changes requested):

strawberry/schema/schema_converter.py
BryceBeagle: Yeah I'm not sure if instantiating the converter _with_ the schema is a good idea. imo the converter should be a separate entity. I do see the problem that as implemented, `from_scalar` needs to call `Schema.get_scalar`. :thinking:

tests/schema/test_custom_scalar.py
BryceBeagle: Two situations to consider here:
1. Two scalars with same name for two _different_ schemas
2. Two scalars with same name for the _same_ schema

jkimbo: > what happens now if we create a scalar with the same name twice?

I (think) the schema creation would fail in the same way that it would fail if there were 2 object types that had the same name. Will add a PR to verify it though.

strawberry/schema/schema_converter.py
jkimbo: Yeah don't know how to do this any other way. We need ways to define hooks on the GraphQL convertor so that we can change functionality. We could pass the "get_scalar" hook as an extension instead of subclassing the schema?

RELEASE.md
jkimbo: I don't follow. We're not creating the scalar here, we're just processing the scalars that are already defined in the schema.

strawberry/schema/schema.py
jkimbo: I doesn't have to be. If you process the scalar before calling `super` then you're just dealing with the raw python type.

strawberry/schema/schema_converter.py
jkimbo: I've changed the approach here so that we don't have this dependency. Now you pass a map of scalar overrides to the schema which then gets passed to the schema convertor. I think this is a better approach.

jkimbo:
@patrick91 @BryceBeagle this PR is ready for review again.

RELEASE.md
BryceBeagle: I think my concern was that we first define the custom scalar by saying that it works _for_ `datetime`:
```python
EpochDateTime = strawberry.scalar(
    datetime,
    ...,
)
```
and then when we want to _use_ the scalar with a schema, we again say that the scalar is being used _for_ `datetime`:
```python
schema = strawberry.Schema(
  Query,
  scalar_overrides={
    datetime: EpochDateTime,
  }
)
```

Can the `scalar_overrides` just take a list of Scalars and figure out what they should be applied to by looking inside the custom scalar?
 ```python
schema = strawberry.Schema(
  Query,
  scalar_overrides=[EpochDateTime],
)
```

I guess the disadvantage of this is that you can't, for example, make a custom scalar that works for all numbers (i.e. `float`), and then tell the schema to use it only for `int`s?

File: RELEASE.md [hunk 6]
```diff
@@ -0,0 +1,37 @@
...
+implementation. Out of the box Strawberry provides you with custom scalars for
+common Python types like `datetime` and `Decimal`. If you require a custom
+implementation of one of these built in scalars you can now pass a map of
+overrides to your schema:
+
+```python
+from datetime import datetime, timezone
+import strawberry
+
+EpocDateTime = strawberry.scalar(
```
BryceBeagle, line 17:
```suggestion
EpochDateTime = strawberry.scalar(
```

RELEASE.md
```diff
@@ -0,0 +1,37 @@
...
+@strawberry.type
+class Query:
+    @strawberry.field
+    def current_time(self) -> datetime:
+        return datetime.now()
+
+schema = strawberry.Schema(
+  Query,
+  scalar_overrides={
+    datetime: EpocDateTime,
```
BryceBeagle, line 32:
```suggestion
    datetime: EpochDateTime,
```

docs/types/scalars.md
```diff
@@ -160,3 +160,38 @@ query ExampleDataQuery {
...
+
+Here is a full example of replacing the built in `DateTime` scalar with one that
+serializes all datetimes as unix timestamps:
+
+```python
+from datetime import datetime, timezone
+import strawberry
+
+# Define your custom scalar
+EpocDateTime = strawberry.scalar(
```
BryceBeagle, line 177:
```suggestion
EpochDateTime = strawberry.scalar(
```

docs/types/scalars.md
```diff
@@ -160,3 +160,38 @@ query ExampleDataQuery {
...
+@strawberry.type
+class Query:
+    @strawberry.field
+    def current_time(self) -> datetime:
+        return datetime.now()
+
+schema = strawberry.Schema(
+  Query,
+  scalar_overrides={
+    datetime: EpocDateTime,
```
BryceBeagle, line 192:
```suggestion
    datetime: EpochDateTime,
```

tests/schema/test_scalars.py
```diff
@@ -94,3 +97,63 @@ def uuid_input(self, input_id: UUID) -> str:
     assert result.data == {
         "uuidInput": "e350746c-33b6-4469-86b0-5f16e1e12232",
     }
+
+
+def test_override_built_in_scalars():
+    EpocDateTime = strawberry.scalar(
```
BryceBeagle, line 103:
```suggestion
    EpochDateTime = strawberry.scalar(
```

tests/schema/test_scalars.py
```diff
@@ -94,3 +97,63 @@ def uuid_input(self, input_id: UUID) -> str:
...
+            return datetime(2021, 8, 11, 12, 0, tzinfo=timezone.utc)
+
+        @strawberry.field
+        def isoformat(self, input_datetime: datetime) -> str:
+            return input_datetime.isoformat()
+
+    schema = strawberry.Schema(
+        Query,
+        scalar_overrides={
+            datetime: EpocDateTime,
```
BryceBeagle, line 122:
```suggestion
            datetime: EpochDateTime,
```

BryceBeagle (dismissed):

RELEASE.md
jkimbo: Oh I see. The issue is that you don't have to provide a first param to `strawberry.scalar`, it's mainly a helper function: https://github.com/strawberry-graphql/strawberry/blob/71e6adb743df4d2830b5a3b046d33adf40daddc3/strawberry/custom_scalar.py#L90-L96

So I'm not sure we could reliably replace the correct scalar types.

strawberry/custom_scalar.py
```diff
@@ -19,6 +20,10 @@ class ScalarDefinition(StrawberryType):
     parse_value: Optional[Callable]
     parse_literal: Optional[Callable]
 
+    # Optionally store the GraphQLScalarType instance so that we don't get
+    # duplicates
+    implementation: Optional[GraphQLScalarType] = None
```
patrick91, line 25: we should find a way to get rid of this, as we don't want to couple Strawberry types to GraphQL-core, but we can do this in another PR :)

patrick91 (approved):
This looks ready to go me 😊 @jkimbo can you add a TWEET.md file, or shall I do it :D

RELEASE.md
patrick91: strawberry.scalar should return the class, no? so this should be doable I think 😊

strawberry/custom_scalar.py
jkimbo: Yeah it’s not ideal…

RELEASE.md
jkimbo: Hmm possibly. I think the mapping is clearer though

patrick91: yes, I think to too :)

I think we can add some error checking in another PR, for example in this case:

```
schema = strawberry.Schema(
  Query,
  scalar_overrides={
    int: EpochDateTime,
  }
)
```

strawberry/custom_scalar.py
patrick91: I think when we introduce StrawberryScalar we can change this :)

jkimbo:
> This looks ready to go me 😊 @jkimbo can you add a TWEET.md file, or shall I do it :D

@patrick91 done!

TWEET.md
```diff
@@ -0,0 +1,3 @@
+This release improves how we deal with custom scalars. Instead of being
+global they are now scoped to the current schema. You can also override
+the built in scalars with your own custom implementation.
```
patrick91, line 3:
```suggestion
🆕 Release $version is out! Thanks to @jkimbo for this great new feature! 
We are now able to override the default scalars, so you don't have to stick
with the defaults provided by us 😊

Get it here 👉 $release_url

---

This release improves how we deal with custom scalars. Instead of being
global they are now scoped to the current schema. You can also override
the built in scalars with your own custom implementation.
```

PR #1150, opened 2021-08-15
Fix issue when returning non strawberry types when return type extends interface

### LINKED ISSUE #1106: Broken duck typing when type implements an interface
This is currently broken (and it has been for a while apparently) due to how we handle `is_type_of`.

If we change it to none it works: https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/schema/schema_converter.py#L265-L269

```python
from dataclasses import dataclass
from typing import List

import strawberry
from strawberry.types import Info


@strawberry.interface
class Node:
    id: int


@strawberry.type
class Anime(Node):
    anime_type: str


@dataclass
class AnimeORM:
    id: int
    anime_type: str


anime_list = [
    AnimeORM(
        id=1,
        anime_type="OVA",
    )
]


@strawberry.type
class Query:
    # @SAPaginatedField
    @strawberry.field
    def get_anime(self, info: Info) -> List[Anime]:
        return anime_list


print(
    schema.execute_sync(
        """{
            getAnime { id }
       }""",
        root_value=Query(),
    ).data,
)
```

patrick91:
Closes #1106

This is not the perfect solution, but it is a start. I think in future we need to check the shape of the returned object matches the shape of one of the types that implement that interface.


botberry:



---

This releases fixes an issue where you were not allowed
to use duck typing and return a different type that the
type declared on the field when the type was implementing
an interface. Now this works as long as you return a type
that has the same shape as the field type.


---




Here's the tweet text: 
```

```

strawberry/schema/schema_converter.py
```diff
@@ -222,11 +222,17 @@ def get_graphql_fields() -> Dict[str, GraphQLField]:
 
             return graphql_fields
 
+        def resolve_type(obj, info, type_):
+            # TODO: this will probably break when passing dicts
+            # or even non strawberry types
+            return self.type_map[obj.__class__._type_definition.name].implementation
```
BryceBeagle, line 228: Can we add type annotations here?

tests/schema/test_interface.py
```diff
@@ -93,3 +94,35 @@ def always_error(self) -> Error:
...
+    @dataclass
+    class AnimeORM:
+        id: int
+        name: str
+
+    @strawberry.type
+    class Query:
+        @strawberry.field
+        def anime(self) -> Anime:
+            return AnimeORM(id=1, name="One Piece")  # type: ignore
```
BryceBeagle, line 117: Are we sure we want to support this kinda of duck typing? This feels like potential for bugs in user code.

strawberry/schema/schema_converter.py
```diff
@@ -222,11 +222,17 @@ def get_graphql_fields() -> Dict[str, GraphQLField]:
 
             return graphql_fields
 
+        def resolve_type(obj, info, type_):
+            # TODO: this will probably break when passing dicts
+            # or even non strawberry types
```
BryceBeagle, line 227: Should this PR be resolved before merge? Or xfail tests added?

tests/schema/test_interface.py
jkimbo: IMO we shouldn't support it. I think it leads to more issues than it solves. It's related to this discussion: https://github.com/strawberry-graphql/strawberry/discussions/515

patrick91: do we want to support returning dictionaries though?

strawberry/schema/schema_converter.py
patrick91: We should add a test for sure :)

tests/schema/test_interface.py
patrick91: I see strawberry types as protocols, we should be able to allow users to returns objects (and dictionary) that match the shape :)

We can also add options to disable this behaviour though (maybe enabled by default)

jkimbo: > I see strawberry types as protocols, we should be able to allow users to returns objects (and dictionary) that match the shape :)

I like the idea but IMO it doesn't fit with the ability to define resolver functions on the type directly or with the dataclasses API generally.

If the Strawberry API was something like this:

```python
MyType = strawberry.type(name='MyType', fields={
	"name": str,
	"age": strawberry.field(str, resolve=lambda obj: 20),
})
```

then it makes sense to allow resolvers to return dictionaries (this is basically the graphql-core API)

The conclusion we came to in https://github.com/strawberry-graphql/strawberry/discussions/515 (which I still agree with) is that we don't disallow returning other things than StrawberryTypes from a resolver, but it's only an advanced option and the docs should always return StrawberryTypes.

patrick91: what about:

```python
@strawberry.type
class MyType:
    age: strawberry.field(resolver=some_resolver)
```

?

> The conclusion we came to in #515 (which I still agree with) is that we don't disallow returning other things than StrawberryTypes from a resolver, but it's only an advanced option and the docs should always return StrawberryTypes.

I think we should write it somewhere that we support this (either by default or with an option) :)

jkimbo: > what about:
>
> @strawberry.type
> class MyType:
>     age: strawberry.field(resolver=some_resolver)

What are you suggesting here? That that is the only way to define resolvers?

> I think we should write it somewhere that we support this (either by default or with an option) :)

Yep totally agree.

patrick91: > What are you suggesting here? That that is the only way to define resolvers?

No, I was wondering if that does fall into the "issue" mentioned here https://github.com/strawberry-graphql/strawberry/discussions/515

jkimbo: Well the issue is what should the call signature of `some_resolver` look like?

It could be:

```python
def some_resolver(obj: MyType) -> int
```

or it could be:

```python
def some_resolver(obj: dict) -> int
```

it is completely dependant on what the resolver for `MyType` returns in the parent type. If we support returning whatever you want from resolvers that return object types then your child resolvers need to take into account all the possible options. If we say that object type resolvers only return an instance of the type, then the resolver signature can only look like the first option. IMO it reduces a whole bunch of confusion.

patrick91: right, I didn't consider that, though it would be ok for Protocols/duck-typing. Not for dicts. I wonder if we could add some magic, but let's leave for the future :D

patrick91:
@jkimbo I've added a (failing) test for dictionaries, do you think we can merge this?

jkimbo (approved):

Issue #841, opened 2021-04-12
printing info.context leads to %d format: a number is required, not NoneType


Speedy1991:
strawberry-graphql==0.56.2
django==3.1.7

schema2.py:
```
@strawberry.type
class Mutation:
    @strawberry.mutation
    def create_bla_1(self, info) -> str:
        print(info.context.request)
        print(info.context)
        return "abc"


@strawberry.type
class Query:
    @strawberry.field
    def hij(self, info) -> str:
        print(info.context.request)
        print(info.context)
        return "hij"


schema = strawberry.Schema(query=Query, mutation=Mutation)
```

urls.py
```
from strawberry.django.views import GraphQLView
...
path('schema2/', csrf_exempt(GraphQLView.as_view(schema=schema2, graphiql=True))),
```

example query (graphiql):
`{ hij }`

Leads to
```
{
  "data": null,
  "errors": [
    {
      "message": "%d format: a number is required, not NoneType",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "hij"
      ]
    }
  ]
}
```

Maybe this is only django related

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/strawberry-graphql) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/strawberry-graphql/strawberry/issues/841">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/841/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/841/pledge.svg">
</picture>
</a>

PR #1163, opened 2021-08-23
Allow printing of info.context in a resolver used by a Django view.

ddirkson:
## Description

Calling print on the StrawberryDjangoContext class called into the __repr__
method of Django's HttpResponse class. If print was called on the context
class before a response was instantiated it caused an exception since the
response class has a custom __repr__ method that expects certain attributes
like status_code to be set on the class.

To fix the issue, this commit overrides the __repr__ method in
StrawberryDjangoContext to just return the __class__ property.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #841 

## Checklist


botberry:



---

Allow printing of info.context in a resolver used by a Django view.


---




Here's the tweet text: 
```
Version (next) is out! This time with a new cool feature by @NucleonJohn! ✨

Now our extensions support async methods! And you can mix and match async and
non-async methods so this should make it easier to adopt async for your
extensions 😊
```

tests/django/test_views.py
```diff
@@ -329,3 +329,31 @@ def abc(self, info: Info) -> str:
...
+    # Swallow the print statements to keep the logs clean.
+    mock = mocker.patch("builtins.print")  # noqa: F841
+    factory = RequestFactory()
+
+    @strawberry.type
+    class Query:
+        @strawberry.field
+        def abc(self, info) -> str:
+            print(info.context.request)
+            print(info.context)
```
jkimbo, line 344: Can you add some assertions on the mock print to check that it's behaving as expected?

jkimbo (changes requested):
Thanks @ddirkson . Minor comment around the test.

tests/django/test_views.py
ddirkson: Sure thing, I'll update it accordingly

ddirkson: Updated, thanks for the feedback!

PR #1169, opened 2021-08-24
Update opentelemetry dependency

joefreeman:
## Description

This updates the (optional) `opentelemetry` dependency from `0.17b0` to `1.4.1`.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,8 +29,8 @@ graphql-core = "~3.1.0"
 asgiref = {version = "^3.2", optional = true}
 flask = {version = "^1.1", optional = true}
 typing_extensions = "^3.7.4"
-opentelemetry-api = {version = "^0.17b0", optional = true}
-opentelemetry-sdk = {version = "^0.17b0", optional = true}
+opentelemetry-api = {version = "^1.4.1", optional = true}
+opentelemetry-sdk = {version = "^1.4.1", optional = true}
 python-dateutil = "^2.7.0"
 cached-property = "^1.5.2"
 pydantic = {version = "<2", optional = true}
@@ -59,8 +59,8 @@ requests = "^2.26.0"
 pre-commit = "^2.14.0"
 pytest-benchmark = "^3.4.1"
 freezegun = "^1.1.0"
-opentelemetry-api = "^0.17b0"
-opentelemetry-sdk = "^0.17b0"
+opentelemetry-api = "^1.4.1"
+opentelemetry-sdk = "^1.4.1"
 flake8-isort = "^4.0.0"
 flake8-black = "^0.2.1"
 django = {version = ">=2,<4", optional = false}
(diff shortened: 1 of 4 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Joe Freeman for the PR 👏

```

strawberry/extensions/tracing/opentelemetry.py
```diff
@@ -76,7 +76,7 @@ async def resolve(self, _next, root, info, *args, **kwargs):
 
             return result
 
-        with self._tracer.use_span(self._root_span):
+        with trace.use_span(self._root_span):
```
patrick91, line 79: why was this changed? 😊

joefreeman: Looks like it got moved in open-telemetry/opentelemetry-python#1668 :shrug:

patrick91: gotcha! thanks for explaing! 😊

pyproject.toml
```diff
@@ -59,8 +59,8 @@ requests = "^2.26.0"
 pre-commit = "^2.14.0"
 pytest-benchmark = "^3.4.1"
 freezegun = "^1.1.0"
-opentelemetry-api = "^0.17b0"
-opentelemetry-sdk = "^0.17b0"
+opentelemetry-api = "^1.4.1"
+opentelemetry-sdk = "^1.4.1"
```
joefreeman, line 63: I guess this might be an annoying upgrade for anyone who is using `opentelemetry==0.17b0`?

patrick91: it could be, maybe we can handle both versions? we only have two places where we had to change our code, so maybe we can do it

joefreeman:
> it could be, maybe we can handle both versions? we only have two places where we had to change our code, so maybe we can do it

Something like this?

I'm not sure how to specify the version to allow the `0.17b0` pre-release, but prevent from using pre-releases after `1.4.1` (i.e., `1.10a0`). Any ideas?

patrick91:
> I'm not sure how to specify the version to allow the `0.17b0` pre-release, but prevent from using pre-releases after `1.4.1` (i.e., `1.10a0`). Any ideas?

mhm, we could allow any version before 2? `<2`?

joefreeman:
> mhm, we could allow any version before 2? `<2`?

I think that still installs `1.10a0` though. Maybe not a big deal, but I think I've found what I was looking for: `"^0.17b0 || ~1.4"` - I think that requires either the specific version we want to keep for backwards compatibility, or something 1.4ish.

patrick91:
@joefreeman sounds good to me 😊 thanks!

patrick91 (approved):
💯

PR #1170, opened 2021-08-24
Fix opentelemetry version

patrick91:
Unfortunately PyPI doesn't support passing "^0.17b0 || ~1.4"

```
HTTP Error 400: Invalid value for requires_dist. Error: Invalid requirement: 'opentelemetry-api (>=0.17b0,<0.18 || >=1.4,<1.5); extra == "opentelemetry"'.
```


joefreeman:
Ah, that's why I couldn't find it in the PEP then :roll_eyes: Thanks for fixing!

patrick91:
@joefreeman no worries, I thought that would work too :D

Issue #1146, opened 2021-08-12
Add query string to Info


jkimbo:
In some circumstances it would be useful to be able to access the query inside the resolver. We should add it to the info object that we pass to resolvers.

Vichoko:
FYI I managed to achieve this access with the following lines:
```
raw_query = info.context["request"]._json["query"]
raw_variables = info.context["request"]._json["variables"]
```

Ckk3:
I will work on this 😉

PR #1176, opened 2021-08-26
Return query string from info

MAM-SYS:
**Hello Friends**
First of all Thanks for this amazing package
In our company we build APIs with strawberry, and we are all happy because of this handy package
I add some lines of code to add query string into the info object for issue #1146 
I also add an small test for this patch
Hope it can help
If anything needs to be fixed about my codes, just let me know
***With Respect***

I just doubt I have to updated the documentation for these changes or not

## Description
Technically, i just extract the sent query from request body from HttpRequest object and return it as info property

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

#1146 

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,13 @@
+Release type: patch
+
+This release add query string into the info object that is sent to the resolver
+
+```python
+import strawberry
+
+@strawberry.type
+class Query:
+    @strawberry.field
+    def query_string(self, info: Info) -> str:
+        return info.query
+```
--- a/strawberry/types/info.py
+++ b/strawberry/types/info.py
@@ -69,4 +69,8 @@ class Info(Generic[ContextType, RootValueType]):
     def path(self) -> Path:
         return self._raw_info.path
 
+    @property
+    def query(self) -> str:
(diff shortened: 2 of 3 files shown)


patrick91:
@MAM-SYS I think this might be useful for extensions, but I’m not sure about getting the string from the request. The core of Strawberry shouldn’t really know about the HTTP layer :)

I’d rather have a wrapper for GraphQLResolveInfo.operation -> https://graphql-core-3.readthedocs.io/en/latest/modules/type.html?highlight=resolveinfo#graphql.type.GraphQLResolveInfo.operation

So that we don’t need to parse queries when needed. Does this make sense? Sorry to make it more complicated!

/cc @jkimbo for sanity check, maybe I misunderstood the issue you wrote :D

MAM-SYS:
@patrick91 You are completely right. All of your reasons  make sense
Maybe i misunderstood the issue. If there is no reason for this patch right now, we can put it away
But this project is awesome and i try to contribute more

patrick91:
I'm going to close this for now, if someone wants this feature we can think of a better way to make it work for all HTTP integrations 😊

PR #1184, opened 2021-08-29
Only run tests when python files have changed

jkimbo:
## Description

This PR updates the GitHub tests workflow to only run if any Python files (or dependency changes) have changed.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,6 +5,11 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
+    paths:
+      - "strawberry/**"
+      - "tests/**"
+      - "pyproject.toml"
+      - "poetry.lock"
 
 jobs:
   unit-tests:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,6 +10,7 @@ on:
       - "tests/**"
       - "pyproject.toml"
       - "poetry.lock"
+      - ".github/workflows/test.yml"
 
 jobs:
   unit-tests:


patrick91 (approved):
Looks good ^^ thank you!

.github/workflows/test.yml
```diff
@@ -5,6 +5,11 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
+    paths:
+      - "strawberry/**"
+      - "tests/**"
+      - "pyproject.toml"
+      - "poetry.lock"
```
patrick91, line 12: oh, shall we put the action itself so if we change it we also test it in the same PR?

jkimbo: Oh yeah good idea

jkimbo:
Ah one problem with this @patrick91 is that we'll have to remove the branch protection rules around running tests before merging. Is that ok?

patrick91:
> Ah one problem with this @patrick91 is that we'll have to remove the branch protection rules around running tests before merging. Is that ok?

yes, maybe in future we can add a new check that fails if we tests did run and fail, but doesn't fail if no tests did run. For now we're good I'd say :)

jkimbo:
OK I've removed the protection rules.

BryceBeagle:
Do you think it would have been better to keep the branch protection rule in-place, and then just skipped the _job_ not the workflow for specific paths?

jkimbo:
@BryceBeagle We could do that but it would make the workflow more complicated. Since we require at least one person to review a PR before merging I don't think it's a problem.

PR #1187, opened 2021-08-30
Add custom dataclass init function

jkimbo:
## Description

This PR adds a custom dataclass init function that enforces that all args get passed as keyword arguments. This allows us to avoid the problem where a type cannot define a field with a default value before a field that doesn't have a default value.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist


strawberry/utils/dataclasses.py
```diff
@@ -0,0 +1,91 @@
...
+    We create a custom __init__ function for the dataclasses that back
+    Strawberry object types to only accept keyword arguments. This allows us to
+    avoid the problem where a type cannot define a field with a default value
+    before a field that doesn't have a default value.
+
+    An example of the problem:
+    https://stackoverflow.com/questions/51575931/class-inheritance-in-python-3-7-dataclasses
+
+    Code is adapted from:
+    https://github.com/python/cpython/blob/v3.9.6/Lib/dataclasses.py#L489-L536
```
jkimbo, line 27: Not sure how to include the license. @BryceBeagle any suggestions?

botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

patrick91 (commented):
This might be good for now, I wan't to see what @BryceBeagle thinks before merging though 😊

I think we should still vendorize in future :)

strawberry/utils/dataclasses.py
```diff
@@ -0,0 +1,91 @@
...
+        _init_params = ["*"] + _init_params
+
+    return _create_fn(
+        "__init__",
+        [self_name] + _init_params,
+        body_lines,
+        locals=locals,
+        globals=globals,
+        return_type=None,
+    )
```
BryceBeagle, line 70:
```suggestion
def dataclass_init_fn(fields, frozen, has_post_init, self_name, globals_):
    """
    We create a custom __init__ function for the dataclasses that back
    Strawberry object types to only accept keyword arguments. This allows us to
    avoid the problem where a type cannot define a field with a default value
    before a field that doesn't have a default value.

    An example of the problem:
    https://stackoverflow.com/questions/51575931/class-inheritance-in-python-3-7-dataclasses

    Code is adapted from:
    https://github.com/python/cpython/blob/v3.9.6/Lib/dataclasses.py#L489-L536

    Note: in Python 3.10 and above we use the `kw_only` argument to achieve the
    same result.
    """
    # fields contains both real fields and InitVar pseudo-fields.

    locals_ = {f"_type_{f.name}": f.type for f in fields}
    locals_.update(
        {
            "MISSING": MISSING,
            "_HAS_DEFAULT_FACTORY": _HAS_DEFAULT_FACTORY,
        }
    )

    body_lines = []
    for f in fields:
        line = _field_init(f, frozen, locals_, self_name)
        # line is None means that this field doesn't require
        # initialization (it's a pseudo-field).  Just skip it.
        if line:
            body_lines.append(line)

    # Does this class have a post-init function?
    if has_post_init:
        params_str = ",".join(f.name for f in fields if f._field_type is _FIELD_INITVAR)
        body_lines.append(f"{self_name}.{_POST_INIT_NAME}({params_str})")

    # If no body lines, use 'pass'.
    if not body_lines:
        body_lines = ["pass"]

    _init_params = [_init_param(f) for f in fields if f.init]
    if len(_init_params) > 0:
        _init_params = ["*"] + _init_params

    return _create_fn(
        "__init__",
        [self_name] + _init_params,
        body_lines,
        locals=locals_,
        globals=globals_,
        return_type=None,
    )
```

BryceBeagle, line 70:
```suggestion
def dataclass_init_fn(fields, frozen, has_post_init, self_name, globals_):
    """
    We create a custom __init__ function for the dataclasses that back
    Strawberry object types to only accept keyword arguments. This allows us to
    avoid the problem where a type cannot define a field with a default value
    before a field that doesn't have a default value.

    An example of the problem:
    https://stackoverflow.com/questions/51575931/class-inheritance-in-python-3-7-dataclasses

    Code is adapted from:
    https://github.com/python/cpython/blob/v3.9.6/Lib/dataclasses.py#L489-L536

    Note: in Python 3.10 and above we use the `kw_only` argument to achieve the
    same result.
    """
    # fields contains both real fields and InitVar pseudo-fields.

    locals_ = {f"_type_{f.name}": f.type for f in fields}
    locals_.update(
        {
            "MISSING": MISSING,
            "_HAS_DEFAULT_FACTORY": _HAS_DEFAULT_FACTORY,
        }
    )

    body_lines = []
    for f in fields:
        line = _field_init(f, frozen, locals_, self_name)
        # line is None means that this field doesn't require
        # initialization (it's a pseudo-field).  Just skip it.
        if line:
            body_lines.append(line)

    # Does this class have a post-init function?
    if has_post_init:
        params_str = ",".join(f.name for f in fields if f._field_type is _FIELD_INITVAR)
        body_lines.append(f"{self_name}.{_POST_INIT_NAME}({params_str})")

    # If no body lines, use 'pass'.
    if not body_lines:
        body_lines = ["pass"]

    _init_params = [_init_param(f) for f in fields if f.init]
    if len(_init_params) > 0:
        _init_params = ["*"] + _init_params

    return _create_fn(
        "__init__",
        [self_name] + _init_params,
        body_lines,
        locals=locals_,
        globals=globals_,
        return_type=None,
    )
```

BryceBeagle, line 70:
```suggestion
def dataclass_init_fn(fields, frozen, has_post_init, self_name, globals_):
    """
    We create a custom __init__ function for the dataclasses that back
    Strawberry object types to only accept keyword arguments. This allows us to
    avoid the problem where a type cannot define a field with a default value
    before a field that doesn't have a default value.

    An example of the problem:
    https://stackoverflow.com/questions/51575931/class-inheritance-in-python-3-7-dataclasses

    Code is adapted from:
    https://github.com/python/cpython/blob/v3.9.6/Lib/dataclasses.py#L489-L536

    Note: in Python 3.10 and above we use the `kw_only` argument to achieve the
    same result.
    """
    # fields contains both real fields and InitVar pseudo-fields.

    locals_ = {f"_type_{f.name}": f.type for f in fields}
    locals_.update(
        {
            "MISSING": MISSING,
            "_HAS_DEFAULT_FACTORY": _HAS_DEFAULT_FACTORY,
        }
    )

    body_lines = []
    for f in fields:
        line = _field_init(f, frozen, locals_, self_name)
        # line is None means that this field doesn't require
        # initialization (it's a pseudo-field).  Just skip it.
        if line:
            body_lines.append(line)

    # Does this class have a post-init function?
    if has_post_init:
        params_str = ",".join(f.name for f in fields if f._field_type is _FIELD_INITVAR)
        body_lines.append(f"{self_name}.{_POST_INIT_NAME}({params_str})")

    # If no body lines, use 'pass'.
    if not body_lines:
        body_lines = ["pass"]

    _init_params = [_init_param(f) for f in fields if f.init]
    if len(_init_params) > 0:
        _init_params = ["*"] + _init_params

    return _create_fn(
        "__init__",
        [self_name] + _init_params,
        body_lines,
        locals=locals_,
        globals=globals_,
        return_type=None,
    )
```

BryceBeagle, line 70: (Tried adding a diff to the full function but GitHub wouldn't process it. There are more things to change)

We should avoid shadowing the `locals` and `globals` builtins

```suggestion
    return _create_fn(
        "__init__",
        [self_name] + _init_params,
        body_lines,
        locals=locals_,
        globals=globals_,
        return_type=None,
    )
```

strawberry/utils/dataclasses.py
```diff
@@ -0,0 +1,91 @@
...
+        cls,
+        "__init__",
+        dataclass_init_fn(
+            fields=fields,
+            frozen=False,
+            has_post_init=hasattr(cls, _POST_INIT_NAME),
+            self_name="__dataclass_self__" if "self" in fields else "self",
+            globals=globals,
+        ),
+    )
```
BryceBeagle, line 91:
```suggestion
    globals_ = sys.modules[cls.__module__].__dict__

    _set_new_attribute(
        cls,
        "__init__",
        dataclass_init_fn(
            fields=fields,
            frozen=False,
            has_post_init=hasattr(cls, _POST_INIT_NAME),
            self_name="__dataclass_self__" if "self" in fields else "self",
            globals=globals_,
        ),
    )
```

BryceBeagle (changes requested):
I think this is a pretty good solution for now, just a few minor changes I see here

strawberry/utils/dataclasses.py
BryceBeagle: I think the best method would be to extract the stuff from cpython as best as we can to it's own directory+file and then place the cpython license in that directory. Then add a comment to the top of the file saying the code in the file abides by a different license. That way there's no confusion to which files the license applies to, and on the flip-side what license applies to that file.

jkimbo:
@patrick91 @BryceBeagle this is ready for a review again. Sorry for the delay!

RELEASE.md
```diff
@@ -0,0 +1,32 @@
+Release type: minor
+
+This release changes how dataclasses are created to make use of the new
+`kw_only` argument in Python 3.10 so that fields without a default value can now
+follow a field with a default value. This feature is also backported to all other
+supported Python versions.
```
patrick91, line 6: 💯

patrick91 (approved):
Looks good to me, I think we need to make this release more visible somehow, to make sure the breaking change is seen by everyone 😊

jkimbo:
@patrick91 yep I agree. Any suggestions on how to do that?

patrick91:
> @patrick91 yep I agree. Any suggestions on how to do that? 

Maybe we can send a newsletter for the end of the year and include it there! I can write something in the next few days ☺️

patrick91:
Reminder for this: https://github.com/strawberry-graphql/strawberry/pull/1503/files#diff-2531d8c5ff7eb7dec3a3cf591953651ebc6cca96089ed9c7215485db4de64ed6R539 😊

I can do it, but I'll try to write the email first :)

BryceBeagle (approved):

botberry:
# Pre-release
:wave:

Releasing commit [50d746383e4dfc1bbf66ca255dad75647c6ca27d] to PyPi as pre-release! :package:

botberry:
# Pre-release
:wave:

Pre-release **0.94.0-dev.1642180413** [50d746383e4dfc1bbf66ca255dad75647c6ca27d] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.94.0-dev.1642180413
```

jkimbo:
@patrick91 is this still relevant?

patrick91:
> @patrick91 is this still relevant?

yes, I need to merge it! sorry about that, will rebase it tonight and see what we need to do before merging it :)

patrick91 (approved):
Hey, so sorry it took me so long to merge this! Thanks for the nudge and all the work @jkimbo 😊

jceresini:
This introduced a breaking change for us when upgrading from 0.119.1 to 0.131.1. We're using federation, and defined a type that is handled by another subgraph:

```
@strawberry.federation.type(extend=True, keys=["id"])
class Resource:
    id: strawberry.ID = strawberry.federation.field(external=True)

# [...]

@strawberry.federation.type(keys=["id"], name="Event" )
class Event:
    id: strawberry.ID
    resource_id: strawberry.Private[str]

    @strawberry.field
    def resource(self) -> Resource:
        return Resource(self.resource_id)
```

After an update we started seeing this error in our logs

```
        return Resource(self.resource_id)
TypeError: __init__() takes 1 positional argument but 2 were given
```

It was a simple fix (passing the id as a kwarg) but tracking it down was a bit frustrating, and unexpected for a minor release. Sharing this in case others run into similar issues.

jceresini:
I tracked down this PR while addressing the issue in our code, and commented without reading the above. The PR clearly explains that its a breaking change. I'll pay more attention to release notes when approving changes that bump the strawberry version.

jkimbo:
@jceresini apologies for the inconvenience. We try hard to not make breaking changes but sometimes it's unavoidable. 

I think this is also an artifact of the quick release cycle that Strawberry has which often results in people upgrading between quite large version ranges and so it's hard to identify all the potential breaking changes. We might benefit from something on the website that allows you show the difference between versions and view all the changes at once. What do you think @patrick91 ?

patrick91:
> @jceresini apologies for the inconvenience. We try hard to not make breaking changes but sometimes it's unavoidable.
> 
> I think this is also an artifact of the quick release cycle that Strawberry has which often results in people upgrading between quite large version ranges and so it's hard to identify all the potential breaking changes. We might benefit from something on the website that allows you show the difference between versions and view all the changes at once. What do you think @patrick91 ?

yes, I'd love to have something like that! were you thinking of a page with a list of breaking changes and pending deprecations? or something else entirely?

@jceresini sorry about this! Ideally we should have done this in a major release, but unfortunately we haven't made the jump to v1 just yet

jkimbo:
> yes, I'd love to have something like that! were you thinking of a page with a list of breaking changes and pending deprecations? or something else entirely?

Highlighting breaking changes would be ideal but I'm not sure how we would extract that. As a first step it would be great to just list all the titles of the changes and you can open each one to find out more. More like traditional release notes. I'm imagining something like this: https://react-native-community.github.io/upgrade-helper/

PR #1212, opened 2021-09-05
Ensure scalar overrides are used everywhere

### LINKED ISSUE #1204: Regression using custom scalars with schema scoping.
v0.75 changed the scope of custom scalars from global to schema specific, but it's caused a subtle regression. Using the given example:
```python
    EpochDateTime = strawberry.scalar(
        datetime,
        serialize=lambda value: int(value.timestamp()),
        parse_value=lambda value: datetime.fromtimestamp(int(value), timezone.utc),
    )
```
the user could always `EpochDateTime` as an annotation, but the point of the example was to override `datetime` behavior - still using `datetime` as the annotation.

The problem now is using builtins _only_ works for overrides. I think it was implied that using an existing type should work, and would be more Pythonic as well. The following example fails in a surprising way:
```python
Long = strawberry.scalar(NewType("Long", int), description="64-bit int")
Binary = strawberry.scalar(
    bytes,
    name="Binary",
    serialize=lambda b: base64.b64encode(b).decode("utf8"),
    parse_value=base64.b64decode,
)
Duration = strawberry.scalar(
    timedelta,
    name="Duration",
    serialize=timedelta.total_seconds,
    parse_value=lambda s: timedelta(seconds=s),
)

@strawberry.type
class Query:
    @strawberry.field
    def long(self, value: Long) -> Long:
        return value

    @strawberry.field
    def binary(self, value: bytes) -> bytes:
        return value

    @strawberry.field
    def duration(self, value: timedelta) -> timedelta:
        return value

schema = strawberry.Schema(Query, scalar_overrides={bytes: Binary, timedelta: Duration})
```
* `Long` works - because it is the scalar
* `bytes` works accidentally - because it happens to be listed as a scalar in a global `SCALAR_TYPES` registry
* `timedelta` fails - because it's not an "override" (but neither is `bytes`)

Basically the schema converter's `scalar_registry` isn't being used as the source of truth on what is-a scalar.

jkimbo:
## Description

Fixes a bug where passing scalars in the `scalar_overrides` parameter wasn't being applied consistently. This PR makes sure that the schema specific `scalar_registry` is taken into account whenever a scalar is resolved.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* Fixes #1204

## Checklist

### DIFF
--- a/strawberry/annotation.py
+++ b/strawberry/annotation.py
@@ -22,7 +22,6 @@ except ImportError:  # pragma: no cover
 from strawberry.custom_scalar import ScalarDefinition
 from strawberry.enum import EnumDefinition
 from strawberry.lazy_type import LazyType
-from strawberry.scalars import SCALAR_TYPES
 from strawberry.type import (
     StrawberryList,
     StrawberryOptional,
@@ -82,8 +81,6 @@ class StrawberryAnnotation:
             return self.create_list(evaled_type)
         elif self._is_optional(evaled_type):
             return self.create_optional(evaled_type)
-        elif self._is_scalar(evaled_type):
-            return evaled_type
         elif self._is_union(evaled_type):
             return self.create_union(evaled_type)
         elif is_type_var(evaled_type):
@@ -198,15 +195,6 @@ class StrawberryAnnotation:
         annotation_origin = getattr(annotation, "__origin__", None)
 
         return annotation_origin == list
-
-    @classmethod
-    def _is_scalar(cls, annotation: Any) -> bool:
-        type_ = getattr(annotation, "__supertype__", annotation)
-
-        if type_ in SCALAR_TYPES:
-            return True
-
-        return hasattr(annotation, "_scalar_definition")
 
     @classmethod
     def _is_strawberry_type(cls, evaled_type: Any) -> bool:
--- a/strawberry/arguments.py
+++ b/strawberry/arguments.py
@@ -6,6 +6,7 @@ from typing import Any, Dict, Iterable, List, Mapping, Optional, Union, cast
 from typing_extensions import Annotated, get_args, get_origin
 
 from strawberry.annotation import StrawberryAnnotation
+from strawberry.custom_scalar import ScalarDefinition, ScalarWrapper
 from strawberry.enum import EnumDefinition
 from strawberry.type import StrawberryList, StrawberryOptional, StrawberryType
 from strawberry.utils.mixins import GraphQLNameMixin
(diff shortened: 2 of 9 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

jkimbo:
@strawberry-graphql/core can I get a review?

strawberry/annotation.py
```diff
@@ -82,8 +81,6 @@ def resolve(self) -> Union[StrawberryType, type]:
             return self.create_list(evaled_type)
         elif self._is_optional(evaled_type):
             return self.create_optional(evaled_type)
-        elif self._is_scalar(evaled_type):
-            return evaled_type
```
BryceBeagle, line 86: Can you explain why the annotation resolver no longer needs to check if an object is a scalar type? How will this work when `StrawberryScalar` is implemented?

jkimbo: This check wasn't really doing much anyway because all that happens now is that scalar types fall to the bottom of this function and just get returned. The `is_scalar` check now relies on the schema context to resolve correctly so I think any kind of conversion should be pushed to schema conversion stage IMO. 

More broadly I can see us relying more and more on schema context when converting the raw python annotations into Strawberry types (and then into GraphQL-core types) so I think conversion should happen as late as possible. I came across this problem when doing #1100 and I think it will also affect issues that we are seeing with deferred annotations.

strawberry/arguments.py
```diff
@@ -103,13 +107,16 @@ def convert_argument(
...
         value_list = cast(Iterable, value)
-        return [convert_argument(x, type_.of_type) for x in value_list]
+        return [
+            convert_argument(x, type_.of_type, scalar_registry, auto_camel_case)
+            for x in value_list
+        ]
 
-    if is_scalar(type_):
+    if is_scalar(type_, scalar_registry):
         return value
```
patrick91, line 120: do you think we could remove this check entirely? as we did above? so we don't need to pass the scalar registry?

strawberry/scalars.py
```diff
@@ -1,17 +1,16 @@
...
 
-def is_scalar(annotation: Any) -> bool:
-    type_ = getattr(annotation, "__supertype__", annotation)
 
-    if type_ in SCALAR_TYPES:
+def is_scalar(
+    annotation: Any,
+    scalar_registry: Dict[object, Union[ScalarWrapper, ScalarDefinition]],
+) -> bool:
+    if annotation in scalar_registry:
```
patrick91, line 13: we should probably move this to the schema converter, not sure it makes a lot of sense to keep it here now that we have to pass the scalar_registry

PR #1225, opened 2021-09-09
Add GraphQLTestClient class for asgi and Django

estyxx:
Add a Strawberry's `GraphQLTestClient` class ready to use in the test. 
Should provide one for integration 
- Django 
- Flask
- Asgi 

- Requests sync -> for e2e test is useful 
- aiohttp asyn 

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

*

## Checklist


botberry:



---

This release adds a `GraphQLTestClient` for Django. It hides the http request's details and asserts that there are no errors in the response (you can always disable this behavior by passing `asserts_errors=False`). This makes it easier to test queries and makes your tests cleaner.

If you are using `pytest` you can add a fixture in `conftest.py`

```python
import pytest

from django.test.client import Client

from strawberry.django.test import GraphQLTestClient


@pytest.fixture()
def graphql_client():
    yield GraphQLTestClient(Client())
```

And use it everywere in your test methods

```
def test_strawberry(graphql_client):
    query = """query Hi($name: String!) {
        hi(name: $name)
    }
    """

    result = graphql_client.query(query, variables={"name": "Marcotte"})

    assert result.data == {"hi": "Hi Marcotte!"}
```


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Etty for the PR 👏

```

RELEASE.md
```diff
@@ -0,0 +1,32 @@
...
+```
+
+And use it everywere in your test methods
+
+```
+def test_strawberry(graphql_client):
+    query = """query Hi($name: String!) {
+        hi(name: $name)
+    }
+    """
```
BryceBeagle, line 27:
```suggestion
def test_strawberry(graphql_client):
    query = """
        query Hi($name: String!) {
            hi(name: $name)
        }
    """
```

strawberry/django/test/client.py
```diff
@@ -0,0 +1,66 @@
+import json
+from typing import Any, Dict, Optional
+
+from typing_extensions import Literal
+
+from strawberry.test import BaseGraphQLTestClient, Response
+
+
+class GraphQLTestClient(BaseGraphQLTestClient):
```
BryceBeagle, line 9: Do we want to name this specific to Django? Or will each integration's client have the same name here?

strawberry/django/test/client.py
```diff
@@ -0,0 +1,66 @@
...
+
+            if len(kwargs) == 1:
+                file_map = json.dumps({ref_variable: [f"variables.{ref_variable}"]})
+            else:
+                file_map = json.dumps(
+                    {
+                        k: [f"variables.{ref_variable}.{index}"]
+                        for index, k in enumerate(kwargs)
+                    }
+                )
```
BryceBeagle, line 39: There's a bit much going on here without comments. Can you add a couple?

BryceBeagle (changes requested):

strawberry/django/test/client.py
estyxx: I didn't check how to do with the other integrations... If possible I would keep the same parameters/features...

strawberry/django/test/client.py
estyxx: Added some.
basically, these lines cover the test cases I found in the `test_upload.py` file. 
Thinking now, I have to test more complicated mappings, with folders and single files together...

strawberry/test/client.py
```diff
@@ -0,0 +1,34 @@
...
+    variables: Optional[Dict[str, Any]]
+
+
+class BaseGraphQLTestClient:
+    def __init__(self, client):
+        self._client = client
+
+    @abstractmethod
+    def query(
+        self,
```
BryceBeagle, line 26: `BaseGraphQLTestClient` needs to inherit from `ABC` to use `@abstractmethod` properly.

BryceBeagle (changes requested):

strawberry/django/test/client.py
```diff
@@ -0,0 +1,70 @@
...
+    ) -> Response:
+        body: Any = {"query": query}
+
+        if variables:
+            body["variables"] = variables
+
+        if format == "multipart":
+            assert variables is not None
+
+            # We have fo map the variables keywords with the files provide in `kwargs`
```
BryceBeagle, line 27:
```suggestion
            # We have to map the variables' keys to the files provided in `kwargs`
```

strawberry/django/test/client.py
```diff
@@ -0,0 +1,70 @@
...
+            # If the variable is an array of files we must number the keys
+            if len(kwargs) == 1:
+                file_map = json.dumps({ref_variable: [f"variables.{ref_variable}"]})
+            else:
+                file_map = json.dumps(
+                    {
+                        k: [f"variables.{ref_variable}.{index}"]
+                        for index, k in enumerate(kwargs)
+                    }
+                )
```
BryceBeagle, line 43: I think theres's a bit too much going on on some of these lines, even with the comments. Can you split things into multiple lines with some intermediary variables? e.g.

```python
if isinstance(variables.get(ref_variable), dict):
    var_with_descriptive_name = next(iter(variables[ref_variable]))
    ref_variable += f".{var_with_descriptive_name}"
```

BryceBeagle (changes requested):

strawberry/django/test/client.py
estyxx: I put this logic into a separate method and added comments... it's still a bit complicated but hope now is at least a bit more clear 😒

estyxx:
Hey @BryceBeagle I change a few things... (don't hate me 😅 )
Because speaking with @marcoacierno most of the logic in the previous Django's `GraphQLTestClient` can be reused, so I moved most of the logic in the `BaseGraphQLTestClient` and split into simple methods, so every client can override the part of logic that is slightly different:

For Example: in asgi, so Starlette's `TestClient` the `files` are not sent in the body, instead of for Django and Flask they are included... for Flask we will have to do the request with `content_type="multipart/form-data"` instead of `format="multipart"`... 
For asgi, as far I understand,  the response will always be in `json` format (so just `response.json()`), instead for the other clients if it's multipart we must decode `data = json.loads(response.data.decode())` instead if it's json we can use `response.json()`...

So tiny little differences... 
With this change, I hope that writing the other Client (in another PR :D ) will be more simpler...

BryceBeagle (approved):

patrick91 (approved):
Awesome stuff!

Issue #1192, opened 2021-08-30
Update Pydantic integration to match Django's integration


patrick91:
@la4de proposed a great way of integrating strawberry with Django:

```
@strawberry.django.type(models.Fruit)
class Fruit:
    id: auto
    name: auto
    color: 'Color'
```

This is better than what we have with Pydantic as you might get a better autocomplete, better type definition (if we add mypy support for it) and it makes it easier to override fields (even when you only want to override the name or description, as you can use `auto as the type`).

We should do the same thing[1] for Pydantic and promote it be a non-experimental feature.

Also worth mentioning that adding Pydantic support might mean that we also get support for https://github.com/tiangolo/sqlmodel which is probably going to be super helpful!

[1] We should keep the old way of passing fields as deprecated for now, and remove in version 1 or 2 :)

Matt343:
I'm wondering if it would be possible to provide a shortcut for types where we just want to copy all of the fields without modification. Right now using the existing method I can sort of do this by passing `list(MyPydanticType.schema()["properties"].keys())` to the fields argument in the decorator, and that's removed quite a bit of boilerplate from my project where I have lots of models I just want to expose without modification.

Perhaps we could use an optional argument to the decorator to tell it to copy the existing fields as is, that way the user could pick between the more explicit version with better autocomplete or the implicit version with less boilerplate?

patrick91:
> Perhaps we could use an optional argument to the decorator to tell it to copy the existing fields as is, that way the user could pick between the more explicit version with better autocomplete or the implicit version with less boilerplate?

yes, I think we can allow that 😊

```python
@strawberry.django.type(models.Fruit, all_fields=True)
class Fruit:
    ...
```

something like this, or maybe more explicit `expose_all_fields=True` :)

jkimbo:
Another option:

```python
@strawberry.django.type(models.Fruit, fields="__all__")
class Fruit:
    ...
```

patrick91:
would that allow this:

```python
@strawberry.django.type(models.Fruit, fields=["id", "type"])
class Fruit:
    ...
```

?

jkimbo:
Yeah, `fields` take either a list or a string.

patrick91:
> Yeah, `fields` take either a list or a string.

This is what wee do already though 😊 not sure I like the idea how having two ways of declaring fields (auto vs fields list)

jkimbo:
[...]

PR #1231, opened 2021-09-10
Fix Pydantic Unions

### LINKED ISSUE #1230: Pydantic conversion fails for Union types
If you include a field in a Pydantic object with a `Union[A, B]` type, where A and B are Pydantic objects, then conversion fails with `AttributeError: 'StrawberryUnion' object has no attribute '_type_definition'`. We need to handle the branches of the union separately to get the right behavior here.

```
class BranchA(pydantic.BaseModel):
    field_a: str

class BranchB(pydantic.BaseModel):
    field_b: int

    age: int
    union_field: Union[BranchA, BranchB]

@strawberry.experimental.pydantic.type(BranchA, fields=["field_a"])
class BranchAType:
    pass

@strawberry.experimental.pydantic.type(BranchB, fields=["field_b"])
class BranchBType:
    pass

@strawberry.experimental.pydantic.type(User, fields=["age", "union_field"])
class UserType:
    pass

origin_user = User(age=1, union_field=BranchA(field_a="abc"))
user = UserType.from_pydantic(origin_user)
# raises AttributeError, should return UserType(age=1, union_field=BranchAType(field_a="abc"))
```

Matt343:
## Description

In order to convert Union types from Pydantic models into Strawberry models, we need to check which branch of the Union we have and proceed with the conversion according to that type definition. I had to add a `_pydantic_type` field to our `StrawberryType` instances that are created through the converter so that we can get a reference to the original type of each branch for comparison. This seems a little hacky, but I couldn't think of a better way.

This fix only works for Unions of object types, but that should be fine as GraphQL doesn't support other types of unions.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* Closes #1230 

## Checklist

### DIFF
--- a/strawberry/experimental/pydantic/object_type.py
+++ b/strawberry/experimental/pydantic/object_type.py
@@ -144,6 +144,7 @@ def type(
         )
 
         model._strawberry_type = cls  # type: ignore
+        cls._pydantic_type = model  # type: ignore
 
         def from_pydantic(instance: Any, extra: Dict[str, Any] = None) -> Any:
             return convert_pydantic_model_to_strawberry_class(
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+Fixes a bug in the Pydantic conversion code around `Union` values.
--- a/strawberry/experimental/pydantic/conversion.py
+++ b/strawberry/experimental/pydantic/conversion.py
@@ -27,7 +27,6 @@ def _convert_from_pydantic_to_strawberry_type(
                 return _convert_from_pydantic_to_strawberry_type(
                     option_type, data_from_model=data, extra=extra
                 )
-        return None
     if isinstance(type_, StrawberryList):
         items = []
         for index, item in enumerate(data):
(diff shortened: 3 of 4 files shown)


botberry:



---

Fixes a bug in the Pydantic conversion code around `Union` values.



Here's the tweet text: 
```

```

patrick91:
@Matt343 thanks! I'll try to review this over the weekend ^^

strawberry/experimental/pydantic/object_type.py
```diff
@@ -144,6 +144,7 @@ def wrap(cls):
         )
 
         model._strawberry_type = cls  # type: ignore
+        cls._pydantic_type = model  # type: ignore
```
patrick91, line 147: this is a patter we want to move away from :)

I think we can keep this for now though, we still need to do some refactoring anyway ^^

patrick91 (approved):
Looks good! I've only added a note, but it is nothing to worry about for now, we'll deal with it when we finalise some refactoring and when we tackle this issue: https://github.com/strawberry-graphql/strawberry/issues/1192

PR #1241, opened 2021-09-13
Pydantic enums and interfaces

### LINKED ISSUE #1238: Enum values don't convert correctly from Pydantic models
Calling `from_pydantic` on a class with an enum field returns `AttributeError: 'EnumDefinition' object has no attribute '_type_definition'`

### LINKED ISSUE #1239: Interface values don't convert correctly from Pydantic models
When calling `from_pydantic` on a class with an interface field, the field value is always converted into an instance of the base class, no matter what its starting type is. The expected behavior should probably be to convert to the corresponding subtype class instead. See here for an example: https://gist.github.com/Matt343/fbce0cdffe1523bb22016bed6f65473f

Matt343:
## Description

Fix the Pydantic conversion method for Enum values, and add a mechanism to specify an interface type when converting from Pydantic. The Pydantic interface is really just a base dataclass for the subclasses to extend. When you do the conversion, you have to use `strawberry.experimental.pydantic.interface` to let us know that this type is an interface. You also have to use your converted interface type as the base class for the sub types as normal.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [x] Documentation

* Closes #1238 
* Closes #1239 

## Checklist

### DIFF
--- a/docs/integrations/pydantic.md
+++ b/docs/integrations/pydantic.md
@@ -38,7 +38,7 @@ from .models import User
     'name',
     'friends'
 ])
-class User:
+class UserType:
     pass
 ```
 
@@ -64,6 +64,48 @@ from .models import User
     'friends'
 ])
 class UserInput:
+    pass
+```
+
+## Interface types
+
+Interface types are similar to normal types; we can create one by using the
+`strawberry.experimental.pydantic.interface` decorator:
+
+```python
+import strawberry
+from pydantic import BaseModel
+from typing import List
+
+# pydantic types
+class User(BaseModel):
+    id: int
+    name: str
+
+class NormalUser(User):
+    friends: List[int] = []
+
+class AdminUser(User):
+    role: int
+
+# strawberry types
+@strawberry.experimental.pydantic.interface(model=User, fields=[
+    'id',
+    'name',
+])
+class UserType:
+    pass
+
+@strawberry.experimental.pydantic.type(model=NormalUser, fields=[
+    'friends',
+])
+class NormalUserType(UserType):  # note the base class
+    pass
+
+@strawberry.experimental.pydantic.type(model=AdminUser, fields=[
+    'role',
+])
+class AdminUserType(UserType):
     pass
 ```
 
--- a/strawberry/experimental/pydantic/conversion.py
+++ b/strawberry/experimental/pydantic/conversion.py
@@ -1,5 +1,6 @@
 from typing import Union, cast
 
+from strawberry.enum import EnumDefinition
 from strawberry.field import StrawberryField
 from strawberry.scalars import is_scalar
 from strawberry.type import StrawberryList, StrawberryOptional, StrawberryType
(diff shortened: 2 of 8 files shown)


Matt343:
Looks like one of the release file jobs timed out, I'm not sure how to restart it

botberry:



---

Fix the Pydantic conversion method for Enum values, and add a mechanism to specify an interface type when converting from Pydantic. The Pydantic interface is really a base dataclass for the subclasses to extend. When you do the conversion, you have to use `strawberry.experimental.pydantic.interface` to let us know that this type is an interface. You also have to use your converted interface type as the base class for the sub types as normal.


---




Here's the tweet text: 
```

```

docs/integrations/pydantic.md
```diff
@@ -67,6 +67,48 @@ class UserInput:
...
+    'id',
+    'name',
+])
+class UserType:
+    pass
+
+@strawberry.experimental.pydantic.type(model=NormalUser, fields=[
+    'friends',
+])
+class NormalUserType(UserType):  # note the base class
```
patrick91, line 102: do you think we should automatically detect interfaces?

strawberry/experimental/pydantic/conversion.py
```diff
@@ -42,6 +45,10 @@ def _convert_from_pydantic_to_strawberry_type(
     elif is_scalar(type_):
         return data
     else:
+        # in the case of an interface, the concrete type may be more specific
+        # than the type in the field definition
```
patrick91, line 49: uhm, not sure I understand the comment here 🤔

tests/experimental/pydantic/test_conversion.py
```diff
@@ -92,7 +93,7 @@ class WorkModel(pydantic.BaseModel):
         name: str
 
     @strawberry.experimental.pydantic.type(WorkModel, fields=["name"])
-    class Work(pydantic.BaseModel):
+    class Work:
```
patrick91, line 96: 👍

patrick91 (commented):
nice work! I've added a couple of comments, nothing crazy though 😊

what's your use case for defining interfaces with pydantic btw?

strawberry/experimental/pydantic/conversion.py
Matt343: So in the case of an interface, the field definition will have some type, which is the base class of the hierarchy. If we just use the type definition to figure out which conversion function to call, we will always convert the subclass to that base type, losing any fields that are only present on the subclass. That's how I found this issue, my objects were getting converted from `PydanticSpecificClass` to `StrawberryBaseClass` instead of `StrawberrySpecificClass`.

docs/integrations/pydantic.md
Matt343: I'm not sure how you would do that, as the base class in pydantic isn't really distinguishable from a normal pydantic class. It just has subclasses that are relevant, but we don't know that when we're inspecting the base class.

Matt343:
> what's your use case for defining interfaces with pydantic btw?

I'm trying to define a pretty complicated conceptual model from an existing API. It has some cases where we could use a union, but there are common fields so it would be nice to be able to define an interface over them. Specifically we're talking about "event" data, where every event has things like timestamps, user ids, and type info, but then they also have type specific fields.

docs/integrations/pydantic.md
patrick91: gotcha :)

patrick91 (approved):

Issue #374, opened 2020-07-10
Move context out from info and expose it as a parameter


patrick91:
Being able to request info (and root) only when needed is a nice feature, but considering we haven't exposed a type for info yet, it might be nice to all to request for context as well

So, from this:

```python
@strawberry.federation.type(extend=True)
class Query:
    @strawberry.field
    def strawberry(self, info) -> str:
        return info.context.strawberry_repo.get_one()
```

to this:

```python
@strawberry.federation.type(extend=True)
class Query:
    @strawberry.field
    def strawberry(self, context: MyContextType) -> str:
        return context.strawberry_repo.get_one()
```

Speedy1991:
Ah this would be awesome!
It would solve the problem with PyCharm not autocompleting custom contextes without the mypy plugin also!

patrick91:
I was thinking about this today, and I think it could be interesting to create a `StrawberryContext` type that we can use to understand which argument is the context

patrick91:
maybe we do:

```python
@strawberry.federation.type(extend=True)
class Query:
    @strawberry.field
    def strawberry(self, context: strawberry.Context[MyContextType]) -> str:
        return context.strawberry_repo.get_one()
```

PR #1243, opened 2021-09-13
Add strawberry.context

jkimbo:
## Description

This PR introduces `strawberry.context` which gives you access to the current context through the magic of [Context Variables](https://docs.python.org/3/library/contextvars.html). I propose that becomes the recommend way to access the context rather than through `info`. What do you think @strawberry-graphql/core ?

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/docs/guides/dataloaders.md
+++ b/docs/guides/dataloaders.md
@@ -171,6 +171,6 @@ class MyGraphQL(GraphQL):
 @strawberry.type
 class Query:
     @strawberry.field
-    async def get_user(self, info: Info, id: strawberry.ID) -> User:
+    async def get_user(self, id: strawberry.ID) -> User:
         return await strawberry.context["user_loader"].load(id)
 ```
--- a/docs/guides/dataloaders.md
+++ b/docs/guides/dataloaders.md
@@ -172,5 +172,5 @@ class MyGraphQL(GraphQL):
 class Query:
     @strawberry.field
     async def get_user(self, info: Info, id: strawberry.ID) -> User:
-        return await info.context["user_loader"].load(id)
+        return await strawberry.context["user_loader"].load(id)
 ```
--- a/docs/guides/permissions.md
+++ b/docs/guides/permissions.md
@@ -18,7 +18,7 @@ class IsAuthenticated(BasePermission):
     message = "User is not authenticated"
 
     # This method can also be async!
-    def has_permission(self, source: typing.Any, **kwargs) -> bool:
+    def has_permission(self, source: typing.Any, info: Info, **kwargs) -> bool:
         return False
 
 @strawberry.type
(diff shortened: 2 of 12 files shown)


docs/guides/dataloaders.md
```diff
@@ -171,6 +171,6 @@ class MyGraphQL(GraphQL):
 @strawberry.type
 class Query:
     @strawberry.field
-    async def get_user(self, info: Info, id: strawberry.ID) -> User:
-        return await info.context["user_loader"].load(id)
+    async def get_user(self, id: strawberry.ID) -> User:
+        return await strawberry.context["user_loader"].load(id)
```
patrick91, line 175: do you think we can:

1. type it somehow?
2. use context.user_loader instead of context["user_loader"] 

?

jkimbo: > 1. type it somehow?

I'm not sure how we could type it unfortunately. You can always cast it when you import it but that's not great. I think that's one of the downsides of this proposal. 

> 2. use context.user_loader instead of context["user_loader"]

That should already work. See tests.

AndrewIngram:
Pros:
1. Most common reason for needing the info arg is gone, so more more consistency in resolvers in general
2. Doesn't feel like a mismatch given context is typically fairly static per query (mutations might update it, but regular queries shouldn't), whilst the rest of info differs per query+resolver.

Cons:
1. Less explicit 
2. There's nothing stopping code further down the call stack from using `strawberry.context` either, which could result in some nasty coupling.

The second con is the one that most concerns me. I've seen Flask codebases where its magic `request` object is accessed at every conceivable point in the call stack and it's an absolute nightmare of tangled layers. As such, we'd want to discourage people from accessing `strawberry.context` outside of the "GraphQL" concern. Arguably the deepest it should go is Dataloaders (if that). There'd definitely need to be some "here be dragons" warnings needed in the docs.

But the fact that this guidance would make its utility mostly about simpler resolver signatures, it makes me wonder if there's enough of an upside to this? Perhaps there is though.

jkimbo:
@AndrewIngram I'm beginning to wonder if there is much benefit to this as well. Especially because you can't type it effectively and the first Pro you list could be solved by #374

docs/guides/dataloaders.md
patrick91: yeah, I wonder if doing:

```
ctx: Context[Something] = strawberry.context
```

is annoying or not.

Do we get benefits by using ContextVars and still using parameters (and dependency injection)?

jkimbo: > Do we get benefits by using ContextVars and still using parameters (and dependency injection)?

You start defining (and exporting from Strawberry) stand alone functions that get useful things from the context. E.g. `current_user()` The only issue with that is there wouldn't be a way for those functions to be known about early in the execution to register things into the context before resolution happens. E.g. you couldn't do `get_dataloader(UserDataloader)` inside a resolver without also registering an extension on the schema that would inject the dataloader into the context. Dependency injection would allow you to that.

jkimbo: Oh it would make the data validation proposal (https://github.com/strawberry-graphql/strawberry/discussions/820#discussioncomment-1081049) a bit cleaner. Instead of having to pass info to the `.validate()` function you could grab the context from the context variable.

jkimbo:
As discussed in our recent community meeting, I'm going to close this PR since we don't think it provides enough value to be worth the potential confusion.

PR #1237, opened 2021-09-12
refactor: move and merge resolver types to types.resolver

scratchmex:
I was struggling to understand where are the types for the resolvers and end up seeing a `type_resolvers.py` and `fields/resolver.py`. I tried to put the things in the most intuitively place.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Enhancement/optimization
- [ ] Documentation

## Issues Fixed or Closed by This PR

*

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the CONTRIBUTING document.
- [x] I have added tests to cover my changes.
- [x] I have tested the changes and verified that they work and don't break anything (as well as I can manage).

PR #1246, opened 2021-09-14
[WIP] refactor: move arguments types to types.arguments 

scratchmex:
Trying to follow up this proposal: https://github.com/strawberry-graphql/strawberry/pull/489 at the point "Convert arguments to StrawberryArgument"

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/tests/schema/test_mutation.py
+++ b/tests/schema/test_mutation.py
@@ -1,4 +1,3 @@
-import dataclasses
 import typing
 
 import strawberry
@@ -214,12 +213,10 @@ def test_converting_to_dict_with_unset():
     class Mutation:
         @strawberry.mutation
         def say(self, input: Input) -> str:
-            data = dataclasses.asdict(input)
-
-            if is_unset(data["name"]):
+            if is_unset(input.name):
                 return "Hello 🤨"
 
-            return f"Hello {data['name']}!"
+            return f"Hello {input.name}!"
 
     schema = strawberry.Schema(query=Query, mutation=Mutation)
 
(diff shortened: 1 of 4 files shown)


tests/schema/test_mutation.py
```diff
@@ -214,12 +213,10 @@ class Input:
     class Mutation:
         @strawberry.mutation
         def say(self, input: Input) -> str:
-            data = dataclasses.asdict(input)
-
-            if is_unset(data["name"]):
+            if is_unset(input.name):
```
BryceBeagle, line 216: Why this change?

BryceBeagle (changes requested):

tests/schema/test_mutation.py
scratchmex: when you run `dataclass.asdict(input)`, for some strange reason (idk if this is intented) it initializes a new class for each type, that is, as `name` field is of the type `UNSET`, it initializes a new instance of the class `_Unset` and therefore the function `is_unset` is false. Basically, is the for the test case to pass as I did this change:
https://github.com/strawberry-graphql/strawberry/blob/429faeec5db2ce061fba208c150e36314c58294c/strawberry/types/arguments.py#L31-L33

and previously it was https://github.com/strawberry-graphql/strawberry/blob/ba6f79a6cb17efb4918551225bd6c39a53e117ef/strawberry/arguments.py#L29-L30

I did that change to avoid multiple instances of the constant `UNSET` which is actually an instance of the class `_Unset`

patrick91:
same as #1237 I think we should revisit both PRs in future

PR #1254, opened 2021-09-15
feat(interface): input type cannot inherit an interface type

### LINKED ISSUE #1236: Interfaces should not work with input types
based on this [stackoverflow question](https://stackoverflow.com/questions/41921137/can-a-graphql-input-type-inherit-from-another-type-or-interface), graphql doesn't support interfaces with input types, and yet strawberry allows this:

```py
import strawberry

@strawberry.interface
class SomeInterface:
    some_arg: str

@strawberry.input
class SomeInput(SomeInterface): # this should throw an error
    another_arg: str
```

scratchmex:
## Description
```python
@strawberry.interface
    class SomeInterface:
        some_arg: str

@strawberry.input
class SomeInput(SomeInterface):  # this should throw an error
    another_arg: str
```
raises the exception `InvalidSuperclassInterface`:
```python
InvalidSuperclassInterface("An Input class 'SomeInput' cannot inherit from an Interface(s) 'SomeInterface'")
```

also works when inheriting from multiple interfaces:
```python
InvalidSuperclassInterface("An Input class 'SomeInput' cannot inherit from an Interface(s) 'SomeInterface, SomeOtherInterface'")
```

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* fix #1236

## Checklist


botberry:



---

Starting with this release, Strawberry will throw an error if one of your input
types tries to inherit from one or more interfaces. This new error enforces the
GraphQL specification that input types cannot implement interfaces.

The following code, for example, will now throw an error:

```python
import strawberry


@strawberry.interface
class SomeInterface:
    some_field: str


@strawberry.input
class SomeInput(SomeInterface):
    another_field: int
```



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Ivan Gonzalez for the PR 👏

```

tests/schema/test_interface.py
```diff
@@ -161,3 +162,27 @@ def anime(self) -> Anime:
 
     assert not result.errors
     assert result.data == {"anime": {"name": "One Piece"}}
+
+
+def test_input_cannot_inherit_from_interface():
```
BryceBeagle, line 167: Can these tests cover the actual messages of the errors as well?

strawberry/exceptions.py
```diff
@@ -176,3 +176,13 @@ def __init__(self, field_name: str, argument_name: str, argument_type: str):
...
         super().__init__(message)
+
+
+class InvalidSuperclassInterface(Exception):
+    def __init__(self, input_name: str, interface_names=List[str]):
+        interface_names = ", ".join(interface_names)
+        message = (
+            f"An Input class {input_name!r} cannot inherit "
+            f"from an Interface(s) {interface_names!r}"
+        )
```
BryceBeagle, line 187: Why do these `fstring` expressions use `!r`? Neither a string or a list of strings should have a difference between `str()` and `repr()`.

BryceBeagle (changes requested):

strawberry/exceptions.py
scratchmex: I put them so they are printed with ticks: `"InterfaceName"` `"['ClassOne', 'ClassTwo']"`

BryceBeagle: Oh I guess there is a difference.

tests/schema/test_interface.py
scratchmex: Do yo know how to catch the exception message with pytest? Or I need to try, catch, save, raise, assert?

BryceBeagle: You can use 
```python
with pytest.raises(ExceptionType) as exc:
    ...
assert exc.value == "something"
```

BryceBeagle: Also, note that `pytest.raises` takes a `match` argument that may be enough here

scratchmex: Let me have a look, thanks

strawberry/exceptions.py
```diff
@@ -176,3 +176,13 @@ def __init__(self, field_name: str, argument_name: str, argument_type: str):
         message = f'Argument "{argument_name}" on field "{field_name}" cannot be of type\
             "{argument_type}"'
         super().__init__(message)
+
+
+class InvalidSuperclassInterface(Exception):
+    def __init__(self, input_name: str, interface_names=List[str]):
```
BryceBeagle, line 182: Actually, now that I think about it, wouldn't it be better to take a `List[type]` instead? Let the Exception handle everything related to output?

BryceBeagle (dismissed):

strawberry/exceptions.py
scratchmex: You are right

patrick91:
Hi @scratchmex would you be able to rebase this PR and address the comments made by Bryce? 😊 no worries if not I can't take a look at some point in future :)

strawberry/exceptions.py
DoctorJohn: Addressed in 4654c7cd8acf92ef3fc4acdc7bc4e103c4db4df8

tests/schema/test_interface.py
DoctorJohn: Addressed in f24892d46e89c676ce69b935ca4a7c3ef33475e2

DoctorJohn:
@patrick91 I rebased this PR, addressed all feedback, and turned the exception into a `StrawberryException`.

PR #1266, opened 2021-09-21
If field is a simple field return a simple resolver

jkimbo:
## Description

This PR improves Strawberry performance by returning a simple resolver for fields without a custom resolver or any permission classes.

Benchmarks:

strawberry-simple-resolvers

Note: "simple" fields won't call the `get_result` function on the field. I've implemented the `is_simple_field` check as a property on the field so that custom fields can override it if they always need to post process results from the resolver.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -35,7 +35,7 @@ from .types.types import FederationFieldParams, TypeDefinition
 _RESOLVER_TYPE = Union[StrawberryResolver, Callable]
 
 
-def default_resolver(name: str, source: Any):
+def default_resolver(name: str, source: Any, *args, **kwargs):
     return getattr(source, name)
 
 
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -299,6 +299,18 @@ class StrawberryField(dataclasses.Field, GraphQLNameMixin):
     def is_async(self) -> bool:
         return self._has_async_permission_classes or self._has_async_base_resolver
 
+    @property
+    def is_simple_field(self):
+        """
+        Flag indicating if this is a "simple" field that just returns the
+        relevant attribute from the source object. If it is a simple field we
+        can avoid constructing an `Info` object and running any permission
+        checks in the resolver which improves performance.
+
+        Note: if a field is "simple" then `get_result` is not called.
+        """
+        return not self.base_resolver and not self.permission_classes
+
 
 T = TypeVar("T")
 
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -34,9 +34,7 @@ from .types.types import FederationFieldParams, TypeDefinition
 
 _RESOLVER_TYPE = Union[StrawberryResolver, Callable]
 
-
-def default_resolver(name: str, source: Any, *args, **kwargs):
-    return getattr(source, name)
+default_resolver = StrawberryResolver.default_resolver()
 
 
 class StrawberryField(dataclasses.Field, GraphQLNameMixin):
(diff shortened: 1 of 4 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

strawberry/field.py
```diff
@@ -287,6 +291,18 @@ def _has_async_base_resolver(self) -> bool:
...
 
+    @property
+    def is_simple_field(self):
+        """
+        Flag indicating if this is a "simple" field that just returns the
+        relevant attribute from the source object. If it is a simple field we
+        can avoid constructing an `Info` object and running any permission
+        checks in the resolver which improves performance.
+
+        Note: if a field is "simple" then `get_result` is not called.
```
patrick91, line 302: does this mean I can't do:

class ACustomField(...):
   def get_result(...):
     return '123'

```
@strawberry.type
class User:
    something: str = ACustomField()
```

jkimbo: You would need to override the property. Like this:

```python
class ACustomField(...):
	@property
	def is_simple_field(self):
		return False

	def get_result(...):
    	return '123'

@strawberry.type
class User:
    something: str = ACustomField()
```

strawberry/field.py
```diff
@@ -34,6 +34,10 @@
 _RESOLVER_TYPE = Union[StrawberryResolver, Callable]
 
 
+def default_resolver(name: str, source: Any, *args, **kwargs):
+    return getattr(source, name)
```
BryceBeagle, line 38: Can this be `StrawberryResolver`? Potentially constructed using a `classmethod`?

BryceBeagle (changes requested):

strawberry/field.py
jkimbo: I guess it can be though what would the benefit be?

BryceBeagle: So that `Field.get_result` always deals with the same datatype

jkimbo: Ok I see. Will change.

strawberry/schema/schema_converter.py
```diff
@@ -53,6 +55,9 @@
 from .types.concrete_type import ConcreteType
 
 
+default_resolver = StrawberryResolver.default_resolver()
```
BryceBeagle, line 58: This is defined twice, here and in `field.py`.

strawberry/types/fields/resolver.py
```diff
@@ -18,6 +18,10 @@
 T = TypeVar("T")
 
 
+def default_resolver(name: str, source: Any, *args, **kwargs):
```
BryceBeagle, line 21: I think this should live as an attribute on `StrawberryResolver`, probably private

strawberry/types/fields/resolver.py
```diff
@@ -38,6 +42,10 @@ def __init__(
     def __call__(self, *args, **kwargs) -> T:
         return self.wrapped_func(*args, **kwargs)
 
+    @classmethod
+    def default_resolver(cls):
```
BryceBeagle, line 46:
```suggestion
    def default_resolver(cls) -> StrawberryResolver:
```

BryceBeagle (changes requested):

strawberry/types/fields/resolver.py
jkimbo: Is mypy not clever enough to figure that out since I’m returning `cls`?

BryceBeagle: Maybe it is, but I tend towards being explicit

jkimbo:
Because this PR is now quite out of date I've recreated it here: #2194

Issue #566, opened 2020-11-11
Merging schemas, schema stitching


kristofgilicze:
I have seen the issue https://github.com/strawberry-graphql/strawberry/issues/38 which addresses remote schemas, i believe this is slightly different.

Most Django projects written in the pattern where the project is split into several smaller 'apps'..

**What I do right now**
```
- models.py
- admin.py
- typing.py
- mutations.py
- resolvers.py
- etc...
```
create these for every app, then assemble my schema at the project level.

**Proposal**
Ideally I would like to have one schema defined per app, and merge them into a main schema.
**Is there a way to merge schemas, that I overlooked ?**
( ? Probably schema stitching is the correct terminology, not sure ? )

This is something that is implemented by Graphene, but in every other way I prefer Strawberry.
It feels very clean and Pythonic indeed, thanks ! :)

patrick91:
> I have seen the issue #38 which addresses remote schemas, i believe this is slightly different.
> 
> Most Django projects written in the pattern where the project is split into several smaller 'apps'..

Oh, interesting, at work we are doing one big app that contains all the API code 😊
but that's just another way of doing it :)

> **Proposal**
> Ideally I would like to have one schema defined per app, and merge them into a main schema.
> **Is there a way to merge schemas, that I overlooked ?**

What I do is this:

```python
import strawberry

from app_a.query import Query as AppAQuery
from app_b.query import Query as AppBQuery

@strawberry.type
class Query(AppAQuery, AppBQuery):
    ...

schema = strawberry.Schema(Query)
```

not great, but it works. But I think we could provide a way for merging/stitching schemas, shouldn't be too complicated[1]

we could provide an API like:

```
schema = strawberry.merge_schemas(a, b, c)
```

or something better, @jkimbo do you have any suggestion? :)

> This is something that is implemented by Graphene
How do you do that in Graphene? :)

> but in every other way I prefer Strawberry.
> It feels very clean and Pythonic indeed, thanks ! :)

Thank you for the feedback! 😊

[1] though we might need to figure out how to solve name clashes and similar potential issues

jkimbo:
> Oh, interesting, at work we are doing one big app that contains all the API code 😊
but that's just another way of doing it :)

This is the way I do it as well but it's interesting to see other ways of structuring projects.

> schema = strawberry.merge_schemas(a, b, c)

This would essentially be the same as `class Query(AppAQuery, AppBQuery):` but with extra checking for collisions @patrick91 ?

[...]

PR #1273, opened 2021-09-23
feat(tools): implement a helper function to merge multiple types e.g. queries, mutations

alexei:
## Description

Schemas are initialised with a query and possibly a mutation. A larger project can't afford to put everything inside the same query/mutation, thus there is a need to be able to compose queries/mutations. Assuming two queries:

```python
@strawberry.type
class QueryA:
    ...

@strawberry.type
class QueryB:
    ...
```

Currently the way to combine them is via inheritance:

```python
@strawberry.type
class ComboQuery(QueryB, QueryA):
    ...

schema = strawberry.Schema(query=ComboQuery)
```

This PR implements a shortcut for the above:

```python
ComboQuery = merge_types((QueryB, QueryA))
schema = strawberry.Schema(query=ComboQuery)
```

Note I didn't update the docs as I wasn't sure whether this should be in General / Queries or General / Schema Basics (or some other place).

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* https://github.com/strawberry-graphql/strawberry/issues/566

## Checklist

### DIFF
--- a/strawberry/tools/merge_types.py
+++ b/strawberry/tools/merge_types.py
@@ -1,0 +1,22 @@
+from typing import Tuple, Type
+
+import strawberry
+
+
+def merge_types(types: Tuple[Type]) -> Type:
+    """Merge multiple Strawberry types into one
+
+    For example, given two queries `A` and `B`, one can merge them as follows:
+
+        merge_types((B, A))
+
+    This is essentially the same as:
+
+        class Query(B, A):
+            ...
+    """
+
+    if not types:
+        raise ValueError("Can't merge types if none are supplied")
+
+    return strawberry.type(type("MegaType", types, {}))
--- a/tests/tools/test_merge_types.py
+++ b/tests/tools/test_merge_types.py
@@ -1,0 +1,78 @@
+from operator import attrgetter
+from textwrap import dedent
+
+import pytest
+
+import strawberry
+from strawberry.tools import merge_types
+
+
+@strawberry.type
+class Person:
+    @strawberry.field
+    def name(self) -> str:
+        return "Eve"
+
+    @strawberry.field
+    def age(self) -> int:
+        return 42
+
+    @strawberry.field
+    def hi(self) -> str:
+        return "Hi"
+
+
+@strawberry.type
+class Greeter:
+    @strawberry.field
+    def hi(self, name: str = "world") -> str:
+        return f"Hello, {name}!"
+
+    @strawberry.field
+    def bye(self, name: str = "world") -> str:
+        return f"Bye, {name}!"
+
+
+def test_inheritance():
+    """It should merge multiple types following the regular inheritance rules"""
+
+    ComboQuery = merge_types((Greeter, Person))
+
+    definition = ComboQuery._type_definition
+    assert len(definition.fields) == 4
+
+    actuals = [attrgetter("python_name", "type")(f) for f in definition.fields]
+    expected = [("hi", str), ("bye", str), ("name", str), ("age", int)]
+    assert actuals == expected
+
+
+def test_empty_list():
+    """It should raise when the `types` argument is empty"""
+
+    with pytest.raises(ValueError):
+        merge_types(())
+
+
+def test_schema():
+    """It should create a valid, usable schema based on a merged query"""
+
+    ComboQuery = merge_types((Greeter, Person))
+    schema = strawberry.Schema(query=ComboQuery)
+
+    sdl = """
+    schema {
+      query: MegaType
+    }
+
+    type MegaType {
+      hi(name: String! = "world"): String!
+      bye(name: String! = "world"): String!
+      name: String!
+      age: Int!
+    }
+    """
+    assert dedent(sdl).strip() == str(schema)
+
+    result = schema.execute_sync("query { hi }")
+    assert not result.errors
+    assert result.data == {"hi": "Hello, world!"}
--- a/strawberry/tools/__init__.py
+++ b/strawberry/tools/__init__.py
@@ -1,6 +1,8 @@
 from .create_type import create_type
+from .merge_types import merge_types
 
 
 __all__ = [
     "create_type",
+    "merge_types",
 ]


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Alexandru Mărășteanu for the PR 👏

```

strawberry/tools/merge_types.py
```diff
@@ -0,0 +1,22 @@
...
+    This is essentially the same as:
+
+        class Query(B, A):
+            ...
+    """
+
+    if not types:
+        raise ValueError("Can't merge types if none are supplied")
+
+    return strawberry.type(type("MegaType", types, {}))
```
jkimbo, line 22: `merge_types` should probably accept a `name` argument so that you can customise the type name. I do like "MegaType" though so maybe that could be the default?

tests/tools/test_merge_types.py
```diff
@@ -0,0 +1,77 @@
...
+      bye(name: String! = "world"): String!
+      name: String!
+      age: Int!
+    }
+    """
+    assert dedent(sdl).strip() == str(schema)
+
+    result = schema.execute_sync("query { hi }")
+    assert not result.errors
+    assert result.data == {"hi": "Hello, world!"}
```
jkimbo, line 77: Nice tests 👍

jkimbo (changes requested):
This looks great, thanks @alexei ! 

What do you think about adding a check that the types don't contain any duplicate fields? That way you won't accidentally override fields defined in another query type. It's the main benefit of using this api rather than inheritance IMO. 

Also could you add some docs to the tools doc?

strawberry/tools/merge_types.py
alexei: Added a name arg, see https://github.com/strawberry-graphql/strawberry/pull/1273/commits/765c92313411c81cb03b056cb11af127424bac65

alexei:
Thanks for the review!

> What do you think about adding a check that the types don't contain any duplicate fields? That way you won't accidentally override fields defined in another query type. It's the main benefit of using this api rather than inheritance IMO.

Fair. What do you think about putting it behind a flag? When it's on (default), Strawberry raises a <ins>warning</ins> about the duplicates; when it's off - well, it doesn't.

> Also could you add some docs to the tools doc?

Done, see https://github.com/strawberry-graphql/strawberry/pull/1273/commits/c00deafb27477d45749ae7b7435a8bf202317960

jkimbo:
> Fair. What do you think about putting it behind a flag? When it's on (default), Strawberry raises a warning about the duplicates; when it's off - well, it doesn't.

Sure it could be behind a flag but when would it make sense to not warn about duplicates? I can't think of a use case for that.

alexei:
> > Fair. What do you think about putting it behind a flag? When it's on (default), Strawberry raises a warning about the duplicates; when it's off - well, it doesn't.
> 
> Sure it could be behind a flag but when would it make sense to not warn about duplicates? I can't think of a use case for that.

Sorry, i was just about to comment that it doesn't make sense to put it behind a flag if it's just a warning i.e. as opposed to an error.

I'll update the function to raise a warning if fields are overriden.

strawberry/tools/merge_types.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+
+    if not types:
+        raise ValueError("Can't merge types if none are supplied")
+
+    fields = chain(*(t._type_definition.fields for t in types))
+    counter = Counter(f.name for f in fields)
+    dupes = [f for f, c in counter.most_common() if c > 1]
+    if dupes:
+        raise Warning("{} has overridden fields: {}".format(name, ", ".join(dupes)))
```
jkimbo, line 33: I think we should raise an exception here instead of a "Warning" since we're stopping execution  (maybe a TypeError 🤷 ). In my mind warnings are informational and shouldn't stop code execution but I think this warrants being an actual error.

alexei: Dynamic languages usually allow one to redeclare variables, override properties, methods etc. Programs still run, and do so with no side effects. Raising an error would stop program execution, which is both unwarranted as the program would still run just fine, and surprising as it goes against the rules of the language. A warning, on the other hand, makes no assumptions about user intent, but still does its duty.

jkimbo: Ok sure that's fine but you are raising an error here which will stop program execution. Did you mean to log the warning instead?

alexei: > Did you mean to log the warning instead?

😬 That's correct - see https://github.com/strawberry-graphql/strawberry/pull/1273/commits/7261ddbf7d6c0fabc4f1be5cb7d250a716b3476c

jkimbo: Perfect 😃

alexei:
@jkimbo 

I think I found a bug: if both query and mutation are compund types, when executing any mutation you'll get an error saying `"Cannot query field _mutation_name_ on type 'MegaType'. Did you mean _query_names_?"` I think the name of the new type needs to be unique - since you are far more familiar with the codebase, could you confirm that? One way to fix this is to default to `'MegaType{"".join(t.__name__ for t in types)}'` or something like that. What do you think?

On another note - do you have any idea why the tests got cancelled the other day? I think I read somewhere GH had some problems yesterday, but I can;t re-run them either.

jkimbo:
@alexei 

> I think the name of the new type needs to be unique - since you are far more familiar with the codebase, could you confirm that?

Ah yes they do need to be unique. Making the name argument required is the simplest solution so let's do that. 

Also no idea why the tests got cancelled. If you make the change above we can see if they run again.

alexei:
@jkimbo it seems the Python 3.7 + Django 2.2.20 test got cancelled because of a `SIGTERM`. I'm curious does it happen again when you re-run the test?

alexei:
> I've added two minor comments, otherwise looks good to me!

Good find, thanks!

> FYI we might change the internal implementation of merge_types in future, when we finalise the internal refactoring 😊

LMK how I can help.

PR #1280, opened 2021-09-28
Add all_fields and auto support in the pydantic integration

### LINKED ISSUE #1192: Update Pydantic integration to match Django's integration

Matt343:
## Description

This adds support for the `auto` type annotation described in #1192 to the Pydantic integration, which allows a user to define the list of fields without having to re-specify the type themselves. This gives better editor and type checker support (although I didn't update the Mypy integration).

I haven't marked the `fields` list as deprecated, I wasn't sure if we wanted to do that in this release or not, but that's an easy change.

I also did a bit of cleanup on the error_type section, since it was missing a lot of the logic in the type function that allows for more complex input types.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

* Closes #1192 

## Checklist

### DIFF
--- a/docs/integrations/pydantic.md
+++ b/docs/integrations/pydantic.md
@@ -30,23 +30,38 @@ We can create a Strawberry type by using the
 
 ```python
 import strawberry
+from strawberry.experimental.pydantic import auto
 
 from .models import User
 
-@strawberry.experimental.pydantic.type(model=User, fields=[
-    'id',
-    'name',
-    'friends'
-])
+@strawberry.experimental.pydantic.type(model=User)
 class UserType:
-    pass
+    id: auto
+    name: auto
+    friends: auto
 ```
 
 The `strawberry.experimental.pydantic.type` decorator accepts a Pydantic model
-and a list of fields that we want to expose on our GraphQL API.
+and wraps a class that contains dataclass style fields with `auto` as the type
+annotation. The fields marked with `auto` will inherit their types from the
+Pydantic model.
 
-> **Note** specifying the list of field is required to prevent accidentally
-> exposing fields that weren't meant to be exposed on a API
+If you want to include all of the fields from your Pydantic model, you can
+instead pass `all_fields=True` to the decorator.
+
+-> **Note** Care should be taken to avoid accidentally exposing fields that
+-> weren't meant to be exposed on an API using this feature.
+
+```python
+import strawberry
+from strawberry.experimental.pydantic import auto
+
+from .models import User
+
+@strawberry.experimental.pydantic.type(model=User, all_fields=True)
+class UserType:
+    pass
+```
 
 ## Input types
 
@@ -55,16 +70,15 @@ Input types are similar to types; we can create one by using the
 
 ```python
 import strawberry
+from strawberry.experimental.pydantic import auto
 
 from .models import User
 
-@strawberry.experimental.pydantic.input(model=User, fields=[
-    'id',
-    'name',
-    'friends'
-])
+@strawberry.experimental.pydantic.input(model=User)
 class UserInput:
-    pass
+    id: auto
+    name: auto
+    friends: auto
 ```
 
 ## Interface types
@@ -74,6 +88,7 @@ Interface types are similar to normal types; we can create one by using the
 
 ```python
 import strawberry
+from strawberry.experimental.pydantic import auto
 from pydantic import BaseModel
 from typing import List
 
@@ -89,24 +104,18 @@ class AdminUser(User):
     role: int
 
 # strawberry types
-@strawberry.experimental.pydantic.interface(model=User, fields=[
-    'id',
-    'name',
-])
+@strawberry.experimental.pydantic.interface(model=User)
 class UserType:
-    pass
+    id: auto
+    name: auto
 
-@strawberry.experimental.pydantic.type(model=NormalUser, fields=[
-    'friends',
-])
+@strawberry.experimental.pydantic.type(model=NormalUser)
 class NormalUserType(UserType):  # note the base class
-    pass
+    friends: auto
 
-@strawberry.experimental.pydantic.type(model=AdminUser, fields=[
-    'role',
-])
+@strawberry.experimental.pydantic.type(model=AdminUser)
 class AdminUserType(UserType):
-    pass
+    role: auto
 ```
 
 ## Error Types
@@ -118,6 +127,7 @@ Pydantic errors in GraphQL. Let's see an example:
 ```python+schema
 import pydantic
 import strawberry
+from strawberry.experimental.pydantic import auto
 
 class User(BaseModel):
     id: int
@@ -125,13 +135,11 @@ class User(BaseModel):
     signup_ts: Optional[datetime] = None
     friends: List[int] = []
 
-@strawberry.experimental.pydantic.error_type(model=User, fields=[
-    'id',
-    'name',
-    'friends'
-])
+@strawberry.experimental.pydantic.error_type(model=User)
 class UserError:
-    pass
+    id: auto
+    name: auto
+    friends: auto
 
 ---
 
@@ -151,6 +159,7 @@ GraphQL type that aren't defined in the pydantic model
 
 ```python+schema
 import strawberry
+from strawberry.experimental.pydantic import auto
 import pydantic
 
 from .models import User
@@ -159,11 +168,10 @@ class User(BaseModel):
     id: int
     name: str
 
-@strawberry.experimental.pydantic.type(model=User, fields=[
-    'id',
-    'name',
-])
+@strawberry.experimental.pydantic.type(model=User)
 class User:
+    id: auto
+    name: auto
     age: int
 
 ---
@@ -186,6 +194,7 @@ To convert a Pydantic instance to a Strawberry instance you can use
 
 ```python
 import strawberry
+from strawberry.experimental.pydantic import auto
 from typing import List, Optional
 from pydantic import BaseModel
 
@@ -195,12 +204,10 @@ class User(BaseModel):
     name: str
 
 
-@strawberry.experimental.pydantic.type(model=User, fields=[
-    'id',
-    'name',
-])
+@strawberry.experimental.pydantic.type(model=User)
 class UserType:
-    pass
+    id: auto
+    name: auto
 
 instance = User(id='123', name='Jake')
 
@@ -213,6 +220,7 @@ specify the values to assign to them.
 
 ```python
 import strawberry
+from strawberry.experimental.pydantic import auto
 from typing import List, Optional
 from pydantic import BaseModel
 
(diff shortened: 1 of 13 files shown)


strawberry/experimental/pydantic/__init__.py
```diff
@@ -1,6 +1,14 @@
 from .error_type import error_type
 from .exceptions import UnregisteredTypeException
 from .object_type import input, interface, type
+from .utils import auto
```
Matt343, line 4: We probably want to promote this up to a more accessible package eventually, but I didn't want to leave the `experimental` namespace without input from a maintainer.

strawberry/experimental/pydantic/utils.py
```diff
@@ -24,3 +26,15 @@ def get_strawberry_type_from_model(type_: Any):
...
+
+def get_private_fields(cls: Type) -> List[dataclasses.Field]:
+    private_fields: List[dataclasses.Field] = []
+    for field in dataclasses.fields(cls):
+        if isinstance(field.type, Private):
+            private_fields.append(field)
+    return private_fields
+
+
+class auto:
```
Matt343, line 39: I wasn't sure if this was the best way to do this, but it does seem to work and gives us some options for adding things to the object if needed.

botberry:



---

Adds support for the `auto` type annotation described in #1192 to the Pydantic
integration, which allows a user to define the list of fields without having to
re-specify the type themselves. This gives better editor and type checker
support. If you want to expose every field you can instead pass
`all_fields=True` to the decorators and leave the body empty.

```python
import pydantic
import strawberry
from strawberry.experimental.pydantic import auto

    age: int
    password: str


@strawberry.experimental.pydantic.type(User)
class UserType:
    age: auto
    password: auto
```


---




Here's the tweet text: 
```

```

strawberry/experimental/pydantic/error_type.py
```diff
@@ -44,26 +49,57 @@ def field_type_to_type(type_):
...
 
-        cls = dataclasses.make_dataclass(
-            cls.__name__,
-            [
+        existing_fields = getattr(cls, "__annotations__", {})
+        fields_set = fields_set.union(
+            set(name for name, typ in existing_fields.items() if typ == auto)
+        )
+
+        if all_fields:
```
jkimbo, line 67: Can we add a warning here if `fields_set` is not empty? We'll be overriding any explicitly defined fields and that is probably a bug in the application code.

strawberry/experimental/pydantic/error_type.py
```diff
@@ -44,26 +49,57 @@ def field_type_to_type(type_):
...
-                    dataclasses.field(default=None),  # type: ignore
+                    field.name,
+                    field.type,
+                    field,
                 )
-                for name, field in model_fields.items()
-                if name in fields_set
-            ],
+                for field in extra_fields + private_fields
+                if field.type != auto
```
jkimbo, line 95: Why the check against auto here? Should we let private fields and strawberry fields use the auto type. It doesn't make much sense to use auto outside of the integrations IMO.

strawberry/experimental/pydantic/exceptions.py
```diff
@@ -5,7 +5,11 @@
 
 class MissingFieldsListError(Exception):
     def __init__(self, type: Type[BaseModel]):
-        message = f"List of fields to copy from {type} is empty"
+        message = (
+            f"List of fields to copy from {type} is empty. Either pass a "
+            f"`fields` list, set `all_fields` to True, or add fields with the "
```
jkimbo, line 10: I think we can just recommend the `auto` api here since that's what we're changing the docs to use.

strawberry/experimental/pydantic/object_type.py
```diff
@@ -58,32 +58,33 @@ def get_type_for_field(field: ModelField):
...
         model_fields = model.__fields__
-        fields_set = set(fields)
+        fields_set = set(fields) if fields else set([])
+
+        existing_fields = getattr(cls, "__annotations__", {})
+        fields_set = fields_set.union(
+            set(name for name, typ in existing_fields.items() if typ == auto)
+        )
+
+        if all_fields:
```
jkimbo, line 81: Same comment as above

strawberry/experimental/pydantic/object_type.py
```diff
@@ -103,16 +104,17 @@ def wrap(cls):
...
-        all_fields.extend(
+        all_model_fields.extend(
             (
                 (
                     field.name,
                     field.type,
                     field,
                 )
                 for field in extra_fields + private_fields
+                if field.type != auto
```
jkimbo, line 117: Same comment as above

strawberry/experimental/pydantic/utils.py
jkimbo: I think we should start using the new sentinel class: https://www.python.org/dev/peps/pep-0661/

It's not currently in the standard lib but they have a reference implementation that should be easy to include: https://github.com/taleinat/python-stdlib-sentinels

jkimbo (changes requested):
This is looking great thanks @Matt343 

Left a couple of suggestions.

strawberry/experimental/pydantic/error_type.py
Matt343: The check against auto is to make sure we're not including the fields that we already captured above. If we don't do this we get duplication of the fields, as the get captured as extra fields and through the initial checks.

jkimbo: Ah I see that makes sense, thanks.

jkimbo (approved):
Thanks @Matt343 . This looks great.

Would like to get someone else from @strawberry-graphql/core to have a look at it as well before we merge it.

docs/integrations/pydantic.md
```diff
@@ -30,23 +30,38 @@ We can create a Strawberry type by using the
 
 ```python
 import strawberry
+from strawberry.experimental.pydantic import auto
```
patrick91, line 33: considered we will use this for django, should we move it somewhere else? 🤔

docs/integrations/pydantic.md
```diff
@@ -30,23 +30,38 @@ We can create a Strawberry type by using the
...
-and a list of fields that we want to expose on our GraphQL API.
+and wraps a class that contains dataclass style fields with `auto` as the type
+annotation. The fields marked with `auto` will inherit their types from the
+Pydantic model.
+
+If you want to include all of the fields from your Pydantic model, you can
+instead pass `all_fields=True` to the decorator.
+
+-> **Note** Care should be taken to avoid accidentally exposing fields that
+-> weren't meant to be exposed on an API using this feature.
```
patrick91, line 53: 💯

docs/integrations/pydantic.md
```diff
@@ -30,23 +30,38 @@ We can create a Strawberry type by using the
...
 
-> **Note** specifying the list of field is required to prevent accidentally
-> exposing fields that weren't meant to be exposed on a API
+```python
+import strawberry
+from strawberry.experimental.pydantic import auto
```
patrick91, line 57:
```suggestion
```

strawberry/experimental/pydantic/__init__.py
patrick91: did you have any id where this could live? :D

strawberry/experimental/pydantic/error_type.py
```diff
@@ -44,26 +50,63 @@ def field_type_to_type(type_):
 def error_type(
     model: Type[BaseModel],
     *,
-    fields: List[str],
+    fields: List[str] = None,
```
patrick91, line 53: should we start deprecating this?

:D

strawberry/experimental/pydantic/exceptions.py
```diff
@@ -5,7 +5,10 @@
 
 class MissingFieldsListError(Exception):
     def __init__(self, type: Type[BaseModel]):
-        message = f"List of fields to copy from {type} is empty"
+        message = (
+            f"List of fields to copy from {type} is empty. Add fields with the "
+            f"`auto` type annotation"
+        )
```
patrick91, line 11: 👏

strawberry/sentinel.py
```diff
@@ -0,0 +1,95 @@
+"""Reference implementation of [PEP-0661](https://www.python.org/dev/peps/pep-0661/)
+
+See https://github.com/taleinat/python-stdlib-sentinels
+"""
```
patrick91, line 4: isn't there a package for this?

if so we should use, otherwise we need to include the license :)

tests/experimental/pydantic/schema/test_basic.py
```diff
@@ -333,9 +413,6 @@ class BranchBType(BaseType):
     class UserType:
         pass
 
-    print(BranchAType._type_definition)
-    print(BaseType._type_definition)
```
patrick91, line 337: thanks!

patrick91 (commented):
wow this is an amazing PR! 👏 

I've left a couple of minor comments, other than that it looks awesome!

docs/integrations/pydantic.md
Matt343: Yeah I'm happy to move it, just wasn't sure where it belonged. Ideally I think we'd expose it from the top level namespace for ease of use, but where the code should go I'm not sure

strawberry/sentinel.py
Matt343: There's a https://pypi.org/project/sentinel/ and a https://pypi.org/project/sentinels/ library, neither of them are connected to the PEP as far as I can tell, but they both claim to have very similar features. I don't see a license of the reference implementation github, but the PEP itself has "This document is placed in the public domain or under the CC0-1.0-Universal license, whichever is more permissive."

I can switch this to one of the libraries mentioned if you want, I agree it doesn't make much sense to maintain our own implementation.

strawberry/experimental/pydantic/error_type.py
Matt343: Yup, I'll add a warning for that

strawberry/sentinel.py
BryceBeagle: Licensing for the `python-stdlib-sentinels` might be a bit tricky because it doesn't include an explicit LICENSE in the repo

Matt343: Yeah I'll go ahead and swap to one of the libraries

PR #1159, opened 2021-08-19
FastAPI integration

Skeen:
## Description

This PR introduces a shell integration around the ASGI integration.

It can be tested using `example.py`:
```Python
from typing import List

from uvicorn.main import run

import strawberry
from fastapi import FastAPI
from strawberry.fastapi import GraphQLRouter


@strawberry.type
class Book:
    title: str
    author: str


def get_books():
    return [
        Book(
            title="The Great Gatsby",
            author="F. Scott Fitzgerald",
        ),
    ]


@strawberry.type
class Query:
    books: List[Book] = strawberry.field(resolver=get_books)


schema = strawberry.Schema(Query)

app = FastAPI()
app.include_router(
    GraphQLRouter(schema, graphiql=True),
    prefix="/graphql",
    tags=["GraphQL"],
)
if __name__ == "__main__":
    run("example:app", port=5011, reload=True)
```

The main argument for existence of this shell integration is having GraphQL exposed in the `openapi.json` generated by FastAPI, and thus in the swagger / redoc documentation:

image

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] Enhancement/optimization
- [x] Documentation

## Checklist

- [ ] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the CONTRIBUTING document.
- [ ] I have added tests to cover my changes.
- [x] I have tested the changes and verified that they work and don't break anything (as well as I can manage).

PR #1291, opened 2021-10-03
FastAPI integration

jiribires:
## Description

FastAPI integration with suport for FastAPI dependency injection into Strawberry context and root value.

Updated fastapi.md contains examples for using context, root value and background task.

Test suite is taken from asgi integration.

## Comparison with #1159 Fast API integration

#1159 adds support for APIRouter with OpenAPI docs. This PR use APIRouter too, but allows usage of FastAPI dependency injection, supports file uploads and contains large test suite. The disadvantage is that my solution does not use the existing ASGI integration and therefore increases maintenance cost.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

## Checklist

### DIFF
--- a/docs/integrations/fastapi.md
+++ b/docs/integrations/fastapi.md
@@ -4,21 +4,151 @@ title: FastAPI
 
 # FastAPI
 
-Strawberry provides support for [FastAPI](https://fastapi.tiangolo.com/) with the ASGI integration.
+Strawberry provides support for [FastAPI](https://fastapi.tiangolo.com/) with the APIRouter integration.
 
 See below example for integrating FastAPI with Strawberry:
 
 ```python
 from fastapi import FastAPI
-from strawberry.asgi import GraphQL
+from strawberry.fastapi import GraphQL
 
 from api.schema import schema
 
 graphql_app = GraphQL(schema)
 
 app = FastAPI()
-app.add_route("/graphql", graphql_app)
-app.add_websocket_route("/graphql", graphql_app)
+app.include_router(graphql_app, prefix="/graphql")
 ```
 
-For more information about Strawberry ASGI refer to [asgi.md](./asgi.md)
+## Options
+
+The `GraphQL` app accepts following options:
+
+- schema: mandatory, the schema created by `strawberry.Schema`.
+- graphiql: optional, defaults to `True`, whether to enable the GraphiQL
+  interface.
+- context_getter: optional FastAPI dependency for prividing custom context value
+- root_value_getter: optional FastAPI dependency for prividing custom root value
+
+## context_getter
+
+`context_getter` option allows to provide a custom context object that can be used in your
+resolver. You can return anything here, by default we return a dictionary with
+the request and background tasks.
+
+`context_getter` is [FastAPI dependency](https://fastapi.tiangolo.com/tutorial/dependencies/)
+and can inject another dependencies.
+
+```python
+from fastapi import Depends, Request, WebSocket, BackgroundTasks
+
+def custom_context_dependency() -> str:
+    return "Hi!"
+
+
+async def get_context(
+    background_tasks: BackgroundTasks,
+    request: Request = None,
+    ws: WebSocket = None,
+    custom_value=Depends(custom_context_dependency),
+):
+    return {
+        "custom_value": custom_value,
+        "request": request or ws,
+        "background_tasks": background_tasks,
+    }
+
+
+graphql_app = GraphQL(
+  schema,
+  context_getter=get_context
+)
+
+@strawberry.type
+class Query:
+    @strawberry.field
+    def example(self, info: Info) -> str:
+        return str(info.context["custom_value"])
+```
+
+Here we are returning a custom context dictionary that contains one extra item
+called "custom_value", which is injected from `custom_context_dependency`.
+
+Then we use the context in a resolver, the resolver will return "Hi!" in this
+case.
+
+### Setting background tasks
+
+Similarly, [background tasks](https://fastapi.tiangolo.com/tutorial/background-tasks/?h=background) can be added via the context:
+
+```python
+from starlette.background import BackgroundTask
+
+async def notify_new_flavour(name: str):
+    ...
+
+@strawberry.type
+class Mutation:
+    @strawberry.mutation
+    def create_flavour(self, name: str, info: Info) -> bool:
+        info.context["background_tasks"].add_task(notify_new_flavour)
+
+        info.context["response"].background = BackgroundTask(notify_new_flavour, name)
+```
+
+## root_value_getter
+
+`root_value_getter` option allows to provide a custom root value for your schema, this is
+probably not used a lot but it might be useful in certain situations.
+
+Here's an example:
+
+```python
+from fastapi import Request
+
+async def get_root_value(request: Request):
+    return Query(name="Patrick")
+
+
+graphql_app = GraphQL(
+    schema,
+    root_value_getter=get_root_value,
+)
+
+@strawberry.type
+class Query:
+    name: str
+```
+
+Here we are returning a Query where the name is "Patrick", so we when requesting
+the field name we'll return "Patrick" in this case.
+
+## process_result
+
+`process_result` allows to customize and/or process results before they are sent
+to the clients. This can be useful logging errors or hiding them (for example to
+hide internal exceptions).
+
+It needs to return an object of `GraphQLHTTPResponse` and accepts the request
+and the execution results.
+
+```python
+from strawberry.http import GraphQLHTTPResponse
+from strawberry.types import ExecutionResult
+
+from graphql.error import format_error as format_graphql_error
+
+class MyGraphQL(GraphQL):
+    async def process_result(
+        self, request: Request, result: ExecutionResult
+    ) -> GraphQLHTTPResponse:
+        data: GraphQLHTTPResponse = {"data": result.data}
+
+        if result.errors:
+            data["errors"] = [format_graphql_error(err) for err in result.errors]
+
+        return data
+```
+
+In this case we are doing the default processing of the result, but it can be
+tweaked based on your needs. the result, but it can be tweaked based on your needs.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,7 @@ pydantic = {version = "<2", optional = true}
 python-multipart = "^0.0.5"
 sanic = {version = ">=20.12.2,<22.0.0", optional = true}
 aiohttp = {version = "^3.7.4.post0", optional = true}
+fastapi = {version = "^0.68.1", optional = true}
 
 [tool.poetry.dev-dependencies]
 pytest = "^6.2"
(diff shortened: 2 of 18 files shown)


botberry:



---

This release adds support for FastAPI integration using APIRouter.

```python
from fastapi import FastAPI
from strawberry.fastapi import GraphQL

from api.schema import schema

graphql_app = GraphQL(schema)

app = FastAPI()
```


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Jiří Bireš for the PR 👏

```

patrick91 (commented):
This looks great, I need a bit more time to test it though! 😊

I'll also ask some people to double check the code as I'm not super familiar with FastAPI :)

strawberry/fastapi/__init__.py
```diff
@@ -0,0 +1,188 @@
...
+        graphiql: bool = True,
+        keep_alive: bool = False,
+        keep_alive_interval: float = 1,
+        debug: bool = False,
+        root_value_getter=get_root_value,
+        context_getter=get_context,
+        subscription_protocols=(GRAPHQL_TRANSPORT_WS_PROTOCOL, GRAPHQL_WS_PROTOCOL),
+        connection_init_wait_timeout: timedelta = timedelta(minutes=1),
+    ):
+        super().__init__(prefix=prefix)
```
Kludex, line 56: Not sure if [`redirect_slashes`](https://github.com/encode/starlette/blob/28088573aebc70807ef2aa422000cddb93b57a70/starlette/routing.py#L527-L541) makes sense as well? 
```suggestion
        connection_init_wait_timeout: timedelta = timedelta(minutes=1),
        default: Optional[ASGIApp] = None,
        on_startup: Optional[Sequence[Callable[[], Any]]] = None,
        on_shutdown: Optional[Sequence[Callable[[], Any]]] = None,
    ):
        super().__init__(
            prefix=prefix,
            default=default,
            on_startup=on_startup,
            on_shutdown=on_shutdown
        )
```

strawberry/fastapi/__init__.py
```diff
@@ -0,0 +1,188 @@
...
+        self.keep_alive = keep_alive
+        self.keep_alive_interval = keep_alive_interval
+        self.debug = debug
+        self.root_value_getter = root_value_getter
+        self.context_getter = context_getter
+        self.protocols = subscription_protocols
+        self.connection_init_wait_timeout = connection_init_wait_timeout
+
+        @self.get(
+            "",
```
Kludex, line 68: Can this be set via Router? Like `graphiql_url`? And be analogous to `docs_url` and `redoc_url` FastAPI has?

In any case, this cannot be an empty string, can it? There's an issue when both the prefix and the endpoint have empty string:
```
Exception: Prefix and path cannot be both empty (path operation: home)
```
I mean, you can only see this error if you don't change the prefix. But I assume, as a user, that the default will work.

strawberry/fastapi/__init__.py
```diff
@@ -0,0 +1,188 @@
...
+        self.graphiql = graphiql
+        self.keep_alive = keep_alive
+        self.keep_alive_interval = keep_alive_interval
+        self.debug = debug
+        self.root_value_getter = root_value_getter
+        self.context_getter = context_getter
+        self.protocols = subscription_protocols
+        self.connection_init_wait_timeout = connection_init_wait_timeout
+
+        @self.get(
```
Kludex, line 67: There's no need to register this endpoint if the `self.graphiql` is `False` to begin with, and you don't need to add the logic around 404, as it will be automatically handled.

There's no need for `response_class` parameter if you are returning it anyway.
```python
        if self.graphiql:
            self.add_api_route(
                "",
                self.get_graphiql_response,
                methods=["GET"],
                responses={
                    200: {
                        "description": "The GraphiQL integrated development environment.",
                    },
                },
            )
```
I think `add_api_route` is cuter on cases like this, but you can also use the `get` decorator.

Kludex: If the endpoint doesn't work because of the `self`, you can turn it into a `classmethod`.

strawberry/fastapi/handlers/graphql_ws_handler.py
```diff
@@ -0,0 +1,62 @@
...
+
+class GraphQLWSHandler(BaseGraphQLWSHandler):
+    def __init__(
+        self,
+        schema: BaseSchema,
+        debug: bool,
+        keep_alive: bool,
+        keep_alive_interval: float,
+        context,
+        root_value,
```
Kludex, line 20: Is it common for those to not have type hint?

Kludex (commented):
I think it would make more sense to be `GraphQLRouter` then `GraphQL`, so I've made some comments taking that in consideration, if decided to be `GraphQL`, just ignore them.

My review was made in good faith, I don't have any familiarity with Strawberry. :)

strawberry/fastapi/__init__.py
jiribires: Unless `static/graphiql.html` is changed to support custom fetch URL graphql and graphiql endpoints must be same. I'm not sure if this is standart in other GraphQL frameworks too but other extensions use same endpoint so I would prefer not to change this.

strawberry/fastapi/__init__.py
jiribires: GraphiQL is registered on same endpoint as HTTP and websocket handlers, so GET request would responded with 405 - method not allowed. ASGI, Flask and other extensions are using 404 status so to keep it consistent I'm using 404 too, although I think 405 would be more appropriate.

docs/integrations/fastapi.md
```diff
@@ -4,21 +4,151 @@ title: FastAPI
...
 app = FastAPI()
-app.add_route("/graphql", graphql_app)
-app.add_websocket_route("/graphql", graphql_app)
+app.include_router(graphql_app, prefix="/graphql")
 ```
 
-For more information about Strawberry ASGI refer to [asgi.md](./asgi.md)
+## Options
+
+The `GraphQL` app accepts following options:
```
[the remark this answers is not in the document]
jiribires: GraphQLRouter is definitely a more appropriate name :-). I'm going to rename it.

strawberry/fastapi/__init__.py
jiribires: I agree that `add_api_route` is better in this case.

artiombell (approved):

strawberry/fastapi/__init__.py
```diff
@@ -0,0 +1,188 @@
+import json
```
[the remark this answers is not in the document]
jiribires: GraphQLRouter is moved to router.py

pyproject.toml
```diff
@@ -44,7 +44,7 @@ pydantic = {version = "<2", optional = true}
 python-multipart = "^0.0.5"
 sanic = {version = ">=20.12.2,<22.0.0", optional = true}
 aiohttp = {version = "^3.7.4.post0", optional = true}
-fastapi = {version = "^0.68.1", optional = true}
+fastapi = {version = ">=0.65.0", optional = true}
```
Kludex, line 47: https://github.com/tiangolo/fastapi/security/advisories/GHSA-8h2j-cgx8-6xv7
```suggestion
fastapi = {version = ">=0.65.2", optional = true}
```

strawberry/fastapi/handlers/graphql_ws_handler.py
jiribires: Yes

docs/integrations/fastapi.md
```diff
@@ -4,21 +4,153 @@ title: FastAPI
...
 See below example for integrating FastAPI with Strawberry:
 
 ```python
 from fastapi import FastAPI
-from strawberry.asgi import GraphQL
+from strawberry.fastapi import GraphQL
 
 from api.schema import schema
 
 graphql_app = GraphQL(schema)
```
Skeen, line 17: Shouldn't all references to `GraphQL` be `GraphQLRouter` instead, both here and in `RELEASE.md`?

docs/integrations/fastapi.md
```diff
@@ -4,21 +4,153 @@ title: FastAPI
...
+        data: GraphQLHTTPResponse = {"data": result.data}
+
+        if result.errors:
+            data["errors"] = [format_graphql_error(err) for err in result.errors]
+
+        return data
+```
+
+In this case we are doing the default processing of the result, but it can be
+tweaked based on your needs. the result, but it can be tweaked based on your needs.
```
Skeen, line 156: There seems to be some issues with the sentence

docs/integrations/fastapi.md
jiribires: Yes, I'm updating docs according to latest code changes and to make them working when copie. And I'm make code examples in docs to run as are, without changes.

strawberry/fastapi/router.py
```diff
@@ -0,0 +1,196 @@
...
+            if not self.graphiql:
+                return Response(status_code=status.HTTP_404_NOT_FOUND)
+            return self.get_graphiql_response()
+
+        @self.post("")
+        async def handle_http_query(
+            request: Request,
+            context=Depends(self.context_getter),
+            root_value=Depends(self.root_value_getter),
+        ):
```
Skeen, line 91:
```suggestion
        ) -> Response:
```

strawberry/fastapi/router.py
```diff
@@ -0,0 +1,196 @@
...
+                    ws=websocket,
+                ).handle()
+            else:
+                await websocket.close(code=4406)
+
+    def pick_preferred_protocol(self, ws: WebSocket) -> Optional[str]:
+        protocols = ws["subprotocols"]
+        intersection = set(protocols) & set(self.protocols)
+        sorted_intersection = sorted(intersection, key=protocols.index)
+        return next(iter(sorted_intersection), None)
```
Skeen, line 173: Could the last two lines not be replaced with a check for the empty set and the `min` function?
I mean, there is no reason to sort the entire set (`O(n*log(n))`) if we just want to find the smallest element (`O(n)`)?

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -182,10 +182,10 @@ async def handle_async_results(
         except asyncio.CancelledError:
             # CancelledErrors are expected during task cleanup.
             return
-        except Exception as error:
+        except Exception as err:
             # GraphQLErrors are handled by graphql-core and included in the
             # ExecutionResult
-            error = GraphQLError(str(error), original_error=error)
+            error = GraphQLError(str(err), original_error=err)
```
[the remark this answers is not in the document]
jiribires: I'm sorry this shouldn't be commited, to be honest I even don't know when and why I changed this 😕.

jiribires:
> Awesome stuff! Thanks @Skeen, @Kludex and @artiombell for the additional reviews!
> 
> @jiribires can you merge master into this PR so we resolve the conflicts? I'll merge after that :D

I've allredy merged main to this PR two hours ago and there are no new changes in main so far.

Thank you all for reviews and suggestions :-)

patrick91:
I'll check the mypy issue and merge this when fixed 😊

botberry:

Issue #681, opened 2021-01-23
Add safety checks around strawberry.type, strawberry.input and strawberry.interface


patrick91:
I found a place where I had this:

```
from .enums import ExampleEnum

ABC = strawberry.type(ExampleEnum)
```

I think we should add a safety check as we already for enums: https://github.com/strawberry-graphql/strawberry/blob/master/strawberry/enum.py#L24-L25

We should only allow to decorate classes with strawberry.type, strawberry.input and strawberry.interface

estyxx:
*Think the code change a bit meanwhile, here there is now the reference:
https://github.com/strawberry-graphql/strawberry/blob/50684a7ca74af876441438351daf6de5ca22a394/strawberry/enum.py#L40-L41

marcoacierno:
I think we can close this? :)

PR #1292, opened 2021-10-03
Add safety checks (#681)

estyxx:
## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #681

## Checklist


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Etty for the PR 👏

```

strawberry/exceptions.py
```diff
@@ -14,6 +14,13 @@ def __init__(self):
         super().__init__(message)
 
 
+class NotAClass(Exception):
+    def __init__(self, method: str):
+        message = f"strawberry.{method} can only be used with classes"
+
+        super().__init__(message)
```
BryceBeagle, line 21: This should just take the object itself, and not the `str` (i.e. the exception should be in charge of the output formatting)

tests/types/test_object_types.py
```diff
@@ -121,3 +124,30 @@ class WishfulThinking:
     field: StrawberryField = WishfulThinking._type_definition.fields[0]
 
     assert field.type is EU
+
+
+def test_raises_error_when_using_type_with_a_not_class_object():
```
BryceBeagle, line 129: This file (and directory of tests) is used for testing the `StrawberryType` logic of fields. We need to revisit the organization of tests in the future, but for now I would create a new test file in `tests/objects` for these

strawberry/exceptions.py
```diff
@@ -14,6 +14,13 @@ def __init__(self):
         super().__init__(message)
 
 
+class NotAClass(Exception):
```
BryceBeagle, line 17: I think a better name for this exception might be something along the lines of `ObjectIsNotTypeError` (especially with the `Error` at the end)

BryceBeagle (changes requested):
I think this check is a really good idea!

strawberry/exceptions.py
estyxx: Mmm but if I give the `cls` object to the exception, this will be the non-class object, I need to pass the method the user is using: `type`|`input`|`interface`... 🤔

BryceBeagle: Oh right, I misread how this error was to be used. Hmmm, in that case I'm not a fan of passing the strings directly. Maybe we can use some `classmethod`s here?

```python
from enum import Enum

class NotAClass(Exception):
    class MethodType(Enum):
        INPUT = "input"
        INTERFACE = "interface"
        TYPE = "type"

    def __init__(self, obj: object, method_type: METHOD_TYPE):
        message = (
            f"strawberry.{method_type} can only be used with class types. Provided object "
            f"{obj} is not a type."
        )

        super().__init__(message)

   @classmethod
   def input(cls, obj: object) -> NotAClass:
       return cls(obj, cls.MethodType.INPUT)

   @classmethod
   def interface(cls) -> NotAClass:
       return cls(obj, cls.MethodType.INTERFACE)

   @classmethod
   def type(cls) -> NotAClass:
       return cls(obj, cls.MethodType.TYPE)
```
```python
a = object()  # example object that's not a type

raise NotAClass.interface(a)
```

estyxx: Nice! much better :)

strawberry/exceptions.py
```diff
@@ -1,19 +1,50 @@
...
 
+from enum import Enum
 from typing import List, Set, Union
 
 from graphql import GraphQLObjectType
 
 from strawberry.type import StrawberryType
 
 
+# TODO: add links to docs
```
BryceBeagle, line 11: This TODO still needs to be resolved

strawberry/object_type.py
```diff
@@ -142,8 +147,17 @@ def type(
...
-        wrapped = _wrap_dataclass(cls)
+        if not inspect.isclass(cls):
+            exc = (
+                ObjectIsNotClassError.input
+                if is_input
+                else ObjectIsNotClassError.interface
+                if is_interface
+                else ObjectIsNotClassError.type
+            )
+            raise exc(cls)
```
BryceBeagle, line 158: I think it would be cleaner to just use an `if/elif/else` tree

BryceBeagle (changes requested):
It also appears were missing some coverage in the tests -- but I'm not sure why. Some of the codecov alerts are for code that looks like it should be executed

strawberry/exceptions.py
estyxx: We should add a section in the docs that shows all these exceptions we can thrown?

estyxx:
Think I figure out why the coverage dropped: pytest doesn't allow to have two files with the identical name in two different directories.

Also, could you please add the Oktoberfest-accepted label? tnks :)

strawberry/exceptions.py
BryceBeagle: Yes :+1: Can we open an issue and link to the issue number here in the comment?

estyxx: Here there is :) #1298

BryceBeagle: Excellent. In the case that I was ambiguous, I meant in the source comment too :slightly_smiling_face:

tests/objects/test_types.py
```diff
@@ -0,0 +1,40 @@
+import pytest
+
+import strawberry
+from strawberry.exceptions import ObjectIsNotClassError
+
+
+def test_raises_error_when_using_type_with_a_not_class_object():
+    expected_error = (
+        r"strawberry.type can only be used with class types. Provided "
+        "object .* is not a type."
```
BryceBeagle, line 10: It doesn't end up changing things here, but you probably wanted to do this to each of these expected errors:
```suggestion
        r"strawberry.type can only be used with class types. Provided "
        r"object .* is not a type."
```

BryceBeagle (changes requested):

BryceBeagle (approved):

tests/objects/test_types.py
estyxx: Yeah the match is working so think python is converting them to be a rstring both... but agree, it's also looks better in the IDE :)

BryceBeagle: It's working because the second string has no special characters. Note that `r""` doesn't make a regex string, it makes a _raw_ string, where escape characters (i.e. those prefixed by `\`s) are ignored and interpreted literally. However, people often use `r""` strings to indicate regex patterns even if not using special characters

estyxx: Oh, you are right, I can omit now the `r`... Was not working at some point, so I added the `r`... I gues there was something else wrong :D

BryceBeagle: I tend to use the `r""` strings for regex even if there's no escape sequences. I think it adds a little expressiveness to the code saying "hey, this is a pattern used as a regex". Could help future maintainers. But I'll leave the choice to use it or not up to you (or @patrick91 if he has an opinion :P)

estyxx: Yeah totally agree! I will leave it there :)

tests/objects/test_types.py
```diff
@@ -0,0 +1,40 @@
...
+            pass
+
+
+def test_raises_error_when_using_interface_with_a_not_class_object():
+    expected_error = (
+        r"strawberry.interface can only be used with class types. Provided "
+        r"object .* is not a type."
+    )
+    with pytest.raises(ObjectIsNotClassError, match=expected_error):
+
```
patrick91, line 37:
```suggestion

    with pytest.raises(ObjectIsNotClassError, match=expected_error):
```

tests/objects/test_types.py
```diff
@@ -0,0 +1,40 @@
...
+
+def test_raises_error_when_using_input_with_a_not_class_object():
+    expected_error = (
+        r"strawberry.input can only be used with class types. Provided "
+        r"object .* is not a type."
+    )
+    with pytest.raises(ObjectIsNotClassError, match=expected_error):
+
+        @strawberry.input
```
patrick91, line 26:
```suggestion

    with pytest.raises(ObjectIsNotClassError, match=expected_error):
        @strawberry.input
```

tests/objects/test_types.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def test_raises_error_when_using_type_with_a_not_class_object():
+    expected_error = (
+        r"strawberry.type can only be used with class types. Provided "
+        r"object .* is not a type."
+    )
+    with pytest.raises(ObjectIsNotClassError, match=expected_error):
+
+        @strawberry.type
```
patrick91, line 14:
```suggestion

    with pytest.raises(ObjectIsNotClassError, match=expected_error):
        @strawberry.type
```

patrick91 (approved):
Nice one! I've only requested a couple of nits, will commit them now :)

PR #1306, opened 2021-10-04
Use non dev version of python 3.10

patrick91:
Python 3.10 was just released 🚀


cclauss:
Try running the tests again because Py3.10 is now available on GitHub Actions.

patrick91:
@cclauss thanks for the heads up! that was faster that I expected :D

PR #1313, opened 2021-10-06
Add SyncToAsync extension to enable use of the Django ORM in an async context

jkimbo:
## Description

This PR adds a new extension `SyncToAsync` that wraps all customer resolvers in the `sync_to_async` decorator so that you can use the Django ORM when using the async view. This also means that you can use dataloaders with Django.

This PR also adds a `create_model_load_fn` to make it easier to make a dataloader for a Django model.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

*

## Checklist


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

Ambro17:
Do i understand correctly that this solution of dataloaders in sync contexts would only work with django models? Is there other issue to track how to solve it if we're using a sync server but not django?

patrick91:
@bellini666 do you want to take a look at this? do we still need it?

bellini666:
> @bellini666 do you want to take a look at this? do we still need it?

strawberry-graphql-django already solves this same issue by making all their fields have a "safe resolve", which will get wrapped in `sync_to_async` when inside running ASGI. You just need to use `strawberry.django.field` instead of `strawberry.field` in those

But the idea here is really interesting and allows using strawberry without strawberry_django for ASGI. Although using `strawberry.django.field` has other advantages than this, specially considering the "cache" resolving that is getting added to my open PR and the optimizer stuff that will get ported from strawberry-django-plus.

The question here is if we want to have 2 ways of solving the same issue for strawberry? I'm not against it, but if we do we would probably want to modify strawberry_django to make use of this extension, maybe "enforcing" its usage, or at least the "safe_resolver" would need to be aware that it is already running inside a `sync_to_async` context to avoid doing it again (we can use `contextvars` for that)

Since this PR is from 2021, maybe it was created in a time were strawberry_django didn't exist yet (or wasn't known)?

What do you think of that all all well @jkimbo ?

patrick91:
I'll close this, since we have it in Strawberry Django 😊

Issue #1314, opened 2021-10-06
opentelemetry fails to register inputs that are not primitives


omarzouk:
When the input to a GQL Query is an object, telemetry fails to register the `graphql.param` attribute because this line doesn't account for the input being an object instead of a primitive. I suggest we consider applying a `json.dump` if `value` is seen to be a dict:

https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/extensions/tracing/opentelemetry.py#L100

I get this error: 
```Invalid type dict for attribute value. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or a sequence of those types```

Full example to reproduce:
```python
from __future__ import annotations

import typing
from enum import Enum

import strawberry
from opentelemetry import trace
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk.trace.export import (
    ConsoleSpanExporter,
    BatchSpanProcessor
)
from starlette.requests import Request
from starlette.websockets import WebSocket
from strawberry.asgi import GraphQL
from strawberry.extensions.tracing import OpenTelemetryExtension
from strawberry.schema import BaseSchema
from strawberry.types import Info

trace.set_tracer_provider(TracerProvider())
trace.get_tracer_provider().add_span_processor(
        BatchSpanProcessor(ConsoleSpanExporter())
)


class API(GraphQL):
    def __init__(self, config, schema: BaseSchema, *args, **kwargs):
        super().__init__(schema, *args, **kwargs)
        self.config = config

    async def get_context(self, request: typing.Union[Request, WebSocket], *args, **kwargs) -> typing.Any:
        return {
            'config': self.config,
        }


@strawberry.type
class Greeting:
    message: str


@strawberry.enum
class Mood(Enum):
    Happy = "Happy"
    Sad = "Sad"


@strawberry.input
class GreetInput:
    name: str
    mood: Mood


@strawberry.federation.type(extend=True)
class Query:
    @strawberry.field
    async def greet(self, info: Info, input: GreetInput) -> Greeting:
        config = info.context['config']

        return Greeting(
                message="Hi {}. Glad to hear you're glad!".format(input.name)
        ) if input.mood == Mood.Happy else Greeting(
                message="Hi {}, I hope you feel better !".format(input.name))

    @strawberry.field
    async def secret_greet(self, info: Info, input: GreetInput) -> Greeting:
        config = info.context['config']

        return Greeting(message="{}, {}".format(input.name, config['secret_message']))


[...]

PR #1315, opened 2021-10-06
Fix: opentelemetry extension should handle non-primitive input parameters

omarzouk:
json dump input parameter if its not a primitive

- [X] Bugfix

* https://github.com/strawberry-graphql/strawberry/issues/1314

## Checklist


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Omar Marzouk for the PR 👏

```

ossareh (commented):
good catch! However, your solution will result in leaking potentially sensitive data. It's not clear to me what the actual correct solution is. 

Here are my thoughts:

Consider a case where a resolver takes a User object, perhaps that object looks something like this:

```
User:
  id: ID
  name: str
  password: str
```

The dev has to choose whether to log the sensitive information (`password`) or no information at all, including insensitive things that may help with debugging - such as the user ID or username.

I'm not convinced that logging a dict like this is at all helpful for the domain that tracing satisfies, the LightStep folks have a short piece in their Best Practices about this: https://opentelemetry.lightstep.com/best-practices/using-attributes/ (see "Leverage Cardinality"). 

I'm going to think about this for a bit, but my suggestion is to not merge this due to the inadvertent risk it offers.

ossareh:
> 
> In a similar example to yours current code will leak a password if the input is just:
> 
> ```
> my_query(id:"abc", password:"123")
> ```
> 

However, in this case, you can filter password like this: https://strawberry.rocks/docs/extensions/opentelemetry#more-examples

The same solution could also be used to strike out a password in my prior example too I suppose.

I can absolutely see value in logging certain parameters, I think this should be a "what parameters to include" list rather than a filter list. I'll bring this up in our discord - consider joining the conversation: https://discord.gg/NsxZAWbX

patrick91:
@omarzouk would you be able to add a test case for this? 😊

omarzouk:
@patrick91 I'm really sorry I've been incredibly busy the fast months, yes I would be happy to add a test for it. I will try to do it next week

omarzouk:
@patrick91 Hi, sorry it took much more than a week 😅  I got covid and then busy at work. However, I'm up and running again, and just pushed a test. Do u think it is enough?

omarzouk:
I have also learned that nulls should not be passed in to the `set_attribute` function as its behaviour is undefined, so I have added a line to skip arguments that are null https://github.com/open-telemetry/opentelemetry-specification/pull/992

omarzouk:
@patrick91 Hi! is this solution still relevant/ok? can we maybe move forward with it to get rid of the errors? or what do u think?

adambabik:
@omarzouk would you mind adding a file `RELEASE.md` to your PR so that the CI pipeline passes? An instruction is in [the first comment](https://github.com/strawberry-graphql/strawberry/pull/1315#issuecomment-936840953).

I would love this to get merged as I have quite many warnings of form `Invalid type NoneType for attribute value. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or a sequence of those types`.

botberry:
# Pre-release
:wave:

Releasing commit [14c5c15a71606b62960b09474903d8bb7231957f] to PyPi as pre-release! :package:

patrick91:
@adambabik would you be ok with trying this pre-release? `strawberry-graphql==0.131.1.dev.1663258832` if that works for you I can merge the PR 😊

adambabik:
@patrick91 thanks! I took a closer look and tried the patch but it may result in an exception when `json.dumps()` fails.

To merge this PR, at the very least, we need to wrap the `json.dumps()` call:
```py
try:
    v = json.dumps(value)
    span.set_attribute(f"graphql.param.{kwarg}", v)
except (TypeError, OverflowError):
    pass
```

With this change, we will start seeing more attributes but not all of them and we still won't have control over serialization. Having said that and also looking at the previous discussion about leaking sensitive data, I wonder if the best strategy wouldn't be to actually deny any params by default. By using `arg_filter` one can add params they actually need. We could also introduce `arg_serializer` to allow handling complex data types.

Alternatively, if we want to can keep adding params by default, we can still add `arg_serializer` and possibly provide a more robust default than `json.dumps()`. For example, put dict keys into attribute's name: `f"graphql.param.{kwarg}.{key_level_1}.{key_level_2}"`. More about attribute's name can be read [in the OpenTelemetry spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-naming.md).

patrick91:
> @patrick91 thanks! I took a closer look and tried the patch but it may result in an exception when `json.dumps()` fails.
> 
> To merge this PR, at the very least, we need to wrap the `json.dumps()` call:
> 
> ```python
> try:
>     v = json.dumps(value)
>     span.set_attribute(f"graphql.param.{kwarg}", v)
> except (TypeError, OverflowError):
>     pass
> ```
> 
> With this change, we will start seeing more attributes but not all of them and we still won't have control over serialization. Having said that and also looking at the previous discussion about leaking sensitive data, I wonder if the best strategy wouldn't be to actually deny any params by default. By using `arg_filter` one can add params they actually need. We could also introduce `arg_serializer` to allow handling complex data types.
> 
> Alternatively, if we want to can keep adding params by default, we can still add `arg_serializer` and possibly provide a more robust default than `json.dumps()`. For example, put dict keys into attribute's name: `f"graphql.param.{kwarg}.{key_level_1}.{key_level_2}"`. More about attribute's name can be read [in the OpenTelemetry spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-naming.md).

I think this is a better approach, I’m not sure if I’ll have time this week to make updates, would you be able to make a PR for it? ☺️

adambabik:
@patrick91 sure, I should be able to submit a PR early next week :)

Issue #1259, opened 2021-09-17
Errors in subscriptions don't get logged


jkimbo:
At the moment if an error is raised in a subscription it doesn't go through the `process_error` function on the schema and so doesn't get logged. The `process_error` function should be called whenever there is an error in subscriptions.

jkimbo:
Fixed by #1316

PR #1316, opened 2021-10-06
Log errors encountered during subscriptions

ossareh:
This refactors the `process_errors` call out of `schema.schema` and creates a mixin. This mixin is then worked into the websocket protocol handlers that recently landed and called from `schema.schema`.

## Description

I took the decision to move `process_errors` to a mixin because that felt the most correct way to handle it. Another option would have been to move it to `schema.base` since the main consumers of this function get a reference to a schema. I considered both options for a while.

Also, I took the option of marking `ExecutionContext` (arg to `process_errors`) as `Optional` since it is not (yet?) used in that function.

Finally, I didn't see a succinct way to write a test for this. Given the various cases that arise which result in errors I opted to punt on this. 

If any of the above decisions feel incorrect or icky lmk and I'll whip up a solution 👍

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* https://github.com/strawberry-graphql/strawberry/issues/1259

## Checklist

### DIFF
--- a/strawberry/utils/logging.py
+++ b/strawberry/utils/logging.py
@@ -2,17 +2,19 @@ import logging
 import sys
 from typing import Any, Optional
 
+from typing_extensions import Final
+
 from graphql.error import GraphQLError
 
 from strawberry.types import ExecutionContext
 
 
-logger = logging.getLogger("strawberry.execution")
-
-
 class StrawberryLogger:
-    @staticmethod
+    logger: Final[logging.Logger] = logging.getLogger("strawberry.execution")
+
+    @classmethod
     def error(
+        cls,
         error: GraphQLError,
         execution_context: Optional[ExecutionContext] = None,
         # https://www.python.org/dev/peps/pep-0484/#arbitrary-argument-lists-and-default-argument-values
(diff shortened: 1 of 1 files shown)


botberry:



---

Errors encountered in subscriptions will now be logged to the `strawberry.execution` logger as errors encountered in Queries and Mutations are. <3


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Michael Ossareh for the PR 👏

```

ossareh:
@jkimbo good question! This change added the error logger to two new places; the web socket implementations. It left in place the existing schema integration. So, at least in the latter case, one overrides it the same way they do now, i.e.: https://strawberry.rocks/docs/types/schema#handling-execution-errors

The web socket implementations could, instead of calling `strawberry.utils.logging.error_logger()` directly (as they do now) call `process_errors` on their `schema` member. To make that consistent we should move the definition of `process_errors` from `strawberry.schema.schema` to `strawberry.schema.base`. 

Another option would be to define the error logger in each module, and allow folks to monkey patch their own logger into it.

Ultimately I agree with the point you raised about moving errors reporting to a different part of the request cycle. I especially like the idea of errors being passed in to extensions' methods to allow us to, as an example, annotate tracing spans with the error information. I don't feel ready to bite off that level of work; there are many parts of the request cycle I do not feel well versed in currently :)

Would either of the other suggestions work better for you? Moving `process_errors` to `base.BaseSchema` would probably result in the most consistent behavior of the two options.

jkimbo:
> Do you see any issues with making Schema a subclass of BaseSchema? I was surprised that it wasn't already honestly.

Putting it on the `BaseSchema` makes much more sense. That was probably my fault when I created the initial PR for this.

Issue #160, opened 2019-09-26
Move CI to GitHub actions and add tests on windows 


patrick91:
There's a PR open for this, might be worth picking it up again: https://github.com/strawberry-graphql/strawberry/pull/148

So we can use GitHub Actions to run tests on windows as well as linux and Mac OS :)

TODO

- [x] Run release on github actions (currently it is on circle ci)
- [ ] Run tests on windows

zhidelev:
Hello, @patrick91 , I would like to work on this issue.  I decided to run tests on Windows and mypy tests failed because of the pytest_mypy plugin using https://pypi.org/project/capturer which does not support Windows. So I will think about a workaround for this case. Other test cases passed on Windows.

patrick91:
@zhidelev thanks! that's really appreciated!

patrick91:
Migration from circle ci to GitHub actions is done here: https://github.com/strawberry-graphql/strawberry/pull/954

marcoacierno:
The only thing left here is running tests on windows?

PR #1321, opened 2021-10-08
Support for Windows in CI

ossareh:
CI tests can now run on windows 🪟🎉

## Description

There are two types of issues tackled here:
 
1. Solutions for code that _doesn't work at all_ on windows
2. Solutions for code that _doesn't work well on_ windows

Code that doesn't work at all on windows has been circumvented; specifically I have added a pytestmark for `tests/pyright/test_*.py` which handles this.

Code that doesn't work well on windows turned out to be in and around the cli test server and its functions:

* introduced `pytest-xprocess` to manage starting the test server
* removed the use of our 🍓 emoji when on windows
* added a new `--log-level` option to `cli.commands.server`, which is passed to uvicorn.

**...but our 🍓 emoji...**

By default windows does not support UTF8 characters on the terminal. This was causing `subprocess.Popen` to fail, which unfortunately was running the code in the context manager, and returning a misleading result from requests; notably that the server was actively rejecting the request (there was no server to `recv()`!).

[1]. I opted to remove 🍓 when on windows. This test could be improved to instead determine whether printing the emoji could work, but I don't have a windows system to hand and verifying that via CI runs would take a while. Ultimately I would prefer a check early in the run process which advises a user to enable the right thing to make it work.

**new `--log-level` option to the server**

I found the best way to detect whether uvicorn had reloaded the schema was to inspect its log file. This in response to my thoughts that I documented here: https://github.com/strawberry-graphql/strawberry/pull/1321#issuecomment-939127034

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* Closes https://github.com/strawberry-graphql/strawberry/issues/160

## Checklist

### Notes

[1]. An earlier version of this description mentioned UTF8 issues in relation to registry updates. I realized later that I was conflating two issues and updated the copy above.

### DIFF
--- a/strawberry/cli/commands/server.py
+++ b/strawberry/cli/commands/server.py
@@ -15,7 +15,7 @@ from strawberry.utils.importer import import_module_symbol
 @click.option(
     "--log-level",
     default="error",
-    type=str,
+    type=click.Choice(["debug", "info", "warning", "error"], case_sensitive=False),
     help="passed to uvicorn to determine the log level",
 )
 @click.option(


botberry:



---

We now run our tests against Windows during CI!


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Michael Ossareh for the PR 👏

```

ossareh:
This is a good example of the type of failure that I've been tackling:

* Test starts ([link](https://github.com/strawberry-graphql/strawberry/pull/1321/checks?check_run_id=3842869396#step:10:179))
* pytest-xprocess detects server has started ([link](https://github.com/strawberry-graphql/strawberry/pull/1321/checks?check_run_id=3842869396#step:10:183))
* Subsequent request fails ([link](https://github.com/strawberry-graphql/strawberry/pull/1321/checks?check_run_id=3842869396#step:10:184))
  * Connection was reset ([link](https://github.com/strawberry-graphql/strawberry/pull/1321/checks?check_run_id=3842869396#step:10:1426))
  * While running this chunk of code ([link](https://github.com/strawberry-graphql/strawberry/pull/1321/checks?check_run_id=3842869396#step:10:2509))
  
So, in this case, it seems there is a raciness which may result in a network request being killed while attempting to verify the new schema has triggered a file reload by the dev server.

I'll noodle on the "right way" to detect this.

PR #1324, opened 2021-10-10
Fix bug with errors thrown in the on_parse_* extension hooks

jkimbo:
## Description

Before they would have been swallowed and just returned in the results
object. This change means that they get propagated up as you would
expect.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/strawberry/schema/execute.py
+++ b/strawberry/schema/execute.py
@@ -66,27 +66,27 @@ async def execute(
         # Note: In graphql-core the schema would be validated here but in
         # Strawberry we are validating it at initialisation time instead
 
-        try:
-            async with extensions_runner.parsing():
+        async with extensions_runner.parsing():
+            try:
                 if not execution_context.graphql_document:
                     execution_context.graphql_document = parse_document(query)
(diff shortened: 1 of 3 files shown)


strawberry/schema/execute.py
```diff
@@ -66,27 +66,27 @@ async def execute(
...
+
+            except Exception as error:  # pragma: no cover
+                error = GraphQLError(str(error), original_error=error)
+
+                execution_context.errors = [error]
+                return ExecutionResult(
+                    data=None,
+                    errors=[error],
+                    extensions=await extensions_runner.get_extensions_results(),
+                )
```
DoctorJohn, line 89: Noticed while investigating #3204:

Looks like this block should not have been moved or kept at all. Now it's catching `Exception`s raised by `parse_document`, which there shouldn't be any.

Before it was moved, it used to catch errors raised inside extensions during the parsing step and add those to the execution result. Such errors are usually progamming errors which we might prefer to cause an internal server error instead.

PR #1348, opened 2021-10-13
changed merge_type's types type hint, from Tuple[Type] to Tuple[Type, ...]

### LINKED ISSUE #1347: `merge_type` `types` type hint
https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/tools/merge_types.py#L9

The current `Tuple[Type]` produces:
```
*.py:15:5: error: Argument 2 to "merge_types" has incompatible type "Tuple[Type[QueryA], Type[QueryB], Type[QueryC]]"; expected "Tuple[Type[Any]]"  [arg-type]
```

According to [mypy](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#tuple-types), we should either change it to `Tuple[Type, ...]` or follow mypy's suggestion and go with a generic `Sequence`.

g-as:
## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

Resolves #1347 

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+This release fixes the `merge_types` type signature.


BryceBeagle:
> had to tell mypy to ignore `type._type_definitions.field`. Should there be a protocol to represent `strawberry.type` wrapped types?

In the future, we aim to stop monkeypatching `.type_definition` onto the wrapped type and instead have the function return entirely new `StrawberryObject` types. (When this happens, this function will no longer take `Tuple[type, ...]`, but instead `Tuple[StrawberryObject, ...]`.)

But for now, I think the mypy `type: ignore` comment you added is fine. Not sure why mypy didn't complain before this change though.

PR #1394, opened 2021-10-30
Fix generic types with lists

### LINKED ISSUE #1385: 'StrawberryList' object has no attribute '__name__'
The following code...

```python
T = TypeVar('T')

@strawberry.type
class Value(Generic[T]):
    valuation_date: datetime.date
    value: T

@strawberry.type
class Foo:
    strings: Value[List[str]]
```

...causes this error:

```
TypeError: Foo fields cannot be resolved. 'StrawberryList' object has no attribute '__name__'
```

If I change that to just `strings: Value[str]` it generates a `StrValue` type in the GraphQL schema.
`List[str]` results in a `[String!]` and doesn't have a name on its own. (Would probably be `StrList` if it needed a name?)
So I guess the correct behavior would be for `Value[List[str]]` to generate a `StrListValue` type?

patrick91:
This PR fixes #1385 as we were not checking for StrawberryOptional and StrawberryList when generating names from Generics :)

Closes #1385


botberry:



---

This release fixes an issue when creating concrete types from generic when passing list objects.


---




Here's the tweet text: 
```

```

jkimbo (approved):

tests/objects/test_generic_objects.py
```diff
@@ -123,6 +124,26 @@ class Connection(Generic[T]):
...
+    @strawberry.type
+    class Foo:
+        strings: Value[List[str]]
+
+    definition = Foo._type_definition
+    assert definition.name == "Foo"
+    assert not definition.is_generic
+    [field] = definition.fields
+    assert field.python_name == "strings"
+    assert field.type._type_definition.name == "StrValue"
```
bikeshedder, line 144: I would expect the name to be of type `StrListValue` since the name of `List[str]` should be `StrList`, no?

What would happen if I defined my `Foo` class like that:
```python
    @strawberry.type
    class Foo:
        strings: Value[List[str]]
        string: Value[str]
```

I guess both types `strings` and `string` be called `StrValue` then. I'm not a GraphQL expert but I'd expect this to cause problems, no?

bikeshedder:
I just created a PR for this branch #1395

tests/objects/test_generic_objects.py
bikeshedder: Fixed by #1395

botberry:
# Pre-release
:wave:

Releasing commit [9f974d81943b0638d3676b797f94c23e6eb63ffb] to PyPi as pre-release! :package:

botberry:
# Pre-release
:wave:

Pre-release **0.86.0-dev.1635977468** [9f974d81943b0638d3676b797f94c23e6eb63ffb] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.86.0-dev.1635977468
```

patrick91:
@bikeshedder I'll take some time to double check how we generate names, hope that's ok :) I want to sure we break the api only once 😊

bikeshedder:
I just checked this branch and found yet another issue with generics and `strawberry.scalar`. e.g.:

```python
JSON = strawberry.scalar(NewType("JSON", Any))
WrapperJSON = Wrapper[JSON]
```

That WrapperJSON should have the name WrapperJSON but it's just called Wrapper.

I started writing a unit test for it but since you're at it anyways maybe it's easier if you add it yourself. I've yet to understand the new generic tests.

bikeshedder (approved):

.github/workflows/test.yml
```diff
@@ -48,15 +48,15 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ${{ matrix.cache-path }}
-          key: ${{ runner.os }}-py-${{ matrix.python-version }}-poetry-${{ hashFiles('poetry.lock') }}
+          key: ${{ runner.os }}-py-${{ matrix.python-version }}-poetry-${{ hashFiles('poetry.lock') }}-2
```
patrick91, line 51: check if reverting this works

RELEASE.md
```diff
@@ -0,0 +1,8 @@
+Release type: minor
+
+This release fixes an issue when creating concrete types from generic when
+passing list objects.
+
+It also changes how type names are generated from generic types, now
+`Value[Optional[List[str]]]` generates `ValueOptionalListStr` instead of
+`StrListOptionalValue` which was unnatural to read.
```
patrick91, line 8: Update this now that we allow to customise type names

strawberry/schema/config.py
```diff
@@ -1,6 +1,17 @@
...
 
 @dataclass
 class StrawberryConfig:
-    auto_camel_case: bool = True
+    name_converter: NameConverter
+
+    def __init__(
+        self,
+        auto_camel_case: bool = True,
+        name_converter: NameConverter = None,
```
patrick91, line 15: we might deprecate `auto_camel_case` in future (for this class)

we should add a docstring here explaining that auto_camel_case is ignored if passing auto_camel_case

patrick91: or use post init

strawberry/type.py
```diff
@@ -108,11 +110,11 @@ def is_generic(self) -> bool:
 
 
 class StrawberryList(StrawberryContainer):
-    ...
+    name = "List"
```
patrick91, line 113: remove this and move it to name converter's generic name handling

strawberry/union.py
```diff
@@ -43,7 +43,7 @@ def __init__(
         type_annotations: Tuple["StrawberryAnnotation", ...] = tuple(),
         description: Optional[str] = None,
     ):
-        self._name = name
+        self.name = name
```
patrick91, line 46: rename this to `graphql_name`

tests/federation/test_schema.py
```diff
@@ -179,17 +179,17 @@ def top_products(self, first: int) -> ListOfProducts[Product]:
     assert not result.errors
 
     sdl = """
-        type Product {
-          upc: String!
+        type ListOfProductsProduct {
```
patrick91, line 182: maybe this should be `ProductListOfProducts` /cc @BryceBeagle

strawberry/schema/config.py
```diff
@@ -1,6 +1,18 @@
...
-    auto_camel_case: bool = True
+    auto_camel_case: InitVar[bool] = None  # type: ignore
+    name_converter: NameConverter = field(default_factory=NameConverter)
+
+    def __post_init__(
+        self,
+        auto_camel_case: bool,
+    ):
+        if auto_camel_case is not None:
+            self.name_converter.auto_camel_case = auto_camel_case
```
patrick91, line 18: @BryceBeagle can you check this? 😊

tests/objects/generics/test_names.py
```diff
@@ -0,0 +1,52 @@
...
+    "types,expected_name",
+    [
+        ([StrawberryList(str)], "ListStrExample"),
+        ([StrawberryList(StrawberryList(str))], "ListListStrExample"),
+        ([StrawberryList(Enum)], "ListEnumExample"),
+        ([StrawberryUnion("Union", (TypeA, TypeB))], "UnionExample"),  # type: ignore
+        ([TypeA], "TypeAExample"),
+        ([CustomInt], "CustomIntExample"),
+        ([TypeA, TypeB], "TypeATypeBExample"),
+        ([TypeA, LazyType["TypeB", "test_names"]], "TypeATypeBExample"),  # type: ignore
```
patrick91, line 40: @BryceBeagle I've changed the naming to be the old one, so now we generate names like so:

`Paginated[Person]` -> `PersonPaginated`

Which makes sense as we are putting related types close to each other (at least alphabetically). Person always lives near PersonPaginated.

When using optional and string we get this at the moment:

`Paginated[Optional[Person]]` -> `OptionalPersonPaginated`
`Paginated[List[Person]]` -> `ListPersonPaginated`

shall we change it to be:

`Paginated[Optional[Person]]` -> `PersonOptionalPaginated`
`Paginated[List[Person]]` -> `PersonListPaginated`

?

patrick91: updated :)

patrick91:
I've tested this on Pollen's repo and it works fine 😊

tests/objects/generics/test_names.py
bikeshedder: Good thing you decided against mixing prefix and postfix name generation. Otherwise `Optional[List[str]]` and `List[Optional[str]]` would result in the same type name being generated.

patrick91: uh, I should have added a test for this!

PR #1395, opened 2021-10-30
Fix type name generation for List[T]

bikeshedder:


bikeshedder:
PRs for PRs are weird. :rofl:

strawberry/types/types.py
```diff
@@ -120,6 +125,8 @@ def get_name_from_type(self, type_: Union[StrawberryType, type]) -> str:
             return type_.name
         elif isinstance(type_, StrawberryUnion):
             return type_.name
+        elif isinstance(type_, StrawberryList):
+            return self.get_name_from_type(type_.of_type) + "List"
```
bikeshedder, line 129: I wonder if there are other `StrawberryContainer` subclasses that require a similiar treatment.

Maybe it would be better to add some kind of `name` property/function to the container type itself so it is in one place and not scattered over the code base?

patrick91: that’s a good point, for now I think we only have List and Optional as containers :)

Can you add that too? ^^

tests/objects/test_generic_objects.py
```diff
@@ -136,15 +136,26 @@ class Value(Generic[T]):
...
+        optional_strings_field,
+    ] = definition.fields
     assert string_field.python_name == "string"
     assert string_field.type._type_definition.name == "StrValue"
     assert strings_field.python_name == "strings"
     assert strings_field.type._type_definition.name == "StrListValue"
+    assert optional_string_field.python_name == "optional_string"
+    assert optional_string_field.type._type_definition.name == "StrOptionalValue"
+    assert optional_strings_field.python_name == "optional_strings"
+    assert optional_strings_field.type._type_definition.name == "StrListOptionalValue"
```
bikeshedder, line 158: The name of this is weird. A human would probably call it a `OptionalStrListValue` but generating it this way could introduce ambiguous type names. e.g.`Optional[Value[List[str]]]` and `Value[List[Optional[str]]]` would result in the same type name being generated.

strawberry/types/types.py
bikeshedder: Done. :+1:

tests/objects/test_generic_objects.py
BryceBeagle: This could potentially be solved by having the name generate inside->out instead of outside->in, _except_ for the outermost element, which would go at the end. This would result in:
- `Value[str]` -> `StrValue`
- `Value[List[str]]` -> `ListStrValue`
- `Value[Optional[str]]` -> `OptionalStrValue`
- `Value[Optional[List[str]]]` -> `OptionalListStrValue`

bikeshedder: I'd rather leave it all `inside->out` or all `outside->in` as you already show some weird behavior otherwise. Why is a `Value` of `StrList` called `ListStrValue`. That doesn't make much sense.

tbh. maybe it makes sense to change this to a all `inside->out` name generation in the future?

- Value[str] -> ValueStr
- Value[List[str]] -> ValueListStr
- Value[Optional[str]] -> ValueOptionalStr
- Value[Optional[List[str]]] -> ValueOptionalListStr

BryceBeagle: > Why is a Value of `StrList` called `ListStrValue`

I think that makes sense, actually. In English, adjectives usually go before the noun. Consider "green tree". It's a `tree` of type `green`. Same with `StrListValue`. It's a `Value` of type `StrList`.

bikeshedder: `ListStrValue != `StrListValue`

I had to read your comment several times to see your mistake. You misread my comment.

I guess we can all agree that either  all `inside->out` or `outside->in` makes sense. Mixing those two just gets confusing a lot as you just showed as well. :grinning: :+1:

patrick91: I'd say let's do this:

```
Value[str] -> ValueStr
Value[List[str]] -> ValueListStr
Value[Optional[str]] -> ValueOptionalStr
Value[Optional[List[str]]] -> ValueOptionalListStr
```

as you suggested @bikeshedder 

I'll open an issue later to allow us to override the names for generic (maybe using Annotated).

Let me know if you can do this @bikeshedder, otherwise I'll try to do it tonight (I'm on holidays so I'm not using my laptop often :))

patrick91: ok, did the change, I want to get @jkimbo's and @AndrewIngram's opinion too, as this is now a breaking change as it changed all the concrete types made from generics, which might be annoying for people using generics.

Generics aren't documented yet, but this might still affect quite a few people

bikeshedder: I'm happy with both styles. Though I have to agree that the `outside->in` order makes it way easier to reason about since the generic reads the same as the generated name just leaving out the square brackets.

jkimbo: I think this makes sense @patrick91 👍

Issue #1405, opened 2021-11-05
No way to resolve type of non-strawberry types returned for an interface


AlecRosenbaum:
This comment is correct: https://github.com/strawberry-graphql/strawberry/blob/feb6edee0df8cac3aeb663f67a169035a0271f40/strawberry/schema/schema_converter.py#L235-L244

It's correctly [reflected in the docs](https://strawberry.rocks/docs/types/interfaces#resolving-an-interface), but still does pose a pretty bad limitation for implementations that resolve to non-strawberry types.

The following test fails (using a similar example to https://github.com/strawberry-graphql/strawberry/pull/1150):

```python
def test_interface_type_resolution():
    @strawberry.interface
    class Node:
        id: int

    @strawberry.type
    class Anime(Node):
        name: str

    @dataclass
    class AnimeORM:
        id: int
        name: str

    @strawberry.type
    class Query:
        @strawberry.field
        def node(self) -> Node:
            return AnimeORM(id=1, name="One Piece")  # type: ignore

    schema = strawberry.Schema(query=Query)

    query = "{ node { __typename, id } }"
    result = schema.execute_sync(query)

    assert not result.errors
    assert result.data == {"node": {"__typename": "Anime", "id": 1}}
    # errors with: "type object 'AnimeORM' has no attribute '_type_definition'"
```

In `graphene`, the converter allows you to define `is_type_of` classmethods on concrete types or a `resolve_type` classmethod on the interface. I think the equivalent within strawberry would be to optionally include `is_type_of` [here](https://github.com/strawberry-graphql/strawberry/blob/feb6edee0df8cac3aeb663f67a169035a0271f40/strawberry/schema/schema_converter.py#L300-L305), and to gracefully fail type resolution for non-strawberry objects (by returning `None`) within [`resolve_type`](https://github.com/strawberry-graphql/strawberry/blob/feb6edee0df8cac3aeb663f67a169035a0271f40/strawberry/schema/schema_converter.py#L235-L248).

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/strawberry-graphql) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/strawberry-graphql/strawberry/issues/1405">
<picture>
[...]

PR #1406, opened 2021-11-05
Implement explicit interface type resolution

AlecRosenbaum:
## Description

The behavioral changes included in this PR are:
a) duck typing on fields that return interfaces or types implementing an interface now requires an explicit `is_type_of` opt-in
b) adding an explicit `is_type_of` opt-in is now implemented with a `is_type_of` classmethod

This seems like it might require a change to the documentation mentioning the classmethod, which I can make assuming this approach looks good.

Also, what release type should this be? Technically it is a breaking change, though it's quite minor.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* https://github.com/strawberry-graphql/strawberry/issues/1405

## Checklist

I did dev work within a codespace based on your template (which was very convenient), so at the very least all the pre-commit scripts ran correctly. I also ran the test suite, which passes.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,9 @@
+Release type: patch
+
+This releases fixes an issue where you were not allowed
+to return a non-strawberry type for fields that return
+and interface. Now this works as long as each type
+implementing the interface implements an `is_type_of`
+classmethod. Previous automatic duck typing on types
+that implement an interface now requires explicit
+resolution using this classmethod.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,7 +2,7 @@ Release type: patch
 
 This releases fixes an issue where you were not allowed
 to return a non-strawberry type for fields that return
-and interface. Now this works as long as each type
+an interface. Now this works as long as each type
 implementing the interface implements an `is_type_of`
 classmethod. Previous automatic duck typing on types
 that implement an interface now requires explicit
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,4 +1,4 @@
-Release type: patch
+Release type: minor
 
 This releases fixes an issue where you were not allowed
 to return a non-strawberry type for fields that return
--- a/strawberry/experimental/pydantic/object_type.py
+++ b/strawberry/experimental/pydantic/object_type.py
@@ -147,10 +147,17 @@ def type(
 
         sorted_fields = missing_default + has_default
 
+        # Implicitly define `is_type_of` to support interfaces/unions that use
+        # pydantic objects (not the corresponding strawberry type)
+        @classmethod
+        def is_type_of(cls, obj, _info) -> bool:
+            return isinstance(obj, (cls, model))
+
         cls = dataclasses.make_dataclass(
             cls.__name__,
             sorted_fields,
             bases=cls.__bases__,
+            namespace={"is_type_of": is_type_of},
         )
 
         _process_type(
--- a/strawberry/experimental/pydantic/object_type.py
+++ b/strawberry/experimental/pydantic/object_type.py
@@ -7,6 +7,8 @@ from typing import Any, Dict, List, Optional, Sequence, Tuple, Type, cast
 from pydantic import BaseModel
 from pydantic.fields import ModelField
 
+from graphql import GraphQLResolveInfo
+
 import strawberry
 from strawberry.arguments import UNSET
 from strawberry.experimental.pydantic.conversion import (
@@ -149,8 +151,8 @@ def type(
 
         # Implicitly define `is_type_of` to support interfaces/unions that use
         # pydantic objects (not the corresponding strawberry type)
-        @classmethod
-        def is_type_of(cls, obj, _info) -> bool:
+        @classmethod  # type: ignore
+        def is_type_of(cls: Type, obj: Any, _info: GraphQLResolveInfo) -> bool:
             return isinstance(obj, (cls, model))
 
         cls = dataclasses.make_dataclass(
--- a/strawberry/object_type.py
+++ b/strawberry/object_type.py
@@ -106,6 +106,7 @@ def _process_type(
 
     interfaces = _get_interfaces(cls)
     fields = _get_fields(cls)
+    is_type_of = getattr(cls, "is_type_of", None)
 
     cls._type_definition = TypeDefinition(
         name=name,
@@ -117,6 +118,7 @@ def _process_type(
         origin=cls,
         extend=extend,
         _fields=fields,
+        is_type_of=is_type_of,
     )
 
     # dataclasses removes attributes from the class here:
--- a/strawberry/schema/schema_converter.py
+++ b/strawberry/schema/schema_converter.py
@@ -283,8 +283,10 @@ class GraphQLCoreConverter:
         if object_type.is_type_of:
             is_type_of = object_type.is_type_of
         elif object_type.interfaces:
+
             def is_type_of(obj: Any, _info: GraphQLResolveInfo) -> bool:
                 return isinstance(obj, object_type.origin)
+
         else:
             is_type_of = None
 
--- a/strawberry/schema/schema_converter.py
+++ b/strawberry/schema/schema_converter.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from enum import Enum
-from typing import Any, Callable, Dict, List, Tuple, Type, Union, cast
+from typing import Any, Callable, Dict, List, Optional, Tuple, Type, Union, cast
 
 from graphql import (
     GraphQLArgument,
@@ -229,27 +229,11 @@ class GraphQLCoreConverter:
 
             return graphql_fields
 
-        def resolve_type(
-            obj: Any,
-            info: GraphQLResolveInfo,
-            type_: Union[GraphQLInterfaceType, GraphQLUnionType],
-        ) -> GraphQLObjectType:
-            # TODO: this will probably break when passing dicts
-            # or even non strawberry types
-            resolved_type = self.type_map[
-                obj.__class__._type_definition.name
-            ].implementation
-
-            assert isinstance(resolved_type, GraphQLObjectType)
-
-            return resolved_type
-
         graphql_interface = GraphQLInterfaceType(
             name=interface_name,
             fields=get_graphql_fields,
             interfaces=list(map(self.from_interface, interface.interfaces)),
             description=interface.description,
-            resolve_type=resolve_type,
         )
 
         self.type_map[interface_name] = ConcreteType(
@@ -295,11 +279,23 @@ class GraphQLCoreConverter:
 
             return graphql_fields
 
+        is_type_of: Optional[Callable[[Any, GraphQLResolveInfo], bool]]
+        if object_type.is_type_of:
+            is_type_of = object_type.is_type_of
+        elif object_type.interfaces:
+
+            def is_type_of(obj: Any, _info: GraphQLResolveInfo) -> bool:
+                return isinstance(obj, object_type.origin)
+
+        else:
+            is_type_of = None
+
         graphql_object_type = GraphQLObjectType(
             name=object_type_name,
             fields=get_graphql_fields,
             interfaces=list(map(self.from_interface, object_type.interfaces)),
             description=object_type.description,
+            is_type_of=is_type_of,
         )
 
         self.type_map[object_type_name] = ConcreteType(
--- a/strawberry/schema/schema_converter.py
+++ b/strawberry/schema/schema_converter.py
@@ -283,10 +283,8 @@ class GraphQLCoreConverter:
         if object_type.is_type_of:
             is_type_of = object_type.is_type_of
         elif object_type.interfaces:
-
             def is_type_of(obj: Any, _info: GraphQLResolveInfo) -> bool:
                 return isinstance(obj, object_type.origin)
-
         else:
             is_type_of = None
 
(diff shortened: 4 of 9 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Alec Rosenbaum for the PR 👏

```

AlecRosenbaum:
I've just rebased/force pushed to resolve some conflicts with `main`. None of the content should have changed, it was just imports.

patrick91:
@AlecRosenbaum thanks! does this work for unions too?

I'll take a proper look during the weekend 😊

AlecRosenbaum:
It does now 🙂

RELEASE.md
```diff
@@ -0,0 +1,9 @@
+Release type: patch
```
```suggestion
Release type: minor
```

I think this deserves a minor release as it is a brand new functionality ^^

strawberry/object_type.py
```diff
@@ -117,6 +118,7 @@ def _process_type(
         origin=cls,
         extend=extend,
         _fields=fields,
+        is_type_of=is_type_of,
```
patrick91, line 121: I think we should do this for pydantic too :)

strawberry/schema/schema_converter.py
```diff
@@ -295,11 +279,23 @@ def get_graphql_fields() -> Dict[str, GraphQLField]:
 
             return graphql_fields
 
+        is_type_of: Optional[Callable[[Any, GraphQLResolveInfo], bool]]
+        if object_type.is_type_of:
+            is_type_of = object_type.is_type_of
+        elif object_type.interfaces:
+
```
patrick91, line 286:
```suggestion
```

strawberry/schema/schema_converter.py
```diff
@@ -295,11 +279,23 @@ def get_graphql_fields() -> Dict[str, GraphQLField]:
...
+            def is_type_of(obj: Any, _info: GraphQLResolveInfo) -> bool:
+                return isinstance(obj, object_type.origin)
+
```
patrick91, line 289:
```suggestion
```

strawberry/schema/schema_converter.py
```diff
@@ -295,11 +279,23 @@ def get_graphql_fields() -> Dict[str, GraphQLField]:
 
             return graphql_fields
 
+        is_type_of: Optional[Callable[[Any, GraphQLResolveInfo], bool]]
```
patrick91, line 282: `GraphQLResolveInfo` <- we should use StrawberryInfo, not sure we can do this easily here though, maybe we can do it after we do this issue: https://github.com/strawberry-graphql/strawberry/issues/1425

strawberry/union.py
```diff
@@ -135,9 +135,16 @@ def _resolve_union_type(
...
-            # Make sure that the type that's passed in is an Object type
+            # If the type given is not an Object type, try resolving using `is_type_of`
+            # defined on the union's inner types
             if not hasattr(root, "_type_definition"):
-                # TODO: If root=python dict, this won't work
+                for inner_type in type_.types:
+                    if inner_type.is_type_of is None:
+                        break
+                    elif inner_type.is_type_of(root, info):
+                        return inner_type.name
```
patrick91, line 145: could we simplify this to:
```suggestion
                    if inner_type.is_type_of is not None and inner_type.is_type_of(root, info):
                        return inner_type.name
```
?

tests/schema/test_interface.py
```diff
@@ -107,6 +107,10 @@ class Entity:
     class Anime(Entity):
         name: str
 
+        @classmethod
+        def is_type_of(cls, obj, _info) -> bool:
+            return isinstance(obj, AnimeORM)
```
patrick91, line 112: is it required now? 🤔

patrick91 (commented):
This looks great! ✨ I've left a few comments. Before merging this I'd like to get @BryceBeagle and @jkimbo's opinions on the API:

with this PR we allow to pass a classmethod `is_type_of` like so:

```python
@strawberry.type
 class Anime(Node):
     name: str

     @classmethod
     def is_type_of(cls, obj, _info) -> bool:
         return isinstance(obj, AnimeORM)
```

which is a bit different that what we do at the moment, I mean, at the moment
we pass all the configuration options to `@strawberry.type`.

I'm fine with having `is_type_of` being a classmethod, but I want to see what you all think :)

@alecrosenbaum we should add some documentation too :)

strawberry/schema/schema_converter.py
AlecRosenbaum: I think this spacing came from black

tests/schema/test_interface.py
AlecRosenbaum: Yeah, with this change you'd have to explicitly resolve the type for non-strawberry types 🙁

I wasn't able to figure out how to support a way that duck typing works implicitly.

Having any `is_type_of` set means that [this branch](https://github.com/graphql-python/graphql-core/blob/f145dc9d0ceb6943b8adcbf86ece3c6403b34bb3/src/graphql/execution/execute.py#L915-L935) is taken within `graphql-core` while executing. It does extra validation on the return type and would now error, since the implicit `is_type_of` definition only works with strawberry types.

Previously, `is_type_of` was `None`. If we change the implicit `is_type_of` definition to `None` instead of checking against the strawberry type, the duck typing test passes (since `Anime` is an object type and not an interface) but the above tests with fields that return interfaces begin to fail (since they can't be resolved to object types anymore).

Supporting both usage patterns I think would mean only defining `resolve_type` on the interface if none of the implementing types have an `is_type_of` defined. That's kind of hard though, since the object types are defined after the interface and I didn't see a good way to un-define `resolve_type` after the fact. Having a callable there at all changes the behavior and makes it ignore all the `is_type_of` definitions on the object type.

strawberry/object_type.py
AlecRosenbaum: Done in [`fdc290a` (#1406)](https://github.com/strawberry-graphql/strawberry/pull/1406/commits/fdc290a53088969581c7dffd8313e57bf3247ef3)

strawberry/experimental/pydantic/object_type.py
```diff
@@ -147,10 +149,17 @@ def wrap(cls):
 
         sorted_fields = missing_default + has_default
 
+        # Implicitly define `is_type_of` to support interfaces/unions that use
+        # pydantic objects (not the corresponding strawberry type)
+        @classmethod  # type: ignore
```
AlecRosenbaum, line 154: Unfortunately, `# type: ignore` is the suggested solution here: https://github.com/python/mypy/issues/4153

tests/schema/test_interface.py
patrick91: thanks for the in-depth explanation :) I'll think about this in the next few days and see if we can find a solution, otherwise this becomes a big breaking change

I guess we can automatically provide a `is_type_of` for interfaces maybe (if none is defined)

BryceBeagle:
If we're having users implement a function to indicate duck typing is possible... why not just have them implement a function that does the cast/conversion instead? I'm not sure we want to go this deep into supporting duck-typing

patrick91:
> If we're having users implement a function to indicate duck typing is possible... why not just have them implement a function that does the cast/conversion instead? I'm not sure we want to go this deep into supporting duck-typing

This is something I used to do at my previous job, conversion requires a lot of typing and I don't think it's worth the trouble in all cases, if you think a strawberry object as a protocol it would make sense being allowed to return any type that conform the protocol :)

Casting might be an option, but I don't really know how we can do that (other than using some hacks). The other issue with it is that you'd require to do it in every resolver. 

We already support duck typing by the way, it's just not well supported when using interfaces/unions (since GraphQL needs to know the actual return type)

jkimbo:
> conversion requires a lot of typing and I don't think it's worth the trouble in all cases

@patrick91 Can you give an example of this? In my experience conversion doesn't require any extra typing.

patrick91:
> > conversion requires a lot of typing and I don't think it's worth the trouble in all cases
> 
> @patrick91 Can you give an example of this? In my experience conversion doesn't require any extra typing. 

Oh I meant writing code, not type hints :D

```python
@strawberry.type
class User: 
    name: str

    @classmethod
    def from_model(cls, model) -> User:
        return User(name=model.name)
```

imagine this with 10 fields :)

jkimbo:
> Oh I meant writing code, not type hints :D
> 
> ```python
> @strawberry.type
> class User: 
>     name: str
> 
>     @classmethod
>     def from_model(cls, model) -> User:
>         return User(name=model.name)
> ```
> 
> imagine this with 10 fields :)

Ok yeah that does make sense. IMO I still prefer that to duck typing but I can appreciate why it's painful.

Separately I've just discovered that graphql-core will check for a `__typename` attribute on a value to map an abstract type (union or interface) to a concrete type: https://github.com/graphql-python/graphql-core/blob/v3.1.6/src/graphql/execution/execute.py#L1254-L1272

Could that be a better way of dealing with this?

patrick91:
> Ok yeah that does make sense. IMO I still prefer that to duck typing but I can appreciate why it's painful.

I feel like there might be cases where you're dealing with dictionaries coming from something like mongodb or an API. I don't think we should force people to convert them to instances of Strawberry objects. 

In addition to that there's also other usecases, like getting partial objects. For example:

```python
@strawberry.type
class User:
    name: str 
    age: int

@strawberry.type
class Query:
    @strawberry.field
    def user(self, info: Info) -> User:
        selected_fields = fields_from_info(info)
        return models.User.first().only(selected_fields)
```

This can't be done at all with plain Strawberry Objects, at least for now 😊

That's why I think we should allow duck typing, especially to people that know what they are doing :)

> Separately I've just discovered that graphql-core will check for a __typename attribute on a value to map an abstract type (union or interface) to a concrete type: graphql-python/graphql-core@v3.1.6/src/graphql/execution/execute.py#L1254-L1272

This is neat! Didn't know that 😊 not sure it would be the solution, as it still requires doing some work in the resolvers, see this ugly example here:

```python
@strawberry.type
class Query:
    @strawberry.field
    def user(self, info: Info) -> SomeInterface:
        items = SomePolymorficModel.all()[:10]
        return add_type_name_to_each_item(items)
```

I think having a `is_type_of` function is still preferable to this, especially in context of using other frameworks with Strawberry (django, pydantic, etc)

Sorry if it looks like I'm insisting on this, I'm not 100% liking the idea of having to defined a classmethod on object types, but I'm trying to see it as a potential user of strawberry, and I wouldn't be happy to force to convert compatibile types to other types 😊

BryceBeagle:
Dictionaries are one thing, but I don't think we should be encouraging duck typing between objects between separate libraries (e.g. an ORM and Strawberry). This seems incredibly fragile and prone to breakage/unexpected failure. Feels way too javascript-y to me. At minimum, I think we should have a "strict" mode toggle that disables/enables this loose (and imo dangerous) behavior.

patrick91:
> Dictionaries are one thing
what makes a dictionary special? I'd argue it is better to support duck typing rather than supporting dictionaries :)


> but I don't think we should be encouraging duck typing between objects between separate libraries (e.g. an ORM and Strawberry). 
We don't encourage, but we shouldn't disallow it. People will want to be able to do that, and again I think we should treat GraphQL types are protocols more than strict classes. 

> This seems incredibly fragile and prone to breakage/unexpected failure. 
We should make errors clearer then, type checkers already help with this :) I've personally used duck typing without any issue ☺️

> Feels way too javascript-y to me. At minimum, I think we should have a "strict" mode toggle that disables/enables this loose (and imo dangerous) behavior.
Would this run isinstance on all returned values?

AlecRosenbaum:
What are the next steps for this PR then?

I'm happy to implement an alternative solution as long as it would address the issue of resolving interfaces on duck-typed objects. At the very least, not supporting this would be a big pain point in our gradual transition away from `graphene`, where we already have a relatively large schema that is 100% reliant on this pattern. We currently have grahpene/strawberry type definitions intermixed but all resolved using duck typing.

jkimbo:
I'm happy for this to be merged. I still have reservations about the api but I also don't have a better solution so we should merge it since it's blocking @AlecRosenbaum

jkimbo (approved):

patrick91:
I think we only need to prevent from having to always pass `is_type_of` now :) https://github.com/strawberry-graphql/strawberry/pull/1406#discussion_r755453955

tests/schema/test_interface.py
AlecRosenbaum: There is one that's automatically defined to resolve strawberry types correctly, it just doesn't handle duck types:

https://github.com/strawberry-graphql/strawberry/pull/1406/files#diff-9db4ecb0b6a9104731da4af6d847961cbc88e4085f4f23f4fcfceb1b22e2fd35R287-R288

```python
def is_type_of(obj: Any, _info: GraphQLResolveInfo) -> bool:
    return isinstance(obj, object_type.origin)
```

Currently the other test cases in this file pass without an explicit classmethod, since they all only use strawberry types.

How would you determine whether a non-strawberry object matches the strawberry type without any extra information from the implementer?

patrick91: > How would you determine whether a non-strawberry object matches the strawberry type without any extra information from the implementer? 

I think I did misunderstand your comment then! I thought it was always required!

Cool! I'll take a final look tonight and I'll merge ^^

patrick91 (approved):
Thank you for doing this PR, and sorry it took a while to get it merged 😊

botberry:

PR #1430, opened 2021-11-19
Allow using a static method as a resolver

illia-v:
## Description

In a project I had to define a resolver that gets some data from the root argument.
It was not possible to use `self` because of some mypy warnings. Therefore, I wanted to remove the `self` argument.
It was possible to remove it without using `@staticmethod`, but mypy complained that `Self argument missing for a non-static method (or an invalid type for self)`.

So, I had to add `@staticmethod`.
I tried adding it before `@strawberry.field`, but Python did not add the field to `__dataclass_fields__` and a type was not created correctly.
Then, I tried adding it after `@strawberry.field`, but Strawberry did not like that because `AttributeError: 'staticmethod' object has no attribute '__annotations__'`. The changes fix this problem.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,6 +1,6 @@
 Release type: patch
 
-Allow using staticmethods and classmethods as resolvers
+This release fixes an issue that prevented using classmethod and staticmethods as resolvers
 
 ```python
 import strawberry
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,7 +1,6 @@
 Release type: patch
 
-Allow using a static method as a resolver when the `self` argument is not
-needed.
+Allow using a staticmethods and classmethods as resolvers
 
 ```python
 import strawberry
@@ -12,4 +11,9 @@ class Query:
     @staticmethod
     def static_text() -> str:
         return "Strawberry"
+
+    @strawberry.field
+    @classmethod
+    def class_name(cls) -> str:
+        return cls.__name__
 ```
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,15 @@
+Release type: patch
+
+Allow using a static method as a resolver when the `self` argument is not
+needed.
+
+```python
+import strawberry
+
+@strawberry.type
+class Query:
+    @strawberry.field
+    @staticmethod
+    def static_text() -> str:
+        return "Strawberry"
+```
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,6 +1,6 @@
 Release type: patch
 
-This release fixes an issue that prevented using classmethod and staticmethods as resolvers
+This release fixes an issue that prevented using `classmethod`s and `staticmethod`s as resolvers
 
 ```python
 import strawberry
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,6 +1,6 @@
 Release type: patch
 
-Allow using a staticmethods and classmethods as resolvers
+Allow using staticmethods and classmethods as resolvers
 
 ```python
 import strawberry
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -37,7 +37,7 @@ if TYPE_CHECKING:
     from .object_type import TypeDefinition
 
 
-_RESOLVER_TYPE = Union[StrawberryResolver, Callable]
+_RESOLVER_TYPE = Union[StrawberryResolver, Callable, staticmethod]
 
 
 class StrawberryField(dataclasses.Field):
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -37,7 +37,7 @@ if TYPE_CHECKING:
     from .object_type import TypeDefinition
 
 
-_RESOLVER_TYPE = Union[StrawberryResolver, Callable, staticmethod]
+_RESOLVER_TYPE = Union[StrawberryResolver, Callable, staticmethod, classmethod]
 
 
 class StrawberryField(dataclasses.Field):
@@ -48,7 +48,7 @@ class StrawberryField(dataclasses.Field):
         python_name: Optional[str] = None,
         graphql_name: Optional[str] = None,
         type_annotation: Optional[StrawberryAnnotation] = None,
-        origin: Optional[Union[Type, Callable]] = None,
+        origin: Optional[Union[Type, Callable, staticmethod, classmethod]] = None,
         is_subscription: bool = False,
         description: Optional[str] = None,
         base_resolver: Optional[StrawberryResolver] = None,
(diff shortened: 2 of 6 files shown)


strawberry/field.py
```diff
@@ -37,7 +37,7 @@
     from .object_type import TypeDefinition
 
 
-_RESOLVER_TYPE = Union[StrawberryResolver, Callable]
+_RESOLVER_TYPE = Union[StrawberryResolver, Callable, staticmethod]
```
patrick91, line 40: is staticmethod a type? :O

tests/fields/test_resolvers.py
```diff
@@ -47,6 +47,25 @@ def name(self) -> str:
...
+
+    definition = Query._type_definition
+
+    assert definition.name == "Query"
+    assert len(definition.fields) == 1
+
+    assert definition.fields[0].python_name == "name"
+    assert definition.fields[0].graphql_name is None
+    assert definition.fields[0].type == str
+    assert definition.fields[0].base_resolver(None) == Query().name()
```
patrick91, line 66: should we add a test for this inside schema? so we check that the execution works fine?

strawberry/field.py
illia-v: Yes, `staticmethod` is a class and its objects are not `Callable`

patrick91: oh, interesting! thanks 😊

tests/fields/test_resolvers.py
illia-v: > should we add a test for this inside schema? so we check that the execution works fine?

Good idea, I added one in b1e3284757593b1bf2799f7fc1c5a5beff374893.

illia-v: It looks like this test fails on this line because of an existing bug: `StrawberryResolver.__call__` does not expect that `self.wrapped_func` can have no `self` argument.

```python
tests/fields/test_resolvers.py:66: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <strawberry.types.fields.resolver.StrawberryResolver object at 0x7fab13d045e0>, args = (None,), kwargs = {}

    def __call__(self, *args, **kwargs) -> T:
>       return self.wrapped_func(*args, **kwargs)
E       TypeError: test_staticmethod_resolver_fields.<locals>.Query.name() takes 0 positional arguments but 1 was given

strawberry/types/fields/resolver.py:39: TypeError
```

illia-v: The same error can be reproduced even without `staticmethod`:
```python
@strawberry.type
class Query:
    @strawberry.field
    def name() -> str:
        return "Name"

definition = Query._type_definition
assert definition.fields[0].base_resolver(None) == Query().name()
```

patrick91: interesting! I'll take a look at it tomorrow. /cc @BryceBeagle

strawberry/field.py
BryceBeagle: > its objects are not `Callable`

I don't think this is true:
```pycon
>>> class Blah:
...     @staticmethod
...     def foo(): ...
... 
>>> callable(Blah.foo)
True
>>> callable(Blah().foo)
True
```

Mypy is perfectly happy with `staticmethod`s as `Callable` too:
https://mypy-play.net/?mypy=latest&python=3.10&gist=c907cc3700c23323978d2ec618a724dc

BryceBeagle: Oh, I think I understand. The issue is during class evaluation, where the function is actually an instance of `staticmethod` and not just of `function`

illia-v: > > its objects are not `Callable`
> 
> I don't think this is true:

Looks like static methods were made callables in 3.10: https://bugs.python.org/issue43682.

I tested in this way:
```python
>>> from collections.abc import Callable
>>> isinstance(staticmethod(int), Callable)
```

This returns `True` for Python 3.10 and `False` for previous versions.

tests/fields/test_resolvers.py
BryceBeagle: I'm trying to remember/investigate _why_ this is the case, but this is because `Query.name` != `Query().name`. I think this was something we wanted to get around to fixing at some point

patrick91: @BryceBeagle what was the resolution of this? can we merge it? :)

BryceBeagle:
I have a more robust fix for this that also supports `classmethod`s working right now :+1:. Just need to get mypy checks to comply. Hopefully we can get this merged by tonight

BryceBeagle:
After much deep-diving into (and frustration with!) the underlying mechanics of `staticmethod` and `classmethod` binding, I _think_ this solution should be very robust and satisfy most, if not all, edge cases. I took the liberty of committing to your branch @illia-v. 

This should be ready for review @patrick91. Let me know if I should add any more comments/documentation anywhere.

RELEASE.md
```diff
@@ -0,0 +1,19 @@
+Release type: patch
+
+Allow using staticmethods and classmethods as resolvers
```
```suggestion
This release fixes an issue that prevented using classmethod and staticmethods as resolvers
```

strawberry/types/fields/resolver.py
```diff
@@ -1,10 +1,10 @@
-from __future__ import annotations
+from __future__ import annotations as _
```
patrick91, line 1: is this done on purpose? :)

tests/fields/test_resolvers.py
```diff
@@ -47,6 +50,52 @@ def name(self) -> str:
...
+
+def test_classmethod_resolver_fields():
+    @strawberry.type
+    class Query:
+        my_val: ClassVar[str] = "thingy"
+
+        @strawberry.field
+        @classmethod
+        def val(cls) -> str:
+            return cls.my_val
```
patrick91, line 83: shall we add tests for staticmethods and classmethods defined on other classes? I know they work, but it is worth testing them 😊

strawberry/types/fields/resolver.py
BryceBeagle: Yeah, I defined a method-scope variable called `annotations` and my linter was complaining about symbol shadowing

tests/fields/test_resolvers.py
BryceBeagle: We already have one for `classmethod`s:
 https://github.com/strawberry-graphql/strawberry/blob/c5a1b6154810d2c186d916346d1ef77578060fc4/tests/schema/test_resolvers.py#L205

patrick91: let's do one for static methods too :D

patrick91 (approved):

botberry:

PR #1438, opened 2021-11-23
Improve type checker support for return value for strawberry.union

patrick91:
This PR improves support of strawberry.union types for type checkers. It basically tells we are returning a Union type :)

~~I need to check if this works on MyPy, as I only tested Pyright at the moment :)~~

We have a plugin for MyPy so we are good :)

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,18 @@
+Release type: minor
+
+This release improves type checking support for `strawberry.union` and now allows
+to use unions without any type issues, like so:
+
+```python
+@strawberry.type
+class User:
+    name: str
+
+@strawberry.type
+class Error:
+    message: str
+
+UserOrError = strawberry.union("UserOrError", (User, Error))
+
+x: UserOrError = User(name="Patrick")
+```
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,7 +1,7 @@
 Release type: minor
 
 This release improves type checking support for `strawberry.union` and now allows
-to use unions without any type issues, like so:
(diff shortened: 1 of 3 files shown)


botberry:



---

This release improves type checking support for `strawberry.union` and now allows
to use unions without any type issues, like so:

```python
@strawberry.type
class User:
    name: str

@strawberry.type
class Error:
    message: str

UserOrError = strawberry.union("UserOrError", (User, Error))

x: UserOrError = User(name="Patrick")
```


---




Here's the tweet text: 
```

```

strawberry/union.py
```diff
@@ -171,9 +171,12 @@ def _resolve_union_type(
...
 
 
+Types = TypeVar("Types", bound=Type)
+
+
 def union(
-    name: str, types: Tuple[Type, ...], *, description: str = None
-) -> StrawberryUnion:
+    name: str, types: Tuple[Types, ...], *, description: str = None
+) -> Union[Types]:  # type: ignore
```
BryceBeagle, line 179: Add a comment explaining that the typing of `types` is perfect right now, but will change in the future with [Variadic Generics](https://www.python.org/dev/peps/pep-0646/).

BryceBeagle (changes requested):

RELEASE.md
```diff
@@ -0,0 +1,18 @@
+Release type: minor
+
+This release improves type checking support for `strawberry.union` and now allows
+to use unions without any type issues, like so:
```
```suggestion
to use unions without any type issue, like so:
```

PR #1445, opened 2021-11-24
Fix check of Literal in replace_pydantic_types

nina-j:
## Description

Make sure replace_pydantic_types returns early if the type found is a Literal - otherwise, the issubclass check fails later, because the `__args__` in Literals are vars and not types. 
Literal will fail to resolve later because Strawberry does not support them. 

Perhaps this should fail fast for unsupported Strawberry types that do not originate in Pydantic.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Nina for the PR 👏

```

nina-j:
`AssertionError: assert typing_extensions.Literal['field'] is typing_extensions.Literal['field']`

Ah yes. :(

strawberry/experimental/pydantic/object_type.py
```diff
@@ -24,6 +25,9 @@
 
 
 def replace_pydantic_types(type_: Any):
+    origin = getattr(type_, "__origin__", None)
+    if origin is Literal:
+        return type_
```
patrick91, line 30: can we add a small comment on why we do this? ^^

patrick91 (commented):
Looks great! I've only added a minor comment :)

RE: 

> Literal will fail to resolve later because Strawberry does not support them.

This is perfect, as we still allow devs to override the pydantic field type :)

nina-j:
> ...
> This is perfect, as we still allow devs to override the pydantic field type :)

Oh, you do? Maybe I overlooked it in the documentation, but it hasn't been immediately clear to me that this was possible!

strawberry/experimental/pydantic/object_type.py
nina-j: https://github.com/strawberry-graphql/strawberry/pull/1445/commits/2d43f7649491009ff359c976f500b76ffb12670d - hope this is sufficient!

patrick91:
> Oh, you do? Maybe I overlooked it in the documentation, but it hasn't been immediately clear to me that this was possible!

Oh, we should document it better then!

you can do this:

```python
class Model(pydantic.BaseModel):
     field: Literal["field"]

 @strawberry.experimental.pydantic.type(Model)
 class Type:
     field: str
```

:)

patrick91 (approved):

PR #1447, opened 2021-11-25
Standardize Info object in extensions with deprecation path

### LINKED ISSUE #1425: Introduce custom GraphQL-core ExecutionContext
I'm opening this issue to introduce a custom `ExecutionContext`. The goal is to remove some of our logic around converting `GraphQLResolveInfo` to our `Info` object and also standardise the `info` object, as current `info` inside extensions is `GraphQLResolveInfo` and not `Info`.

I think we can introduce a custom `ExecutionContext` and override `build_resolve_info` as shown here: https://github.com/graphql-python/graphql-core/blob/f145dc9d0ceb6943b8adcbf86ece3c6403b34bb3/src/graphql/execution/execute.py#L482-L488

I haven't checked how many time this is being called by graphql-core, hopefully we don't have to override a lot of other methods

patrick91:
## Summary

This PR addresses #1425 by ensuring extensions receive `strawberry.Info` instead of raw `GraphQLResolveInfo`.

### Changes

- **Duck-typing support**: Added `_raw_info` property to `_OperationContextAwareGraphQLResolveInfo` that returns `self`, enabling duck-type compatibility with `strawberry.Info`

- **Deprecation wrapper**: Added `__init_subclass__` hook in `SchemaExtension` that detects extensions using the old `GraphQLResolveInfo` type hint and wraps them with a deprecation warning

- **Updated internal extensions**: Migrated directives, Apollo tracing, Datadog tracing, and OpenTelemetry extensions to use `Info` type hint with `get_raw_info()` helper when raw info is needed

### Backwards Compatibility

Extensions using the old signature will continue to work but will emit a deprecation warning:

```
DeprecationWarning: Extension MyExtension uses the deprecated GraphQLResolveInfo 
type hint for the 'info' parameter in resolve(). Please update to use strawberry.Info 
instead. GraphQLResolveInfo support will be removed in a future version.
```

### Migration Guide

**Before:**
```python
from graphql import GraphQLResolveInfo

    def resolve(self, _next, root, info: GraphQLResolveInfo, *args, **kwargs):
        # info is GraphQLResolveInfo
        return _next(root, info, *args, **kwargs)
```

**After:**
```python
from strawberry.types.info import Info

    def resolve(self, _next, root, info: Info, *args, **kwargs):
        # info is strawberry.Info with access to _raw_info if needed
        return _next(root, info, *args, **kwargs)
```

## Test plan

- [x] All existing schema tests pass (582 tests)
- [x] Deprecation warnings correctly issued for old-style extensions
- [x] Internal extensions work with new `Info` type

### DIFF
--- a/docs/breaking-changes.md
+++ b/docs/breaking-changes.md
@@ -4,6 +4,7 @@ title: List of breaking changes and deprecations
 
 # List of breaking changes and deprecations
 
+- [Version 0.288.0 - Extension Info Parameter Deprecation](./breaking-changes/0.288.0.md)
 - [Version 0.285.0 - 10 November 2025](./breaking-changes/0.285.0.md)
 - [Version 0.285.0 - 6 October 2025](./breaking-changes/0.283.0.md)
 - [Version 0.279.0 - 19 August 2025](./breaking-changes/0.279.0.md)
--- a/docs/breaking-changes/0.288.0.md
+++ b/docs/breaking-changes/0.288.0.md
@@ -1,0 +1,136 @@
+---
+title: 0.288.0 Breaking Changes
+slug: breaking-changes/0.288.0
+---
+
+# v0.288.0 Deprecation: Extension resolve() Info Parameter
+
+This release deprecates the use of `GraphQLResolveInfo` as the type hint for the
+`info` parameter in extension `resolve()` methods. Extensions should now use
+`strawberry.Info` instead.
+
+## What Changed
+
+Extensions now receive `strawberry.Info` instead of raw `GraphQLResolveInfo` in
+their `resolve()` methods. This standardizes the info object across resolvers
+and extensions.
+
+```python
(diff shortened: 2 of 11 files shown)


docs/breaking-changes.md
```diff
@@ -4,6 +4,7 @@ title: List of breaking changes and deprecations
 
 # List of breaking changes and deprecations
 
+- [Version 0.288.0 - Extension Info Parameter Deprecation](./breaking-changes/0.288.0.md)
```
patrick91, line 7: We need to update this with the date of release, or maybe we stop adding the dates? :D

bellini666: maybe stop adding dates? version info is more important IMO

bellini666 (commented):
Love this! Will help strawberry-django simplify the optimizer extension a little bit 🤏

Issue #1448, opened 2021-11-25
Resolvers imported from another modules break the type of fields


bikeshedder:
Strawberry breaks when you try to use a resolver from a different module than where the type is defined in. e.g. the following code does not work:

```python
import strawberry
from my_resolvers import value_resolver

@strawberry.type
class Foo:
    bar: 'Bar' = strawberry.field(resolver=value_resolver)

@strawberry.type
class Bar:
    name: str
```

It fails with the following error:
```
TypeError: Foo fields cannot be resolved. Unexpected type 'None'
```

uvicorn:
Is problem solved?

bikeshedder:
Indeed. #1449 has been merged and this got fixed.

PR #1449, opened 2021-11-25
Fix cross-module type resolving for fields and resolvers

bikeshedder:
Strawberry breaks when you try to use a resolver from a different module than where the type is defined in. e.g. the following code does not work:

```python
import strawberry
from my_resolvers import value_resolver

@strawberry.type
class Foo:
    bar: 'Bar' = strawberry.field(resolver=value_resolver)

@strawberry.type
class Bar:
    name: str
```

It fails with the following error:
```
TypeError: Foo fields cannot be resolved. Unexpected type 'None'
```

## Types of Changes

- [ ] Core
- [X] Bugfix
- [ ] New feature
- [ ] Documentation

* #1448

## Checklist


strawberry/types/type_resolver.py
```diff
@@ -113,7 +113,7 @@ class if one is not set by either using an explicit strawberry.field(name=...) o
             # or the class where this field was defined, so we always have
             # the correct origin for determining field types when resolving
             # the types.
-            field.origin = field.origin or cls
+            field.origin = cls
```
bikeshedder, line 116: I couldn't figure out why the original code preferred `field.origin` over `cls`. I didn't find a any place where this is needed.

botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Michael P. Jung for the PR 👏

```

bikeshedder:
After talking to `Doctor` in Discord I also think that the field type should have precedence over the return type of the resolver. This PR does not change this behavior. It just makes sure that a resolver without a return type does not break the return type of a field if they are both defined in different python modules.

patrick91:
> After talking to Doctor in Discord I also think that the field type should have precedence over the return type of the resolver. 

I agree with that! Maybe we should change it in this PR too, what do you think? :)

bikeshedder:
> > After talking to Doctor in Discord I also think that the field type should have precedence over the return type of the resolver.
> 
> I agree with that! Maybe we should change it in this PR too, what do you think? :)

The current PR is just a bugfix. Changing the precedence could break existing code and will require a new major (or minor version, since you're still 0.x) version.

strawberry/types/type_resolver.py
BryceBeagle: I think the comment above this line may need some updating? We're always setting the field origin to the class now?

bikeshedder: Do you remember why you introduced that `field.origin or cls` in the first place?

BryceBeagle: I believe it's a holdover from when we used to monkeypatch field information onto the resolver function instead of having dedicated `StrawberryField` and `StrawberryResolver` objects.

bikeshedder: I just realized that this fixed one use case but breaks another one: class inheritance

Previously a field would know it's origin and look up it's type relative to the origin. Now a subclass will set the origin of all fields, also the inherited ones breaking the lookup for those.

Until this is resolved this PR should not be merged.

patrick91: do we have a test for this? :)

bikeshedder: I seam to be wrong about this. I just added tests for this and my change doesn't break inheritance. So yes. Now there is a test for this and to my surprise it doesn't fail.

bikeshedder: Gotcha. That issue with inheritance does indeed arise when inheriting from a class that has some fields defined as `x: int = strawberry.field()`.

bikeshedder: This line was indeed correct so I revered the change. The culprit was in the `base_resolver` setter which did set the `origin`. Removing that line was enough to fix all test cases. :tada:

bikeshedder:
This PR is ready to review. :+1:

tests/types/test_modules.py
```diff
@@ -0,0 +1,155 @@
...
+# be imported into the namespaces under the same name and all type names
+# must be unique. e.g. when importing `a.AObject` into the `c` module it
+# must be renamed to something like `C_AObject`. That way wrong resolves
+# like `C_AObject` being loaded from `a` or `AObject` being loaded from
+# `c` can be detected.
+
+
+from typing import List
+
+from modules import a_mod, b_mod, c_mod, x_mod
```
BryceBeagle, line 16: Would it be possible to rename these a/b modules to be a bit more self-describing? It's not clear just looking a the test names what the difference between a/b/c/x are and what the exact purpose of each test is.

BryceBeagle: Also, what do you think about moving the test file to be within the `modules` directory, so that the "importing resolvers from other modules" tests are right with their respective files?

RELEASE.md
```diff
@@ -0,0 +1,3 @@
+Release type: patch
+
+Fix resolvers imported from another modules
```
BryceBeagle, line 3: Can we add a bit more detail for other readers on what was broken (and is now fixed)

BryceBeagle (changes requested):
Thanks for this PR! All this work for a single-line change, haha. 

The changes I've requested are entirely just about documentation so the new tests make sense for people looking at them in the future. Other than that, great work!

tests/types/test_modules.py
bikeshedder: A and B are standalone modules. C has the types doing composition and inheritance (using types from A and B). And X ... well... it's the unknown containing the resolver without typing. My first instinct was to go with `foo`, `bar` and `baz` but I soon realized that `a`, `b` and `c` is actually easier to reason about especially as all types are prefixed with that letter. Going with something more descriptive would just increase the already long type names. Do you have a better idea for the module names and type names? Feel free to change it. I don't have a better idea. :see_no_evil: 

I'm fine with moving the `test_modules.py` into the `modules` directory. I just didn't do it as it felt like it was just a supplement to the `test_modules.py` file.

tests/types/test_modules.py
```diff
@@ -0,0 +1,159 @@
+# The following tests ensure that the resolver is able to resolve all types
```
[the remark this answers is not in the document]
bikeshedder: I reworded it:
> The following tests ensure that the types are resolved using the correct
> module. Concrete types should be non-problematic and are only included
> here for completeless. The problematic cases are where a `str` is given
> as type which can only be resolved at schema construction.

RELEASE.md
bikeshedder: Yes. As the tests show the removed line did not only break fields with a resolver from a different module but also inheritance. I'm looking into that right now.

tests/types/test_modules.py
BryceBeagle: Hmm, I see what you're saying. Not really good options for naming. Can you add `docstring` at the top of each file explaining what's inside, as an alternative.

> I'm fine with moving the test_modules.py into the modules directory. I just didn't do it as it felt like it was just a supplement to the test_modules.py file.

I think making the test entirely self-contained in a directory would be best.

bikeshedder: I just added a docstring to the `__init__.py` explaining the files. I can move those strings to every module if that's better.

bikeshedder: How about a `README.md` file explaining the whole purpose of the module in the tests?

That could actually also include the part of the `test_modules.py` docstring explaining the reason why the types are named that way.

BryceBeagle: > I just added a docstring to the __init__.py explaining the files. I can move those strings to every module if that's better.

I think that would be best. I don't know if we even need to have the `__init__.py`. The tests work fine without it

bikeshedder: That `c.C_AObject` alias for the `a.AObject` surely isn't self explaining and someone might want to refactor this in the future as it's just a "silly rename".

BryceBeagle: > How about a README.md file explaining the whole purpose of the module in the tests?

I have no preference here, so I'll leave the discretion to you.

bikeshedder:
@BryceBeagle I just updated the release message and changed the title of this PR accordingly. I hope I didn't write it too verbose, this time.

tests/types/test_modules.py
BryceBeagle: Did we decide to not move the tests here to a self-enclosed directory?

bikeshedder: I can move the `test_modules.py` into the `modules` directory, but it will require two changes:

1. The directory `modules` needs a `__init__.py` file
2. The imports from `test_modules.py` need to be changed to `from . import a_mod, b_mod, c_mod, x_mod`

OR

All `*._mod.py` files are moved into their own directory and the content is moved into their `__init__.py`.

It seams that absolute module imports at root level only work for directories when running tests.

bikeshedder: This happens when I try to run it using an absolute import.
```
$ poetry run pytest tests/types/modules/test_modules.py 
(...)
_________________________ ERROR collecting tests/types/modules/test_modules.py _____________________________________________________________________________
ImportError while importing test module '/home/bikeshedder/projects/foss/strawberry/tests/types/modules/test_modules.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../../.asdf/installs/python/3.8.6/lib/python3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/types/modules/test_modules.py:11: in <module>
    import a_mod, b_mod, c_mod, x_mod
E   ModuleNotFoundError: No module named 'a_mod'
```
Only when switching to a relative import and adding a `__init__.py` I'm able to run the tests.

patrick91: I think it is fine to use relative and `__init__.py` 😄

patrick91 (approved):

patrick91:
Bryce will take a final look at this and merge soon ^^

thanks for working on this @bikeshedder!

tests/types/test_modules.py
bikeshedder: Now the file lives in the `tests/types/models` directory. :+1:

tests/types/cross_module_resolvers/test_cross_module_resolvers.py
```diff
@@ -0,0 +1,152 @@
...
+"""
+
+from typing import List
+
+import a_mod
+import b_mod
+import c_mod
+import x_mod
+
+import strawberry
```
BryceBeagle, line 15: Not sure why isort sorted this this way....

BryceBeagle (approved):

patrick91:
Didn't notice this was approved by Bryce too :D I'll merge soon :)

botberry:

PR #1455, opened 2021-11-26
Add support for | to StrawberryUnion

patrick91:
This PR adds support for the new union operator to `StrawberryUnion` :)

Supports:
1. Merging with `None` (making a union optional) -> `MyUnion | None`
2. Merging with another union -> `MyUnion | AnotherUnion`
3. Merging with another type -> `MyUnion | AnotherType`
4. Merging with a union -> `MyUnion | Union[AnotherType, YetAnotherType]`

### DIFF
--- a/strawberry/union.py
+++ b/strawberry/union.py
@@ -28,7 +28,7 @@ from strawberry.exceptions import (
     UnallowedReturnTypeForUnion,
     WrongReturnTypeForUnion,
 )
-from strawberry.type import StrawberryType
+from strawberry.type import StrawberryOptional, StrawberryType
 
 
 if TYPE_CHECKING:
@@ -62,6 +62,14 @@ class StrawberryUnion(StrawberryType):
     def __hash__(self) -> int:
         # TODO: Is this a bad idea? __eq__ objects are supposed to have the same hash
         return id(self)
+
+    def __or__(self, other: Union[StrawberryType, type]) -> StrawberryType:
+        # TODO: shall we add support to merge unions with other unions?
+        # TODO: shall we add support to merge unions with other types?]
+        # current use case is to allow SomeUnion | None, as a nicer way to
+        # mark a field as optional when using a union type
+
+        return StrawberryOptional(of_type=self)
 
     @property
     def types(self) -> Tuple[StrawberryType, ...]:
--- a/strawberry/union.py
+++ b/strawberry/union.py
@@ -64,10 +64,6 @@ class StrawberryUnion(StrawberryType):
         return id(self)
 
     def __or__(self, other: Union[StrawberryType, type]) -> StrawberryType:
-        # TODO: shall we add support to merge unions with other unions?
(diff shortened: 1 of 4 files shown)


strawberry/union.py
```diff
@@ -63,6 +63,32 @@ def __hash__(self) -> int:
...
+
+        if isinstance(other, StrawberryUnion):
+            return StrawberryUnion(
+                type_annotations=tuple(
+                    set(
+                        (
+                            *self.type_annotations,
+                            *other.type_annotations,
+                        )
+                    )
```
patrick91, line 83: shall we make type_annotations always a set? 🤔

patrick91:
@BryceBeagle waht do you think of this? :D

strawberry/union.py
```diff
@@ -63,6 +64,35 @@ def __hash__(self) -> int:
...
         return id(self)
 
+    def __or__(self, other: Union[StrawberryType, type]) -> StrawberryType:
+        if hasattr(other, "_type_definition"):
+            return StrawberryUnion(
+                type_annotations=(
+                    *self.type_annotations,
+                    StrawberryAnnotation(other),
+                )
+            )
```
BryceBeagle, line 74: Maybe we can make a factory method for this? `StrawberryUnion.extend` or something?

strawberry/union.py
BryceBeagle: According to [PEP 604](https://www.python.org/dev/peps/pep-0604/):
> The order of the items in the Union should not matter for equality.
> ```
> (int | str) == (str | int)
> (int | str | float) == typing.Union[str, float, int]
> ```
so I think that's a good idea. Only issue could be hashability of the annotations

tests/types/resolving/test_union_pipe.py
```diff
@@ -0,0 +1,176 @@
...
+    UserOrError = strawberry.union("UserOrError", (User, Error))
+    ExampleOrError = strawberry.union("ExampleOrError", (Example, Error))
+    annotation = StrawberryAnnotation(UserOrError | ExampleOrError)
+    resolved = annotation.resolve()
+
+    assert isinstance(resolved, StrawberryUnion)
+
+    assert len(resolved.type_annotations) == 3
+
+    # doing this to avoid errors due to order of types
```
BryceBeagle, line 131: This shouldn't be an issue if we make the annotations a set, right?

tests/types/resolving/test_union_pipe.py
```diff
@@ -0,0 +1,176 @@
+import sys
```
BryceBeagle, line 1: Rename the file to `test_unions_pipe.py` (or maybe `test_union_pipes.py`) for plurality consistency with the other files

BryceBeagle (changes requested):

patrick91:
I'm going to close this since we'll support unions with Annotated soon!

PR #1463, opened 2021-12-01
Fix issue with enums as params of generic inside unions

### LINKED ISSUE #1403: Error validating return types when using a Generic in Union
There seems to be an issue with types.
Generics worked fine and dandy until I've tried to use them in a Union.

```python
@strawberry.type
class FaultyType:
    code_sent: bool
    result: Union[Pet, ErrorNode[Codes]]
```
Here returning an ErrorNode in "result" field fails type check.
Fails with strawberry.exceptions.UnallowedReturnTypeForUnion: The type "<class 'strawberry_sample.ErrorNode'>" of the field "result" is not in the list of the types of the union: "['CodesErrorNode', 'Pet']"

Returning the same generic without unions in type declaration works fine.

Full sample:
https://gist.github.com/MeRuslan/5bd179f28fc5ae05e815429ee02ebdf6

patrick91:
This should fix #1403

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,4 @@
+Release type: patch
+
+This release fixes an issue that prevented using enums as
+arguments for generic types inside unions.
--- a/strawberry/types/types.py
+++ b/strawberry/types/types.py
@@ -152,6 +152,13 @@ class TypeDefinition(StrawberryType):
 
             # Check if the expected type matches the type found on the type_map
             real_concrete_type = type(getattr(root, generic_field.name))
+
+            # TODO: uniform type var map, at the moment we map object types
+            # to their class (not to TypeDefinition) while we map enum to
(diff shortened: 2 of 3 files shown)


botberry:



---

This release fixes an issue that prevented using enums as
arguments for generic types inside unions.


---




Here's the tweet text: 
```

```

patrick91:
> As the core issue was inside of `TypeDefinition.is_implemented_by`, can we narrow the scope of the provided test? Maybe we can test the functionality of `union.get_type_resolver(self.type_map)` better?

sure, but I'd do that in another PR (unless you have time to do that now), we shouldn't hold out bug fixes for too long 😊
but I think we should keep the new test anyway even if we add more proper unit tests :)

BryceBeagle (approved):

PR #1467, opened 2021-12-02
Make the annotation parser aware about _Unset

sarahhenkens:
## Description

The PR makes the annotation parser aware of the `_Unset` class type inside the Union with None to allow for strict argument typing and still correctly work with strawberry's type generation.

### Why is this change needed

When specifying a union field arg like this the following:

```python
name: Union[_Unset, Optional[str]] = UNSET,
age: Union[int, None, _Unset] = UNSET,
```

Strawberry will incorrectly try to convert `Union[_Unset, str]` into a GraphQL enum. Preventing the usage of stricter type annotations to ensure type checkers are aware of the third `_Unset` type that is set by defaulting to `UNSET`

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

Relates to #872.

## Checklist

### DIFF
--- a/strawberry/arguments.py
+++ b/strawberry/arguments.py
@@ -24,19 +24,11 @@ from strawberry.type import StrawberryList, StrawberryOptional, StrawberryType
 from .exceptions import MultipleStrawberryArgumentsError, UnsupportedTypeError
 from .scalars import is_scalar
 from .types.types import TypeDefinition
+from .unset import _Unset
 
 
 if TYPE_CHECKING:
     from strawberry.schema.config import StrawberryConfig
-
-
-class _Unset:
-    def __str__(self):
-        return ""
-
-    def __bool__(self):
-        return False
-
 
 UNSET: Any = _Unset()
 
--- a/tests/schema/test_arguments.py
+++ b/tests/schema/test_arguments.py
@@ -1,10 +1,10 @@
 from textwrap import dedent
-from typing import Optional
+from typing import Optional, Union
 
 from typing_extensions import Annotated
 
 import strawberry
-from strawberry.arguments import UNSET, is_unset
+from strawberry.arguments import UNSET, _Unset, is_unset
 
 
 def test_argument_descriptions():
@@ -105,6 +105,29 @@ def test_optional_argument_unset():
     assert result.data == {"hello": "Hi there"}
 
 
+def test_optional_argument_unset_with_exhaustive_typing():
+    @strawberry.type
+    class Query:
+        @strawberry.field
+        def hello(
+            self,
+            name: Union[_Unset, Optional[str]] = UNSET,
+            age: Union[int, None, _Unset] = UNSET,
+        ) -> str:
+            if is_unset(name):
+                return "Hi there"
+            return f"Hi {name}"
+
+    schema = strawberry.Schema(query=Query)
+
+    assert str(schema) == dedent(
+        """\
+        type Query {
+          hello(name: String, age: Int): String!
+        }"""
+    )
+
+
 def test_optional_input_field_unset():
     @strawberry.input
     class TestInput:
@@ -146,3 +169,36 @@ def test_optional_input_field_unset():
     )
     assert not result.errors
     assert result.data == {"hello": "Hi there"}
+
+
+def test_optional_input_field_unset_with_exhaustive_typing():
+    @strawberry.input
+    class TestInput:
+        name: Union[_Unset, Optional[str]] = UNSET
+        age: Union[_Unset, None, int] = UNSET
+
+    @strawberry.type
+    class Query:
+        @strawberry.field
+        def hello(self, input: TestInput) -> str:
+            if is_unset(input.name):
+                return "Hi there"
+            return f"Hi {input.name}"
+
+    schema = strawberry.Schema(query=Query)
+
+    assert (
+        str(schema)
+        == dedent(
+            """
+        type Query {
+          hello(input: TestInput!): String!
+        }
+
+        input TestInput {
+          name: String
+          age: Int
+        }
+        """
+        ).strip()
+    )
--- a/tests/schema/test_arguments.py
+++ b/tests/schema/test_arguments.py
@@ -103,3 +103,46 @@ def test_optional_argument_unset():
     )
     assert not result.errors
     assert result.data == {"hello": "Hi there"}
+
+
+def test_optional_input_field_unset():
+    @strawberry.input
+    class TestInput:
+        name: Optional[str] = UNSET
+        age: Optional[int] = UNSET
+
+    @strawberry.type
+    class Query:
+        @strawberry.field
+        def hello(self, input: TestInput) -> str:
+            if is_unset(input.name):
+                return "Hi there"
+            return f"Hi {input.name}"
+
+    schema = strawberry.Schema(query=Query)
+
+    assert (
+        str(schema)
+        == dedent(
+            """
+        type Query {
+          hello(input: TestInput!): String!
+        }
+
+        input TestInput {
+          name: String
+          age: Int
+        }
+        """
+        ).strip()
+    )
+
+    result = schema.execute_sync(
+        """
+        query {
+            hello(input: {})
+        }
+    """
+    )
+    assert not result.errors
+    assert result.data == {"hello": "Hi there"}
--- a/tests/schema/test_arguments.py
+++ b/tests/schema/test_arguments.py
@@ -1,10 +1,10 @@
 from textwrap import dedent
-from typing import Optional, Union
+from typing import Optional
 
 from typing_extensions import Annotated
 
 import strawberry
-from strawberry.arguments import UNSET, _Unset, is_unset
+from strawberry.arguments import UNSET, is_unset
 
 
 def test_argument_descriptions():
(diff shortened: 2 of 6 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Sarah Henkens for the PR 👏

```

strawberry/arguments.py
```diff
@@ -24,20 +24,12 @@
...
 from .scalars import is_scalar
 from .types.types import TypeDefinition
+from .unset import _Unset
 
 
 if TYPE_CHECKING:
     from strawberry.schema.config import StrawberryConfig
 
-
-class _Unset:
```
sarahhenkens, line 33: I hade to move this to its own file because of a circular dependency

BryceBeagle:
Why are you adding `_Unset` to the `Union`? We set `UNSET` as `Any` (i.e. `UNSET: Any = _Unset()`), so should be able to use it for any assignment without adding it to the annotations, even with `--strict` enabled.

```python
from typing import Any

class _Unset:
    ...
        
UNSET: Any = _Unset()

foo: int = UNSET
```
```
Success: no issues found in 1 source file
```

https://mypy-play.net/?mypy=latest&python=3.10&flags=strict&gist=179e8124da44a9949b94b5b1ec46a4c2

tests/schema/test_arguments.py
```diff
@@ -105,6 +105,29 @@ def hello(self, name: Optional[str] = UNSET, age: Optional[int] = UNSET) -> str:
     assert result.data == {"hello": "Hi there"}
 
 
+def test_optional_argument_unset_with_exhaustive_typing():
```
BryceBeagle, line 108: Can these test be narrowed in scope and moved to be with the other type resolving tests for annotations? I don't think we need to going all the way to query execution to test this.

tests/schema/test_arguments.py
```diff
@@ -1,10 +1,10 @@
 from textwrap import dedent
-from typing import Optional
+from typing import Optional, Union
 
 from typing_extensions import Annotated
 
 import strawberry
-from strawberry.arguments import UNSET, is_unset
+from strawberry.arguments import UNSET, _Unset, is_unset
```
BryceBeagle, line 7: Not sure I like exporting a ~private symbol like this. @patrick91 should we just rename it `Unset`?

BryceBeagle (changes requested):

tests/schema/test_arguments.py
patrick91: I'd say yes. But I wonder if we could find a way to make `UNSET`	work as a type, this:

`x: Unset | SomethingElse = UNSET`

feels a bit strange, what do you think @BryceBeagle @sarahhenkens ?

sarahhenkens: Having `UNSET` being set to `any` is the root cause of failures in type checking to occur. Imagine the following scenario:

```python
@strawberry.field
async def jobs(
    self,
    status: Union[_Unset, Optional[Status]] = UNSET,
    first: Optional[int] = UNSET,
    after: Optional[str] = UNSET,
    last: Optional[int] = UNSET,
    before: Optional[str] = UNSET,
) -> Connection[types.Job]:
    result = await Container.job_service().get_paginated(
        status=status, first=first, after=after, last=last, before=before
    )
 ```

From a type checking perspective `UNSET` is technically not valid value for `first: Optional[int] = UNSET,` because its an instance of `_Unset`, but because `UNSET` has its type overwritten to `any`, type checkers will just ignore it.

This gives code a false sense of security, this is why my code example for `status` has an explicit type for it to ensure downstream code is aware of it.

If `Unset` has its correct type set to `_Unset`, type checkers will prevent you from doing  `first: Optional[int] = UNSET,` which is in my opinion a much better default behavior than having `UNSET` be an `any`.

I wanted to keep backward compatibility since other projects might assume it to be any so I just made changes to the internals to support this special type without making changes to the public API and its types.

Comment (sarahhenkens) [reply tests/schema/test_arguments.py:7]: > I'd say yes. But I wonder if we could find a way to make UNSET work as a type, this:
> x: Unset | SomethingElse = UNSET
> feels a bit strange, what do you think @BryceBeagle @sarahhenkens ?

At the core, `x: Optional[T] = UNSET` is essential an 3-way value, "null, not-set, T". In my original ticket (#872) I recommended a new type to represent this tri-state:

```python
T = TypeVar("T")
UnsetOrOptional = Union[_Unset, Optional[T]]
```

This would make the interface much cleaner:
```python
status: UnsetOrOptional[Status]] = UNSET,
```

tests/schema/test_arguments.py
sarahhenkens: Great call out!!

I choose this test since it's the closest to an integration test that ensures the annotation parser works and it correctly still considers the schema output to be a nullable field.

I've updated the tests in `tests/types/resolving/test_optionals.py` as well. Should I remove the ones in `test_arguments`? Or would it be useful to keep them in here as well?

tests/schema/test_arguments.py
patrick91: I think I was suggesting to be able to do this:

```
x: UNSET | SomethingElse = UNSET
```

treating UNSET almost like a literal, so wouldn't need to expose `_Unset` ^^

But I'm not sure this is easy to do at the moment!

sarahhenkens: Having a clearer API sound great! 

My changes are just in the core logic to not have it break on its own `_Unset` though.

Thoughts on getting this merged to unblock any projects that want the strict typing while we work in the future on a better general `UNSET` behavior? 

I wanted to keep this PR's scope minimal by just fixing the annotation parser bug.

patrick91: sounds sensible! I'll open an issue for the (potential) improvement :)

patrick91 (approved):
@sarahhenkens looks good to me! Are you ok with keeping using `_Unset` for now, or do you want to rename it to `Unset`?

For me we can keep `_Unset` until we deal with https://github.com/strawberry-graphql/strawberry/issues/1473 so it is clear that this class is not really meant to be used :)

sarahhenkens:
@patrick91 I think we should keep it as `_Unset` to be 100% we do not break compatibility for projects that might be using it.

tests/schema/test_arguments.py
sarahhenkens: @BryceBeagle can you take another 👀  at this PR?

BryceBeagle: With the tests in tests/types/resolving I think these tests here are redundant. No codepath should escape those tests and make it here undetected.

tests/types/resolving/test_optionals.py
```diff
@@ -16,6 +17,28 @@ def test_basic_optional():
...
+    resolved = annotation.resolve()
+
+    assert isinstance(resolved, StrawberryOptional)
+    assert resolved.of_type is str
+
+    assert resolved == StrawberryOptional(of_type=str)
+    assert resolved == Optional[str]
+
+
+def test_optional_with_unset_as_union():
```
BryceBeagle, line 31: Can we add a test like this for normal `Union`s (without `None`; e.g. `Union[SomeType, OtherType, _Unset]`)?

BryceBeagle (changes requested):

tests/types/resolving/test_optionals.py
sarahhenkens: Great idea!!

I was writing a test case for this but then I realized that if I specify a `Union[str, int, _Unset]`, it does not get stripped out (existing behavior) because the change in this PR is only occurring in the code path that hits optionals.

Optional types are detected by strawberry when it's a union that contains a `None`. We would need to add a `None` in this test case which doesn't make sense to add it to the union tests.

Instead, I've added another assertion into the `test_optionals`:

`test_optional_union_containing_a_real_union_and_unset`

This asserts that `Union[_Unset, None, Foo, Bar]` is resolved as a proper optional `Union[Foo, Bar]` without a default.

tests/schema/test_arguments.py
sarahhenkens: Updated!

tests/types/resolving/test_optionals.py
BryceBeagle: Hmm that makes me wonder if this logic should instead be moved (or mirrored) to to `.create_union`.

Also, what happens if someone does `Optional[_Unset]` right now? I think that should probably throw an exception.

BryceBeagle (changes requested):

tests/types/resolving/test_optionals.py
BryceBeagle: Hmm that makes me wonder if this logic should instead be moved (or mirrored) to to `.create_union`.

Also, what happens if someone does `Optional[_Unset]` right now? I think that should probably throw an exception.

sarahhenkens: IMO that is too much of an edge case to handle here since `_Unset` is already by design an internal type.

BryceBeagle: Hmm, @patrick91 what do you think?

patrick91: sorry I missed this comment, I think it is fine to leave it! we can deal with this case when working on https://github.com/strawberry-graphql/strawberry/issues/1473

BryceBeagle (approved):

botberry:

PR #1477, opened 2021-12-07
WIP: Automatically derive nested pydantic class

thejaminator:
Could we automatically derive nested pydantic classes?
Suppose someone has a repo using pydantic and wants to start using strawberry.
It is troublesome to manually create a new class for each pydantic basemodel and decorate it with `@strawberry.experimental.pydantic.type`. (Thats the issue i am facing now)

Could we automatically (optionally) derive strawberry types for nested pydantic classes?
My initial MR works for this basic example. I'm opening one early so i can find out what issues may occur.

```
from typing import List
import strawberry


class Child(BaseModel):
    value: str


class Parent(BaseModel):
    children: List[Child]


class GrandParent(BaseModel):
    children: List[Parent]


@strawberry.experimental.pydantic.type(model=GrandParent, all_fields=True)
class GrandParentStrawberry: # only need to decorate GrandParentStrawberry
    ...


@strawberry.type
class Query:
    @strawberry.field
    def hello(self) -> GrandParentStrawberry:
        return GrandParentStrawberry.from_pydantic(
            GrandParent(children=[Parent(children=[Child(value="123")])])
        )

```

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

- [ x] My code follows the code style of this project.

### DIFF
--- a/strawberry/experimental/pydantic/object_type.py
+++ b/strawberry/experimental/pydantic/object_type.py
@@ -1,6 +1,7 @@
 import builtins
 import dataclasses
 import warnings
+from enum import Enum
 from functools import partial
 from typing import Any, Dict, List, Optional, Sequence, Tuple, Type, cast
 
@@ -25,6 +26,8 @@ from strawberry.types.types import TypeDefinition
 
 from .exceptions import MissingFieldsListError, UnregisteredTypeException
 
+create_type = type
+
 
 def replace_pydantic_types(type_: Any):
     origin = getattr(type_, "__origin__", None)
@@ -48,11 +51,20 @@ def replace_pydantic_types(type_: Any):
 
         return new_type
 
+    if issubclass(type_, Enum):
+        new_type = strawberry.enum(type_)
+        return new_type
+
     if issubclass(type_, BaseModel):
         if hasattr(type_, "_strawberry_type"):
             return type_._strawberry_type
         else:
-            raise UnregisteredTypeException(type_)
+            # register nested class
+            strawberry.experimental.pydantic.type(model=type_, all_fields=True)(
+                create_type(type_.__name__, (), {})
+            )
+
+            return type_._strawberry_type
 
     return type_
 


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

patrick91:
I'm not opposed to this :) I definitely like the idea of making it optional

Let's see what other people think! @la4de (this might be something we could, if we don't already, do for django), @jkimbo, @BryceBeagle, @marcoacierno

strawberry/experimental/pydantic/object_type.py
```diff
@@ -25,15 +26,30 @@
...
 
-def replace_pydantic_types(type_: Any):
+
+@dataclasses.dataclass
+class RegisterNestedOptions:
+    """Options for registering nested pydantic model"""
+
+    format_type: Callable[[str], str] = lambda x: x # Default does not rename the nested model's name
+    # todo: Other options such as exclude all fields that have a certain name
+
```
thejaminator, line 41: i didn't put these arguments in the decorator arguments because these options only apply if register_nested is passed. 
format_type is quite useful in my use case because we frequently have to rename "Type" into "Type{ServiceName}".

Alternatively, we could support a bunch of class decorator for pydantic basemodels. which is what the library [Caliban does for scala](https://ghostdogpr.github.io/caliban/docs/schema.html#annotations).

```
@GQLName("UserGQL") # rename its to UserGQL when it gets automatically registered
     ...
```

thejaminator:
thanks, 

> I'm not opposed to this :) I definitely like the idea of making it optional
> 
> Let's see what other people think! @la4de (this might be something we could, if we don't already, do for django), @jkimbo, @BryceBeagle, @marcoacierno

thanks, i've fixed the tests and added options for registration of nested basemodels. Let me know if that is desirable and i can update the docs

strawberry/experimental/pydantic/object_type.py
```diff
@@ -48,19 +67,38 @@ def replace_pydantic_types(type_: Any):
...
+                    model=type_, all_fields=True, register_nested=register_nested
+                )(create_type(type_name, (), {}))
+
+                assert hasattr(
+                    type_, "_strawberry_type"
+                ), f"Failed to create _strawberry_type for {type_.__name__}. Please report this bug"
+
+                return type_._strawberry_type
+            else:
+                raise UnregisteredTypeException(type_)
```
BryceBeagle, line 91: Logic here can be simplified and # of returns reduced if we reverse the `if/else`, plus I think it reads a bit better:

```python
if not hasattr(type_, "_strawberry_type"):
    ...
return type_._strawberry_type
```

tests/experimental/pydantic/test_basic.py
```diff
@@ -187,6 +189,94 @@ class UserType:
...
+    definition: TypeDefinition = UserType._type_definition
+    assert definition.name == "UserType"
+
+    [field1, field2] = definition.fields
+
+    assert field1.python_name == "age"
+    assert field1.type is int
+
+    assert field2.python_name == "group"
+    assert field2.type.__name__ == "Group", "should not rename type by default"
```
BryceBeagle, line 220: For consistency with the rest of the tests in the repo, can we remove the custom assert messages? If you thin the message is important, let's just make it a comment/docstring on the test.

BryceBeagle (changes requested):

tests/experimental/pydantic/test_basic.py
thejaminator: sure

XChikuX:
I know we are working on first class pydantic support. However, this PR would benefit me highly!
Is there any reasonn this hasn't been commited? Other than a release.md ?

patrick91:
> I know we are working on first class pydantic support. However, this PR would benefit me highly! Is there any reasonn this hasn't been commited? Other than a release.md ?

From my side, I don't use Pydantic much so I haven't been able to test it properly :)

@XChikuX would you be able to test this if I make a pre-release? :)

patrick91:
@XChikuX rebasing this might take me a bit of time :D

thejaminator:
I'm not really sure if this pattern works appropriately in all cases. Like
what happens if you want to manually add fields in the converted type. How
do you make sure the user has control over it.

patrick91:
> I'm not really sure if this pattern works appropriately in all cases. Like what happens if you want to manually add fields in the converted type. How do you make sure the user has control over it.

that's a fair point. We need to think more about this pattern. I also see a future we enable pydantic via something like this:

```
@strawberry.type # or pydantic.type
class MyType(BaseModel):
    ...
```

and then in that case it would work since we are exposing all the fields (we'd basically treat pydantic as a first class citizen).

But for how strawberry.experimental.pydantic works, we probably might want to have a different behaviour. Maybe we can add a flag on type (ie convert_nested=True)? it's a bit ugly though

XChikuX:
I agree we shouldn't throw this into main without covering the edge cases.
Maybe can just leave it as `nested=True` for now, Conversion should be implied in the one word.

I don't mind waiting for pydantic support, if its coming in the next month or two.
Otherwise, I'd like to push for this since its already half implemented. I can try to cover the edge cases like what @thejaminator mentioned when I get time.

My code is quickly getting verbose using the experimental library.

thejaminator:
Sure, go ahead. I'll suggest that instead of a parameter is_nested: bool,
make it something like nested_options: Optional[NestedOptions].
NestedOptions can be a data class that describes the settings for
conversions. That allows us to add more settings in the future.
The None case will indicate our current behavior.


There's a few options that i foresee being useful. Like a callable that
allows you to specify the type renaming behavior. And in the future maybe
something like a set of  Type[BaseModel] that the user can add to exclude
from automatic conversion.

thejaminator:
@XChikuX i'll commit to coming up with something during our strawberry sprint on 18 feb :)

XChikuX:
@patrick91 Aww.. Any reason this isn't planned anymore?

patrick91:
@XChikuX mostly bandwidth, and the first class pydantic integration that I am planning 😊

But also I'm not sure I want the implicit behaviour 🤔

XChikuX:
What downside do you see for the implicit behavior?

patrick91:
> What downside do you see for the implicit behavior?

Mainly exposing things you don't want (or shouldn't expose) 😊

PR #1484, opened 2021-12-07
Official django 4.0 compat

g-as:
This was kind of a mess, since django 4.0's python compat starts at 3.8. This could be simplified by relaxing/removing the bounds on django.

I also had to capitalize Django in `pyproject.toml` in order to properly override the entry in the tests when `poetry add [...]` is called. This could be circumvented by just using `poetry run pip install django~=3.2` rather than `poetry add`.

Lastly, `poetry.lock` had a lot of "stale" deps, so I included the result of `poetry update`. I can easily discard the commit if this isn't the place.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,7 +69,7 @@ jobs:
   django-unit-tests:
     strategy:
       matrix:
-        django: [3.2, 3.1.8, 3.0.14, 2.2.20]
+        django: [4.0, 3.2, 2.2.25]
 
     name: Test Django ${{ matrix.django }}
     runs-on: ubuntu-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,7 +69,7 @@ jobs:
   django-unit-tests:
     strategy:
       matrix:
-        django: [4.0, 3.2, 2.2.25]
+        django: ["4.0", "3.2", "2.2"]
 
     name: Test Django ${{ matrix.django }}
     runs-on: ubuntu-latest
@@ -80,7 +80,8 @@ jobs:
       - uses: actions/setup-python@v2
         id: setup-python
         with:
-          python-version: '3.x'
+          # Pin to 3.9 because Django 2.2 does not support 3.10
+          python-version: '3.9'
           architecture: x64
 
       - run: pip install poetry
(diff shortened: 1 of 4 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Guillaume Andreu Sabater for the PR 👏

```

.github/workflows/test.yml
```diff
@@ -69,7 +69,7 @@ jobs:
   django-unit-tests:
     strategy:
       matrix:
-        django: [3.2, 3.1.8, 3.0.14, 2.2.20]
+        django: ["4.0", "3.2", "2.2"]
```
patrick91, line 72: shall we remove official support for 2.2 since it is not LTS anymore?

.github/workflows/test.yml
```diff
@@ -99,7 +100,8 @@ jobs:
         if: steps.cache-poetry-dependencies.outputs.cache-hit != 'true'
 
       - name: Install django ${{ matrix.django }}
-        run: poetry add django==${{ matrix.django }}
+        # Add --python ^3.9 marker because django 4.0 does not support 3.7
+        run: poetry add --python ^3.9 django@^${{ matrix.django }}
```
patrick91, line 104: oh wow, didn't know they dropped support for 3.7

I was thinking if we should drop support for 3.7, but I think for now we can keep it, we still have a number of users on 3.7:
Screenshot 2021-12-08 at 00 04 50

source: https://pypistats.org/packages/strawberry-graphql

.github/workflows/test.yml
patrick91: what about 3.0 and 3.1?

patrick91:
Thanks for doing this! Didn't notice we didn't support 4.0 by default 😊

.github/workflows/test.yml
g-as: 2.2 is LTS
3.0 was dropped a few month ago.
3.1 reached end of life today.

https://www.djangoproject.com/download/

.github/workflows/test.yml
g-as: Yes, and 3.7 still has 18 months before EOL. https://endoflife.date/python

Here is the discussion about which python version to support for 4.0:
https://groups.google.com/g/django-developers/c/ezUpskYAcyo/m/e0V7Ho0ZAwAJ

g-as:
> Thanks for doing this! Didn't notice we didn't support 4.0 by default 😊

Since the dependency is optional, the current `pyproject.toml` does not prevent from installing django > 3 along strawberry.

patrick91:
@g-as perfect! can you add a release file? 😊

g-as:
Sure. Patch or minor?

RELEASE.md
```diff
@@ -0,0 +1,4 @@
+Release type: patch
+
+Add support for Django 4.0.
+Remove 3.0 and 3.1 from tests.
```
```suggestion
This release adds support for Django 4.0
```

:)

patrick91 (approved):
Awesome! Thanks for doing this 😊

Issue #1493, opened 2021-12-11
Custom context_getter function overrides default context behaviour thus removing 'request'


tsmith023:
Hello strawberry devs. I've recently come across this package when desiring to move away from Graphene+Flask to Strawberry+FastAPI.

However, when I define a custom context_getter function like:
```python
async def get_context(cache: Redis = Depends(yield_cache), g: AsyncGraphTraversal = Depends(yield_traversal)) -> Dict[str, Redis | AsyncGraphTraversal]:
    return {
        'cache': cache,
        'g': g
    }
```
and call it in:
```python
from strawberry.fastapi import GraphQLRouter
gql_app = GraphQLRouter(
    schema = schema,
    context_getter = get_context,
)
```
then the default context values of `'request'` and `'background_tasks'` are overwritten by the custom context. Since I see no other way of passing `request` down to the `info.context` object outside of the `GraphQLRouter` instance through the custom `context_getter` function, is this potentially a bug?

I looked at the source code and saw that line 65 of strawberry/fastapi/router.py reads `self.context_getter = context_getter or self._get_context` where:
```python
@staticmethod
async def __get_context(
    background_tasks: BackgroundTasks, request: Request = None, ws: WebSocket = None
) -> Optional[Any]:
    return {"request": request or ws, "background_tasks": background_tasks}
```
which will naturally ignore the addition of `'request'` into `info.context` when there is a custom `context_getter` function.

Since these are functions at this point, it is not simple to merge their results. However, further down on lines 88 and 136, these functions are injected as dependencies and passed to handlers on lines 123, 153, and 163. If separate `default_context` and `custom_context` variables were defined, could their results be merged at this point?

patrick91:
@tsmith023 sounds like something we should do, especially since we don't have a way to get the request from the context getter.

Do you agree @Kludex?

Kludex:
We'll need to perform some magic here if we want to inject information.

```python
    @staticmethod
    def __get_context_getter(
        custom_getter: Callable[..., Optional[Dict[str, Any]]]
    ) -> Callable[..., Dict[str, Any]]:
        async def dependency(
            background_tasks: BackgroundTasks,
            request: Request = None,
            ws: WebSocket = None,
            custom_getter: Optional[Dict[str, Any]] = None
        )
            return {"request": request or ws, "background_tasks": background_tasks, **custom_getter}
        
        sig = signature(dependency)
        sig = sig.replace(
            parameters=[
[...]

PR #1494, opened 2021-12-13
Inject custom FastAPI dependencies into context object without overriding default context

tsmith023:
## Description
The `__get_context()` dependency injector has been renamed to `__get_context_getter()` and changed to incorporate the injection of a custom `context_getter()` dependency injector.

@Kludex wrote the solution to be in terms of `signature` so that the default and custom context getters can be merged and stored within `self.context_getter`. In this way, the method functions need not be altered by this change. If the `context_getter` object is `None` then a null-closure is passed to the injector instead.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #1493

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+Change `context_getter` in `strawberry.fastapi.GraphQLRouter` to merge, rather than overwrite, default and custom getters.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,6 @@
 Release type: patch
 
 Change `context_getter` in `strawberry.fastapi.GraphQLRouter` to merge, rather than overwrite, default and custom getters.
+
+This mean now you can always access the `request` instance from `info.context`, even when using
+a custom context getter.
--- a/strawberry/fastapi/router.py
+++ b/strawberry/fastapi/router.py
@@ -45,13 +45,14 @@ class GraphQLRouter(APIRouter):
                 "background_tasks": background_tasks,
                 **custom_getter,
             }
-            
 
         sig = signature(dependency)
         sig = sig.replace(
             parameters=[
                 *list(sig.parameters.values())[:-1],
-                sig.parameters["custom_getter: Dict[str, Any]"].replace(default=Depends(custom_getter)),
+                sig.parameters["custom_getter: Dict[str, Any]"].replace(
+                    default=Depends(custom_getter)
+                ),
             ]
         )
         dependency.__signature__ = sig
--- a/strawberry/fastapi/router.py
+++ b/strawberry/fastapi/router.py
@@ -85,7 +85,8 @@ class GraphQLRouter(APIRouter):
         @self.post(path)
         async def handle_http_query(
             request: Request,
-            context=Depends(self.__get_context) | Depends(self.context_getter) if self.context_getter is not None else Depends(self.__get_context),
+            custom_context=Depends(self.context_getter),
+            default_context=Depends(self.__get_context),
             root_value=Depends(self.root_value_getter),
         ) -> Response:
             content_type = request.headers.get("content-type", "")
@@ -117,7 +118,8 @@ class GraphQLRouter(APIRouter):
                     "No GraphQL query found in the request",
                     status_code=status.HTTP_400_BAD_REQUEST,
                 )
-
+            
+            context = default_context if custom_context == default_context else default_context | custom_context
             result = await self.execute(
                 request_data.query,
                 variables=request_data.variables,
--- a/strawberry/fastapi/router.py
+++ b/strawberry/fastapi/router.py
@@ -62,7 +62,7 @@ class GraphQLRouter(APIRouter):
         self.keep_alive_interval = keep_alive_interval
         self.debug = debug
         self.root_value_getter = root_value_getter or self.__get_root_value
-        self.context_getter = context_getter
+        self.context_getter = context_getter or self.__get_context
         self.protocols = subscription_protocols
         self.connection_init_wait_timeout = connection_init_wait_timeout
 
--- a/strawberry/fastapi/router.py
+++ b/strawberry/fastapi/router.py
@@ -37,15 +37,19 @@ class GraphQLRouter(APIRouter):
             background_tasks: BackgroundTasks,
             request: Request = None,
             ws: WebSocket = None,
-            custom_getter: Optional[Dict[str, Any]] = None
+            custom_getter: Optional[Dict[str, Any]] = None,
         ):
-            return {"request": request or ws, "background_tasks": background_tasks, **custom_getter}
+            return {
+                "request": request or ws,
+                "background_tasks": background_tasks,
+                **custom_getter,
+            }
 
         sig = signature(dependency)
         sig = sig.replace(
             parameters=[
                 *list(sig.parameters.values())[:-1],
-                sig.parameters["custom_getter"].replace(default=Depends(custom_getter))
+                sig.parameters["custom_getter"].replace(default=Depends(custom_getter)),
             ]
         )
         dependency.__signature__ = sig
--- a/strawberry/fastapi/router.py
+++ b/strawberry/fastapi/router.py
@@ -61,7 +61,7 @@ class GraphQLRouter(APIRouter):
         self.keep_alive = keep_alive
         self.keep_alive_interval = keep_alive_interval
         self.debug = debug
-        self.root_value_getter = root_value_getter | self.__get_root_value
+        self.root_value_getter = root_value_getter or self.__get_root_value
         self.context_getter = context_getter
         self.protocols = subscription_protocols
         self.connection_init_wait_timeout = connection_init_wait_timeout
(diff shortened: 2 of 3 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

patrick91 (commented):
Nice <3, can you also add a couple of tests? 😊

tsmith023:
Apologies for sending so many commits and requests to run workflows. I didn't anticipate your testing harness being so comprehensive and opinionated, a lesson learnt for next time!

strawberry/fastapi/router.py
```diff
@@ -30,10 +31,30 @@ async def __get_root_value():
...
+            background_tasks: BackgroundTasks,
+            request: Request = None,
+            ws: WebSocket = None,
+            custom_getter: Dict[str, Any] = {},
+        ):
+            return {
+                "request": request or ws,
+                "background_tasks": background_tasks,
+                **custom_getter,
+            }
```
Kludex, line 47: It really doesn't matter if it's a positional parameter or _keyworded_ one, as you'll modify the signature below.

The [mutable default value](https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html) is unnecessary on the dependency.

```suggestion
        async def dependency(
            custom_getter: Dict[str, Any],
            background_tasks: BackgroundTasks,
            request: Request = None,
            ws: WebSocket = None,
        ):
            return {
                "request": request or ws,
                "background_tasks": background_tasks,
                **custom_getter,
            }
```
The order also doesn't matter, jfyk.

strawberry/fastapi/router.py
```diff
@@ -30,10 +31,30 @@ async def __get_root_value():
...
+
+        sig = signature(dependency)
+        sig = sig.replace(
+            parameters=[
+                *list(sig.parameters.values())[:-1],
+                sig.parameters["custom_getter"].replace(default=Depends(custom_getter)),
+            ]
+        )
```
Kludex, line 55: There's an alternative solution to basically build all the function. Something between those:

- https://github.com/Kludex/fastapi-authorization/blob/67765b28bd350bfef83288c4e4c36c55725f255a/fastapi_authorization/abac.py#L32-L45
- https://github.com/Kludex/fastapi-health/blob/f012f5a1e7a2e184fd7a51a17cf4c26215bb605e/fastapi_health/route.py#L103-L122 

But the current one is fine. :+1:

strawberry/fastapi/router.py
tsmith023: Wow, I never knew about that dictionary anti-pattern. Thanks for informing me!

Kludex: Be aware that your code would have never reached the antipattern behavior. I'm just mentioning because it is not really necessary, and it avoids future PRs about "fixing" what is actually not doing anything :kissing:

tsmith023: This change causes: `msg        = "duplicate parameter name: 'custom_getter'"` due to `parameters = [<Parameter "background_tasks: starlette.background.BackgroundTasks">, <Parameter "custom_getter: Dict[str, Any]">, <P...eter "request: starlette.requests.Request = None">, <Parameter "custom_getter: Dict[str, Any] = Depends(get_context)">]`, so a default value is required. I'll set it to `None` though to avoid that anti-pattern.

tsmith023: There were errors caused by this: `parameters = [<Parameter "background_tasks: starlette.background.BackgroundTasks">, <Parameter "custom_getter: Dict[str, Any]">, <P...eter "request: starlette.requests.Request = None">, <Parameter "custom_getter: Dict[str, Any] = Depends(get_context)">]`, so line 57 of `router.py` should read
```python
sig.parameters["custom_getter: Dict[str, Any]"].replace(default=Depends(custom_getter))`.
```

Is this unusual that the type signatures are being supplied to `signature()` by `dependency()`?

Kludex: ups! I'm sorry, the order matters because we're pasting the n - 1 parameters back to the signature :sweat_smile: 

image

tsmith023: Aha, that's what that cryptic line is doing! Your method makes sense to me now, thanks for the help :)

strawberry/fastapi/router.py
```diff
@@ -62,7 +83,11 @@ def __init__(
         self.keep_alive_interval = keep_alive_interval
         self.debug = debug
         self.root_value_getter = root_value_getter or self.__get_root_value
-        self.context_getter = context_getter or self.__get_context
+        self.context_getter = (
+            self.__get_context_getter(context_getter)
+            if context_getter is not None
+            else self.__get_context_getter(lambda: {})
+        )
```
BryceBeagle, line 90:
```suggestion
        self.context_getter = self.__get_context_getter(context_getter or lambda: {})
```

strawberry/fastapi/router.py
```diff
@@ -30,10 +31,30 @@ async def __get_root_value():
...
-        return {"request": request or ws, "background_tasks": background_tasks}
+    def __get_context_getter(
+        custom_getter: Callable[..., Optional[Dict[str, Any]]]
+    ) -> Callable[..., Dict[str, Any]]:
+        async def dependency(
+            custom_getter: Dict[str, Any],
+            background_tasks: BackgroundTasks,
+            request: Request = None,
+            ws: WebSocket = None,
+        ) -> Callable[..., Dict[str, Any]]:
```
BryceBeagle, line 42:
```suggestion
    def __get_context_getter(
        custom_getter: Callable[..., Optional[Dict[str, Any]]]
    ) -> Callable[..., Dict[str, Any]]:
        async def dependency(
            custom_getter: Dict[str, Any],
            background_tasks: BackgroundTasks,
            request: Optional[Request] = None,
            ws: Optional[WebSocket] = None,
        ) -> Callable[..., Dict[str, Any]]:
```

BryceBeagle: Can the `Any`s here be narrowed to `object`s?

strawberry/fastapi/router.py
```diff
@@ -30,10 +31,30 @@ async def __get_root_value():
...
+            return {
+                "request": request or ws,
+                "background_tasks": background_tasks,
+                **custom_getter,
+            }
+
+        sig = signature(dependency)
+        sig = sig.replace(
+            parameters=[
+                *list(sig.parameters.values())[1:],
```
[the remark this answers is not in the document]
BryceBeagle: If you could add a little comment explaining why the first item in the dict is dropped, that'd be good too. It's not immediately obvious

BryceBeagle (changes requested):
Some minor suggestions/requests :+1:

strawberry/fastapi/router.py
tsmith023: Do you mean the `Any`s provided to `custom_getter` or just those within the `Callable[]`s?

strawberry/fastapi/router.py
tsmith023: I tried this before and was getting a syntax error. However, looking at it again with fresh eyes, it works if the `lambda` is enclosed in parentheses, so:
```python
self.context_getter = self.__get_context_getter(context_getter or (lambda: {}))
```

strawberry/fastapi/router.py
tsmith023: The `Optional`s here trigger typing errors within the fastAPI testing suite so can't be included. Is it an issue if these types are omitted since they have default values of `None`?

Kludex: I don't think you have an option.

strawberry/fastapi/router.py
BryceBeagle: Now that I think about it, would
```python
self.context_getter = self.__get_context_getter(context_getter or dict)
```
work?

tsmith023: It needs to be a lambda closure as `context_getter` is a `Callable` that injects dependencies. However, I've also decided it's easier to have `context_getter or (lambda: None)` here and then `**(custom_getter or {})` on line 46. These are all requirements for the MyPy checks to pass.

tsmith023:
I think that this is in a finalised state now that can be reviewed. I have it passing the test suite on my local machine and working within my projects so it just needs the sign-off of any or all of @patrick91, @Kludex, @BryceBeagle :)

strawberry/fastapi/router.py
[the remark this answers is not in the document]
tsmith023: > Suggestion, but you can ignore if you don't like it. Benefit of `itertools.islice` here is pretty minimal:

I feel like leaving this as is since, as you say, the benefit is minimal and appears less 'pythonic' to me

BryceBeagle (approved):

RELEASE.md
```diff
@@ -0,0 +1,3 @@
+Release type: patch
+
+Change `context_getter` in `strawberry.fastapi.GraphQLRouter` to merge, rather than overwrite, default and custom getters.
```
```suggestion
Change `context_getter` in `strawberry.fastapi.GraphQLRouter` to merge, rather than overwrite, default and custom getters.

This mean now you can always access the `request` instance from `info.context`, even when using
a custom context getter.
```

patrick91 (approved):
Awesome! Thanks you 😊

botberry:

PR #1498, opened 2021-12-15
Lazily get the loop in dataloaders

patrick91:
Dataloaders could be created in a different context,
we shouldn't assume that the current loop is the one
that the dataloader is going to be used in 😊

The error can be reproduce with this snippet:

```python
import asyncio
import strawberry
from aiohttp import web
from strawberry.aiohttp.views import GraphQLView
from strawberry.dataloader import DataLoader

@strawberry.type
class User:
    id: strawberry.ID

async def load_users(keys: list[strawberry.ID]) -> list[User]:
    return [User(id=key) for key in keys]

user_loader = DataLoader(load_fn=load_users)

@strawberry.type
class Query:
    @strawberry.field
    async def get_user(self, user_id: strawberry.ID) -> User:
        return await user_loader.load(user_id)

app = web.Application()
app.add_routes([web.view("/", GraphQLView(schema))])
web.run_app(app)
```

I need to write a test for this before merging it, I might have missed some edge cases too :) but feel free to review it :)

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,4 @@
+Release type: patch
+
+This release changes when we fetch the event loop in dataloaders
+to prevent using the wrong event loop in some occasions.
--- a/strawberry/dataloader.py
+++ b/strawberry/dataloader.py
@@ -47,12 +47,19 @@ class DataLoader(Generic[K, T]):
         self.load_fn = load_fn
         self.max_batch_size = max_batch_size
 
-        self.loop = loop or get_event_loop()
+        self._loop = loop
 
         self.cache = cache
 
         if self.cache:
             self.cache_map = {}
+
+    @property
+    def loop(self) -> AbstractEventLoop:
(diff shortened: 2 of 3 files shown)


botberry:
# Pre-release
:wave:

Releasing commit [eb45126c318ad743abe395cbb5b999690b288216] to PyPi as pre-release! :package:

botberry:
# Pre-release
:wave:

Pre-release **0.93.5-dev.1639595516** [eb45126c318ad743abe395cbb5b999690b288216] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.93.5-dev.1639595516
```

botberry:



---

This release changes when we fetch the event loop in dataloaders
to prevent using the wrong event loop in some occasions.


---




Here's the tweet text: 
```

```

BryceBeagle (approved):

PR #1503, opened 2021-12-17
Bump mypy and fix compatibility issue with 0.920

Apakottur:
Bump mypy to 0.920, and fix compatibility issues with the new version.

## Details

Mypy introduced a breaking change in it's API in https://github.com/python/mypy/pull/9951, where `TypeVarDef` and `TypeVarType` were merged into one class.

In addition to bumping the mypy version, we fix the compatibility issue with a simple check which preserves support for older versions of mypy.

This fix is similar to the one in Pydantic - https://github.com/samuelcolvin/pydantic/pull/3175

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Yossi Rozantsev for the PR 👏

```

strawberry/utils/await_maybe.py
```diff
@@ -8,7 +8,7 @@
 
 
 async def await_maybe(value: AwaitableOrValue):
-    if inspect.iscoroutine(value):
+    if inspect.isawaitable(value):
```
BryceBeagle, line 11: Is this change related to the version bump?

strawberry/ext/mypy_plugin.py
```diff
@@ -507,6 +520,7 @@ def collect_attributes(self) -> Optional[List[DataclassAttribute]]:
                 line=stmt.line,
                 column=stmt.column,
                 type=sym.type,
+                kw_only=False,
```
BryceBeagle, line 523: Is this change related to the version bump?

BryceBeagle: Unrelated to this PR, but why do we use the `dict` constructor here instead of using a `{}` literal, @patrick91?

RELEASE.md
```diff
@@ -0,0 +1,3 @@
+Release type: minor
```
```suggestion
Release type: patch
```

This should be a patch I think 😊

strawberry/ext/mypy_plugin.py
patrick91: not 100% sure why this looks different than the mypy plugin: https://github.com/python/mypy/blob/master/mypy/plugins/dataclasses.py#L342

but if I did the change I used dict because I didn't want to change all the params to have quotes :D

strawberry/utils/await_maybe.py
Apakottur: Yeah, mypy 0.910 is ok with this but mypy 0.920 gives:
`Incompatible types in "await" (actual type "CoroutineType", expected type "Awaitable[Any]")`.

We could alter the typing of the function to only accepts coroutines, instead of changing to `isawaitable`.

strawberry/ext/mypy_plugin.py
Apakottur: Yeah the mypy dataclasses plugin added support for `kw_only` in https://github.com/python/mypy/pull/10867, so we need it here as well.
I just put `False` here to see if things work. I'll compare our plugin to theirs and match the implementation.

RELEASE.md
Apakottur: Done.

strawberry/ext/mypy_plugin.py
```diff
@@ -509,23 +558,22 @@ def collect_attributes(self) -> Optional[List[DataclassAttribute]]:
...
-            try:
-                attribute = DataclassAttribute(**params)  # type: ignore
-            except TypeError:
+            # Support the addition of `info` in mypy 0.800 and `kw_only` in mypy 0.920
+            # without breaking backwards compatibility.
+            if _Version.VERSION >= Decimal("0.800"):
                 params["info"] = cls.info
-                attribute = DataclassAttribute(**params)  # type: ignore
+            if _Version.VERSION >= Decimal("0.920"):
+                params["kw_only"] = is_kw_only
```
Apakottur, line 566: @patrick91 @BryceBeagle 

Retaining backwards compatibility becomes difficult as a lot has been added to the dataclasses plugin in `0.920`, see all the functions below.
We can add guards like this all over the place but maybe it's better to just keep multiple files for different versions? Or drop support for old versions? What do you think?

strawberry/utils/await_maybe.py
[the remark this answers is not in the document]
Apakottur: I agree, but this line now becomes a mypy error, following the version bump.
See https://github.com/strawberry-graphql/strawberry/pull/1503#discussion_r771814580

patrick91: would you be come sending a separate PR just for this? 😊 

I think it is worth isolating this change 😊 I also want to make sure there's no breaking change (even though our tests are passing)

strawberry/ext/mypy_plugin.py
patrick91: we don't support kwonly, so we could drop support for it entirely :)

strawberry/utils/await_maybe.py
Apakottur: Sure, here's a separate PR - https://github.com/strawberry-graphql/strawberry/pull/1505 :)

strawberry/ext/mypy_plugin.py
```diff
@@ -42,13 +42,19 @@
...
     UnionType,
     get_proper_type,
 )
 
 
+# Backwards compatible with the removal of `TypeVarDef` in mypy 0.920.
+try:
+    from mypy.types import TypeVarDef
+except ImportError:
+    TypeVarDef = TypeVarType
```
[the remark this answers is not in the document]
Apakottur: Done

strawberry/ext/mypy_plugin.py
```diff
@@ -262,33 +274,42 @@ def is_dataclasses_field_or_strawberry_field(expr: Expression) -> bool:
...
     """
-
-    if is_dataclasses_field_or_strawberry_field(expr):
-        expr = cast(CallExpr, expr)
-
+    if (
+        isinstance(expr, CallExpr)
+        and isinstance(expr.callee, RefExpr)
+        and expr.callee.fullname == "dataclasses.field"
+    ):
```
patrick91, line 288: I think we should keep using `is_dataclasses_field_or_strawberry_field` otherwise it won't work for `strawberry.field` :)

Apakottur: I undid all the changes to the function, seems like they were just adding an error message which seems irrelevant.

strawberry/ext/mypy_plugin.py
```diff
@@ -509,16 +531,14 @@ def collect_attributes(self) -> Optional[List[DataclassAttribute]]:
...
-            try:
-                attribute = DataclassAttribute(**params)  # type: ignore
-            except TypeError:
+            # Support the addition of `info` in mypy 0.800 and `kw_only` in mypy 0.920
+            # without breaking backwards compatibility.
+            if MypyVersion.VERSION >= Decimal("0.800"):
                 params["info"] = cls.info
-                attribute = DataclassAttribute(**params)  # type: ignore
+            if MypyVersion.VERSION >= Decimal("0.920"):
+                params["kw_only"] = False
```
patrick91, line 539: we'll need to change this after we merge: https://github.com/strawberry-graphql/strawberry/pull/1187

patrick91 (approved):
Looks great! thank you so much 😊

strawberry/ext/mypy_plugin.py
Apakottur: Good point. I removed all the `kw_only` logic and just set it to `False` as before, if the version of mypy is >=0.920

Apakottur:
I was able to run the plugin with mypy versions `0.920`, `0.910` and `0.800`. Should we try an older version of mypy in the tests, or is manual testing enough here?

patrick91:
> I was able to run the plugin with mypy versions `0.920`, `0.910` and `0.800`. Should we try an older version of mypy in the tests, or is manual testing enough here?

should be fine I think, no need to make our ci more complicated 😊

PR #1506, opened 2021-12-20
Overview of type hints as used by strawberry

tsmith023:
## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation

- https://github.com/strawberry-graphql/strawberry/issues/328

## Checklist

### DIFF
--- a/docs/concepts/typings.md
+++ b/docs/concepts/typings.md
@@ -12,7 +12,6 @@ When using Strawberry to build graphQL APIs, as was shown in [Schema basics](htt
 
 The complete mapping of the required type hints for the relevant graphQL types is as follows:
 
-<<<<<<< HEAD
 | GraphQL         | Python                        |
 | --------------- | ----------------------------- |
 | `ID`            | `strawberry.ID`               |
@@ -33,25 +32,6 @@ There are a few different ways in which these Python type hints can be used to e
 - For versions of Python >= 3.10, it is possible to annotate an array of types with `list[Type]`. However, for all previous versions, `typing.List[Type]` must be used instead.
 - The annotation `|` is shorthand for `typing.Union[]`, allowing either of `typing.Union[TypeA, TypeB]` or `TypeA | TypeB` interchangably.
 - The annotation `typing.Optional[Type]` is shorthand for `typing.Union[None, Type]`, which is itself equivalent to `None | Type`.
-=======
-| GraphQL         | Python            |
-| --------------- | ----------------- |
-| `ID`            | `strawberry.ID`   |
-| `String`        | `str`             |
-| `Integer`       | `int`             |
-| `Float`         | `float`           |
-| `Array` or `[]` | `typing.List`     |
-| `Union`         | `typing.Union`    |
-| `Nullable`      | `typing.Optional` |
-| `Date`          | `datetime.date`   |
-| `Decimal`       | `decimal.Decimal` |
-
-where `typing`, `datetime`, and `decimal` are all part of the Python standard library.
-
-The usefulness of using `typing.List` instead of the keyword `list` is so that the types held in the list can be notated. A list of integers is notated as `typing.List[int]` and not as `list[int]`. The only effective equivalence that can be understood by the machine is that `typing.List[typing.Any]` is equivalent to `list`.
-
-The `typing.Optional` hint signifies that the variable being hinted can be its type or `None`. As a result, `Optional[int]` is equivalent to `Union[None, int]`.
->>>>>>> 5614cb72c82a2b8550dfb0ca5b52e5d45e0724fd
 
 ## Example
 
@@ -114,7 +94,6 @@ class Group:
 
 ## Motivation
 
-<<<<<<< HEAD
 Python, much like Javascript and Ruby, is a _dynamically typed_ language that allows for high-level programming where the fundamental types of variables, e.g. integers, arrays, hash-maps, _etc._, are understood by the machine at _runtime_ through Just-in-Time compilation.
 
 Yet, much like the low-level languages of C, Java, and Rust, the graphQL query language is _statically typed_ since the data types defined by the schema must be known prior to compiling the API code in order to define a definite schema to query against.
@@ -122,12 +101,3 @@ Yet, much like the low-level languages of C, Java, and Rust, the graphQL query l
 In the low-level _statically typed_ languages mentioned above, every function must have the types of both their arguments and returns explicitly declared so that the compiler can interpret their behaviours correctly and ensure type safety and consistency.
 
 Strawberry takes inspiration from these languages by requiring that all of its types, fields, resolvers, and mutations declare the types of their arguments and returns. Through this, the schema is generated in a standard and efficient way that aligns with the style-direction of Python and programming as a whole.
-=======
-Python, much like Javascript and Ruby, is a _dynamically typed_ language that allows for high-level programming where the fundamental types of variables, e.g. integers, arrays, hash-maps, etc., are understood by the machine at _runtime_ through Just-in-Time compilation.
-
-Yet, much like the low-level languages of C, Java, and Rust, the graphQL query language is _statically typed_ since the data types defined by the schema must be known prior to running the API in order to define a definite schema to query against.
-
-In the low-level _statically typed_ languages in the above, every function must have the types of both their arguments and returns explicitly declared so that the compiler can interpret their behaviours correctly and ensure type safety and consistency.
-
-Strawberry takes inspiration from these languages by requiring that all Strawberry types, fields, resolvers, and mutations declare the types of their arguments and returns. Through this, the schema is generated in a standard and efficient way that aligns with the style-direction of Python and programming as a whole.
->>>>>>> 5614cb72c82a2b8550dfb0ca5b52e5d45e0724fd
--- a/docs/concepts/typings.md
+++ b/docs/concepts/typings.md
@@ -4,4 +4,77 @@ title: Type hints
 
 # Type hints
 
-...
+Type hints are a modern feature of Python available since 3.5 whose existence is heavily influenced by the features of type-safe languages such as Rust. To learn more about type hints in Python, see [Real Python’s Type hinting walkthrough](https://realpython.com/lessons/type-hinting/).
+
+When using Strawberry to build graphQL APIs, as was shown in [Schema basics](https://strawberry.rocks/docs/general/schema-basics), type hints are required within classes decorated by `@strawberry.type` & `@strawberry.input` and functions decorated by `@strawberry.field` & `strawberry.mutation`. These type hints are sourced as the keywords `str`, `int`, `float`, and from packages imported directly from the Python standard libraries `typing`, that has been available since Python 3.5, `datetime`, and `decimal`.
+
+## Mapping to graphQL types
+
+The complete mapping of the required type hints for the relevant graphQL types is as follows:
+
+| GraphQL         | Python            |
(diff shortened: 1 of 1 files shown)


docs/concepts/typings.md
```diff
@@ -4,4 +4,78 @@ title: Type hints
...
+```python
+import datetime
+import typing
+import strawberry
+
+@strawberry.type
+class Book:
+    title: str
+    author: 'Author'
+    date_published: datetime.Date
```
g-as, line 46: `datetime.date`

docs/concepts/typings.md
```diff
@@ -4,4 +4,78 @@ title: Type hints
...
+| GraphQL         | Python            |
+| --------------- | ----------------- |
+| `ID`            | `strawberry.ID`   |
+| `String`        | `str`             |
+| `Integer`       | `int`             |
+| `Float`         | `float`           |
+| `Array` or `[]` | `typing.List`     |
+| `Union`         | `typing.Union`    |
+| `Nullable`      | `typing.Optional` |
+| `Date`          | `datetime.date`   |
```
patrick91, line 24: we also map datetime and time 😊

docs/concepts/typings.md
```diff
@@ -4,4 +4,78 @@ title: Type hints
...
+| `Decimal`       | `decimal.Decimal` |
+
+where `typing`, `datetime`, and `decimal` are all part of the Python standard library.
+
+The usefulness of using `typing.List` instead of the keyword `list` is so that the types held in the list can be notated. A list of integers is notated as `typing.List[int]` and not as `list[int]`. The only effective equivalence that can be understood by the machine is that `typing.List[typing.Any]` is equivalent to `list`.
```
patrick91, line 29: from python 3.10 you can use `list[str]` as an equivalent to `typing.List[str]` :)

I think we should make a not for that, or maybe use it as the default 

There's also `int | None` for optional types and `TypeA | TypeB` for union types

docs/concepts/typings.md
```diff
@@ -4,4 +4,78 @@ title: Type hints
...
+
+@strawberry.type
+class Group:
+    name: Optional[str] # groups of authors don't necessarily need a name
+    authors: typing.List['Author']
+
+    @strawberry.field
+    async def books(self, info: strawberry.types.Info) -> typing.List['Book']:
+        names = [author.name for author in self.authors]
+        return await info.context['loaders'].load_books_by_author_names.load(names)
```
patrick91, line 66: can we make this full example work without having to import anything else? 😊

patrick91 (commented):
Added some initial comments 😊 thank you so much for starting this!

docs/concepts/typings.md
tsmith023: Yes, I had a feeling that linking two pages like this is sub-optimal. Will change!

docs/concepts/typings.md
g-as: 🎉

patrick91:
https://github.com/python/cpython/pull/30222/files

docs/concepts/typings.md
```diff
@@ -4,4 +4,103 @@ title: Type hints
...
+| `String`      | `str`                      |
+| `Integer`     | `int`                      |
+| `Float`       | `float`                    |
+| `Decimal`     | `decimal.Decimal`          |
+| `Array`, `[]` | `typing.List` or `list`    |
+| `Union`       | `typing.Union` or `        | `   |
+| `Nullable`    | `typing.Optional` or `None | `   |
+| `date`        | `datetime.date`            |
+| `timetz`      | `datetime.time`            |
+| `timestamptz` | `datetime.datetime`        |
```
BryceBeagle, line 27: I think this table's formatting is broken

docs/concepts/typings.md
```diff
@@ -4,4 +4,103 @@ title: Type hints
 
 # Type hints
 
-...
+Type hints are a modern feature of Python available since 3.5 whose existence is heavily influenced by the features of type-safe languages such as Rust. To learn more about type hints in Python, see [Real Python’s Type hinting walkthrough](https://realpython.com/lessons/type-hinting/).
```
BryceBeagle, line 7: Can we add explicit line wrapping (probably ~80 or ~120 chars) to this file.

docs/concepts/typings.md
```diff
@@ -4,4 +4,103 @@ title: Type hints
...
+
+When using Strawberry to build graphQL APIs, as was shown in [Schema basics](https://strawberry.rocks/docs/general/schema-basics), type hints are required within classes decorated by `@strawberry.type` & `@strawberry.input` and functions decorated by `@strawberry.field` & `strawberry.mutation`. These type hints are sourced as the keywords `str`, `int`, `float`, and from packages imported directly from the Python standard libraries `typing`, that has been available since Python 3.5, `datetime`, and `decimal`.
```
BryceBeagle, line 9: > These type hints are sourced as the keywords `str`, `int`, `float`, and from packages imported directly from the Python standard libraries `typing`, that has been available since Python 3.5, `datetime`, and `decimal`.

The wording here is a bit confusing. I think what you're saying is that GraphQL types are directly created using the types you listed, but it sounds like these are the only allowed hints. We support more (esp. using types defined with `@strawberry.type`) so we should make this more clear.

docs/concepts/typings.md
```diff
@@ -4,4 +4,103 @@ title: Type hints
...
+| GraphQL       | Python                     |
+| ------------- | -------------------------- | --- |
+| `ID`          | `strawberry.ID`            |
+| `String`      | `str`                      |
+| `Integer`     | `int`                      |
+| `Float`       | `float`                    |
+| `Decimal`     | `decimal.Decimal`          |
+| `Array`, `[]` | `typing.List` or `list`    |
+| `Union`       | `typing.Union` or `        | `   |
+| `Nullable`    | `typing.Optional` or `None | `   |
```
BryceBeagle, line 24: Is `Nullable` a real GraphQL type? I don't think it should be in backticks here

docs/concepts/typings.md
```diff
@@ -4,4 +4,103 @@ title: Type hints
...
+
+| GraphQL       | Python                     |
+| ------------- | -------------------------- | --- |
+| `ID`          | `strawberry.ID`            |
+| `String`      | `str`                      |
+| `Integer`     | `int`                      |
+| `Float`       | `float`                    |
+| `Decimal`     | `decimal.Decimal`          |
+| `Array`, `[]` | `typing.List` or `list`    |
+| `Union`       | `typing.Union` or `        | `   |
```
BryceBeagle, line 23: We might want to make it clear that `list` and ` | ` are only available in recent versions of Python here. You made a note before but we should draw the connection to it directly (e.g. using an `*`)

docs/concepts/typings.md
```diff
@@ -4,4 +4,103 @@ title: Type hints
...
+A complete example of this, extending upon [Schema basics](https://strawberry.rocks/docs/general/schema-basics), might be the following:
+
+```python
+import datetime
+import typing
+import strawberry
+
+BOOKS_LOOKUP = {
+    'Frank Herbert': [{
+        'title': 'Dune',
```
BryceBeagle, line 48: Great book :+1:

docs/concepts/typings.md
```diff
@@ -4,4 +4,103 @@ title: Type hints
...
+
+@strawberry.type
+class Book:
+    title: str
+    author: 'Author'
+    date_published: datetime.date
+    price: decimal.Decimal
+    isbn: Union[int, str] # could be 1234567890000 or '123-4-56-789000-0'
+
+def get_books_by_author(root: 'Author') -> typing.List['Book']:
```
BryceBeagle, line 63: I think usually in other examples we've been importing _from_ the `typing` namespace and not using the package directly. We should be/continue to be consistent.

docs/concepts/typings.md
```diff
@@ -4,4 +4,103 @@ title: Type hints
...
+@strawberry.type
+class Group:
+    name: Optional[str] # groups of authors don't necessarily have names
+    authors: typing.List['Author']
+
+    @strawberry.field
+    def books(self) -> typing.List['Book']:
+        books = []
+        for author in self.authors:
+            books += get_books_by_author(author)
```
BryceBeagle, line 89:
```suggestion
        books = list(map(get_books_by_author, self.authors))
```
Not sure if we want to be more explicit for an example, though

docs/concepts/typings.md
BryceBeagle: Also, we may want to use `future` annotations

docs/concepts/typings.md
```diff
@@ -4,4 +4,103 @@ title: Type hints
...
+- Type hinting doesn't stop at being a requirement for Strawberry to function, it is also immensely helpful for collaborating developers. By specifying the type of `stored_books` in `get_books_by_author`, an IDE equipped with PyLance will be able to infer that `book` within the list comprehension is a dictionary and so will understand that `.get()` is a method function of the `dict` class. This helps the readability and maintainability of written code.
+
+## Motivation
+
+Python, much like Javascript and Ruby, is a _dynamically typed_ language that allows for high-level programming where the fundamental types of variables, e.g. integers, arrays, hash-maps, _etc._, are understood by the machine at _runtime_ through Just-in-Time compilation.
+
+Yet, much like the low-level languages of C, Java, and Rust, the graphQL query language is _statically typed_ since the data types defined by the schema must be known prior to compiling the API code in order to define a definite schema to query against.
```
BryceBeagle, line 102:
```suggestion
Yet, much like the low-level languages of C, Java, and Rust, the GraphQL query language is _statically typed_ since the data types defined by the schema must be known prior to compiling the API code in order to define a definite schema to query against.
```

This needs adjustment in a couple places in this file

BryceBeagle (changes requested):
Couple quick comments to start this review off. Thanks for working on this :+1:

docs/concepts/typings.md
tsmith023: Will go with 120 for now subject to any further requests

docs/concepts/typings.md
tsmith023: It isn't a type by its own right but it is an important GraphQL concept. The translation of ! from the SDL to Strawberry resolver definitions is important, especially as a code-first library.
I included it here because it was something that I tripped over when I first came to the library. The logic is reversed in Graphene; fields are null by default with `required=True` needed to make it non-null.

docs/concepts/typings.md
tsmith023: Good point, I will try to make that as clear as possible

docs/concepts/typings.md
tsmith023: I have changed the example to use Python 3.10 hinting style inline with @patrick91's comments, so there are no imports from typing

docs/concepts/typings.md
tsmith023: I think I'll leave the more explicit version for clarity

patrick91 (approved):
Nice one! Sorry it took a while to review :)

I've updated the code with some of @BryceBeagle comments

patrick91:
@tsmith023 sorry I didn't see your comments before pushing!

tsmith023:
Oh, I was actually in the process of making some minor changes to it now! But it's all good if you feel it's suitable to merge :)

patrick91:
@tsmith023 feel free to open another PR if you want, I merged too quickly, my fault for not refreshing the tab :D

Issue #1221, opened 2021-09-07
Allow to configure exception catching for schema execution


N-Maas:
Currently, any exception that is raised in a resolver is catched and converted to a `GraphQLError`, which is returned in the errors part of the result (see `strawberry/schema/execute.py`).

However, this is not always desirable. For example, the error message might contain sensitive information about the server that should not be accessible to a client. Instead, the server should return a 500 HTTP error (which would work more or less automatically if the exception is not catched). So we need a possibility to tell strawberry that it should not catch all exceptions, but only specific ones. This could mean catching only `GraphQLError`, or allowing the user to define a set of exceptions that should be catched.

jkimbo:
@N-Maas I there are 2 issues that you have outlined here and they are slightly different:

### 1. Error messages might contain sensitive information about the server

This is true and we'd like to make it easier to hide error messages using extensions (see https://github.com/strawberry-graphql/strawberry/issues/1155). Until then here is a custom extension that only allows certain errors to be included in the response:

```python
from graphql import GraphQLError
from strawberry.extensions import Extension

class VisibleError(Exception):
	pass

class MaskErrors(Extension):
	def on_request_end(self):
		result = self.execution_context.result
		if result.errors:
			processed_errors = []
			for error in result.errors:
				
				if error.original_error and not isinstance(error.original_error, VisibleError):
					processed_errors.append(
						GraphQLError(
							message="Unexpected error.",
							nodes=error.nodes,
							source=error.source,
							positions=error.positions,
							path=error.path,
							original_error=None
						)
					)
				else:
					processed_errors.append(error)

			result.errors = processed_errors

schema = strawberry.Schema(Query, extensions=[MaskErrors])
```

(I'll make a PR for this soon so that you can use it directly from strawberry)

### 2. The server should return a 500 if there are any/some errors

The GraphQL spec actually defines how a server should deal with errors and it explicitly says that errors should be caught and added to the "errors" response: https://spec.graphql.org/June2018/#sec-Errors-and-Non-Nullability 
There is also a draft spec for "GraphQL over HTTP" which defines the status codes that the server should return: https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#status-codes

So it's unlikely that Strawberry will support returning 500 errors if there is an error or selectively catching specific exceptions since it contravenes the spec.

---

I hope thats helpful and that I characterised your question correctly. Let me know if I've missed anything.

[...]

PR #1507, opened 2021-12-20
Parse Decimal/UUID scalar similar to other builtin scalars

silas:
This updates the `Decimal`/`UUID` scalar parser to function similar to the `Date`/`DateTime` scalars, specifically it handles `ValueError` and ensures `original_error` is not defined.

The motivation for making this change is that I'm handling internal server errors similar to #1221 and currently this excludes type errors for the `UUID` scalar.

I'm not sure what the compatibility guarantees this project makes about error messages, but since it's currently just returning the standard library message I assume a change like this won't break anyone's code, but it could technically be considered a breaking change.

~Assuming this change is accepted we might also want to make a similar change for the `Decimal` scalar.~

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/strawberry/schema/types/base_scalars.py
+++ b/strawberry/schema/types/base_scalars.py
@@ -11,7 +11,7 @@ from graphql import GraphQLError
 from strawberry.custom_scalar import scalar
 
 
-def wrap_iso_parser(parser: Callable, type_: str) -> Callable:
+def wrap_parser(parser: Callable, type_: str) -> Callable:
     def inner(value: str):
         try:
             return parser(value)
@@ -19,6 +19,13 @@ def wrap_iso_parser(parser: Callable, type_: str) -> Callable:
             raise GraphQLError(f'Value cannot represent a {type_}: "{value}". {e}')
 
     return inner
+
+
+def parse_decimal(value: str):
+    try:
+        return decimal.Decimal(value)
+    except decimal.DecimalException:
+        raise GraphQLError(f'Value cannot represent a Decimal: "{value}".')
 
 
 isoformat = methodcaller("isoformat")
@@ -29,21 +36,21 @@ Date = scalar(
     name="Date",
     description="Date (isoformat)",
     serialize=isoformat,
-    parse_value=wrap_iso_parser(datetime.date.fromisoformat, "Date"),
+    parse_value=wrap_parser(datetime.date.fromisoformat, "Date"),
 )
 DateTime = scalar(
     datetime.datetime,
     name="DateTime",
     description="Date with time (isoformat)",
     serialize=isoformat,
-    parse_value=wrap_iso_parser(dateutil.parser.isoparse, "DateTime"),
+    parse_value=wrap_parser(dateutil.parser.isoparse, "DateTime"),
 )
 Time = scalar(
     datetime.time,
     name="Time",
     description="Time (isoformat)",
     serialize=isoformat,
-    parse_value=wrap_iso_parser(datetime.time.fromisoformat, "Time"),
+    parse_value=wrap_parser(datetime.time.fromisoformat, "Time"),
 )
 
 Decimal = scalar(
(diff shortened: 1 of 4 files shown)


botberry:



---

This release updates the UUID scalar parser to exclude the original_error exception and format the error message similar to other builtin scalars.


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Silas Sewell for the PR 👏

```

patrick91 (commented):
Sounds great to me! Would you mind doing the same for decimal in this PR? 😊

silas:
@patrick91 Yeah, I can take a look. The decimal [signals](https://docs.python.org/3/library/decimal.html#signals) exceptions aren't very human friendly and will take some work to map to something reasonable (I haven't used the decimal library extensively so there might be a better way).

If you don't mind a generic message we could just handle `DecimalException` and return an invalid decimal error.

Do you have a preference?

Example:

```python
>>> decimal.Decimal('aaa')
  File "<stdin>", line 1, in <module>
decimal.InvalidOperation: [<class 'decimal.ConversionSyntax'>]
```

silas:
I updated the PR to include decimal error handling with a generic error message.

patrick91:
@silas awesome thanks! I'll do a final review in the next few days 😊

strawberry/schema/types/base_scalars.py
```diff
@@ -21,6 +21,13 @@ def inner(value: str):
     return inner
 
 
+def parse_decimal(value: str):
```
BryceBeagle, line 24:
```suggestion
def parse_decimal(value: str) -> decimal.Decimal:
```

tests/schema/types/test_decimal.py
```diff
@@ -31,3 +33,40 @@ def example_decimal(self, decimal: Decimal) -> Decimal:
...
+
+    @strawberry.type
+    class Query:
+        ok: bool
+
+    @strawberry.type
+    class Mutation:
+        @strawberry.mutation
+        def decimal_input(self, decimal_input: Decimal) -> Decimal:
+            assert isinstance(decimal_input, Decimal)
```
BryceBeagle, line 52: Does this assert end up getting hit in this test?

BryceBeagle: (Question also applies to the UUID test)

strawberry/schema/types/base_scalars.py
```diff
@@ -21,6 +21,13 @@ def inner(value: str):
...
+    try:
+        return decimal.Decimal(value)
+    except decimal.DecimalException:
+        raise GraphQLError(f'Value cannot represent a Decimal: "{value}".')
```
BryceBeagle, line 28: We should consider making this string a common asset that both `wrap_parser` and `parse_decimal` use

BryceBeagle (changes requested):

strawberry/schema/types/base_scalars.py
silas: Fixed

tests/schema/types/test_decimal.py
silas: Removed

strawberry/schema/types/base_scalars.py
silas: I didn't want to reuse the same format string because the formatted decimal exceptions aren't very API friend (ex: `[<class 'decimal.ConversionSyntax'>]`).

I could modify `wrap_parser` to something like the following if you wanted:

```python
from typing import Callable

from graphql import GraphQLError


def wrap_parser(
    parser: Callable,
    type_: str,
    classinfo: type = ValueError,
    format_exception: bool = True,
) -> Callable:
    def inner(value: str):
        try:
            return parser(value)
        except Exception as e:
            if not isinstance(e, classinfo):
                raise
            message = f'Value cannot represent a {type_}: "{value}".'
            if format_exception:
                message += f" {e}"
            raise GraphQLError(message)

    return inner


date_parser = wrap_parser(datetime.date.fromisoformat, "Date")
decimal_parser = wrap_parser(
    decimal.Decimal, "Decimal", decimal.DecimalException, False
)
```

BryceBeagle: Nah if it's inconvenient, let's not worry about it :+1:

BryceBeagle (approved):

patrick91 (approved):
Looks great! Thank you! 😊

botberry:

paulo-raca:
Hello,

I just noticed this, and it is really easy to forget to handle ValueErrors when defining new scalars.

I suggest we apply `wrap_parser` automatically to all scalars types (In `@scalar` decorator), what do you think?

PR #1345, opened 2021-10-13
add pagination guide

aryaniyaps:
This PR adds a guide on how to implement pagination in GraphQL

## Description
This PR adds a guide on how to implement pagination in GraphQL. It discusses about the common ways we would do pagination (whether GraphQL is involved or not), like cursor-based or offset-based pagination before implementing the same in GraphQL.

- [x] added general pagination guide
- [x] added offset pagination implementation 
- [x] added cursor pagination implementation 
- [x] merged relay connection docs by @Ambro17 

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement/optimization
- [x] Documentation

## Issues Fixed or Closed by This PR

* #175
* #764 
* #328 
* #268 

## Checklist

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the CONTRIBUTING document.
- [ ] I have added tests to cover my changes.
- [ ] I have tested the changes and verified that they work and don't break anything (as well as I can manage).

PR #1514, opened 2021-12-24
✨ Add docs with example of cursor based pagination

Ambro17:
I add the page with a minimal working example of pagination using strawberry's support for generic types.


docs/guides/pagination.md
```diff
@@ -4,4 +4,124 @@ title: Pagination
 
 # Pagination
 
-Documentation coming soon
+Pagination is a common use case in APIs to efficiently return some results of a relationship instead of all of them, since that may be unperformant.
+
+GraphQL as a spec [recommends cursor based pagination](https://graphql.org/learn/pagination/) and refers to [Relay's Connection Spec](https://relay.dev/graphql/connections.htm) for specific implementation details.
```
BryceBeagle, line 9:
```suggestion
APIs commonly use pagination to efficiently return a portion of a result instead
of every single item, which can have inefficient performance. 

The GraphQL spec [recommends cursor-based pagination](https://graphql.org/learn/pagination/) 
and refers to [Relay's Connection Spec](https://relay.dev/graphql/connections.htm)
for specific implementation details.
```

docs/guides/pagination.md
```diff
@@ -4,4 +4,124 @@ title: Pagination
...
+
+Here we show a minimal example of how you can leverage Strawberry's generic Types
+to build the types required to comply with the relay spec.
+
+```python
+
```
BryceBeagle, line 15:
```suggestion
```

docs/guides/pagination.md
```diff
@@ -4,4 +4,124 @@ title: Pagination
...
+import base64
+from typing import List, Generic, TypeVar, Optional
+
+import strawberry
+from strawberry.arguments import UNSET
+
+
+GenericType = TypeVar("T")  # type: ignore warning as 'T' is too succint to be meaningful for humans
```
BryceBeagle, line 23: I think using `T` is common enough this isn't an issue

docs/guides/pagination.md
```diff
@@ -4,4 +4,124 @@ title: Pagination
...
+
+@strawberry.type
+class Connection(Generic[GenericType]):
+    """Represents a paginated relationship between two entities
+
+    This pattern is using when the relationship itself has attributes.
+    In the facebook domain for example, a friendship between two persons
+    would be a connection that might have a friendshipStartTime
+    """
+    page_info: 'PageInfo'
```
BryceBeagle, line 34: Can you change the strings to use double quotes by default (in line with the `black` formatting we use in the code-base)?

docs/guides/pagination.md
```diff
@@ -4,4 +4,124 @@ title: Pagination
...
+GenericType = TypeVar("T")  # type: ignore warning as 'T' is too succint to be meaningful for humans
+
+
+@strawberry.type
+class Connection(Generic[GenericType]):
+    """Represents a paginated relationship between two entities
+
+    This pattern is using when the relationship itself has attributes.
+    In the facebook domain for example, a friendship between two persons
+    would be a connection that might have a friendshipStartTime
```
BryceBeagle, line 32:
```suggestion
    This pattern is used when the relationship itself has attributes.
    In a Facebook-based domain example, a friendship between two people
    would be a connection that might have a `friendshipStartTime`
```

docs/guides/pagination.md
```diff
@@ -4,4 +4,124 @@ title: Pagination
...
+
+
+
+def build_book_cursor(book: Book):
+    """Adapt this method to build an *opaque* id from an instance"""
+    bookid = f'{id(book)}'.encode('utf-8')
+    return base64.b64encode(bookid).decode()
+
+
+Cursor = str
```
BryceBeagle, line 81:
```suggestion
Cursor: TypeAlias = str
```

BryceBeagle: This will need an import from `typing_extensions` for backwards compat with 3.7

docs/guides/pagination.md
```diff
@@ -4,4 +4,124 @@ title: Pagination
...
+
+
+def get_books(first: int = 10, after: Optional[Cursor] = UNSET) -> Connection[Book]:
+    """
+    A non-trivial implementation should efficiently fetch only
+    the books after the offset and only first ones.
+    For simplicity, here we build the list and then slice it accordingly
```
BryceBeagle, line 88:
```suggestion
    A non-trivial implementation should efficiently query a database to fetch only
    the necessary books after the offset.
    For simplicity, here we build the list and then slice it accordingly
```

docs/guides/pagination.md
```diff
@@ -4,4 +4,124 @@ title: Pagination
...
+
+    @classmethod
+    def from_db_model(cls, instance):
+        """Adapt this method with logic to map your orm instance to a strawberry decorated class"""
+        return cls(title=instance.title, author=instance.title)
+
+
+
+def build_book_cursor(book: Book):
+    """Adapt this method to build an *opaque* id from an instance"""
```
BryceBeagle, line 76:
```suggestion
    """Adapt this method to build an *opaque* ID from an instance"""
```

BryceBeagle: Do we need to explain what "opaque" means?

docs/guides/pagination.md
```diff
@@ -4,4 +4,124 @@ title: Pagination
...
+    page_info: 'PageInfo'
+    edges: list['Edge[GenericType]']
+
+
+@strawberry.type
+class PageInfo:
+    """Pagination context to navigate objects with cursor-based pagination
+
+    Instead of a page and per_page one has a cursor of the last object
+    and fetches item starting from that one
```
BryceBeagle, line 43: Not sure what 
> Instead of a page and per_page one

means

docs/guides/pagination.md
```diff
@@ -4,4 +4,124 @@ title: Pagination
...
+class Query:
+    books: List[Book] = strawberry.field(resolver=get_books)
+
+schema = strawberry.Schema(query=Query)
+```
+
+Name your file `pagination.py` and run `strawberry server pagination`
+
+When you visit the graphql url on your terminal you should see something like this
+<img src="../images/pagination-graphiql-screenshot.png" alt="A view of the GraphiQL interface with a example pagination query"/>
```
BryceBeagle, line 127:
```suggestion
When you visit the GraphQL URL from your terminal output, you should see something like this:
<img src="../images/pagination-graphiql-screenshot.png" alt="A view of the GraphiQL interface with an example pagination query"/>
```

BryceBeagle (changes requested):
Thanks for doing this! I added some phrasing/grammar suggestions, but looks great otherwise.

docs/guides/pagination.md
Ambro17: Let me disagree on this one, it's common when you have dealed with generic types and type hinting, but
for a tutorial aimed for begginers, i prefer to avoid acronyms and be specific on what this line does

docs/guides/pagination.md
Ambro17: I don't think there's much to gain with type hinting this variable

BryceBeagle: At minimum, it makes linters stop complaining about the variable being capitalized

docs/guides/pagination.md
Ambro17: I rephrased it with your suggestion, but one could implement pagination without querying a database directly. For example it could be a grpc call or an http request. I wouldn't suggest here that's the only way to do it and leave it to the user to decide

docs/guides/pagination.md
Ambro17: Perhaps, if they read the graphql reference or the relay one at the top they should know what it means.

docs/guides/pagination.md
Ambro17: Rephrased, let me know if it's more clear

docs/guides/pagination.md
BryceBeagle: Can we turn the word into a link to where it's defined better?

aryaniyaps:
#1345 I had worked on this earlier, just couldn't find the time to finish the PR!

Ambro17:
Hello @aryaniyaps, unfortunately, i saw your docs PR after creating mine. I still think this PR provides value as it complies with Relay Spec. The docs you added are much more complete, but slightly diverge in their approach to relay's spec.
Perhaps we could merge boths PRs and create subsections for relay pagination and offset based pagination (Personally, i wouldn't guide towards offset based pagination as it's not the recommended graphql practice and suffers with large collections)

aryaniyaps:
> Hello @aryaniyaps, unfortunately, i saw your docs PR after creating mine. I still think this PR provides value as it complies with Relay Spec. The docs you added are much more complete, but slightly diverge in their approach to relay's spec.
> Perhaps we could merge boths PRs and create subsections for relay pagination and offset based pagination (Personally, i wouldn't guide towards offset based pagination as it's not the recommended graphql practice and suffers with large collections)

Merging sounds like a great idea! Although offset based pagination isnt recommended, as it doesn't scale for large datasets, it is a quick and easy approach to pagination that can be adopted for small to medium datasets. Maybe this should be made more clearer in the docs 😀

docs/guides/pagination.md
patrick91: I wouldn't care too much for the docs 😊

patrick91 (approved):
Looks pretty great! Going to merge it, we can improve it in future :)

I'll take a look at https://github.com/strawberry-graphql/strawberry/pull/1345/ 
as well and maybe update it based on this PR :)

PR #1538, opened 2022-01-04
fix(pydantic): Fix pydantic default  factory conversion 

thejaminator:
Fixes issue when using a pydantic.Field(default_factory = something)

My bad, forgot to add a test for that

## Types of Changes

- [ ] Core
- [ x] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/strawberry/experimental/pydantic/object_type.py
+++ b/strawberry/experimental/pydantic/object_type.py
@@ -121,7 +121,10 @@ def type(
                     # always unset because we use default_factory instead
                     default=UNSET,
                     default_factory=defaults_into_factory(
-                        default=field.default if not field.required else UNSET,
+                        # pydantic's default and default_factory are None by default
+                        default=field.default
+                        if (not field.required) and (field.default_factory is None)
+                        else UNSET,
                         default_factory=field.default_factory or UNSET,
                     ),
                     type_annotation=get_type_for_field(field),
--- a/strawberry/experimental/pydantic/object_type.py
+++ b/strawberry/experimental/pydantic/object_type.py
@@ -18,7 +18,7 @@ from strawberry.experimental.pydantic.conversion import (
 from strawberry.experimental.pydantic.fields import get_basic_type
 from strawberry.experimental.pydantic.utils import (
     DataclassCreationFields,
-    defaults_into_factory,
+    get_default_factory_for_field,
     get_private_fields,
     sort_creation_fields,
 )
@@ -120,13 +120,7 @@ def type(
                     graphql_name=field.alias if field.has_alias else None,
                     # always unset because we use default_factory instead
                     default=UNSET,
-                    default_factory=defaults_into_factory(
-                        # pydantic's default and default_factory are None by default
-                        default=field.default
-                        if (not field.required) and (field.default_factory is None)
-                        else UNSET,
-                        default_factory=field.default_factory or UNSET,
-                    ),
+                    default_factory=get_default_factory_for_field(field),
                     type_annotation=get_type_for_field(field),
                     description=field.field_info.description,
                 ),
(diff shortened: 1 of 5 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

strawberry/experimental/pydantic/utils.py
```diff
@@ -85,9 +85,11 @@ def defaults_into_factory(
...
     if default is not UNSET:
         if default_factory is not UNSET:
-            raise BothDefaultAndDefaultFactoryDefinedError(
-                default=default, default_factory=default_factory  # type: ignore
-            )
+            # pydantic ModelField default and default_factory are None by default
+            if default is not None and default_factory is not None:
+                raise BothDefaultAndDefaultFactoryDefinedError(
+                    default=default, default_factory=default_factory  # type: ignore
+                )
```
thejaminator, line 92: hacky?

if you looked at pydantic's ModelField, it sets default and default_factory to None by default.
rather than having a value such as UNSET to differentiate between None as the user's default, and having no default.

tests/experimental/pydantic/test_basic.py
```diff
@@ -272,6 +272,46 @@ class UserType:
...
+    class UserType2:
+        friend: strawberry.auto
+
+    assert UserType2().friend is None
+
+    # Test instantiation using default_factory
+
+    class User3(pydantic.BaseModel):
+        friend: Optional[str] = pydantic.Field(default_factory=lambda: "friend")
+
```
thejaminator, line 298: previously defining it like this would fail, because default=None, and we'll think that a factory and default was both defined. 
see https://discord.com/channels/689806334337482765/773519351423827978/927667160250212462

tests/experimental/pydantic/test_conversion.py
```diff
@@ -787,22 +787,6 @@ def test_convert_input_types_to_pydantic_default_and_default_factory():
...
-    hacked_field.default = None
-    with pytest.raises(
-        BothDefaultAndDefaultFactoryDefinedError,
-        match=("Not allowed to specify both default and default_factory."),
-    ):
-
-        @strawberry.experimental.pydantic.input(User2)
-        class UserInput:
-            password: strawberry.auto
-
```
thejaminator, line 805: turns out hacking it like this wasn't possible. 
doing
```
hacked_field.default = None 
```
didn't actually change the default. pydantic under the hood would ignore the change somehow.

the test only worked because it already was None by default

strawberry/experimental/pydantic/object_type.py
```diff
@@ -121,7 +121,10 @@ def wrap(cls):
                     # always unset because we use default_factory instead
                     default=UNSET,
                     default_factory=defaults_into_factory(
-                        default=field.default if not field.required else UNSET,
+                        # pydantic's default and default_factory are None by default
+                        default=field.default
+                        if (not field.required) and (field.default_factory is None)
+                        else UNSET,
```
thejaminator, line 127: turns out pydantic's `ModelField` doesn't use it's Undefined class to mark that the default was undefined. Thats only used in `Field`, when the user defines it. Once the model class is actually instantiated, `Field` gets turned into a `ModelField` and that Undefined distinction gets lost.

thejaminator: since field.default_factory is `Optional[NoArgAnyCallable]`, if it is None and if the field is not required, 
then the default value must come from field.default

patrick91:
@thejaminator I did change this a bit, would you be able to check if the changes make sense? 

Also let me know if you want a pre-release if you have some code to test this 😊

thejaminator:
> @thejaminator I did change this a bit, would you be able to check if the changes make sense?
> 
> Also let me know if you want a pre-release if you have some code to test this 😊

@patrick91  i added a few test statements and corrected a typo. looks good! i tested it locally on my codebase and seems fine, so won't need a pre-release :D

patrick91:
Awesome! Thanks ☺️

patrick91 (approved):

PR #1546, opened 2022-01-06
feat(pydantic): Support use_pydantic_alias parameter for pydantic conversion

thejaminator:
## Description

Adds support for gql_uses_alias parameter in pydantic model conversion.
Decides if the graphql field name should use the alias name or not.

## Current behaviour

```python
from pydantic import BaseModel, Field
class UserModel(BaseModel):
      id: int = Field(..., alias="my_alias_name")
      another_field: int
```

```python
    UserModel
)
class User:
    id: strawberry.auto
    another_field: strawberry.auto
```

On the graphql schema,  You'll see
```javascript
type User {
      myAliasName: Int!
      anotherField: Int!
}
```
What if you manually define the type of  `id` ? 

```python
    UserModel
)
class User:
    id: int
    another_field: strawberry.auto
```
Confusingly, the User type becomes

```javascript
type User {
      id: Int! // instead of myAliasName 😱 
      anotherField: Int!
}
```

Two issues
1. When refactoring, if we first directly specify the type, and then change it to strawberry.auto. It changes graphql schema. Not a pleasant surprise :D 
2. Users don't have a way opt out of using the alias when using strawberry.auto. The default behaviour also not easily visible without looking at the code / observing the schema change.

The first issue is difficult to fix since its the default behaviour already :D 
For the second issue:
This MR adds the option `gql_uses_alias` to allow users to control whether the alias gets used in schema.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [X] New feature
- [ ] Documentation

## Checklist

- [x ] My code follows the code style of this project.


strawberry/experimental/pydantic/object_type.py
```diff
@@ -83,6 +83,7 @@ def type(
     description: Optional[str] = None,
     directives: Optional[Sequence[StrawberrySchemaDirective]] = (),
     all_fields: bool = False,
+    gql_uses_alias: bool = True,
```
thejaminator, line 86: couldn't think of a better name that isn't too long. any ideas?

botberry:

```markdown
Release type: patch

```


Here's the tweet text: 
```

```

strawberry/experimental/pydantic/object_type.py
patrick91: maybe we could use just `use_alias`?

let me /cc @Matt343 as he's probably using the pydantic integration, I don't use it at the moment so I'm not 100% sure if we should do this :)

Matt343: I'm just wondering if this option should be specified at the field level instead of the object level, since you might want to use an alias in one place but not another. I'm also not clear on what affect an alias has if you turn this feature off, where would it actually be used?
For the first issue related to `strawberry.auto`, I don't think that's the intended behavior so maybe we should be open to a bugfix there?

thejaminator: >  I'm also not clear on what affect an alias has if you turn this feature off, where would it actually be used?

Not sure if i understand right. But [the alias is the pydantic alias i'm referring to](https://pydantic-docs.helpmanual.io/usage/model_config/#alias-precedence)


> I'm just wondering if this option should be specified at the field level instead of the object level, since you might want to use an alias in one place but not another. 

interesting. how could we support this?
we could have some Annotation on the pydantic model. But idk how well that would work with pydantic. And personally i don't like mixing strawberry logic in my pydantic models layer.
maybe we can do something like this?

```python
    UserModel
)
class User:
    id: strawberry.auto(rename_to="someOtherName")
    another_field: strawberry.auto
```


> For the first issue related to strawberry.auto, I don't think that's the intended behavior so maybe we should be open to a bugfix there?

cool! though it the default behaviour for sometime so idk if theres people relying on it. and schema changes aren't so obvious :(

Matt343: Ah I see, I thought we were talking about strawberry aliases here. This makes a lot of sense now, and I can see how the current behavior is pretty confusing. What you're actually running into is an undocumented feature/bug in the pydantic integration that exists when you define a field that has the same name as a pydantic field but doesn't use the `auto` type. What that does is sort of copy the field from pydantic but not all the way, which leads to very strange behaviors. I'm not sure what we should do here, but it's certainly not what we do today, even if that behavior is sometimes useful by accident.

As for where to put the setting, I like offering the field level if possible but maybe what you have now is a good starting point. You wouldn't be able to do `strawberry.auto(rename_to="someOtherName")` afaik due to the way that type annotations need to work, but we could potentially add something like:

```
    UserModel
)
class User:
    id: strawberry.auto = strawberry.field(alias="someOtherName")
    another_field: strawberry.auto
```

This would be some work though, so I feel like it might be good as a follow on feature rather than a blocker now, I don't see any reason not to have the object level config anyway.

thejaminator: great! seems like we're on the same page. sorry didn't explain that it was a pydantic alias earlier :) 
@patrick91 could we add the object level config first?

thejaminator: Hi all, matt suggested that we can have an field level alias by `strawberry.field(alias="someOtherName")`.
i think thats reasonable and something we can support in a future MR.

for this MR, would like to continue forward to support a configuration at the decorator level.

 i have changed the decorator parameter to `use_pydantic_alias` to differentiate the pydantic alias from a possible future strawberry.field(alias="someOtherName") parameter.

RELEASE.md
```diff
@@ -0,0 +1,34 @@
+Release type: patch
+
+Adds support for use_pydantic_alias parameter in pydantic model conversion.
+Decides if the graphql field name should use the alias name or not.
```
```suggestion
Adds support for `use_pydantic_alias` parameter in pydantic model conversion.
Decides if the all the GraphQL field names for the generated type should use the alias name or not.
```

RELEASE.md
```diff
@@ -0,0 +1,34 @@
...
+      id: int = Field(..., alias="my_alias_name")
+
+@strawberry.experimental.pydantic.type(
+    UserModel, use_pydantic_alias=False
+)
+class User:
+    id: strawberry.auto
+```
+
+If use_pydantic_alias is `False`, the graphql type User will have the field id
```
patrick91, line 20:
```suggestion
If `use_pydantic_alias` is `False`, the GraphQL type User will use `id` for the name of the `id` field coming from the Pydantic model.
```

RELEASE.md
```diff
@@ -0,0 +1,34 @@
...
+```
+type User {
+      id: Int!
+}
+```
+
+instead of the field myAliasName.
```
patrick91, line 27:
```suggestion
With `use_pydantic_alias` set to `True` (the default behaviour) the GraphQL type user will use `myAliasName` for the `id` field coming from the Pydantic models (since the field has a `alias` specified`)
```

RELEASE.md
```diff
@@ -0,0 +1,34 @@
...
+```
+type User {
+      myAliasName: Int!
+}
+```
+
+Currently use_pydantic_alias is set to `True` for backwards compatibility.
```
patrick91, line 34:
```suggestion
`use_pydantic_alias` is set to `True` for backwards compatibility.
```

strawberry/experimental/pydantic/object_type.py
```diff
@@ -142,7 +143,9 @@ def wrap(cls: Any) -> Type[StrawberryTypeFromPydantic[PydanticModel]]:
                 type_annotation=get_type_for_field(field, is_input),
                 field=StrawberryField(
                     python_name=field.name,
-                    graphql_name=field.alias if field.has_alias else None,
+                    graphql_name=field.alias
+                    if field.has_alias and use_pydantic_alias
+                    else None,
```
patrick91, line 148:
```suggestion
                    graphql_name=(
                        field.alias
                        if field.has_alias and use_pydantic_alias
                        else None
                    ),
```

patrick91 (approved):
I've added a couple of minor changes 😊

I also prefer being able to specify if we should use the alias on a field basis. But I think it is fine to merge this for now, though maybe in future we should see if we can make this a schema configuration option, since it feels a bit weird to have it on a type basis :)

thejaminator:
> I've added a couple of minor changes 😊
> 
> I also prefer being able to specify if we should use the alias on a field basis. But I think it is fine to merge this for now, though maybe in future we should see if we can make this a schema configuration option, since it feels a bit weird to have it on a type basis :)

yes it seems eventually we need a schema level config for pydantic related things. 🤔

PR #1547, opened 2022-01-06
fix(pydantic): Fix TypeError when converting BaseModel with NewType field

thejaminator:
## Description

Previously you would get an error 
```
TypeError: issubclass() arg 1 must be a class
```
as we would call issubclass on the NewType

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/strawberry/experimental/pydantic/fields.py
+++ b/strawberry/experimental/pydantic/fields.py
@@ -3,6 +3,7 @@ from typing import Optional
 from uuid import UUID
 
 import pydantic
+from pydantic.typing import is_new_type, new_type_supertype
 
 from .exceptions import UnsupportedTypeError
 
@@ -75,5 +76,8 @@ def get_basic_type(type_):
 
         if type_ is None:
             raise UnsupportedTypeError()
+
+    if is_new_type(type_):
+        return new_type_supertype(type_)
 
     return type_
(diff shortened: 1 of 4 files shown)


strawberry/experimental/pydantic/object_type.py
```diff
@@ -37,21 +37,21 @@ def replace_pydantic_types(type_: Any):
...
-                new_type.name,
+            replaced_type = builtins.type(
+                replaced_type.name,
                 (),
-                {"_type_definition": new_type},
+                {"_type_definition": replaced_type},
             )
 
-        return new_type
+        return replaced_type
```
thejaminator, line 54: new_type in this file wasn't refering to typing's NewType.
renamed this to replaced_type because it was confusing.

strawberry/experimental/pydantic/fields.py
```diff
@@ -76,4 +77,7 @@ def get_basic_type(type_):
         if type_ is None:
             raise UnsupportedTypeError()
 
+    if is_new_type(type_):
+        return new_type_supertype(type_)
+
```
thejaminator, line 82: this is the only logic we needed to change.

strawberry/experimental/pydantic/object_type.py
patrick91: nice <3

patrick91 (approved):
Amazing! Thank you <3

Issue #1541, opened 2022-01-04
FastAPI router Dependency Context regression - Mapping required.


bbsbb:
Hey there,

We use a typed object as dependency context with FastAPI( https://fastapi.tiangolo.com/tutorial/dependencies/classes-as-dependencies/ )

After the merge of:  #1494, we are affected by a regression - there is an implicit assumption that the context is limited to Mapping ( https://github.com/strawberry-graphql/strawberry/blob/0.93.16/strawberry/fastapi/router.py#L46 ).
This is not actually a FastAPI requirement. Is it possible to be defensive and allow objects as context as well?

@tsmith023

tsmith023:
Hi Boyan,

If I understand you correctly, you have a custom-defined context dependency in the form of a class rather than a simple dictionary. Then the line that you linked causes your code to break with an error like: ``TypeError: argument must be a mapping, not 'CustomContext'``, as a result. Is this analysis correct?

Since the default context object from previous versions was also a dictionary containing `request` and `background_tasks`, I think the structure of the the merged context object should also be as a dictionary. The `**` operator was introduced so that these default keys were integrated into the context object without introducing breaking changes to those using dictionary custom contexts.

I think the way to appease both class-based and dictionary-based custom contexts is to add functionality in router.py to
- add new attributes `request` and `background_tasks` to the custom context if it's a class,
- maintain the current logic if the custom context is a dictionary.

@patrick91, @Kludex, @BryceBeagle do you guys agree?

If this were implemented, would it be an issue for you if `request` and `background_tasks` were injected into your context object @bbsbb?

bbsbb:
Hey Tommy,

This is spot on and your understanding is correct. The default was indeed always dictionary, but it worked seamlessly with objects.

The suggested implementation seems interesting. It would resolve the underlying issue as far as we are concerned!

patrick91:
@tsmith023 that sounds good! would you be able to send a PR for that? 😊

tsmith023:
@bbsbb, I notice in your initial comment that opened this issue that you describe you custom context dependency as a 'typed object'. Do you mean an instance of a class that has been decorated with @strawberry.type or just a normal class, perhaps even a Pydantic one given your recent PR :)?

bbsbb:
Hey @tsmith023 ,

[...]

Issue #1258, opened 2021-09-17
Standardise context creation and access


jkimbo:
At the moment the Django and Sanic integrations return a custom dataclass from the default implementation of `get_context`. This allows gives a nicer developer experience because you can access properties via dict or attribute notation. However if you override the `get_context` method to customise the context and return a dictionary (which is what the docs recommend) it can break assumptions made elsewhere.

I propose that we standardise how the context is created by wrapping whatever is returned from `get_context` in a `StrawberryContext` wrapper class. This would also allow us to store internal data in the context which might be useful in custom fields or extensions.

Something like this:

```python
class MyGraphQLView(GraphQLView):
	def get_context(self, request, response):
		return {
			"request": request,
			"response": response,
			"current_time": datetime.now(),
		}

@strawberry.type
class Query:
	@strawberry.field
	def current_time(self, info) -> str:
		context = info.context  # type is StrawberryContext
		return context.current_time.isoformat()
```

Also I think it should be possible (and encouraged) to modify context using extensions rather than having to override the base view. 

Only concern I have is how we type this. Thoughts @strawberry-graphql/core  ?

BryceBeagle:
I think this is a great idea.

> Only concern I have is how we type this.

Can you elaborate on what makes typing this complicated?

jkimbo:
> Can you elaborate on what makes typing this complicated?

So if we take the example I posted you might want to type your context object like this:

```python
class MyContext(TypedDict):
	request: Request
	response: Response
	current_time: datetime

class MyGraphQLView(GraphQLView):
	def get_context(self, request, response) -> MyContext:
		return {
			"request": request,
			"response": response,
			"current_time": datetime.now(),
		}
```

If you have that it's unclear how we could tell Mypy that this is ok:

```python
@strawberry.type
class Query:
	@strawberry.field
	def current_time(self, info: Info[MyContext]) -> str:
		context = info.context.current_time  # how does Mypy know that `current_time` is an attribute on `context`?
```

jkimbo:
Dataclasses would be good but how would you allow extra things to be added outside of the "get_context" call? For example adding things to context through extensions.

patrick91:
> Dataclasses would be good but how would you allow extra things to be added outside of the "get_context" call? For example adding things to context through extensions.

do you have any use case in mind? You can always add new property, but it might be that clean

patrick91:
[...]

PR #1564, opened 2022-01-07
Introduce BaseContext class from which user CustomContext should inherit

tsmith023:
## Description

FastAPI dependency injection of user-defined custom contexts should work for both dictionary contexts and class contexts. In its current state, Strawberry is only capable of handling custom user dictionary contexts.

To fix this, I propose that we introduce a BaseContext class within fastAPI/router.py that has instance attributes `request`, `background_tasks`, and `response`. The attributes are populated as the context object moves through the methods of GraphQLRouter and is then merged with any custom user context, which itself must be a child of BaseContext.

This PR would seem to resolve an issue unique to fastAPI, however it appears that it might be relevant to the wider library since https://github.com/strawberry-graphql/strawberry/issues/1258 seems to be particularly similar.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [x] New feature
- [x] Documentation

https://github.com/strawberry-graphql/strawberry/issues/1541

## Checklist


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

tests/fastapi/test_context.py
```diff
@@ -41,11 +84,43 @@ def abc(self, info: Info) -> str:
...
+    test_client = TestClient(app)
+    try:
+        response = test_client.post("/graphql", json={"query": "{ abc }"})
+    except TypeError as e:
+        assert (
+            str(e) == "The custom context dependency must be either a "
+            "class that inherits from BaseContext or a dictionary"
+        )
+        response = "Task failed successfully"
+    assert response == "Task failed successfully"
```
BryceBeagle, line 126:
```suggestion
    assert response == "Task failed successfully"

```
Also, missing new line at end of file

BryceBeagle (changes requested):

tests/fastapi/test_context.py
patrick91: this should have been covered by precommit 🤔

patrick91 (commented):
Looks pretty good to me! Other than Bryce's comments we also need a release file 😊

tsmith023:
Made the changes requested by @BryceBeagle, will now add the release file and also a small change to the documentation to note the requirement for custom class contexts to inherit from BaseContext.

docs/integrations/fastapi.md
```diff
@@ -97,10 +110,64 @@ app.include_router(graphql_app, prefix="/graphql")
...
+
+
+class CustomContext(BaseContext):
+    def __init__(self, greeting: str, name: str):
+        self.greeting = greeting
+        self.name = name
+
+
+def custom_context_dependency() -> CustomContext:
+    return CustomContext(greeting = "you rock!", name = "John")
```
BryceBeagle, line 137:
```suggestion
    return CustomContext(greeting="you rock!", name="John")
```

strawberry/exceptions.py
```diff
@@ -220,3 +220,12 @@ def __init__(self):
...
 
         super().__init__(message)
+
+
+class InvalidCustomContext(Exception):
+    """Raised when a custom context object is of the wrong python type"""
+
+    def __init__(self):
+        message = "The custom context must be either a class\
+            that inherits from BaseContext or a dictionary"
```
BryceBeagle, line 230:
```suggestion
        message = (
            "The custom context must be either a class that inherits from "
            "BaseContext or a dictionary"
        )
```

BryceBeagle: Avoid `\` whenever possible. If this were to get merged as-is, the string would look like:
```pycon
>>> message
'The custom context must be either a class            that inherits from BaseContext or a dictionary'
```
Your test, also using the `\`, misses this.

strawberry/fastapi/router.py
```diff
@@ -24,8 +24,15 @@
...
-        self.background_tasks: BackgroundTasks = None
+        self.request: Optional[Union[Request, WebSocket]] = None
+        self.background_tasks: Optional[BackgroundTasks] = None
+        self.response: Optional[Response] = None
+
+
+CustomContext = Union[BaseContext, Dict[str, Any]]
+MergedContext = Union[
+    BaseContext, Dict[str, Union[Any, BackgroundTasks, Request, Response, WebSocket]]
+]
```
BryceBeagle, line 35: 1. @patrick91, do you know the best practices for type aliases? Top of file, or next to usage (between classes in this case)?
2. 
```suggestion
CustomContext: TypeAlias = Union[BaseContext, Dict[str, Any]]
MergedContext: TypeAlias = Union[
    BaseContext, 
    Dict[str, Union[Any, BackgroundTasks, Request, Response, WebSocket]],
]
```
3. What are the implications of `Union[Any, BackgroundTasks, Request, Response, WebSocket]]`? Is the `Any` supposed to be there? I'm unclear how it will affect the hints.

strawberry/fastapi/router.py
```diff
@@ -38,36 +45,34 @@ async def __get_root_value():
...
+            custom_getter: Optional[CustomContext],
             background_tasks: BackgroundTasks,
             request: Request = None,
             response: Response = None,
             ws: WebSocket = None,
-        ) -> Union[
-            BaseContext, Dict[str, Union[Any, BackgroundTasks, Request, WebSocket]]
-        ]:
+        ) -> MergedContext:
+            default_dict = {
```
BryceBeagle, line 57: Perhaps a better var name here would be `default_context`? I was initially confused thinking this was `defaultdict` from the standard library.

strawberry/fastapi/router.py
```diff
@@ -38,36 +45,34 @@ async def __get_root_value():
...
+                    **default_dict,
+                    **custom_getter,
                 }
+            elif custom_getter is None:
+                return default_dict
             else:
-                raise TypeError(
-                    "The custom context dependency is neither a class that inherits from BaseContext nor a dictionary"
-                )
+                raise InvalidCustomContext()
```
BryceBeagle, line 75: Should we pass the invalid context to this exception so that it can use it in the exception string?

tests/fastapi/test_context.py
```diff
@@ -112,9 +118,9 @@ def get_context(value: str = Depends(custom_context_dependency)) -> str:
...
-        response = test_client.post("/graphql", json={"query": "{ abc }"})
-    except TypeError as e:
-        assert str(e) == "The custom context dependency must be either a class or a dictionary"
-        response = "Task failed successfully"
-    assert response == "Task failed successfully"
\ No newline at end of file
+    with pytest.raises(
+        InvalidCustomContext,
+        match="The custom context must be either a class\
+            that inherits from BaseContext or a dictionary",
```
BryceBeagle, line 124: As discussed above, this string is not as-expected because of the `\`

BryceBeagle (changes requested):

strawberry/fastapi/router.py
tsmith023: If they're at the top then `BaseContext` isn't within their scope. Would doing:
```python
CustomContext = Union['BaseContext', Dict[str, Any]]
MergedContext = Union[
    'BaseContext', Dict[str, Union[Any, BackgroundTasks, Request, Response, WebSocket]]
]
```
be viable in this case? Or shall we move BaseContext to its own file, say `context.py` and import it?

tsmith023: @BryceBeagle, the `Any` in that `Union` signifies the completely general type that the user can supply in their context dictionary as their custom value(s).

strawberry/exceptions.py
tsmith023: Sorry about this one, I had it as something else previously but didn't realise they weren't equivalent. Will change!

docs/integrations/fastapi.md
tsmith023: To be clear here, is this a style convention that Strawberry adheres too? Is it also documented in a PEP somewhere? I ask because I prefer whitespace in those instances and would like to know if I'm wrong!

strawberry/fastapi/router.py
tsmith023: It would require `strawberry.exceptions` to import `BaseContext` from `strawberry.fastapi` to correctly type the argument to the exception's `__init__` method. Might that pose a problem of cross-dependency? Perhaps we make an `exceptions.py` file in `strawberry.fastapi`?

docs/integrations/fastapi.md
jkimbo: @tsmith023 We generally follow black formatting in our codebase and in the docs.

BryceBeagle: This is also a PEP-8 recommendation:
https://www.python.org/dev/peps/pep-0008/#other-recommendations

> Don't use spaces around the = sign when used to indicate a keyword argument, or when used to indicate a default value for an unannotated function parameter:
>
> ```python
> # Correct:
> def complex(real, imag=0.0):
>     return magic(r=real, i=imag)
> ```
> ```python
> # Wrong:
> def complex(real, imag = 0.0):
>     return magic(r = real, i = imag)

tsmith023:
@patrick91, @BryceBeagle, sorry for taking so long to make these changes but they're there now! Let me know if they're acceptable or any more are needed. 

There is, however, a bizarre codecov check issue that I simply do not understand. It claims that coverage has been impacted within the `strawberry/experimental/pydantic/object_type.py` file, which has (at least to me)  no relevance to the changes that I have made. Am I missing something here?

strawberry/fastapi/router.py
BryceBeagle: @patrick91 should we be using (enforcing?) `from __future__ import annotation`-style forward references?

BryceBeagle (approved):
Approving, other than the question I left for @patrick91 on a comment

strawberry/fastapi/router.py
patrick91: uhm, let's not for now 😊 I think it is fine to use both styles, at least until we know what will happen to `from __future__ import annotation` :D

patrick91 (approved):

PR #1571, opened 2022-01-11
📚 Add missing imports and extra note on how to run the full example

Ambro17:
Add missing imports and explanation on how to start an ASGI server

## Description

The example at the docs were incomplete and a user gave feedback that she/he didn't manage to get it working.
This PR aims to simplify the setup to have it working just by copy-pasting

### DIFF
--- a/docs/guides/dataloaders.md
+++ b/docs/guides/dataloaders.md
@@ -192,7 +192,7 @@ schema = strawberry.Schema(query=Query)
 app = MyGraphQL(schema)
 ```
 
-Now you can run the server with any async server (ASGI) that you want, for example we can install uvicorn with
+Now you can run the server with any async server ([ASGI](../integrations/asgi.md)) that you want, for example we can install uvicorn with
 ```bash
 pip install uvicorn
 ```
--- a/docs/guides/dataloaders.md
+++ b/docs/guides/dataloaders.md
@@ -88,6 +88,7 @@ Let's see an example of how you can use DataLoaders with GraphQL:
 ```python
 from typing import List
 
+from strawberry.dataloader import DataLoader
 import strawberry
 
 @strawberry.type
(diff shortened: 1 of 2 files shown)


docs/guides/dataloaders.md
```diff
@@ -185,4 +186,17 @@ class Query:
     @strawberry.field
     async def get_user(self, info: Info, id: strawberry.ID) -> User:
         return await info.context["user_loader"].load(id)
+
+
+schema = strawberry.Schema(query=Query)
+app = MyGraphQL(schema)
+```
+
+Now you can run the server with any async server (ASGI) that you want, for example we can install uvicorn with
```
Ambro17, line 195: Is it possible to reference the ASGI section of the docs from here? I think it would be nice to get an updated source of truth in case anything gets outdated

estyxx: I will rather add a note to refer to the Getting Started guide that has the instruction to install strawberry with the debug-server.

Something like this think can be fine :)
```
You can now run the example above with `strawberry server schema:schema`, please refer to  
[Getting Started tutorial](docs/index.md#step-1-create-a-new-project-and-install-strawberry) to understand how to install Strawberry and run the schema. 
```

Or even leave just the example... I can see we don't usually repeat in every section how to install and run the example, hope is pretty obvious maybe? 
What do you think @patrick91 ? 

As in our docs, we use `schema.py` as the schema file's name, I will avoid name it differently for dataloaders.

patrick91: yes, a standard markdown link should work 😊

estyxx: *I guess if you were struggling with that was not so obvious 😅

estyxx: Yeah, on a second thought, it's worth adding a little of repetition, in case somebody missed the getting started page...

patrick91 (approved):
Thanks ✨

PR #1592, opened 2022-01-24
Fix #1504 where the pydantic integration fails when defining input and output types

### LINKED ISSUE #1504: Pydantic  - Usage of nested models with explicitly defined types/inputs- regression in Strawberry 0.92.0
Morning, 

Thanks for the awesome library! :+1: 

As of Strawberry 0.92 defining both, an input and a type for a nested pydantic model produces a confusion error on QUERY side:

>> Expected value of type 'AResponse' but got: <BInput instance>.

e.g. 

```python 
@strawberry.experimental.pydantic.input(model=A)                                                                                                                  
class AInput:                                                                                                                                                     
    a: strawberry.auto                                                                                                                                            
                                                                                                                                                                  
@strawberry.experimental.pydantic.input(model=B)                                                                                                                  
class BInput:                                                                                                                                                     
    b: strawberry.auto                                                                                                                                            
    a: strawberry.auto                                                                                                                                            
                                                                                                                                                                  
@strawberry.experimental.pydantic.type(model=A)                                                                                                                   
class AResponse:                                                                                                                                                     
     a: strawberry.auto                                                                                                                                            
                                                                                                                                                                  
@strawberry.experimental.pydantic.type(model=B)                                                                                                                   
class BResponse:                                                                                                                                                  
    b: strawberry.auto                                                                                                                                            
    a: strawberry.auto                                                                                                                                            
   
```

This may be a regression in behavior(the error got caught and exposed), rather than implementation.

Would be willing to contribute a PR if you could confirm that this is unexpected behaviour and not something that is intentional awaiting a rework down the line. Have a great weekend!

Matt343:
## Description

We were using a single class variable to store both the Strawberry output type and input type associated with a given Pydantic base class. This meant that we would get incorrect type definitions (input types referencing output types in their fields and vice versa) depending on the order the Strawberry types were defined in.

I added a second variable to track the Strawberry input type and updated the usages of the variable to choose the correct one based on whether we are defining an input type or not.

Note that error types have the same problem in main, but they are also just pretty broken right now so I didn't address that here.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #1504 

## Checklist

### DIFF
--- a/strawberry/experimental/pydantic/object_type.py
+++ b/strawberry/experimental/pydantic/object_type.py
@@ -32,14 +32,14 @@ from strawberry.types.types import TypeDefinition
 from .exceptions import MissingFieldsListError, UnregisteredTypeException
 
 
-def replace_pydantic_types(type_: Any):
+def replace_pydantic_types(type_: Any, is_input: bool):
     origin = getattr(type_, "__origin__", None)
     if origin is Literal:
         # Literal does not have types in its __args__ so we return early
         return type_
     if hasattr(type_, "__args__"):
         replaced_type = type_.copy_with(
-            tuple(replace_pydantic_types(t) for t in type_.__args__)
+            tuple(replace_pydantic_types(t, is_input) for t in type_.__args__)
         )
 
         if isinstance(replaced_type, TypeDefinition):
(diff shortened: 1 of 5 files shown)


botberry:



---

Fix bug #1504 in the Pydantic integration, where it was impossible to define
both an input and output type based on the same Pydantic base class.


---




Here's the tweet text: 
```

```

strawberry/experimental/pydantic/object_type.py
```diff
@@ -67,18 +67,24 @@ def replace_pydantic_types(type_: Any):
...
+            if hasattr(type_, "_strawberry_input_type"):
+                return type_._strawberry_input_type
+            else:
+                raise UnregisteredTypeException(type_)
         else:
-            raise UnregisteredTypeException(type_)
+            if hasattr(type_, "_strawberry_type"):
+                return type_._strawberry_type
+            else:
+                raise UnregisteredTypeException(type_)
```
patrick91, line 79: nit; can we do something like: `attr = "_strawberry_input_type" if is_input else "_strawberry_type"`?

tests/experimental/pydantic/test_basic.py
```diff
@@ -539,3 +564,62 @@ class UserType:
...
+        groups: List[GroupOutput]
+
+    @strawberry.type
+    class Mutation:
+        @strawberry.mutation
+        def updateGroup(group: GroupInput) -> GroupOutput:
+            pass
+
+    # This triggers the exception from #1504
+    strawberry.Schema(query=Query, mutation=Mutation)
```
patrick91, line 618: 👍 

shall we also assert the schema string since we create one? :)

patrick91 (commented):
Looks great 👏  I've added two comments, after resolving them we can merge 😊

strawberry/experimental/pydantic/object_type.py
Matt343: Good call

patrick91 (approved):
Awesome! Thanks for updating the PR so quickly!

PR #1601, opened 2022-01-27
Upgrade to graphql-core 3.2

### LINKED ISSUE #1599: Update to graphql-core 3.2

patrick91:
## Description

This PR updates GraphQL core to 3.2. There were some breaking and type changes which we should document better 🤔 

Also, the type update I've done in `info.py` doesn't feel right, I think we should keep list (or the correct actual type)

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* Closes #1599


botberry:



---

This release updates `graphql-core` to `3.2.0`


---




Here's the tweet text: 
```

```

strawberry/types/info.py
```diff
@@ -36,15 +45,15 @@ def schema(self) -> "Schema":
...
+    def field_nodes(self) -> Collection[FieldNode]:  # deprecated
         warnings.warn(
             "`info.field_nodes` is deprecated, use `selected_fields` instead",
             DeprecationWarning,
         )
         return self._raw_info.field_nodes
 
     @cached_property
-    def selected_fields(self) -> List[Selection]:
+    def selected_fields(self) -> Collection[Selection]:
```
DoctorJohn, line 56:
```suggestion
    def selected_fields(self) -> List[Selection]:
```

Our `convert_selections` function returns `List[Selection]`, so we could continue using `List[Selection]` here.

https://github.com/strawberry-graphql/strawberry/blob/b50c87d7d53b4a099151101889d69a4d23c440d5/strawberry/types/nodes.py#L64-L79

strawberry/types/info.py
```diff
@@ -36,15 +45,15 @@ def schema(self) -> "Schema":
         return self._raw_info.schema._strawberry_schema  # type: ignore
 
     @property
-    def field_nodes(self) -> List[FieldNode]:  # deprecated
+    def field_nodes(self) -> Collection[FieldNode]:  # deprecated
```
DoctorJohn, line 48: Not sure why `GraphQLResolveInfo.field_nodes` is typed as `Collection[FieldNode]` in graphql-core 🤔 It appears whenever they construct a `GraphQLResolveInfo` instance their field nodes are lists.

strawberry/extensions/directives.py
```diff
@@ -18,7 +18,7 @@ async def resolve(
     ) -> AwaitableOrValue[Any]:
         result = await await_maybe(_next(root, info, *args, **kwargs))
 
-        for directive in info.field_nodes[0].directives:
+        for directive in info.field_nodes[0].directives:  # type: ignore
```
DoctorJohn, line 21: Do we ignore the type here because `info.field_nodes` now is a `Collection[FieldNode]` and we can't be sure `info.field_nodes[0]` is actually a single `FieldNode`? Maybe we should ask @Cito why they went with `Collection[FieldNode]` instead of `List[FieldNode]` 🤔

strawberry/schema/schema.py
```diff
@@ -72,9 +72,9 @@ def __init__(
             else None
         )
 
-        graphql_directives = [
+        graphql_directives = tuple(
```
DoctorJohn, line 75: can confirm this does what it's supposed to do 👍

strawberry/extensions/directives.py
Cito: Good question @DoctorJohn. I think it was a mistake, and reverted the change in the main branch of graphql-core already.

patrick91: Thanks @Cito for responding and thanks @DoctorJohn for reviewing, will update the PR soon :)

tests/aiohttp/test_graphql_transport_ws.py
```diff
@@ -341,7 +341,7 @@ async def test_subscription_syntax_error(aiohttp_client):
         assert response["type"] == ErrorMessage.type
         assert response["id"] == "sub1"
         assert len(response["payload"]) == 1
-        assert response["payload"][0]["path"] is None
+        assert response["payload"][0].get("path") is None
```
BryceBeagle, line 344: Why did these get changed to `.get()`? There are two ways this can return `None`, so we should make sure we're checking just one of the means.

tests/schema/test_schema_generation.py
```diff
@@ -67,9 +68,11 @@ class Query:
 def test_custom_execution_context():
     class CustomExecutionContext(GraphQLExecutionContext):
         def build_response(
-            self, data: AwaitableOrValue[Optional[Dict[str, Any]]]
+            self,
+            data: AwaitableOrValue[Optional[Dict[str, Any]]],
+            errors: List[GraphQLError],
         ) -> AwaitableOrValue[ExecutionResult]:
-            result = cast(ExecutionResult, super().build_response(data))
+            result = cast(ExecutionResult, super().build_response(data, errors))
```
BryceBeagle, line 75: Not related to this change, but why is this `cast` here?

tests/aiohttp/test_graphql_transport_ws.py
patrick91: `path` is not always available anymore: https://github.com/graphql-python/graphql-core/blob/main/src/graphql/error/graphql_error.py#L222-L237

pyproject.toml
```diff
@@ -101,6 +101,7 @@ types-aiofiles = "^0.8.3"
 sanic-testing = "^0.8"
 fastapi = {version = ">=0.65.0", optional = false}
 pytest-xprocess = "^0.18.1"
+MarkupSafe = "2.0.1"
```
patrick91, line 104: I had to pin this for until we update flask to latest, markupsafe 2.1.0 has some breaking changes and flask 1.0 requires jinja <3 which isn't being maintaned anymore: https://github.com/pallets/jinja/issues/1585

strawberry/subscriptions/protocols/graphql_ws/handlers.py
```diff
@@ -114,14 +114,14 @@ async def handle_start(self, message: OperationMessage) -> None:
             )
         except GraphQLError as error:
             error_payload = format_graphql_error(error)
-            await self.send_message(GQL_ERROR, operation_id, error_payload)
+            await self.send_message(GQL_ERROR, operation_id, error_payload)  # type: ignore # noqa: E501
```
patrick91, line 117: @DoctorJohn can you take a look at these ignores? :)

DoctorJohn: Just took a look at it locally, we don't need those ignores. mypy and pytest agree :)

patrick91: uhm, I do get an error when removing it: 

```
strawberry/subscriptions/protocols/graphql_ws/handlers.py:117: error: Argument 3 to "send_message" of "BaseGraphQLWSHandler" has incompatible type "GraphQLFormattedError"; expected "Union[Dict[str, Any], Dict[str, Any], StartPayload, DataPayload, ErrorPayload, None]"  [arg-type]
```

DoctorJohn: Lemme take another look 👀

DoctorJohn: @patrick91 Silly mistake by me. Just fixed it and pushed to this branch (hope thats okay). I should take some time and refactor the legacy protocols typings. For the new protocol I used dataclasses instead of typed dicts which worked really well.

patrick91: no worries! and thanks for fixing it! 😊 might be worth open an issue for the refactoring, so we know that it will happen at some point 😊

tests/schema/test_schema_generation.py
patrick91: fixed 😊

PR #1604, opened 2022-01-30
Add GraphQLTestClient for aiohttp

estyxx:
## Description

As for `Django` and `asgi` I'm adding the `GraphQLTestClient` for `aiohttp`

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

*

## Checklist


botberry:



---

This release introduce the `GraphQLTestClient` also for `aiohttp` integration (was already availabe for `asgi` and `Django`). It hides the http request's details and asserts that there are no errors in the response (you can always disable this behavior by passing `asserts_errors=False`). This makes it easier to test queries and makes your tests cleaner.

If you are using `pytest` you can add a fixture in `conftest.py`

```python
import pytest

from strawberry.aiohttp.test.client import GraphQLTestClient

@pytest.fixture
def graphql_client(aiohttp_client, myapp):
    yield GraphQLTestClient(aiohttp_client(myapp))
```

And use it everywere in your test methods

```python
def test_strawberry(graphql_client):
    query = """
        query Hi($name: String!) {
            hi(name: $name)
        }
    """

    result = graphql_client.query(query, variables={"name": "🍓"})

    assert result.data == {"hi": "Hi 🍓!"}
```


---




Here's the tweet text: 
```

```

RELEASE.md
```diff
@@ -0,0 +1,30 @@
+Release type: minor
+
+This release introduce the `GraphQLTestClient` also for `aiohttp` integration (was already availabe for `asgi` and `Django`). It hides the http request's details and asserts that there are no errors in the response (you can always disable this behavior by passing `asserts_errors=False`). This makes it easier to test queries and makes your tests cleaner.
```
```suggestion
This release adds an implementation of the `GraphQLTestClient` for the `aiohttp` integration (in addition to the existing `asgi` and `Django` support). It hides the HTTP request's details and verifies that there are no errors in the response (this behavior can be disabled by passing `asserts_errors=False`). This makes it easier to test queries and makes your tests cleaner.
```

Would also suggest wrapping this a bit

RELEASE.md
```diff
@@ -0,0 +1,30 @@
...
+import pytest
+
+from strawberry.aiohttp.test.client import GraphQLTestClient
+
+@pytest.fixture
+def graphql_client(aiohttp_client, myapp):
+    yield GraphQLTestClient(aiohttp_client(myapp))
+```
+
+And use it everywere in your test methods
```
BryceBeagle, line 17:
```suggestion
And use it everywhere in your tests
```

BryceBeagle (changes requested):

strawberry/aiohttp/test/client.py
```diff
@@ -0,0 +1,42 @@
+from typing import Dict, Mapping, Optional
+
+from strawberry.test.client import BaseGraphQLTestClient, Response
+
+
+class GraphQLTestClient(BaseGraphQLTestClient):
+    async def query(
```
DoctorJohn, line 7: mypy is unhappy because `query` and `request` violate the liskov substitution principle. We could fix this by adjusting the return types in the `BaseGraphQLTestClient` :)

DoctorJohn (changes requested):
Looks good overall! I like the idea of a single test client for all integrations. But I must admit I'm still undecided whether I like the 'syntax' we use for file uploads. But maybe I'm a little biased in this regard, since i did it differently in my own graphql client library.

strawberry/aiohttp/test/client.py
estyxx: I should have fixed it! thanks :)

DoctorJohn (approved):

BryceBeagle (approved):

PR #1624, opened 2022-02-02
Custom pydantic conversions

### LINKED ISSUE #1617: Feature Request: provide a mechanism to change the conversion from Pydantic to Strawberry
Currently, we use a registration mechanism and some basic rules to figure out how a type should be converted from Pydantic to Strawberry and vice versa. The rules map scalars to their equivalent scalar using some hard-coded mappings, then recursively build up more complex types from there.

This is all great for automatic conversion, but there are use cases where I'd like to be able to provide a registered Strawberry type that doesn't follow these rules, and instead uses some custom logic to convert to and from the Pydantic type. 

An example of when this would be useful is if I'm defining a custom scalar that I want to be used in some of the fields of the model:

```python
class MyModel(pydantic.BaseModel):
    my_field: int

CustomInt = strawberry.scalar(NewType("CustomInt", int), ...)

@strawberry.experimental.pydantic.type(model=MyModel)
class MyStrawberryModel:
    my_field: strawberry.auto  # I want this to be wrapped in CustomInt during the conversion so it uses my custom serializer logic
```

Another time this would be helpful is when the Pydantic model represents some type or value that I don't want to translate directly to GraphQL types, either because it is cumbersome/undesirable, or because it is actually not representable in GraphQL. An example of the latter case would be anything involving unions of scalar fields, which GraphQL doesn't know how to represent. If I could provide custom conversion logic, I could keep the union type in Pydantic but provide some alternative concrete type in Strawberry land.

```python
class MyModel(pydantic.BaseModel):
    my_field: Union[int, str]

@strawberry.experimental.pydantic.type(model=MyModel)
class MyStrawberryModel:
    my_field: strawberry.auto  # I want this to be mapped to a specific type in Strawberry
```

The overall goal here is to be able to support existing (possibly highly used and constrained already) Pydantic models in Strawberry with more flexibility. There are definitely cases where it's helpful to be able to change the data representation during conversion beyond these, this is just what I've run into during my implementation. The problem today is that if you want to do custom conversions/types for one piece of a deeply nested hierarchy, the entire conversion process from Pydantic fails because the types aren't registered as being related to the Pydantic models.

Matt343:
## Description

Add support for overriding the `from_pydantic` and `to_pydantic` methods on Pydantic converted types. This allows users to have complete control over the behavior of the conversions if needed, while still preserving the type references that we set up when you register a Pydantic type.

I also fixed up the behavior for when you use a field with the same name as one of the Pydantic fields but use a different type. When you do this, we call the constructor of the type annotation on the value we get from Pydantic before returning it in a query. This is a handy shortcut for using scalar types in Pydantic models or handling fields like `Union[int, str]` in a quick way. The problem before was that it was sort of a happy accident that this worked at all, and it didn't respect aliases and didn't work on objects with only one field. I fixed this behavior so it is more consistent in these cases and added documentation so this is an expected feature now.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [x] New feature
- [x] Documentation

* Fixes #1617 

## Checklist


botberry:



---

Add better support for custom Pydantic conversion logic and standardize the
behavior when not using `strawberry.auto` as the type.

See https://strawberry.rocks/docs/integrations/pydantic#custom-conversion-logic for details and examples.


---




Here's the tweet text: 
```

```

tests/experimental/pydantic/test_basic.py
```diff
@@ -658,3 +659,45 @@ def updateGroup(group: GroupInput) -> GroupOutput:
...
+        password: strawberry.auto
+
+    definition: TypeDefinition = User._type_definition
+    assert definition.name == "User"
+
+    [field1, field2] = definition.fields
+
+    assert field1.python_name == "age_"
+    assert field1.type is str
+    assert field1.graphql_name == "age"
```
thejaminator, line 699: i don't think this should block your MR. 

but what if i don't want to use the alias in the graphql name?

i was previously providing the type manually so that it wouldn't do that. but i agree its not nice.

we'll handle the alias setting here? https://github.com/strawberry-graphql/strawberry/pull/1546

docs/integrations/pydantic.md
```diff
@@ -313,3 +313,119 @@ class Query:
...
+
+@strawberry.experimental.pydantic.type(model=User)
+class UserType:
+    id: strawberry.auto
+    # Flatten the content dict into specific fields in the query
+    content_name: Optional[str] = None
+    content_description: Optional[str] = None
+
+    @staticmethod
+    def from_pydantic(user: User, extra: Dict[str, Any] = None) -> "UserType":
```
thejaminator, line 406: i'm not sure whether we should enforce / encourage that the signature should have the keywords "instance", and "extra" 
as specified in the protocol `StrawberryTypeFromPydantic`. 

```suggestion
    def from_pydantic(instance: User, extra: Dict[str, Any] = None) -> "UserType":
```

currently conversion logic wise, it shouldn't affect anything since no code calls from_pydantic. 
but both mypy and pycharm will think that the keyword is "instance" instead of "user" here. 
so people may pass in the wrong keyword.

thejaminator: image

tests/experimental/pydantic/test_basic.py
Matt343: Yeah I think having a dedicated option to control that is much better

docs/integrations/pydantic.md
Matt343: Fixed the example at least, I'm not sure if we should enforce it though since using another name does technically work aside from IDEs.

thejaminator:
@Matt343 theres also the case where the custom conversion is a nested inside another basemodel. 
should the parent call the custom from_pydantic method of the child? (currently it seems like it doesn't)


```python
def test_custom_conversion_functions_nested():
    class User(BaseModel):
        age: int
        password: Optional[str]

    @strawberry.experimental.pydantic.type(User)
    class UserType:
        age: str
        password: strawberry.auto

        @staticmethod
        def from_pydantic(user: User, extra: Dict[str, Any] = None) -> "UserType":
            return UserType(
                age=str(user.age),
                password=base64.b64encode(user.password.encode()).decode()
                if user.password
                else None,
            )

        def to_pydantic(self) -> User:
            return User(
                age=int(self.age),
                password=base64.b64decode(self.password.encode()).decode()
                if self.password
                else None,
            )

    class Parent(BaseModel):
        user: User

    @strawberry.experimental.pydantic.type(Parent)
    class ParentType:
        user: strawberry.auto

    user = User(age=1, password="abc")
    user_strawberry = UserType.from_pydantic(user)

    assert user_strawberry.age == "1"
    assert user_strawberry.password == "YWJj"

    user_pydantic = user_strawberry.to_pydantic()
    assert user == user_pydantic

    parent = Parent(user=user)
    parent_strawberry = ParentType.from_pydantic(parent)
    assert parent_strawberry.user.age == "1" # this fails
    assert parent_strawberry.user.password == "YWJj" # this fails

```

Matt343:
> @Matt343 theres also the case where the custom conversion is a nested inside another basemodel. should the parent call the custom from_pydantic method of the child? (currently it seems like it doesn't)

Good catch, I've added tests for this and revealed a bug in my initial implementation around interfaces. I've changed the method used to override the conversion functions to be more explicit so we can tell the difference between an inherited method and a user supplied one.

thejaminator:
> > @Matt343 theres also the case where the custom conversion is a nested inside another basemodel. should the parent call the custom from_pydantic method of the child? (currently it seems like it doesn't)
> 
> Good catch, I've added tests for this and revealed a bug in my initial implementation around interfaces. I've changed the method used to override the conversion functions to be more explicit so we can tell the difference between an inherited method and a user supplied one.

ah, i liked the previous way of just defining the to_pydantic and from_pydantic in the class rather than supplying in the decorator. was the change solely due to not being able to tell the difference? i can try it out

Matt343:
> > > @Matt343 theres also the case where the custom conversion is a nested inside another basemodel. should the parent call the custom from_pydantic method of the child? (currently it seems like it doesn't)
> > 
> > 
> > Good catch, I've added tests for this and revealed a bug in my initial implementation around interfaces. I've changed the method used to override the conversion functions to be more explicit so we can tell the difference between an inherited method and a user supplied one.
> 
> ah, i liked the previous way of just defining the to_pydantic and from_pydantic in the class rather than supplying in the decorator. was the change solely due to not being able to tell the difference? i can try it out

Yeah basically you can't tell if the decorated class is inheriting the methods (in which case we need to redefine them in order for interfaces to work properly) or if they are user defined. Because of that you don't know if you should override them with the default methods or not.

thejaminator:
> Yeah basically you can't tell if the decorated class is inheriting the methods (in which case we need to redefine them in order for interfaces to work properly) or if they are user defined. Because of that you don't know if you should override them with the default methods or not.

hmmm yes i'm not sure for interfaces because i think we could just disallow users to declare from_pydantic and to_pydantic on interfaces because it doesn't make sense?
but yes inheritance from another strwaberry type is tricky

Matt343:
> > Yeah basically you can't tell if the decorated class is inheriting the methods (in which case we need to redefine them in order for interfaces to work properly) or if they are user defined. Because of that you don't know if you should override them with the default methods or not.
> 
> hmmm yes i'm not sure for interfaces because i think we could just disallow users to declare from_pydantic and to_pydantic on interfaces because it doesn't make sense? but yes inheritance from another strawberry.type is tricky

I think you could use the `__qualname__` property on the function to tell if it comes from a superclass, bu that seems like a bit of a hack

thejaminator:
> > > Yeah basically you can't tell if the decorated class is inheriting the methods (in which case we need to redefine them in order for interfaces to work properly) or if they are user defined. Because of that you don't know if you should override them with the default methods or not.
> > 
> > 
> > hmmm yes i'm not sure for interfaces because i think we could just disallow users to declare from_pydantic and to_pydantic on interfaces because it doesn't make sense? but yes inheritance from another strawberry.type is tricky
> 
> I think you could use the `__qualname__` property on the function to tell if it comes from a superclass, bu that seems like a bit of a hack

if we do that hack, we can raise an exception if the user doesn't redefine the method in the child class. otherwise its not intuitive whether the child class will use the parent's method or the default one. just thinking :)

Matt343:
> if we do that hack, we can raise an exception if the user doesn't redefine the method in the child class. otherwise its not intuitive whether the child class will use the super method or the default one. just thinking :)

Not sure I follow this, you'd want to just detect if the supplied method was defined on the decorated class or a superclass. If it's from a superclass, we can safely override with the default, otherwise we would leave it. Not sure where we would want to raise something.

Overall, that method seems a bit more implicit to me, there's more weird rules about where executed code comes from and that could be confusing.

thejaminator:
> Not sure I follow this, you'd want to just detect if the supplied method was defined on the decorated class or a superclass. If it's from a superclass, we can safely override with the default, otherwise we would leave it. Not sure where we would want to raise something.

i think we'll talking about the same thing. I'm suggesting that to avoid the implicitness. we don't override with the default child classes if the parent implements it. we error out at class definition time saying that the parent class has a manual from_pydantic, so the child class needs to have a manual from_pydantic too.

```python
    @strawberry.experimental.pydantic.type(User)
    class UserType:
        age: str
        password: strawberry.auto
    
        @staticmethod
        def from_pydantic(user: User, extra: Dict[str, Any] = None) -> "UserType":
            ...
        
    class AnotherUserType(UserType):
        another_field: str 
        
    AnotherUserType.from_pydantic(something)
    # am i calling the default from_pydantic or the parent UserType's from_pydantic
    # not clear even though we can leave it as the default
    # so we'll raise an error at class definition time that they should manually define it

```

Matt343:
> i think we'll talking about the same thing. I'm suggesting that to avoid the implicitness. we don't override with the default child classes if the parent implements it. we error out at class definition time saying that the parent class has a manual from_pydantic, so the child class needs to have a manual from_pydantic too.
> 
> which i think is reasonable enough. because if the parent has a manual from_pydantic, the child probably needs one too?
> 
> ```python
>     @strawberry.experimental.pydantic.type(User)
>     class UserType:
>         age: str
>         password: strawberry.auto
>     
>         @staticmethod
>         def from_pydantic(user: User, extra: Dict[str, Any] = None) -> "UserType":
>             ...
>         
>     class AnotherUserType(UserType):
>         another_field: str 
>         
>     AnotherUserType.from_pydantic(something)
>     # am i calling the default from_pydantic or the parent UserType's from_pydantic
>     # not clear even though we can leave it as the default
>     # so we'll raise an error at class definition time that they should manually define it
> ```

Ah interesting, the issue I see is that we don't actually know if the parent class method is the default or a custom override, because if you just define a normal interface with no custom method it will still have the methods defined by the time we get to the child class. I guess we could add a property to the default methods that would let us know when we find them...

PR #1644, opened 2022-02-14
Support using docstrings as GraphQL descriptions

paulo-raca:
## Description

The most "pythonic" way of adding documentation to classes/functions is using docstrings, and it is well supported by all kinds of tools.

This PR modifies Strawberry to use support it too, adding the docstrings as GraphQL descriptions (Unless a description is explicitly provided with strawberry decorators)

It supports object types (type/input/interface), enums and fields with resolvers.

Unfortunately python attributes do not support docstrings, so regular dataclass fields (fields not backed by resolvers) will need to continue using `@strawberry.field(description=...)`

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Enhancement/optimization
- [ ] Documentation

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the CONTRIBUTING document.
- [x] I have added tests to cover my changes.
- [x] I have tested the changes and verified that they work and don't break anything (as well as I can manage).

Issue #653, opened 2021-01-09
Implicit description from class docstring


AadamZ5:
Long story short, in my opinion,

This, (my class used for example)
```python3
...
@strawberry.type(description="Represents data about a device that is currently connected to the application")
class ActiveHdd(HddEntry):
    serial: ID
...
```

Should be equivalent to this,
```python3
...
@strawberry.type
class ActiveHdd(HddEntry):
    """
    Represents data about a device that is currently connected to the application
    """
    serial: ID
...
```

We can gather this from the `__doc__` attribute on the class, or a clean sanitized string using `getdoc` from the `inspect` module,
```python3
from inspect import getdoc

class MyClass:
    """
    My class that should have a document string.
    """
    pass

assert getdoc(MyClass) == 'My class that should have a document string.'
```

I believe the description should be gathered from the docstring if no description parameter is explicitly set.

What do you think? I love the project so far! I hope to contribute more at some point.

Thanks!

patrick91:
Hi @AadamZ5! That's something we did at the very beginning but we removed this functionality because it was easy to forget about and expose implementation details (something that happened at work if I'm honest 😊).

What about using a new property on the class?

```
@strawberry.type
class ActiveHdd:
    __description__ = """
    Represents data about a device that is currently connected to the application
    """
    serial: ID
```

Pinging @jkimbo, @BryceBeagle and @ulgens for more opinions on this :)

Ambro17:
I like the new attribute a little more, since the docstring may be oriented to developers and leak some implementation details.
At work we had a similar discussion and agreed with the special attribute approach.
If you want to take the docstring as the field description you explicitly set `__description__ = __doc__` and you get the desired behavior, without the risk of doing it accidentally

AadamZ5:
The extra attribute would work, but truly I see it as redundant to the `description=` parameter in the decorator. The only reason I thought it would be a good idea was because it is a straightforward and implicit way to specify the description without actually specifying any explicit parameters or attributes. I love me some implicit-ness.

I think that if the description is gathered from `__doc__` as a _fallback_, then we can avoid exposing any implementation details by explicitly setting `description=` in the decorator.

Maybe I should rethink how my class's docstring is being used. Perhaps I am not using it how it _should_ be used. Regardless, this isn't a bug or issue.

What do you think?

BryceBeagle:
> I think that if the description is gathered from `__doc__` as a _fallback_

[...]

PR #1645, opened 2022-02-15
Support generating GraphQL descriptions from docstrings

paulo-raca:
The most "pythonic" way of adding documentation to classes/functions is using docstrings, and it is well supported by all kinds of tools.

This PR modified Strawberry to support it too, as discussed on #653

This PR superseeds #1644

## Description

- In order to maintain backwards compatibility and avoid leaking implementation details, this feature needs to be enabled with `StrawberryConfig(description_from_docstrings=True)`
- Docstrings are parsed with [docstring_parser](https://github.com/rr-/docstring_parser/tree/master/docstring_parser), which allows inferring documentation for each resolver argument / enum value / dataclass field

## Example

Here is an example of using docstrings in types and fields:

```
@strawberry.enum
class EnumType(Enum):
    """
    Example enum

    Attributes:
        FOO: Some description
        BAR: Another description
    """

    FOO = "foo"
    BAR = "bar"

@strawberry.type
class Query:
    """
    The main GraphQL type

    Attributes:
        enum: A dataclass field
    """
    enum: EnumType = EnumType.BAR

    @strawberry.field
    def resolver(self, arg1: enum, arg2: int) -> int:
        """
        A GraphQL field with a resolver and arguments

        Params:
            arg1: An enum argument
            arg2: An int argument
        """
        return 1
```

It produces this GraphQL schema:

```
"""Example enum"""
enum EnumType {
  """Some description"""
  FOO

  """Another description"""
  BAR
}

"""The main GraphQL type"""
type Query {
  """A dataclass field"""
  enum: EnumType!

  """A GraphQL field with a resolver and arguments"""
  resolver(
    """An enum argument"""
    arg1: EnumType!

    """An int argument"""
    arg2: Int!
  ): Int!
}
```

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

* #653

## Checklist

- [ ] I have updated the documentation accordingly. (TODO)

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,73 @@
+Release type: minor
+
+Support using docstrings are GraphQL descriptions
+
+It is now possible to use python docstrings to provide GraphQL descriptions.
+
+## Example
+
+Here is an example of using docstrings in types and fields:
+
+```
+@strawberry.enum
+class EnumType(Enum):
+    """
+    Example enum
+
+    Attributes:
+        FOO: Some description
+        BAR: Another description
+    """
+
+    FOO = "foo"
+    BAR = "bar"
+
+@strawberry.type
+class Query:
+    """
+    The main GraphQL type
+
+    Attributes:
+        enum: A dataclass field
+    """
+    enum: EnumType = EnumType.BAR
+
+    @strawberry.field
+    def resolver(self, arg1: enum, arg2: int) -> int:
+        """
+        A GraphQL field with a resolver and arguments
+
+        Params:
+            arg1: An enum argument
+            arg2: An int argument
+        """
+        return 1
+```
+
+It produces this GraphQL schema:
+
+```
+"""Example enum"""
+enum EnumType {
+  """Some description"""
+  FOO
+
+  """Another description"""
+  BAR
+}
+
+"""The main GraphQL type"""
+type Query {
+  """A dataclass field"""
+  enum: EnumType!
+
+  """A GraphQL field with a resolver and arguments"""
+  resolver(
+    """An enum argument"""
+    arg1: EnumType!
+
+    """An int argument"""
+    arg2: Int!
+  ): Int!
+}
+```
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,6 +50,7 @@ aiohttp = {version = "^3.7.4.post0", optional = true}
 fastapi = {version = ">=0.65.2", optional = true}
 sentinel = "^0.3.0"
 "backports.cached-property" = "^1.0.1"
+docstring-parser = "^0.13"
 
 [tool.poetry.dev-dependencies]
 pytest = "^7.0"
--- a/strawberry/schema/config.py
+++ b/strawberry/schema/config.py
@@ -9,6 +9,7 @@ from .name_converter import NameConverter
 class StrawberryConfig:
     auto_camel_case: InitVar[bool] = None  # type: ignore
     name_converter: NameConverter = field(default_factory=NameConverter)
+    description_from_docstrings: bool = False
 
     def __post_init__(
         self,
--- a/strawberry/schema/schema_converter.py
+++ b/strawberry/schema/schema_converter.py
@@ -39,7 +39,9 @@ from strawberry.type import StrawberryList, StrawberryOptional, StrawberryType
 from strawberry.types.info import Info
 from strawberry.types.types import TypeDefinition
 from strawberry.union import StrawberryUnion
+from strawberry.utils import docstrings
 from strawberry.utils.await_maybe import await_maybe
+from strawberry.utils.docstrings import Docstring
 
 from . import compat
 from .types.concrete_type import ConcreteType
@@ -68,7 +70,9 @@ class GraphQLCoreConverter:
         self.config = config
         self.scalar_registry = scalar_registry
 
-    def from_argument(self, argument: StrawberryArgument) -> GraphQLArgument:
+    def from_argument(
+        self, argument: StrawberryArgument, parent_docstring: Optional[Docstring]
+    ) -> GraphQLArgument:
         argument_type: GraphQLType
 
         if isinstance(argument.type, StrawberryOptional):
@@ -78,10 +82,16 @@ class GraphQLCoreConverter:
 
         default_value = Undefined if argument.default is UNSET else argument.default
 
+        description = argument.description
+        if description is None and self.config.description_from_docstrings:
+            description = docstrings.arg_description(
+                parent_docstring, argument.python_name
+            )
+
         return GraphQLArgument(
             type_=argument_type,
             default_value=default_value,
-            description=argument.description,
+            description=description,
             deprecation_reason=argument.deprecation_reason,
         )
 
@@ -96,10 +106,17 @@ class GraphQLCoreConverter:
             assert isinstance(graphql_enum, CustomGraphQLEnumType)  # For mypy
             return graphql_enum
 
+        description = enum.description
+        docstring = enum.docstring
+        if description is None and self.config.description_from_docstrings:
+            description = docstrings.type_description(docstring)
+
         graphql_enum = CustomGraphQLEnumType(
             name=enum_name,
-            values={item.name: self.from_enum_value(item) for item in enum.values},
-            description=enum.description,
+            values={
+                item.name: self.from_enum_value(item, docstring) for item in enum.values
+            },
+            description=description,
         )
 
         self.type_map[enum_name] = ConcreteType(
@@ -108,15 +125,27 @@ class GraphQLCoreConverter:
 
         return graphql_enum
 
-    def from_enum_value(self, enum_value: EnumValue) -> GraphQLEnumValue:
-        return GraphQLEnumValue(enum_value.value)
+    def from_enum_value(
+        self, enum_value: EnumValue, parent_docstring: Optional[Docstring] = None
+    ) -> GraphQLEnumValue:
+        description = enum_value.description
+        if description is None and self.config.description_from_docstrings:
+            description = docstrings.attribute_description(
+                parent_docstring, enum_value.name
+            )
+        return GraphQLEnumValue(enum_value.value, description=description)
 
     def from_directive(self, directive: StrawberryDirective) -> GraphQLDirective:
+        description = directive.description
+        docstring = directive.docstring
+        if description is None and self.config.description_from_docstrings:
+            description = docstrings.type_description(docstring)
+
         graphql_arguments = {}
 
         for argument in directive.arguments:
             argument_name = self.config.name_converter.from_argument(argument)
-            graphql_arguments[argument_name] = self.from_argument(argument)
+            graphql_arguments[argument_name] = self.from_argument(argument, docstring)
 
         directive_name = self.config.name_converter.from_type(directive)
 
@@ -124,10 +153,12 @@ class GraphQLCoreConverter:
             name=directive_name,
             locations=directive.locations,
             args=graphql_arguments,
-            description=directive.description,
+            description=description,
         )
 
-    def from_field(self, field: StrawberryField) -> GraphQLField:
+    def from_field(
+        self, field: StrawberryField, parent_docstring: Optional[Docstring]
+    ) -> GraphQLField:
         field_type: GraphQLType
 
         if isinstance(field.type, StrawberryOptional):
@@ -142,22 +173,34 @@ class GraphQLCoreConverter:
             subscribe = resolver
             resolver = lambda event, *_, **__: event  # noqa: E731
 
+        description = field.description
+        docstring = field.docstring
+        if description is None and self.config.description_from_docstrings:
+            if docstring is not None:
+                description = docstrings.func_description(docstring)
+            else:
+                description = docstrings.attribute_description(
+                    parent_docstring, field.python_name
+                )
+
         graphql_arguments = {}
         for argument in field.arguments:
             argument_name = self.config.name_converter.from_argument(argument)
-            graphql_arguments[argument_name] = self.from_argument(argument)
+            graphql_arguments[argument_name] = self.from_argument(argument, docstring)
 
         return GraphQLField(
             type_=field_type,
             args=graphql_arguments,
             resolve=resolver,
             subscribe=subscribe,
-            description=field.description,
+            description=description,
             deprecation_reason=field.deprecation_reason,
             extensions={"python_name": field.python_name},
         )
 
-    def from_input_field(self, field: StrawberryField) -> GraphQLInputField:
+    def from_input_field(
+        self, field: StrawberryField, parent_docstring: Optional[Docstring]
+    ) -> GraphQLInputField:
         field_type: GraphQLType
 
         if isinstance(field.type, StrawberryOptional):
@@ -172,10 +215,16 @@ class GraphQLCoreConverter:
         else:
             default_value = field.default_value
 
+        description = field.description
+        if description is None and self.config.description_from_docstrings:
+            description = docstrings.attribute_description(
+                parent_docstring, field.python_name
+            )
+
         return GraphQLInputField(
             type_=field_type,
             default_value=default_value,
-            description=field.description,
+            description=description,
             deprecation_reason=field.deprecation_reason,
         )
 
@@ -190,19 +239,24 @@ class GraphQLCoreConverter:
             assert isinstance(graphql_object_type, GraphQLInputObjectType)  # For mypy
             return graphql_object_type
 
+        description = type_definition.description
+        docstring = type_definition.docstring
+        if description is None and self.config.description_from_docstrings:
+            description = docstrings.type_description(docstring)
+
         def get_graphql_fields() -> Dict[str, GraphQLInputField]:
             graphql_fields = {}
             for field in type_definition.fields:
                 field_name = self.config.name_converter.from_field(field)
 
-                graphql_fields[field_name] = self.from_input_field(field)
+                graphql_fields[field_name] = self.from_input_field(field, docstring)
 
             return graphql_fields
 
         graphql_object_type = GraphQLInputObjectType(
             name=type_name,
             fields=get_graphql_fields,
-            description=type_definition.description,
+            description=description,
         )
 
         self.type_map[type_name] = ConcreteType(
@@ -222,12 +276,17 @@ class GraphQLCoreConverter:
             assert isinstance(graphql_interface, GraphQLInterfaceType)  # For mypy
             return graphql_interface
 
+        description = interface.description
+        docstring = interface.docstring
+        if description is None and self.config.description_from_docstrings:
+            description = docstrings.type_description(docstring)
+
         def get_graphql_fields() -> Dict[str, GraphQLField]:
             graphql_fields = {}
 
             for field in interface.fields:
                 field_name = self.config.name_converter.from_field(field)
-                graphql_fields[field_name] = self.from_field(field)
+                graphql_fields[field_name] = self.from_field(field, docstring)
 
             return graphql_fields
 
@@ -235,7 +294,7 @@ class GraphQLCoreConverter:
             name=interface_name,
             fields=get_graphql_fields,
             interfaces=list(map(self.from_interface, interface.interfaces)),
-            description=interface.description,
+            description=description,
         )
 
         self.type_map[interface_name] = ConcreteType(
@@ -271,13 +330,18 @@ class GraphQLCoreConverter:
             assert isinstance(graphql_object_type, GraphQLObjectType)  # For mypy
             return graphql_object_type
 
+        description = object_type.description
+        docstring = object_type.docstring
+        if description is None and self.config.description_from_docstrings:
+            description = docstrings.type_description(docstring)
+
         def get_graphql_fields() -> Dict[str, GraphQLField]:
             graphql_fields = {}
 
             for field in object_type.fields:
                 field_name = self.config.name_converter.from_field(field)
 
-                graphql_fields[field_name] = self.from_field(field)
+                graphql_fields[field_name] = self.from_field(field, docstring)
 
             return graphql_fields
 
@@ -296,7 +360,7 @@ class GraphQLCoreConverter:
             name=object_type_name,
             fields=get_graphql_fields,
             interfaces=list(map(self.from_interface, object_type.interfaces)),
-            description=object_type.description,
+            description=description,
             is_type_of=is_type_of,
         )
 
--- a/strawberry/utils/docstrings.py
+++ b/strawberry/utils/docstrings.py
@@ -1,0 +1,53 @@
+from typing import Any, Optional
+
+import docstring_parser
+
+
+Docstring = docstring_parser.Docstring
+
+
+def get(obj: Any) -> Optional[Docstring]:
+    if obj is None or obj.__doc__ is None:
+        return None
+    return docstring_parser.parse(obj.__doc__)
+
+
+def type_description(docstring: Optional[Docstring]) -> Optional[str]:
+    if docstring is not None:
+        parts = []
+        if docstring.short_description:
+            parts.append(docstring.short_description)
+        if (
+            docstring.blank_after_short_description
+            and docstring.short_description
+            and docstring.long_description
+        ):
+            parts.append("")
+        if docstring.long_description:
+            parts.append(docstring.long_description)
+
+        # TODO: Expose other docstring bits (returns, raises, examples, etc)
+        # TODO: GraphQL expects descriptions to be formatted as Markdown,
+        # convertion may be needed (See pydoc-markdown)
+        return "\n".join(parts)
+    return None
+
+
+def func_description(docstring: Optional[Docstring]) -> Optional[str]:
+    return type_description(docstring)
+
+
+def arg_description(docstring: Optional[Docstring], name: str) -> Optional[str]:
+    if docstring is not None:
+        for param in docstring.params:
+            if param.args[0] == "param" and param.arg_name == name:
+                return param.description
+    return None
+
+
+def attribute_description(docstring: Optional[Docstring], name: str) -> Optional[str]:
+    if docstring is not None:
+        for param in docstring.params:
+            if param.args[0] == "attribute" and param.arg_name == name:
+                return param.description
+    return None
--- a/tests/schema/test_docstring.py
+++ b/tests/schema/test_docstring.py
@@ -1,0 +1,258 @@
+import textwrap
+from enum import Enum
+
+from graphql import DirectiveLocation
+
+import strawberry
+from strawberry.schema.config import StrawberryConfig
+
+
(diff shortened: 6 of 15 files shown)


botberry:



---

Support using docstrings are GraphQL descriptions

It is now possible to use python docstrings to provide GraphQL descriptions.

## Example

Here is an example of using docstrings in types and fields:

```
@strawberry.enum
class EnumType(Enum):
    """
    Example enum

    Attributes:
        FOO: Some description
        BAR: Another description
    """

    FOO = "foo"
    BAR = "bar"

@strawberry.type
class Query:
    """
    The main GraphQL type

    Attributes:
        enum: A dataclass field
    """
    enum: EnumType = EnumType.BAR

    @strawberry.field
    def resolver(self, arg1: enum, arg2: int) -> int:
        """
        A GraphQL field with a resolver and arguments

        Params:
            arg1: An enum argument
            arg2: An int argument
        """
        return 1
```

It produces this GraphQL schema:

```
"""Example enum"""
enum EnumType {
  """Some description"""
  FOO

  """Another description"""
  BAR
}

"""The main GraphQL type"""
type Query {
  """A dataclass field"""
  enum: EnumType!

  """A GraphQL field with a resolver and arguments"""
  resolver(
    """An enum argument"""
    arg1: EnumType!

    """An int argument"""
    arg2: Int!
  ): Int!
}
```


---




Here's the tweet text: 
```

```

RELEASE.md
```diff
@@ -0,0 +1,73 @@
...
+
+Support using docstrings as GraphQL descriptions
+
+It is now possible to use python docstrings to provide GraphQL descriptions.
+
+## Example
+
+Here is an example of using docstrings in types and fields:
+
+```
```
patrick91, line 11:
```suggestion
```python
```

RELEASE.md
```diff
@@ -0,0 +1,73 @@
...
+        Params:
+            arg1: An enum argument
+            arg2: An int argument
+        """
+        return 1
+```
+
+Produces this GraphQL schema:
+
+```
```
patrick91, line 49:
```suggestion
```graphql
```

docs/general/descriptions.md
```diff
@@ -0,0 +1,127 @@
...
+
+    @strawberry.type(description="The main GraphQL type")
+    class Query:
+        enum: EnumType = strawberry.field(default=EnumType.BAR, description="A dataclass field")
+
+        @strawberry.field(description="A GraphQL field with a resolver and arguments")
+        def resolver(self, arg1: str, arg2: int) -> int:
+            return 1
+
+    schema = strawberry.Schema(query=Query)
```
patrick91, line 31: we don't need these spaces at the beginning 😊

docs/general/descriptions.md
```diff
@@ -0,0 +1,127 @@
...
+## Docstrings
+
+It is also possible to generate descriptions from Python docstrings.
+
+Additionally it is possible to add descriptions to enum values and resolver arguments when
+one of the [supported syntaxes is used](https://pypi.org/project/docstring-parser/)
+
+<Note>
+This needs to be explicitly enabled with `StrawberryConfig(description_from_docstrings=True)`
+</Note>
```
patrick91, line 63:
```suggestion
<Note>

This needs to be explicitly enabled with `StrawberryConfig(description_from_docstrings=True)`

</Note>
```

strawberry/schema/config.py
```diff
@@ -9,6 +9,7 @@
 class StrawberryConfig:
     auto_camel_case: InitVar[bool] = None  # type: ignore
     name_converter: NameConverter = field(default_factory=NameConverter)
+    description_from_docstrings: bool = False
```
patrick91, line 12: @BryceBeagle what do you think of this name? 😊

strawberry/schema/schema_converter.py
```diff
@@ -103,26 +120,40 @@ def from_enum(self, enum: EnumDefinition) -> CustomGraphQLEnumType:
...
+            description = docstrings.attribute_description(
+                parent_docstring, enum_value.name
+            )
+        return GraphQLEnumValue(enum_value.value, description=description)
 
     def from_directive(self, directive: StrawberryDirective) -> GraphQLDirective:
+        description = directive.description
+        docstring = directive.docstring
+        if description is None and self.config.description_from_docstrings:
+            description = docstrings.type_description(docstring)
```
patrick91, line 137: shall we make a a `self._get_description` utility method? :)

strawberry/utils/docstrings.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def get(obj: Any) -> Optional[Docstring]:
+    if obj is None or obj.__doc__ is None:
+        return None
+    return docstring_parser.parse(obj.__doc__)
+
+
+def type_description(docstring: Optional[Docstring]) -> Optional[str]:
+    if docstring is not None:
```
patrick91, line 16: can we do `if docstring is None: return None` so we don't nest the rest? 😊

patrick91: same below :)

patrick91 (commented):
Looks pretty neat! I've added a few comments and will do another review in the next couple of days 😊 👏

docs/general/descriptions.md
paulo-raca: Sorry :sweat_smile:

strawberry/utils/docstrings.py
paulo-raca: Done

strawberry/schema/schema_converter.py
paulo-raca: Good idea

strawberry/schema/config.py
BryceBeagle: I think `descriptions_from_docstrings` would be better. Or maybe even just `docstring_descriptions`. I'm wondering if it could be useful to do something like:

```python
from enum import auto, Flag

class DocstringDescriptions(Flag):
    NONE = auto()
    TYPES = auto()
    RESOLVERS = auto()
    ARGUMENTS = auto()
    ALL = TYPES | RESOLVERS | ARGUMENTS

@dataclass    
class StrawberryConfig:
    docstring_descriptions: DocstringDescriptions = DocstringDescriptions.NONE
```

This would allow users to have a little more granularity with which docstrings are getting enabled

BryceBeagle:
Very cool! 

I'm wondering if it could be useful to do something like:

```python
from enum import auto, Flag

class DocstringDescriptions(Flag):
    NONE = auto()
    TYPES = auto()
    RESOLVERS = auto()
    ARGUMENTS = auto()
    ALL = TYPES | RESOLVERS | ARGUMENTS

@dataclass    
class StrawberryConfig:
    docstring_descriptions: DocstringDescriptions = DocstringDescriptions.NONE
```

This would allow users to have a little more granularity with which docstrings are getting used for descriptions

strawberry/schema/schema_converter.py
paulo-raca: Done

paulo-raca:
> This would allow users to have a little more granularity with which docstrings are getting used for descriptions

:thinking: Interesting idea.

However I don't see a scenario where I would enable it for types but not for resolvers (or something like this -- But I might be wrong).

If we want finer granularity, I think it would be better to specify it in each declaration instead? 

e.g., `@strawberry.type(description_from_docstring=True)`

patrick91:
> Very cool!
> 
> I'm wondering if it could be useful to do something like:
> 
> ```python
> from enum import auto, Flag
> 
> class DocstringDescriptions(Flag):
>     NONE = auto()
>     TYPES = auto()
>     RESOLVERS = auto()
>     ARGUMENTS = auto()
>     ALL = TYPES | RESOLVERS | ARGUMENTS
> 
> @dataclass    
> class StrawberryConfig:
>     docstring_descriptions: DocstringDescriptions = DocstringDescriptions.NONE
> ```
> 
> This would allow users to have a little more granularity with which docstrings are getting used for descriptions

I like this idea, but I think it might be a bit confusing, for example resolvers don't appear in the GraphQL schema and arguments don't support docstrings 😊

I think we can stick with `descriptions_from_docstrings`

patrick91:
@paulo-raca I'm going to think a bit more about the configuration options.

I think @BryceBeagle idea is worth exploring more, mostly because of the resolvers. If we have this:

```python
# in some module.py
def get_age() -> int:
    """Something that might have sensitive information in it"""

    return 43

# schema.py

from module import get_age

@strawberry.type
class Query:
    age: int = strawberry.field(resolver=get_age)
```

the fact that the type and the resolver are in separate files might lead to leaking some internal information 😊

maybe I'm worrying too much :D

paulo-raca:
> The fact that the type and the resolver are in separate files might lead to leaking some internal information blush

:thinking: Maybe it would it make sense to distinguish `inline` resolvers (`@strawberry.field def get_age() -> int`) from external ones? (`age: int = strawberry.field(resolver=get_age)`)?

paulo-raca:
I've updated the code to support various "description sources" -- It is now possible to enable and prioritize various sources, including from strawberry decorators and various types of docstrings. The sources can also be overridden in each element using strawberry decorators.

- Docstrings from parent classes are now taken in account, so that you don't need to re-document inherited fields
- PEP-257 docstrings are now supported to document enum values and object fields

TODO:
 - Update the tests and documentations
 - Support resolver arguments using  PEP-257 docstrings

mnekrasov-iss:
Wanted to check in on this PR. It is super useful to our use case. Was wondering if there was any movement on it.

paulo-raca:
Thanks for pinging me, @mnekrasov-iss.

I have finally taken the time to update make a few more cleanups and update it with the main branch and it is ready for review :+1:

tests/schema/test_docstring.py
```diff
@@ -0,0 +1,293 @@
+import textwrap
```
BryceBeagle, line 1: Great tests! I have a couple of requests tho:
1. Can the tests be made a bit more unit-y, where we test the docstring type of one type at a time? I.e. a test for enums, then a test for resolvers, then a test for objects, etc. 
2. Explicit tests that ensure docstring is/is not exposed for each of the `DescriptionSource` choices.
3. What do you think about adding some for a few of the other formats `docstring_parser` supports? Just so that viewers looking at the tests for examples understand that multiple formats are supported

BryceBeagle (changes requested):
This is really great work @paulo-raca! This is an awesome contribution. I've left some requests for (what are hopefully) some minor changes, but I'd love to get this merged. 

I'll make sure @patrick91 takes a look at this too :+1:

strawberry/utils/pick.py
```diff
@@ -0,0 +1,12 @@
+from typing import Optional, TypeVar
+
+
+T = TypeVar("T")
+
+
+def pick_not_none(*args: Optional[T]) -> T:
```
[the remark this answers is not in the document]
paulo-raca: It is used in schema_converter

strawberry/utils/docstrings.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class Docstring:
+    def __init__(self, target: Any) -> None:
+        self.target = target
+        self._docstring: Optional[docstring_parser.Docstring] = None
+        self._attribute_raw_docstrings: Optional[Dict[str, str]] = None
+        self._attribute_docstrings: Dict[str, Optional[str]] = {}
+
+    @property
```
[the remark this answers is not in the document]
paulo-raca: Good idea!

strawberry/utils/docstrings.py
```diff
@@ -0,0 +1,93 @@
...
+
+            attr_docstring = self._attribute_raw_docstrings.get(name)
+            if attr_docstring:
+                attr_docstring = Docstring(attr_docstring).main_description
+            self._attribute_docstrings[name] = attr_docstring
+
+        return self._attribute_docstrings[name]
+
+    @staticmethod
+    def get(obj: Any) -> Optional["Docstring"]:
```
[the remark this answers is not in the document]
paulo-raca: Neat!

pyproject.toml
```diff
@@ -46,6 +46,8 @@ sanic = {version = ">=20.12.2,<22.0.0", optional = true}
 aiohttp = {version = "^3.7.4.post0", optional = true}
 fastapi = {version = ">=0.65.2", optional = true}
 "backports.cached-property" = "^1.0.1"
+docstring-parser = "^0.14.1"
+attributes-doc = "^0.2.0"
```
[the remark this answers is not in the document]
paulo-raca: Removed, I inlined the function in docstrings.py

strawberry/utils/docstrings.py
```diff
@@ -0,0 +1,93 @@
...
+        parts = []
+        if docstring.short_description:
+            parts.append(docstring.short_description)
+        if docstring.blank_after_short_description:
+            parts.append("")
+        if docstring.long_description:
+            parts.append(docstring.long_description)
+
+        # TODO: Expose other docstring bits (returns, raises, examples, etc)
+        return "\n".join(parts).strip() or None
```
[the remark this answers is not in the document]
paulo-raca: Yes, docstring-parser is not clever enough to remove the empty lines between the main description and the argument section, so we would end with a lot of unnecessary blank lines otherwise

strawberry/utils/pick.py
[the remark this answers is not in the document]
BryceBeagle: Are you sure? I see neither current reference to `pick_not_none` nor one added in this PR.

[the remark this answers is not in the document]
BryceBeagle: Oh nvm, I got tricked. GitHub had completely minimized the diff in the PR because it was a large change and I hadn't noticed. That made `ctrl`+`f` stop working too

strawberry/schema/schema_converter.py
```diff
@@ -190,10 +265,15 @@ def from_schema_directive(self, directive: Any) -> GraphQLDirective:
...
             args=args,
-            description=strawberry_directive.description,
+            description=description,
         )
 
-    def from_field(self, field: StrawberryField) -> GraphQLField:
+    def from_field(
+        self,
+        field: StrawberryField,
+        description_sources: List[DescriptionSource],
```
BryceBeagle, line 274: Hmm looking at this now, I'm unsure how this works and probably will be for users too, potentially leading to accidental leakage.

What happens if I have

```python
@strawberry.type(description_sources=[DescriptionSource.TYPE_DOCSTRINGS])
class Query:
    """AAAAA"""

    @strawberry.resolver(description_sources=[DescriptionSource.RESOLVER_DOCSTRINGS])
    def my_resolver(): ...
    """BBBBB"""

schema = strawberry.Schema(
    query=Query,
    config=StrawberryConfig(description_sources=[DescriptionSource.STRAWBERRY_DESCRIPTIONS]),
)
````

what makes it on the schema? I would definitely like clarification on ownership/nesting and test cases too.

File: docs/general/descriptions.md [hunk 15]
```diff
@@ -0,0 +1,129 @@
...
+## Docstrings
+
+It is also possible to generate descriptions from Python docstrings.
+
+Additionally it is possible to add descriptions to enum values and resolver arguments when
+one of the [supported syntaxes is used](https://pypi.org/project/docstring-parser/)
+
+<Note>
+
+This needs to be explicitly enabled with `StrawberryConfig(descriptions_from_docstrings=True)`
```
BryceBeagle, line 63: This and the other doc never got updated after `descriptions_from_docstrings` was removed

tests/schema/test_docstring.py
```diff
@@ -0,0 +1,293 @@
...
+        """
+        return ""
+
+
+def test_docstrings_enabled():
+    schema = strawberry.Schema(
+        query=Query,
+        mutation=Mutation,
+        directives=[replace],
+        config=StrawberryConfig(description_sources=DescriptionSource),
```
BryceBeagle, line 149: Why are you passing the enum _Type_ to `description_sources` instead of a list of instances?

BryceBeagle (changes requested):

strawberry/utils/docstrings.py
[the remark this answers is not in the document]
paulo-raca: `long_description` is sometimes an empty string, so it would end with a trailing \n

strawberry/utils/pick.py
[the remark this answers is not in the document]
paulo-raca: There is a semantic difference between `description_sources is None` (Don't change it) and `description_sources == []` (Don't add any description), which isn't captured by the `a or b`

strawberry/utils/pick.py
```diff
@@ -0,0 +1,12 @@
...
+    """Returns the first not-None argument"""
+    for arg in args:
+        if arg is not None:
+            return arg
+    raise ValueError("Expected not-None argument")
```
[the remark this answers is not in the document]
paulo-raca: Given the function name, it sounds like a reasonable behavior ;)

Would you rather return None and handle the issue on the caller?

docs/general/descriptions.md
paulo-raca: Good catch!

strawberry/schema/config.py
```diff
@@ -1,14 +1,19 @@
...
-from .name_converter import NameConverter
+from strawberry.description_source import DescriptionSource
+from strawberry.schema.name_converter import NameConverter
 
 
 @dataclass
 class StrawberryConfig:
     auto_camel_case: InitVar[bool] = None
     name_converter: NameConverter = field(default_factory=NameConverter)
+    description_sources: List[DescriptionSource] = field(
```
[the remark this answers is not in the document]
paulo-raca: I considered that, but I've chosen a list so that the ordering can be used to set different priorities:

E.g.:
```python
@strawberry.type(description="Strawberry Description")
class Foo:
   """ Docstring description"""
   bar: int

will return different docstrings for `description_sources = [STRAWBERRY_DESCRIPTIONS, TYPE_DOCSTRINGS]` vs `[TYPE_DOCSTRINGS, STRAWBERRY_DESCRIPTIONS]`

**Comment by paulo-raca** [comment]
@BryceBeagle Thank you very much for all the effort you have been putting on reviewing this PR :pray: 

I have a few more tests to write, but I think this is looking pretty good :rocket:

File: strawberry/schema/schema_converter.py (hunk 14 above)
Comment (BryceBeagle) [reply strawberry/schema/schema_converter.py:277]: Gotcha, so inner attributes' `description_sources` override outer attributes'.

Comment (BryceBeagle) [reply strawberry/schema/schema_converter.py:277]: Could you make this super-duper clear in the description? I don't want users getting confused

File: tests/schema/test_docstring.py (hunk 16 above)
Comment (BryceBeagle) [reply tests/schema/test_docstring.py]: Potentially `DescriptionSource.__members__.values()` would be the most clear. It's the only option that doesn't rely on the `Enum.__iter__` magic, but I don't think `list(DescriptionSource)` is _bad_.

I'm actually surprised that passing the `DescriptionSource` class here (without converting it to a `list`) doesn't fail mypy checks, because everything is annotated as taking `List[DescriptionSource]` and not just `Iterable[DescriptionSource]`. I think we should make the `config=` above more explicit, but I also think the annotation for `description_sources` should be changed to `Iterable` everywhere. What do you think?

File: strawberry/schema/config.py (hunk 18 above)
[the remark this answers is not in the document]
Comment (BryceBeagle) [reply strawberry/schema/config.py:14]: Hmmm, is that desirable? Not sure if having fallback-style ordering is necessary/desirable if attributes are _also_ inheriting from their scope-ings. What do you think @patrick91 ?

File: strawberry/utils/pick.py (hunk 9 above)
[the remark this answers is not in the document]
Comment (BryceBeagle) [reply strawberry/utils/pick.py:7]: Hmm that's a fair point as it's currently implemented. This is again making me question having the list of sources. If we didn't have a list, having a `DescriptionSource.NONE` (as a false-y value) would make perfect sense.

**Comment by BryceBeagle** [comment]
Thanks for adding these great tests! I think we're almost there. Let me know what you think of my replies to the other comments :+1: 

cc @patrick91 to take a look too

File: strawberry/utils/pick.py (hunk 9 above)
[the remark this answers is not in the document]
Comment (paulo-raca) [reply strawberry/utils/pick.py:7]: Having a false-y `DescriptionSource.NONE` results in the same problem: It gets ignored and the code fallbacks to the parent/global config.

```python
config=StrawberryConfig(description_sources=DescriptionSources.ALL)

[...]

@strawberry.type(description_sources=DescriptionSource.NONE)
class UndocumentedClass:
  """ Secrets that shouldn't be exposed """
   ...

[...]

description_sources = type.description_sources or config.description_sources  
# NONE or ALL => ALL
# Secrets are exposed!  O.O
```

This would work if NONE was not False-y, but that goes against normal semantics of `enum.Flag` :sweat: 

I'll keep this method for now

strawberry/schema/config.py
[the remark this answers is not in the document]
paulo-raca: Okay, I've changed to enum.Flags and it does look better.

However we now need to be extra conscious about the priorities predefined in code.

In particular, it currenclt assumes: description > resolver > class > attribute
:thinking: Maybe attribute docstrings have priority over class docstrings?

patrick91:
Hi @paulo-raca sorry for the late reply! I've just had a look at this with @BryceBeagle and I was wondering if we can split this PR into multiple ones. I'm personally not sure if we should add all the features implemented here and it might be worth discussing each of them separately.

Would you be ok with splitting this into the following PRs:

1. getting description for types from docstrings
2. getting description for resolvers from docstrings (could be done with the one above)
3. getting descriptions for attributes from [inline docstrings](https://peps.python.org/pep-0224/) (?) 
4. getting descriptions for arguments/enum-values from docstring in the resolver/type (?)

The last two items are the ones I'm not 100% convinced about, the idea is very interesting, but I think it might be too confusing to have so many ways of defining descriptions 😊

paulo-raca:
Hello, @patrick91, thanks for having a look at this.

Sorry, but I hadn't touched this for a few weeks. I just updated it with the `main` branch.

>  I've just had a look at this with @BryceBeagle and I was wondering if we can split this PR into multiple ones. I'm personally not sure if we should add all the features implemented here and it might be worth discussing each of them separately.
>
> Would you be ok with splitting this into the following PRs:
>
> 1. getting description for types from docstrings
> 2. getting description for resolvers from docstrings (could be done with the one above)
> 3. getting descriptions for attributes from [inline docstrings](https://peps.python.org/pep-0224/) (?)
> 4. getting descriptions for arguments/enum-values from docstring in the resolver/type (?)

I looked into it, and I don't think it is worth the effort: 85% of this PR is plumbing (moving `description_sources` and `docstrings` all the way to `schema_converter._get_description()`) -- which I think shouldn't be split --, and from there the code to handle type docstring (1), resolver docstrings (2) and enum value/argument docstrings (4) are essentially the same.

The support for [inline attributes docstrings](https://peps.python.org/pep-0224/) (3) is the only thing that might be worth moving into a separate PR, since it's implementation is slightly different.

> The last two items are the ones I'm not 100% convinced about

Yeah, I understand (3) is weird and non-standard.
I found it really useful for documenting classes with lots of fields, but I understand if it doesn't make the cut.

(4), OTOH, is essential!
I can gate it with a separate `DescriptionSource`s for Enum values / StrawberryType Fields / Resolver Arguments. Would that suffice?

> The idea is very interesting, but I think it might be too confusing to have so many ways of defining descriptions blush

TBH, I think the usage is pretty straightforward, you either use explicit strawberry annotations or standard python docstrings syntax, and things just work :smiley: 

What I do think is confusing is the large number of values that ended in `DescriptionSource`. I'm divided between having granular configuration (the current code) or just having `DESCRIPTIONS`, `DOCSTRINGS` and `ATTRIBUTE_DOCSTRINGS`

paulo-raca:
Updated with main branch

paulo-raca:
This is not going anywhere :cry: 

However I have an alternative idea on how to achieve this:

- Instead of DescriptionSources being a set of "Flags" for the supported sources, we can make an abstract class for it:
  ```
  DescriptionTarget = Union[StrawberryField, StrawberryArgument, EnumDefinition, EnumValue, StrawberryDirective,   TypeDefinition, ...]

  class DescriptionSource:
      def get_description(self, target: DescriptionTarget) -> Optional[str]
          raise NotImplemented()

  DescriptionSources = List[DescriptionSource]
  ```
- Trim this PR and just leave the plumbing to find the `descriptionSources` on each DescriptionTarget and call them.
- Implement DosctringDescriptionSource in a separate PR (maybe a separate dependency entirely)

Some DescriptionSource implementations may even be implemented to act as "modifiers". E.g., the permissions modifiers in #2335

patrick91:
hi @paulo-raca, sorry I missed your update. I was chatting with @bellini666  about this and I personally think it is worth splitting this PR into multiple ones, starting with supporting standard docstrings (via a flag) and then we can decide if we should add support for pep-0224 (I'm slightly against it).

What do you think?

jkimbo:
@paulo-raca first off a huge thank you for spending so much time on this feature and for keeping it up to date for so long. I think it's a valuable feature however I'm not sure having a static list of description targets is the right API for it.

My 2 cents is that we can achieve this functionality by implementing some more methods on Schema Extensions that would allow the developer to customise types when building a schema. That way we can support a much larger set of use cases.

For example here are some straw man API's to automatically add descriptions to types and fields:

```python
class AutomaticDescriptions(SchemaExtension):
	def prepare_type(self, type: StrawberryObjectType) -> StrawberryObjectType:
		orig_cls = type._cls
		if hasattr(orig_cls, "__doc__"):
			# you can parse the doc string with `docstring_parser` here
			type.description = orig_cls.__doc__
			# you could also ignore doc strings that contained a string like "INTERNAL" in it
		
		return type

	def prepare_field(self, field: StrawberryField) -> StrawberryField:
		resolver = field.resolver
		if resolver and hasattr(resolver, "__doc__"):
			field.description = resolver.__doc__

		return field

	... # other functions for enums, input types, interfaces etc
```

Also here is an example of prefixing all type names that begin with `Book` with `Legacy`:

```python
class RenameBookTypes(SchemaExtension):
	def prepare_type(self, type: StrawberryObjectType) -> StrawberryObjectType:
		if type.name.startswith("Book"):
			type.name = `Legacy{type.name}`
		
		return type
```

Injecting descriptions to all mutations:

```python
class MutationLogging(SchemaExtension):
	def prepare_field(self, field: StrawberryField) -> StrawberryField:
		if field.is_mutation:
			field.description += "This is a mutation"

		return field
```

Adding new fields to the Query type to allow easy third party integration:

```python
class InjectPingField(SchemaExtension):
	def prepare_type(self, type: StrawberryObjectType) -> StrawberryObjectType:
		if type.name == "Query":
			type.fields["ping"] = strawberry.field(graphql_type=str, resolver=lambda: "pong")
		
		return type
```

Those are mostly just toy examples but I think having an extensible way of modifying the schema at build time could be quite powerful and allow for lots of functionality that doesn't have to live in Strawberry itself.

What do you think @paulo-raca ?

Issue #1409, opened 2021-11-09
Cannot use dict or JSON types in Pydantic models as attributes


mecampbellsoup:
## Description

I'm trying to define a GQL Pydantic model w/ a JSON scalar attribute:

```
JSONScalar = strawberry.scalar(
    NewType("JSONScalar", Any),
    serialize=lambda v: v,
    parse_value=lambda v: json.loads(v),
    description="The GenericScalar scalar type represents a generic GraphQL scalar value that could be: List or Object."
)


class ResourcePyd(PydanticModel):
    type: str
    # TODO: Figure out how to use `Json` type for this attribute
    # spec: Optional[Json]
    # spec: Optional[str]
    spec: Optional[JSONScalar]
```

However I can't get Pydantic to comply:

```
[cloud-app] [Tue Nov  9 17:21:12 2021] Running: pytest --durations=3
[cloud-app] ImportError while loading conftest '/cloud-app/tests/conftest.py'.
[cloud-app] tests/conftest.py:24: in <module>
[cloud-app]     import app.gql_mutations as gql_mutations
[cloud-app] app/gql_mutations.py:21: in <module>
[cloud-app]     from app.gql_errors import BadRequest, NotAuthorized, InternalError
[cloud-app] app/gql_errors.py:15: in <module>
[cloud-app]     from app.models import Error, Errors, ResponseCode
[cloud-app] app/models.py:128: in <module>
[cloud-app]     class ResourcePyd(PydanticModel):
[cloud-app] pydantic/main.py:299: in pydantic.main.ModelMetaclass.__new__
[cloud-app]     ???
[cloud-app] pydantic/fields.py:411: in pydantic.fields.ModelField.infer
[cloud-app]     ???
[cloud-app] pydantic/fields.py:342: in pydantic.fields.ModelField.__init__
[cloud-app]     ???
[cloud-app] pydantic/fields.py:456: in pydantic.fields.ModelField.prepare
[cloud-app]     ???
[cloud-app] pydantic/fields.py:670: in pydantic.fields.ModelField.populate_validators
[cloud-app]     ???
[cloud-app] pydantic/validators.py:710: in find_validators
[cloud-app]     ???
[cloud-app] E   RuntimeError: error checking inheritance of <strawberry.custom_scalar.ScalarWrapper object at 0x7ff0045537f0> (type: ScalarWrapper)
```

## Refs

[...]

PR #1647, opened 2022-02-16
Bug in JSONScalar example

paulo-raca:
There was a bug in the JSONScalar example: `parse_value` takes a data structure (dicts/lists/primitives) instead of a JSON string, therefore `json.loads` is not necessary.

This PR started as a simple fix to the documentation, however @patrick91 suggested that we also added it into strawberry sources and tests, which I have done.

While at it, I did the same for the `Base64` example -- Which I unfolded into `Base16`, `Base32` and `Base64`

Finally, I also added support for the `specified_by_url` in scalars

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,7 @@
+Release type: minor
+
+There was a bug with the JSON custom scalar example.
+
+This release fixes the example and adds JSON scalar type into `strawberry.scalars`.
+
+The same happens with `Base64`: We now have Base16, Base32 and Base64 into `strawberry.scalars`
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -19,4 +19,3 @@ class Example:
     c: Base64
     d: JSON
 ```
-
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,7 +1,22 @@
 Release type: minor
 
-There was a bug with the JSON custom scalar example.
+This release adds the following scalar types:
 
-This release fixes the example and adds JSON scalar type into `strawberry.scalars`.
+- `JSON`
+- `Base16`
+- `Base32`
+- `Base64`
 
-The same happens with `Base64`: We now have Base16, Base32 and Base64 into `strawberry.scalars`
+they can be used like so:
+
+```python
+from strawberry.scalar import Base16, Base32, Base64, JSON
+
+@strawberry.type
+class Example:
+    a: Base16
+    b: Base32
+    c: Base64
+    d: JSON
+```
+
--- a/docs/types/scalars.md
+++ b/docs/types/scalars.md
@@ -119,6 +119,7 @@ assert results.data  == {"base64": "aGk="}
 ```
 
 <Note>
+
 The `Base16`, `Base32` and `Base64` scalar types are available in `strawberry.scalars`
 
 ```python
@@ -171,6 +172,7 @@ query ExampleDataQuery {
 ```
 
 <Note>
+
 The `JSON` scalar type is available in `strawberry.scalars`
 
 ```python
--- a/docs/types/scalars.md
+++ b/docs/types/scalars.md
@@ -118,6 +118,15 @@ result = schema.execute_sync("{ base64 }")
 assert results.data  == {"base64": "aGk="}
 ```
 
+<Note>
+The `Base16`, `Base32` and `Base64` scalar types are available in `strawberry.scalars`
+
+```python
+from strawberry.scalars import Base16, Base32, Base64
+```
+
+</Note>
+
 ## Example JSONScalar
 
 ```python
@@ -126,11 +135,11 @@ from typing import Any, NewType
 
 import strawberry
 
-JSONScalar = strawberry.scalar(
-    NewType("JSONScalar", Any),
+JSON = strawberry.scalar(
+    NewType("JSON", object),
+    description="The `JSON` scalar type represents JSON values as specified by ECMA-404",
     serialize=lambda v: v,
     parse_value=lambda v: v,
-    description="The JSONScalar scalar type represents a generic GraphQL scalar value that can be built from: List, Object, String, Number, Boolean or Null"
 )
 
 ```
@@ -141,7 +150,7 @@ Usage:
 @strawberry.type
 class Query:
     @strawberry.field
-    def data(self, info) -> JSONScalar:
+    def data(self, info) -> JSON:
         return {"hello": {"a": 1}, "someNumbers": [1, 2, 3]}
 
 ```
@@ -160,6 +169,15 @@ query ExampleDataQuery {
   }
 }
 ```
+
+<Note>
+The `JSON` scalar type is available in `strawberry.scalars`
+
+```python
+from strawberry.scalars import JSON
+```
+
+</Note>
 
 ## Overriding built in scalars
 
--- a/docs/types/scalars.md
+++ b/docs/types/scalars.md
@@ -129,8 +129,8 @@ import strawberry
 JSONScalar = strawberry.scalar(
     NewType("JSONScalar", Any),
     serialize=lambda v: v,
-    parse_value=lambda v: json.loads(v),
-    description="The GenericScalar scalar type represents a generic GraphQL scalar value that could be: List or Object."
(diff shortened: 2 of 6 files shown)


paulo-raca:
Sure!
Would it make sense to add `JSONScalar` into strawberry, or should it be kept only in the documentation/tests?

patrick91:
> Sure! Would it make sense to add `JSONScalar` into strawberry, or should it be kept only in the documentation/tests?

I think we can add one!

Also see this comment here, the code is slightly different 🤔  https://github.com/strawberry-graphql/strawberry/issues/1409#issuecomment-969261345

paulo-raca:
> Also see this comment here, the code is slightly different thinking [#1409 (comment)](https://github.com/strawberry-graphql/strawberry/issues/1409#issuecomment-969261345)

The default implementation of `parse_literal` is `self.parse_value(value_from_ast_untyped(value))`, so that implementation is basically defining a broken `parse_value` and then skipping it. Hacky, but it does work :+1:

botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

paulo-raca:
> > Sure! Would it make sense to add `JSONScalar` into strawberry, or should it be kept only in the documentation/tests?
> 
> I think we can add one!

I went ahead and did the same for the Base64 example -- Turns out the PR is now far from a one-line documentation fix

strawberry/scalars.py
```diff
@@ -1,10 +1,48 @@
...
+Base32 = scalar(
+    NewType("Base32", bytes),
+    description="Represents binary data as Base32-encoded strings,"
+    " using the standard alphabet.",
+    specified_by_url=("https://datatracker.ietf.org/doc/html/rfc4648.html#section-6"),
+    serialize=lambda v: base64.b32encode(v).decode("utf-8"),
+    parse_value=lambda v: base64.b32decode(v.encode("utf-8"), casefold=True),
+)
+
+Base64 = scalar(
```
paulo-raca, line 37: :thinking: Maybe register it as the default for `bytes`?

patrick91: I'd prefer to make it explicty 😊

strawberry/schema/types/scalar.py
```diff
@@ -32,6 +33,7 @@ def _make_scalar_definition(scalar_type: GraphQLScalarType) -> ScalarDefinition:
     return ScalarDefinition(
         name=scalar_type.name,
         description=scalar_type.name,
+        specified_by_url=scalar_type.specified_by_url,
```
patrick91, line 36: 💯

docs/types/scalars.md
```diff
@@ -118,6 +118,16 @@ result = schema.execute_sync("{ base64 }")
...
 
+<Note>
+
+The `Base16`, `Base32` and `Base64` scalar types are available in `strawberry.scalars`
+
+```python
+from strawberry.scalars import Base16, Base32, Base64
+```
+
+</Note>
```
patrick91, line 129: shall we add an example/reason for when to use them?

RELEASE.md
```diff
@@ -0,0 +1,7 @@
+Release type: minor
+
+There was a bug with the JSON custom scalar example.
+
+This release fixes the example and adds JSON scalar type into `strawberry.scalars`.
+
+The same happens with `Base64`: We now have Base16, Base32 and Base64 into `strawberry.scalars`
```
patrick91, line 7:
``````suggestion
This release adds the following scalar types:

- `JSON`
- `Base16`
- `Base32`
- `Base64`

they can be used like so:

```python
from strawberry.scalar import Base16, Base32, Base64, JSON

@strawberry.type
class Example:
    a: Base16
    b: Base32
    c: Base64
    d: JSON
```

``````

**Review by patrick91 — COMMENTED** [review]
Looks great! I've updated the release notes, I've removed the mention of the docs since they are only used for the changelog 😊

File: strawberry/scalars.py [hunk 5]
```diff
@@ -1,10 +1,48 @@
...
+JSON = scalar(
+    NewType("JSON", object),  # mypy doesn't like `NewType("name", Any)`
+    description="The `JSON` scalar type represents JSON values as specified by"
+    " [ECMA-404](http://www.ecma-international.org"
+    "/publications/files/ECMA-ST/ECMA-404.pdf).",
+    specified_by_url="http://www.ecma-international.org"
+    "/publications/files/ECMA-ST/ECMA-404.pdf",
+    serialize=lambda v: v,
+    parse_value=lambda v: v,
+)
```
BryceBeagle, line 18:
```suggestion
JSON = scalar(
    NewType("JSON", object),  # mypy doesn't like `NewType("name", Any)`
    description=(
        "The `JSON` scalar type represents JSON values as specified by "
        "[ECMA-404]"
        "(http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf)."
    ),
    specified_by_url=(
        "http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf"
    ),
    serialize=lambda v: v,
    parse_value=lambda v: v,
)
```

I'm a bit confused why `description` uses a Markdown (?) formatted link?

strawberry/scalars.py
```diff
@@ -1,10 +1,48 @@
...
+
+Base32 = scalar(
+    NewType("Base32", bytes),
+    description="Represents binary data as Base32-encoded strings,"
+    " using the standard alphabet.",
+    specified_by_url=("https://datatracker.ietf.org/doc/html/rfc4648.html#section-6"),
+    serialize=lambda v: base64.b32encode(v).decode("utf-8"),
+    parse_value=lambda v: base64.b32decode(v.encode("utf-8"), casefold=True),
+)
```
BryceBeagle, line 35:
```suggestion
Base32 = scalar(
    NewType("Base32", bytes),
    description=(
        "Represents binary data as Base32-encoded strings, using the standard alphabet."
    ),
    specified_by_url=("https://datatracker.ietf.org/doc/html/rfc4648.html#section-6"),
    serialize=lambda v: base64.b32encode(v).decode("utf-8"),
    parse_value=lambda v: base64.b32decode(v.encode("utf-8"), casefold=True),
)
```

strawberry/scalars.py
```diff
@@ -1,10 +1,48 @@
...
+
+Base64 = scalar(
+    NewType("Base64", bytes),
+    description="Represents binary data as Base64-encoded strings,"
+    " using the standard alphabet.",
+    specified_by_url="https://datatracker.ietf.org/doc/html/rfc4648.html#section-4",
+    serialize=lambda v: base64.b64encode(v).decode("utf-8"),
+    parse_value=lambda v: base64.b64decode(v.encode("utf-8")),
+)
```
BryceBeagle, line 44:
```suggestion
    NewType("Base64", bytes),
    description=(
        "Represents binary data as Base64-encoded strings, using the standard alphabet."
    ),
    specified_by_url="https://datatracker.ietf.org/doc/html/rfc4648.html#section-4",
    serialize=lambda v: base64.b64encode(v).decode("utf-8"),
    parse_value=lambda v: base64.b64decode(v.encode("utf-8")),
)
```

tests/schema/test_scalars.py
```diff
@@ -99,6 +101,155 @@ def uuid_input(self, input_id: UUID) -> str:
...
+        scalar JSON @specifiedBy(url: "http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf")
+
+        type Query {
+          echoJson(data: JSON!): JSON!
+          echoJsonNullable(data: JSON): JSON
+        }
+    '''  # noqa: E501
+        ).strip()
+    )
+
```
BryceBeagle, line 133: I think this formatting in this file got screwed up by the autoformatter. It can probably be fixed

BryceBeagle (changes requested):
Formatting change requests :+1:

paulo-raca:
> Formatting change requests +1

Al changes addressed :+1:

patrick91 (approved):
Thanks so much for this! 😊

PR #1648, opened 2022-02-16
Support "void" functions

paulo-raca:
## Description

It might not be a best practice, but sometimes a mutation resolver just doesn't need to return anything.

Having a resolver that returns `None` is currently an error in strawberry.

This PR adds the scalar "Void" and uses it automatically when the type is `None` (or `NoneType`, which can be confusing :roll_eyes:)

This scalar mimics the `Void` type provided by [GraphQL Scalars](https://www.graphql-scalars.dev/docs/scalars/void)

## Example:

```
    @strawberry.type
    class Mutation:
        @strawberry.field
        def do_something(self, arg: int) -> None:
            return
```
results in this schema:
```
   type Mutation {
       doSomething(arg: Int!): Void
    }
```
---

## Changes:
- Added a new `Void` scalar, and registered it to handle `None` and `NoneType`
- Allow `None` type annotations
- Merge `from_optional` and `from_non_optional` into `from_maybe_optional`
- Add a special case for NoneType, to avoid wrapping it into `GraphQLNonNull`

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -13,8 +13,10 @@ class Mutation:
     @strawberry.field
     def do_something(self, arg: int) -> None:
         return
+```
 results in this schema:
 ```grapqhl
 type Mutation {
     doSomething(arg: Int!): Void
 }
+```
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -15,8 +15,7 @@ and will always send `null` in the response
             return
 ```
 results in this schema:
-```
-   type Mutation {
-       doSomething(arg: Int!): Void
-    }
-```
+```grapqhl
+type Mutation {
+    doSomething(arg: Int!): Void
+}
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,22 @@
+Release type: minor
+
+Support "void" functions
+
+It is now possible to have a resolver that returns "None". Strawberry will automatically assign the new `Void` scalar in the schema
+and will always send `null` in the response
+
+## Exampe
+
+```
+    @strawberry.type
+    class Mutation:
+        @strawberry.field
+        def do_something(self, arg: int) -> None:
+            return
+```
+results in this schema:
+```
+   type Mutation {
+       doSomething(arg: Int!): Void
+    }
+```
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -7,13 +7,12 @@ and will always send `null` in the response
 
 ## Exampe
 
-```
-    @strawberry.type
-    class Mutation:
-        @strawberry.field
-        def do_something(self, arg: int) -> None:
-            return
-```
+```python
+@strawberry.type
+class Mutation:
+    @strawberry.field
+    def do_something(self, arg: int) -> None:
+        return
 results in this schema:
 ```grapqhl
 type Mutation {
--- a/docs/general/mutations.md
+++ b/docs/general/mutations.md
@@ -77,5 +77,7 @@ type Mutation {
 ```
 
 <Note>
+
 Mutations with void-result go against [GQL best practices](https://graphql-rules.com/rules/mutation-payload)
+
 </Note>
--- a/docs/general/mutations.md
+++ b/docs/general/mutations.md
@@ -57,3 +57,25 @@ example we might want to return an error if the book already exists.
 You can checkout our documentation on
 [dealing with errors](/docs/guides/errors#expected-errors) to learn how to return a
 union of types from a mutation.
+
+## Mutations without returned data
+
+It is also possible to write a mutation that doesn't return anything.
+
+This is mapped to a `Void` GraphQL scalar, and always returns `null`
+
+```python+schema
+@strawberry.type
+class Mutation:
+    @strawberry.mutation
+    def restart() -> None:
+        print(f'Restarting the server')
+---
+type Mutation {
+  restart: Void
+}
+```
+
+<Note>
+Mutations with void-result go against [GQL best practices](https://graphql-rules.com/rules/mutation-payload)
+</Note>
--- a/strawberry/schema/schema_converter.py
+++ b/strawberry/schema/schema_converter.py
@@ -11,14 +11,15 @@ from graphql import (
     GraphQLField,
     GraphQLInputField,
     GraphQLInputObjectType,
+    GraphQLInputType,
     GraphQLInterfaceType,
     GraphQLList,
     GraphQLNonNull,
     GraphQLNullableType,
     GraphQLObjectType,
+    GraphQLOutputType,
     GraphQLResolveInfo,
     GraphQLScalarType,
-    GraphQLType,
     GraphQLUnionType,
     Undefined,
 )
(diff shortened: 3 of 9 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

strawberry/schema/schema_converter.py
```diff
@@ -436,6 +408,17 @@ def from_scalar(self, scalar: Type) -> GraphQLScalarType:
...
+    def from_maybe_optional(
+        self, type_: Union[StrawberryType, type]
+    ) -> Union[GraphQLNullableType, GraphQLNonNull]:
+        NoneType = type(None)
+        if type_ is None or type_ is NoneType:
+            return self.from_type(type_)
+        elif isinstance(type_, StrawberryOptional):
+            return self.from_type(type_.of_type)
+        else:
+            return GraphQLNonNull(self.from_type(type_))
```
patrick91, line 420: I like this! /cc @BryceBeagle

patrick91 (commented):
Uhm, I'm not sure we should add a void type, I personally never needed to do something like that (mostly because I use union types for returning errors).

Maybe it would be better do add some docs about this pattern? Have you seen something like this in other libraries?

/cc @jkimbo what do you think?

ThirVondukr:
@patrick91 I think associating entities would be a valid use case for `Void` type?
```graphql
type Tag {
  id: String!
  slug: String!
}

type Post {
  id: String!
  tags: [Tag!]!
}

mutation tagPost(tagId: String!, postId: String!): Void
```

marcoacierno:
> @patrick91 I think associating entities would be a valid use case for `Void` type?
> 
> ```graphql
> type Tag {
>   id: String!
>   slug: String!
> }
> 
> type Post {
>   id: String!
>   tags: [Tag!]!
> }
> 
> mutation tagPost(tagId: String!, postId: String!): Void
> ```

I think in that case you might want the mutation to return the list of updated tags to that the FE can update the UI without refetching the tags of the post :)

patrick91:
> > Uhm, I'm not sure we should add a void type, I personally never needed to do something like that (mostly because I use union types for returning errors).
> 
> Yes, there is a consensus that this is not the best practice 😅
> 
> However I believe strawberry should still be able to handle it anyway
> 
> > Maybe it would be better do add some docs about this pattern?
> 
> I can add a line about that in the "Mutations" section
> 
> > Have you seen something like this in other libraries?
> 
> TBH I'm quite new to GraphQL and I haven't seen much
> 
> However [GraphQL Scalars](https://www.graphql-scalars.dev/docs/scalars/void) has this and [it is used in many github projects](https://github.com/search?p=2&q=GraphQLVoid&type=Code)

Gotcha, maybe we should add support for it, but not bind it to `None` 🤔 

If more people ask for it to be bound to `None` we could have a configuration option.

What do you think?

jkimbo:
IMO this seems like a good default. You can override the behaviour of None with the scalar overrides anyway if someone doesn't like it.

RELEASE.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Exampe
+
+```
+    @strawberry.type
+    class Mutation:
+        @strawberry.field
+        def do_something(self, arg: int) -> None:
+            return
+```
```
patrick91, line 16:
```suggestion
```python
@strawberry.type
class Mutation:
    @strawberry.field
    def do_something(self, arg: int) -> None:
        return
```
```

File: RELEASE.md [hunk 3]
```diff
@@ -0,0 +1,22 @@
...
+results in this schema:
+```
+   type Mutation {
+       doSomething(arg: Int!): Void
+    }
+```
```
patrick91, line 22:
```suggestion
```grapqhl
type Mutation {
    doSomething(arg: Int!): Void
}
```
```

File: docs/general/mutations.md [hunk 4]
```diff
@@ -57,3 +57,25 @@ example we might want to return an error if the book already exists.
...
+    def restart() -> None:
+        print(f'Restarting the server')
+---
+type Mutation {
+  restart: Void
+}
+```
+
+<Note>
+Mutations with void-result go against [GQL best practices](https://graphql-rules.com/rules/mutation-payload)
```
patrick91, line 80:
```suggestion

Mutations with void-result go against [GQL best practices](https://graphql-rules.com/rules/mutation-payload)

```

strawberry/schema/types/base_scalars.py
```diff
@@ -67,3 +67,17 @@ def parse_decimal(value: str) -> decimal.Decimal:
...
+    if x is not None:
+        raise ValueError(f"Expected 'None', got '{x}'")
+
+
+Void = scalar(
+    type(None),
+    name="Void",
+    serialize=_verify_void,
+    parse_value=_verify_void,
+    description="Represents NULL values",
```
patrick91, line 82:
```suggestion
    description="Represents None values",
```
maybe it is better to use None here? or Void? since it will be used in the schema :)

patrick91 (commented):
Looks pretty good to me! Thanks for the PR and thanks to everyone for sharing their opinions 😊

I've left a couple of minor comments, then this is ready to be merged :)

strawberry/schema/types/base_scalars.py
paulo-raca: I think `NULL` makes sense as it will be exposed on on the GraphQL protocol, while `None` is a Python thing. 

However I don't have strong feeling about that :man_shrugging: 

For whatever it is worth, I have defined this description to be the consistent with [graphql-scalar's Void](https://github.com/Urigo/graphql-scalars/blob/master/src/scalars/Void.ts#L7)

docs/general/mutations.md
```diff
@@ -57,3 +57,27 @@ example we might want to return an error if the book already exists.
...
+        print(f'Restarting the server')
+---
+type Mutation {
+  restart: Void
+}
+```
+
+<Note>
+
+Mutations with void-result go against [GQL best practices](https://graphql-rules.com/rules/mutation-payload)
```
paulo-raca, line 81: :thinking: The link is broken in the [preview documentation](https://strawberry.rocks/docs/pr/1648/general/mutations).

Any idea what is wrong? Is there a different syntax inside `<Note>`?

patrick91: it needs the spaces, now it looks ok! https://strawberry.rocks/docs/pr/1648/general/mutations

RELEASE.md
```diff
@@ -0,0 +1,20 @@
...
+@strawberry.type
+class Mutation:
+    @strawberry.field
+    def do_something(self, arg: int) -> None:
+        return
+results in this schema:
+```grapqhl
+type Mutation {
+    doSomething(arg: Int!): Void
+}
```
patrick91, line 20: I think this file got messed up because of the suggestions I made 😊 can you fix it? I'm unable to push to your branch :)

paulo-raca: Whoops, I didn't notice :facepalm:

docs/general/mutations.md
paulo-raca: Thanks!

botberry:

paulo-raca:
Thanks! :heart:

PR #1683, opened 2022-02-24
Retrigger release

patrick91:
Not 100% sure what happened, but the latest release wasn't sent to PyPI. Error is here:

https://github.com/strawberry-graphql/strawberry/runs/5321083922?check_suite_focus=true

/cc @marcoacierno do you have any clue why this happened?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,43 +4,7 @@ CHANGELOG
 0.98.1 - 2022-02-24
 -------------------
 
-Adds support for converting pydantic conlist.
-Note that constraint is not enforced in the graphql type.
-Thus, we recommend always working on the pydantic type such that the validation is enforced.
-
-```python
-import strawberry
-from pydantic import BaseModel, conlist
-
-class Example(BaseModel):
-    friends: conlist(str, min_items=1)
-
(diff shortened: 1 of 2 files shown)


marcoacierno:
Weird, there are some weird stuff that from a quick look I don't understand

The action you sent says it failed because of commits not pulled (so something pushed between the job start and the push?)

```
hint: Updates were rejected because the remote contains work that you do
hint: not have locally. This is usually caused by another repository pushing
hint: to the same ref. You may want to first integrate the remote changes
hint: (e.g., 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```

I can see there are 2 failed attempts of that workflow

The first attempt https://github.com/strawberry-graphql/strawberry/actions/runs/1893790665/attempts/1 failed because github_release couldn't find the tag 0.98.1, even if it was created correctly: https://github.com/strawberry-graphql/strawberry/runs/5320877091?check_suite_focus=true#step:6:24 which makes me think of a cache/race condition issue?

I wonder if the reason the other workflow failed is because the first workflow pushed the tag, when the workflow was re-attempted the second time it started at the same point as the first attempt so it didn't have the latest tag pushed?

But all of this might be nonsense 🤔

PR #1684, opened 2022-02-25
Make `strawberry.Private` compatible with PEP-563

### LINKED ISSUE #1586: `from __future__ import annotations` incompatible with `strawberry.Private`
Using a minor variation from the [getting started tutorial](https://strawberry.rocks/docs) (addition of `from __future__ import annotations` and the use of `strawberry.Private`) results in a traceback. The code works as expected when commenting out the `from __future__` import.

```

import typing
import strawberry

@strawberry.type
class Book:
    title: str
    author: str
    
    foo: strawberry.Private[int]


def get_books():
    return [
        Book(
            title='The Great Gatsby',
            author='F. Scott Fitzgerald',
            foo=1,
        ),
    ]


@strawberry.type
class Query:
    books: typing.List[Book] = strawberry.field(resolver=get_books)


```

Starting the built-in strawberry server results in the following traceback:
```
$ strawberry server schema
  File "/path/to/project/virtualenv/lib/python3.10/site-packages/graphql/type/definition.py", line 735, in fields
    fields = resolve_thunk(self._fields)
  File "/path/to/project/virtualenv/lib/python3.10/site-packages/graphql/type/definition.py", line 264, in resolve_thunk
    return thunk() if callable(thunk) else thunk
  File "/path/to/project/virtualenv/lib/python3.10/site-packages/strawberry/schema/schema_converter.py", line 280, in get_graphql_fields
    graphql_fields[field_name] = self.from_field(field)
  File "/path/to/project/virtualenv/lib/python3.10/site-packages/strawberry/schema/schema_converter.py", line 136, in from_field
    field_type = self.from_non_optional(field.type)
  File "/path/to/project/virtualenv/lib/python3.10/site-packages/strawberry/schema/schema_converter.py", line 261, in from_non_optional
    of_type = self.from_type(type_)
  File "/path/to/project/virtualenv/lib/python3.10/site-packages/strawberry/schema/schema_converter.py", line 469, in from_type
    raise TypeError(f"Unexpected type '{type_}'")
TypeError: Unexpected type 'typing.Annotated[int, <strawberry.private.StrawberryPrivate object at 0x7f72dcd2b820>]'

During handling of the above exception, another exception occurred:

  File "/path/to/project/virtualenv/bin/strawberry", line 8, in <module>
    sys.exit(run())
  File "/path/to/project/virtualenv/lib/python3.10/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/path/to/project/virtualenv/lib/python3.10/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/path/to/project/virtualenv/lib/python3.10/site-packages/click/core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/path/to/project/virtualenv/lib/python3.10/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/path/to/project/virtualenv/lib/python3.10/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/path/to/project/virtualenv/lib/python3.10/site-packages/strawberry/cli/commands/server.py", line 46, in server
    schema_symbol = import_module_symbol(schema, default_symbol_name="schema")
  File "/path/to/project/virtualenv/lib/python3.10/site-packages/strawberry/utils/importer.py", line 15, in import_module_symbol
    module = importlib.import_module(module_name)
  File "/dls_sw/apps/mamba/0.19.1/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/path/to/project/./schema.py", line 29, in <module>
  File "/path/to/project/virtualenv/lib/python3.10/site-packages/strawberry/schema/schema.py", line 84, in __init__
    self._schema = GraphQLSchema(
  File "/path/to/project/virtualenv/lib/python3.10/site-packages/graphql/type/schema.py", line 208, in __init__
    collect_referenced_types(query)
  File "/path/to/project/virtualenv/lib/python3.10/site-packages/graphql/type/schema.py", line 423, in collect_referenced_types
    collect_referenced_types(field.type)
  File "/path/to/project/virtualenv/lib/python3.10/site-packages/graphql/type/schema.py", line 422, in collect_referenced_types
    for field in named_type.fields.values():
  File "/dls_sw/apps/mamba/0.19.1/lib/python3.10/functools.py", line 970, in __get__
    val = self.func(instance)
  File "/path/to/project/virtualenv/lib/python3.10/site-packages/graphql/type/definition.py", line 737, in fields
    raise TypeError(f"{self.name} fields cannot be resolved. {error}")
TypeError: Book fields cannot be resolved. Unexpected type 'typing.Annotated[int, <strawberry.private.StrawberryPrivate object at 0x7f72dcd2b820>]'
```

Python version:
```
$ python
Python 3.10.1 | packaged by conda-forge | (main, Dec 22 2021, 01:39:05) [GCC 9.4.0] on linux
```

Strawberry version:
```
$ pip show strawberry-graphql
Name: strawberry-graphql
Version: 0.94.0
Summary: A library for creating GraphQL APIs
Home-page: https://strawberry.rocks/
Author: Patrick Arminio
Author-email: patrick.arminio@gmail.com
License: MIT
Location: /path/to/project/virtualenv/lib/python3.10/site-packages
Requires: python-multipart, backports.cached-property, click, graphql-core, python-dateutil, pygments, typing_extensions, sentinel
Required-by: 
```

OS: RHEL7.9

skilkis:
## Description

This commit Closes #1586 by checking for private fields at schema
conversion time. As per [PEP-563], in the future Python annotations will
no longer be evaluated at definition time. As reported by Issue #1586,
this means that `strawberry.Private` is incompatible with postponed
evaluation as the check for a private field requires an evaluated
type-hint annotation to work.

By checking for private fields at schema conversion time, it is
guaranteed that all fields that should be included in the schema are
resolvable using an eval. This ensures that the current approach for
defining private fields can be left intact.

The current filtering for fields annotated with `strawberry.Private` in
`types.type_resolver.py` are left intact to not needlessly instantiate
`StrawberryField` objects when `strawberry.Private` is resolvable.

Summary of Changes:
- Added check for private fields at schema-conversion time
- Added test to check that postponed evaluation with
  `strawberry.Private` functions correctly
- Reduced code duplication in schema_converter.py

[PEP-563]: https://www.python.org/dev/peps/pep-0563/

## Types of Changes

- [x] Core
- [x] Bugfix

Issue #1586

## Checklist

### DIFF
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -18,6 +18,7 @@ from typing import (
     overload,
 )
 
+import sentinel
 from backports.cached_property import cached_property
 from typing_extensions import Literal
 
@@ -38,6 +39,8 @@ if TYPE_CHECKING:
 
 
 _RESOLVER_TYPE = Union[StrawberryResolver, Callable, staticmethod, classmethod]
+
+UNRESOLVED = sentinel.create("UNRESOLVED")
 
 
 class StrawberryField(dataclasses.Field):
@@ -212,7 +215,7 @@ class StrawberryField(dataclasses.Field):
 
             return self.type_annotation.resolve()
         except NameError:
-            return None  # type: ignore
+            return UNRESOLVED
 
     @type.setter
     def type(self, type_: Any) -> None:
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -40,11 +40,7 @@ if TYPE_CHECKING:
 _RESOLVER_TYPE = Union[StrawberryResolver, Callable, staticmethod, classmethod]
 
 
-class UnresolvedType(type):
-    pass
-
-
-UNRESOLVED = UnresolvedType()
+UNRESOLVED = object()
 
 
 class StrawberryField(dataclasses.Field):
@@ -199,7 +195,7 @@ class StrawberryField(dataclasses.Field):
         _ = resolver.arguments
 
     @property  # type: ignore
-    def type(self) -> Union[StrawberryType, type]:  # type: ignore
+    def type(self) -> Union[StrawberryType, type, Literal[UNRESOLVED]]:  # type: ignore
         # We are catching NameError because dataclasses tries to fetch the type
         # of the field from the class before the class is fully defined.
         # This triggers a NameError error when using forward references because
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -18,7 +18,6 @@ from typing import (
     overload,
 )
 
-import sentinel
 from backports.cached_property import cached_property
 from typing_extensions import Literal
 
@@ -40,7 +39,12 @@ if TYPE_CHECKING:
 
 _RESOLVER_TYPE = Union[StrawberryResolver, Callable, staticmethod, classmethod]
 
-UNRESOLVED = sentinel.create("UNRESOLVED")
+
+class UnresolvedType(type):
+    pass
+
+
+UNRESOLVED = UnresolvedType()
 
 
 class StrawberryField(dataclasses.Field):
--- a/strawberry/schema/schema_converter.py
+++ b/strawberry/schema/schema_converter.py
@@ -195,6 +195,7 @@ class GraphQLCoreConverter:
             TypeError: If the type of a field in ``fields`` is `UNRESOLVED`
         """
         thunk_mapping = {}
+
         for f in fields:
             if f.type is UNRESOLVED:
                 raise TypeError(
--- a/strawberry/schema/schema_converter.py
+++ b/strawberry/schema/schema_converter.py
@@ -1,7 +1,18 @@
 from __future__ import annotations
 
 from enum import Enum
-from typing import Any, Callable, Dict, List, Optional, Tuple, Type, Union, cast
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    Tuple,
+    Type,
+    TypeVar,
+    Union,
+    cast,
+)
 
 from graphql import (
     GraphQLArgument,
@@ -34,6 +45,7 @@ from strawberry.exceptions import (
 )
 from strawberry.field import StrawberryField
 from strawberry.lazy_type import LazyType
+from strawberry.private import is_private
 from strawberry.schema.config import StrawberryConfig
 from strawberry.schema.types.scalar import _make_scalar_type
 from strawberry.type import StrawberryList, StrawberryOptional, StrawberryType
@@ -163,6 +175,38 @@ class GraphQLCoreConverter:
             deprecation_reason=field.deprecation_reason,
         )
 
+    FieldType = TypeVar("FieldType", GraphQLField, GraphQLInputField)
+
+    @staticmethod
+    def _get_thunk_mapping(
+        fields: List[StrawberryField],
+        name_converter: Callable[[StrawberryField], str],
+        field_converter: Callable[[StrawberryField], FieldType],
+    ) -> Dict[str, FieldType]:
+        return {
+            name_converter(f): field_converter(f)
+            for f in fields
+            if not is_private(f.type)
+        }
+
+    def get_graphql_fields(
+        self, type_definition: TypeDefinition
+    ) -> Dict[str, GraphQLField]:
+        return self._get_thunk_mapping(
+            fields=type_definition.fields,
+            name_converter=self.config.name_converter.from_field,
+            field_converter=self.from_field,
+        )
+
+    def get_graphql_input_fields(
+        self, type_definition: TypeDefinition
+    ) -> Dict[str, GraphQLInputField]:
+        return self._get_thunk_mapping(
+            fields=type_definition.fields,
+            name_converter=self.config.name_converter.from_field,
+            field_converter=self.from_input_field,
+        )
+
     def from_input_object(self, object_type: type) -> GraphQLInputObjectType:
         type_definition = object_type._type_definition  # type: ignore
 
@@ -174,18 +218,9 @@ class GraphQLCoreConverter:
             assert isinstance(graphql_object_type, GraphQLInputObjectType)  # For mypy
             return graphql_object_type
 
-        def get_graphql_fields() -> Dict[str, GraphQLInputField]:
-            graphql_fields = {}
-            for field in type_definition.fields:
-                field_name = self.config.name_converter.from_field(field)
-
-                graphql_fields[field_name] = self.from_input_field(field)
-
-            return graphql_fields
-
         graphql_object_type = GraphQLInputObjectType(
             name=type_name,
-            fields=get_graphql_fields,
+            fields=lambda: self.get_graphql_input_fields(type_definition),
             description=type_definition.description,
         )
 
@@ -206,18 +241,9 @@ class GraphQLCoreConverter:
             assert isinstance(graphql_interface, GraphQLInterfaceType)  # For mypy
             return graphql_interface
 
-        def get_graphql_fields() -> Dict[str, GraphQLField]:
-            graphql_fields = {}
-
-            for field in interface.fields:
-                field_name = self.config.name_converter.from_field(field)
-                graphql_fields[field_name] = self.from_field(field)
-
-            return graphql_fields
-
         graphql_interface = GraphQLInterfaceType(
             name=interface_name,
-            fields=get_graphql_fields,
+            fields=lambda: self.get_graphql_fields(interface),
             interfaces=list(map(self.from_interface, interface.interfaces)),
             description=interface.description,
         )
@@ -243,16 +269,6 @@ class GraphQLCoreConverter:
             assert isinstance(graphql_object_type, GraphQLObjectType)  # For mypy
             return graphql_object_type
 
-        def get_graphql_fields() -> Dict[str, GraphQLField]:
-            graphql_fields = {}
-
-            for field in object_type.fields:
-                field_name = self.config.name_converter.from_field(field)
-
-                graphql_fields[field_name] = self.from_field(field)
-
-            return graphql_fields
-
         is_type_of: Optional[Callable[[Any, GraphQLResolveInfo], bool]]
         if object_type.is_type_of:
             is_type_of = object_type.is_type_of
@@ -266,7 +282,7 @@ class GraphQLCoreConverter:
 
         graphql_object_type = GraphQLObjectType(
             name=object_type_name,
-            fields=get_graphql_fields,
+            fields=lambda: self.get_graphql_fields(object_type),
             interfaces=list(map(self.from_interface, object_type.interfaces)),
             description=object_type.description,
             is_type_of=is_type_of,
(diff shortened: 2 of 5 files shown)


botberry:



---

Add support for postponed evaluation of annotations
([PEP-563](https://www.python.org/dev/peps/pep-0563/)) to `strawberry.Private`
annotated fields.

## Example

This release fixes Issue #1586 using schema-conversion time filtering of
`strawberry.Private` fields for PEP-563. This means the following is now
supported:

```python

@strawberry.type
class Query:
    foo: "strawberry.Private[int]"
```

Forward references are supported as well:

```python

from dataclasses import dataclass

@strawberry.type
class Query:
    private_foo: strawberry.Private[SensitiveData]

    @strawberry.field
    def foo(self) -> int:
        return self.private_foo.visible

@dataclass
class SensitiveData:

    visible: int
    not_visible int
```


---




Here's the tweet text: 
```

```

strawberry/schema/schema_converter.py
```diff
@@ -163,6 +175,38 @@ def from_input_field(self, field: StrawberryField) -> GraphQLInputField:
             deprecation_reason=field.deprecation_reason,
         )
 
+    FieldType = TypeVar("FieldType", GraphQLField, GraphQLInputField)
+
+    @staticmethod
+    def _get_thunk_mapping(
```
BryceBeagle, line 181: Can you add a (brief) docstring documenting what this function does?

strawberry/schema/schema_converter.py
```diff
@@ -163,6 +175,38 @@ def from_input_field(self, field: StrawberryField) -> GraphQLInputField:
...
+        fields: List[StrawberryField],
+        name_converter: Callable[[StrawberryField], str],
+        field_converter: Callable[[StrawberryField], FieldType],
```
BryceBeagle, line 184: Is there a reason these pull from args instead of using `self.` variables directly (which is where the values come from anywhere). Is this abstraction beneficial?

tests/schema/test_private_field.py
```diff
@@ -55,3 +55,29 @@ def age_in_months(self) -> int:
...
+    class SensitiveData:
+        value: int
+        info: str
+
+    schema = strawberry.Schema(query=Query)
+
+    result = schema.execute_sync(
+        "query { accesibleInfo }", root_value=Query(not_seen=SensitiveData(1, "foo"))
+    )
+    assert result.data == {"accesibleInfo": "foo"}
```
BryceBeagle, line 83: Should we be instead (or in addition) having this test ensure that `not_seen` does not exist on the schema? This test passes exactly the same if the `strawberry.Private` wrapping is simply removed.

Also, a couple nitpicky requests:
* Unless it serves a purpose for this test (which I don't think it does :thinking:), can you move the `dataclass` import to the top of the file?
* `accessible` is spelled wrong

BryceBeagle (changes requested):
Great work! I've added a few (minor) requests, and this should be good to merge :+1:

tests/schema/test_private_field.py
skilkis: Hey @BryceBeagle thanks a lot for the review! You made a good point about the test still passing if `strawberry.Private` is removed. I'll fix that right away

strawberry/schema/schema_converter.py
skilkis: `name_converter` can be pulled from `self` directly since all type definitions seem to use the same converter, but from what I've seen `field_converter` changes based on if it is called from `self.from_input_object` or `self.from_object`. The reason I made an abstraction here is to separate the logic of the filtering and the creation of a GraphQL Core  `ThunkMapping` from that of selecting which converter(s) to use for the field. This is to prevent having to change `_get_thunk_mapping` if new types are added. With this in mind I didn't pull the converter(s) from `self` directly so that other functions (or in the future a mapping of Strawberry types to GraphQL types ) can decide which one to use.

The amount of code here could be reduced further if there is a way to have a single `get_graphql_fields` method that is responsible for selecting the right converters. But right now that's difficult to do since the static type of `type_definition` in `from_input_object` evaluates to `Any`. I couldn't find a way to address that quickly so I left it as is for now.

strawberry/schema/schema_converter.py
skilkis: Added one now 😄

skilkis:
@BryceBeagle I've implemented your suggestions, thanks again for the great feedback.

Adding a negative-engineering test actually revealed an important limitation: this PR works for classes defined on module-scope but not for within a function for example. This happens since `field.type` is only resolved if the class was defined at module level, otherwise it becomes `None`. `StrawberryAnnotation` uses the accepted approach in PEP-563 using an eval with `class.__module.__dict__`. Depending on the acceptance of [PEP-649](https://www.python.org/dev/peps/pep-0649/) there might be support for resolving type-hints of locally-scoped types in the future. However, this means that for now the filtering will not work if a class is defined outside of module-scope. Please see [7d359de](https://github.com/strawberry-graphql/strawberry/pull/1684/commits/7d359de633e9aabd9e7af78b0747fa83d1b53a37) for the failing test.

From here I see two options:
1. Modify `is_private` to optionally raise an error when `type_ is None`. Doing so would make sure that schema conversion fails if an unresolved type exists when performing the file check. This could be implemented as follows: 
 ```python
  def is_private(type_: object, raise_if_unresolved: bool = False) -> bool:
      if get_origin(type_) is Annotated:
          return any(
              isinstance(argument, StrawberryPrivate) for argument in get_args(type_)
          )
      elif type_ is None and raise_if_unresolved:
          raise TypeError(
              f"Could not determine if type '{type_}' is private. Ensure that the "
              "parent class of this field is defined in the module namespace."
          )
      else:
          return False
```

2. Update the documentation / release notes to make the user aware of the potential security vulnerability if they chose to define classes within a local-scope. This is how [Pydantic](https://pydantic-docs.helpmanual.io/usage/postponed_annotations/) handles it, but it would mean that users not reading the documentation might run into this vulnerability with `strawberry.Private`

Let me know how you would like to proceed!

**Review by patrick91 — APPROVED** [review]
Looks good to me! I'll wait for @BryceBeagle's final review before merging this :)

**Comment by patrick91** [comment]
@skilkis I'd go with option 2, but I'd do it in another PR :)

we don't support types defined in functions already and I think it does make sense to add a user friendly error :)

**Review by BryceBeagle — APPROVED** [review]
Hey @skilkis, thanks for the very thorough responses! I'm going to approve it now so it can get merged when needed, but I'll also leave my comments here:

> this PR works for classes defined on module-scope but not for within a function for example

Yeah... this is an unsolved problem we've had for a while with a lot of our typing things. Main use-case has been in tests, so it's not been the worst thing at least.

> From here I see three options:

I like option #2 the best. All of our types are supposed to be considered "unresolved" until schema generation to give room for everything to get defined in one step, then resolved in another. Feel free to add it in this PR, or open a ticket tracking it so it can be performed in a future one.

I don't understand #3, though. What "potential security vulnerability" are you referring to? Is there a potential attack vector here?

**Comment by skilkis** [comment]
@BryceBeagle @patrick91 thanks for approving this PR! I'm excited to be contributing to Strawberry and happy to see that we all agreed on going for Option 2 😄 I would prefer to implement it here quickly before the PR gets merged. It's not a lot of code but can tie up this loose-end with the failing test and more importantly with private fields making their way into the schema.

@BryceBeagle the security vulnerability I mentioned is about leaking sensitive data. I assume it occurs since the type gets resolved as `None` and then makes its way into the schema as a `Void` field. The data is then leaked via en error message when the return of `not_seen` is not `None`. To demonstrate this we can extend the failing test in this PR and query the `notSeen` field in the schema: 

```python
def test_private_field_defined_outside_module_scope():
    """Check compatibility of strawberry.Private when defined outside module scope."""

    @strawberry.type
    class LocallyScopedQuery:
        not_seen: "strawberry.Private[LocallyScopedSensitiveData]"

        @strawberry.field
        def accessible_info(self) -> str:
            return self.not_seen.info

    @dataclass
    class LocallyScopedSensitiveData:
        value: int
        info: str

    @strawberry.type
    class Query:

        @strawberry.field
        def local_query(self) -> LocallyScopedQuery:
            return LocallyScopedQuery(not_seen=LocallyScopedSensitiveData(1, "foo"))

    import uvicorn

    graphql_app = GraphQLRouter(schema)

    app = FastAPI()
    uvicorn.run(app, host="0.0.0.0", port=8000)
```

Query:

```graphql
query LeakingData {
  localQuery {
    notSeen
  }
}
```

Output:

```json
{
  "data": {
    "localQuery": {
      "notSeen": null
    }
  },
  "errors": [
    {
      "message": "Expected 'None', got 'test_private_field_defined_outside_module_scope.<locals>.LocallyScopedSensitiveData(value=1, info='foo')'",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "path": [
        "localQuery",
        "notSeen"
      ]
    }
  ]
}
```

The error message is long, but if you scroll you can see that the sensitive data gets leaked from the  `__repr__` of `LocallyScopedSensitiveData`.

patrick91:
> @BryceBeagle the security vulnerability I mentioned is about leaking sensitive data. I assume it occurs since the type gets resolved as None and then makes its way into the schema as a Void field.

this is probably not the best behaviour, do you know in what other cases it happens?

skilkis:
@patrick91 this is the only one I'm aware of right now. It can be solved by Option 2 since right now the type gets resolved to 'None' and then gets converted to a 'Void' field. I'll push a commit to showcase Option 2, and then you can decide if it belongs in this PR. Otherwise we drop it with a rebase and I make a new PR

strawberry/schema/schema_converter.py
```diff
@@ -190,12 +190,21 @@ def _get_thunk_mapping(
...
-            for f in fields
-            if not is_private(f.type)
-        }
+        thunk_mapping = {}
+        for f in fields:
+            if f.type is UNRESOLVED:
+                raise TypeError(
+                    f"Could not resolve the type of '{f.name}'. Check that the class "
+                    "is accessible from the global module scope."
+                )
```
patrick91, line 203: can we make a custom exception for this? 😊

strawberry/field.py
```diff
@@ -39,6 +40,8 @@
 
 _RESOLVER_TYPE = Union[StrawberryResolver, Callable, staticmethod, classmethod]
 
+UNRESOLVED = sentinel.create("UNRESOLVED")
```
patrick91, line 43: I think we removed the sentinel package recently, can you rebase? 

maybe we create our own sentinel without add the package back again :)

skilkis:
@patrick91 I rebased to the latest release and implemented your suggestions! Should be ready to merge 😄

patrick91:
I've rebased this again and fixed the sentinel 😊

botberry:

PR #1686, opened 2022-02-27
Add support for query params for all integrations

### LINKED ISSUE #834: Support GET requests for queries?
[Apollo client supports](https://www.apollographql.com/docs/react/networking/advanced-http-networking/) using GET instead of POST for GraphQL queries.

As far as I can tell strawberry doesn't seem to support this (at least [in the ASGI wrapper](https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/asgi/__init__.py#L237)). Am I missing something? Is there already a plan to do so?

### LINKED ISSUE #1584: Support GET for fetching queries when GraphiQL is turned off (FastAPI)
The official documentation lists best practices for serving over HTTP [here](https://graphql.org/learn/serving-over-http).
In particular, it mentions support of [GET requests to fetch queries](https://graphql.org/learn/serving-over-http/#get-request).

Strawberry does not currently support this (at least not in the FastAPI integration, unsure about others), and will simply return 404 when GraphiQL is turned off. 

I think it should! For the FastAPI integration, we probably need something along the lines of how POST requests are handled, cf. https://github.com/strawberry-graphql/strawberry/blob/b021481d253f5e6cc56f4bd09c548f97acd9fd84/strawberry/fastapi/router.py#L111-L123

MattExact:
## Description

Support GraphQL operations via GET requests using query parameters.

If a GET request request contains query parameters, they are parsed and executed.
Otherwise, if GraphiQL should be rendered, then  GET request returns the rendered GraphiQL.

I have standardised the conditions when GraphiQL should be rendered.
This is now when GraphiQL is enabled and the Accept headers contain "text/html" or "\*/\*".

While making these changes I also came across some issues with the Chalice integration's error handling, so I have fixed that as part of this PR.

Potential snag points on this PR that I would like some discussion on:

- Handling query params for Sanic - Sanic uses `urllib` to parse query params, and returns a list for each variable, so I am only taking the first value - but there could be a better way to handle this.
- My changes resulted in changes to some of the responses for bad GET requests when GraphiQL is disabled, and there are still inconsistencies between integrations in the response codes returned (at the moment either a 400, 404 or 415). This could potentially be standardised across integrations.
- Documentation may be needed to describe the request formats Strawberry supports. This may be needed on an integration level as there are differences in the formats each supports, i.e. whether multipart form data is supported.
- In relation to the point above, should multipart form data support be added to integrations that support it where it is not there?
- Test coverage - there are a couple of files where I need to improve the coverage

I am going to look at adding the documentation and test coverage now - just wanted to get the PR out so the rest of the code can be critiqued.
There are quite a lot of changes in this, so would appreciate as many eyes as possible and a thorough review!

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* Closes #1584
* Closes #834

## Checklist

- [x] My change requires a change to the documentation. _(not 100% sure if needed but could be good to add)_

### DIFF
--- a/docs/integrations/aiohttp.md
+++ b/docs/integrations/aiohttp.md
@@ -4,8 +4,8 @@ title: AIOHTTP
 
 # AIOHTTP
 
-Strawberry comes with a basic AIOHTTP integration. It provides a view that you can
-use to serve your GraphQL schema:
+Strawberry comes with a basic AIOHTTP integration. It provides a view that you
+can use to serve your GraphQL schema:
 
 ```python
 import strawberry
@@ -30,11 +30,15 @@ app.router.add_route("*", "/graphql", GraphQLView(schema=schema))
 The `GraphQLView` accepts two options at the moment:
 
 - `schema`: mandatory, the schema created by `strawberry.Schema`.
-- `graphiql`: optional, defaults to `True`, whether to enable the GraphiQL interface.
+- `graphiql`: optional, defaults to `True`, whether to enable the GraphiQL
+  interface.
+- `allow_queries_via_get`: optional, defaults to `True`, whether to enable
+  queries via `GET` requests
 
 ## Extending the view
 
-The base `GraphQLView` class can be extended by overriding the following methods:
+The base `GraphQLView` class can be extended by overriding the following
+methods:
 
 - `async get_context(self, request: aiohttp.web.Request, response: aiohttp.web.StreamResponse) -> object`
 - `async get_root_value(self, request: aiohttp.web.Request) -> object`
@@ -42,9 +46,9 @@ The base `GraphQLView` class can be extended by overriding the following methods
 
 ## get_context
 
-By overriding `GraphQLView.get_context` you can provide a custom context object for
-your resolvers. You can return anything here; by default GraphQLView returns a
-dictionary with the request.
+By overriding `GraphQLView.get_context` you can provide a custom context object
+for your resolvers. You can return anything here; by default GraphQLView returns
+a dictionary with the request.
 
 ```python
 import strawberry
@@ -68,12 +72,14 @@ class Query:
 Here we are returning a custom context dictionary that contains only one item
 called `"example"`.
 
-Then we can use the context in a resolver. In this case the resolver will return `1`.
+Then we can use the context in a resolver. In this case the resolver will return
+`1`.
 
 ## get_root_value
 
-By overriding `GraphQLView.get_root_value` you can provide a custom root value for your
-schema. This is probably not used a lot but it might be useful in certain situations.
+By overriding `GraphQLView.get_root_value` you can provide a custom root value
+for your schema. This is probably not used a lot but it might be useful in
+certain situations.
 
 Here's an example:
 
@@ -93,17 +99,17 @@ class Query:
     name: str
 ```
 
-Here we configure a Query where requesting the `name` field will return `"Patrick"`
-through the custom root value.
+Here we configure a Query where requesting the `name` field will return
+`"Patrick"` through the custom root value.
 
 ## process_result
 
-By overriding `GraphQLView.process_result` you can customize and/or process results
-before they are sent to a client. This can be useful for logging errors, or even hiding
-them (for example to hide internal exceptions).
+By overriding `GraphQLView.process_result` you can customize and/or process
+results before they are sent to a client. This can be useful for logging errors,
+or even hiding them (for example to hide internal exceptions).
 
-It needs to return an object of `GraphQLHTTPResponse` and accepts the request and
-execution result.
+It needs to return an object of `GraphQLHTTPResponse` and accepts the request
+and execution result.
 
 ```python
 from aiohttp import web
--- a/docs/integrations/aiohttp.md
+++ b/docs/integrations/aiohttp.md
@@ -15,7 +15,9 @@ from strawberry.aiohttp.views import GraphQLView
 
 @strawberry.type
 class Query:
-    pass
+    @strawberry.field
+    def hello(self, name: str = "World") -> str:
+        return f"Hello, {name}!"
 
 
 schema = strawberry.Schema(query=Query)
--- a/strawberry/aiohttp/handlers/http_handler.py
+++ b/strawberry/aiohttp/handlers/http_handler.py
@@ -38,8 +38,8 @@ class HTTPHandler:
         if request.query:
             try:
                 request_data = parse_request_data(
-                    request.query
-                )  # type:ignore[arg-type]
+                    request.query  # type:ignore[arg-type]
+                )
             except MissingQueryError:
                 raise web.HTTPBadRequest(reason="No GraphQL query found in the request")
 
--- a/strawberry/aiohttp/handlers/http_handler.py
+++ b/strawberry/aiohttp/handlers/http_handler.py
@@ -19,6 +19,7 @@ class HTTPHandler:
         self,
         schema: BaseSchema,
         graphiql: bool,
+        allow_queries_via_get: bool,
         get_context,
         get_root_value,
         process_result,
@@ -26,6 +27,7 @@ class HTTPHandler:
     ):
         self.schema = schema
         self.graphiql = graphiql
+        self.allow_queries_via_get = allow_queries_via_get
         self.get_context = get_context
         self.get_root_value = get_root_value
         self.process_result = process_result
@@ -70,18 +72,23 @@ class HTTPHandler:
         context = await self.get_context(request, response)
         root_value = await self.get_root_value(request)
 
+        allowed_operation_types = set(OperationType.from_http(method))
+
+        if not self.allow_queries_via_get and method == "GET":
+            allowed_operation_types = allowed_operation_types - {OperationType.QUERY}
+
         try:
             result = await self.schema.execute(
                 query=request_data.query,
                 root_value=root_value,
                 variable_values=request_data.variables,
                 context_value=context,
-                allowed_operation_types=OperationType.from_http(method),
+                allowed_operation_types=list(allowed_operation_types),
             )
         except InvalidOperationTypeError as e:
             raise web.HTTPBadRequest(
-                reason=f"{e.operation_type.value}s are not allowed when using {method}"
-            )
+                reason=e.as_http_error_reason(method=method)
+            ) from e
 
         response_data = await self.process_result(request, result)
         response.text = json.dumps(response_data)
--- a/strawberry/aiohttp/handlers/http_handler.py
+++ b/strawberry/aiohttp/handlers/http_handler.py
@@ -1,13 +1,17 @@
 import json
 from io import BytesIO
 from pathlib import Path
-from typing import Any, Dict
+from typing import Any, Dict, Union
+
+from typing_extensions import Literal
 
 from aiohttp import web
 from strawberry.exceptions import MissingQueryError
 from strawberry.file_uploads.utils import replace_placeholders_with_files
 from strawberry.http import GraphQLRequestData, parse_request_data
 from strawberry.schema import BaseSchema
+from strawberry.schema.exceptions import InvalidOperationTypeError
+from strawberry.types.graphql import OperationType
 
 
 class HTTPHandler:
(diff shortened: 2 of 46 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Matt Exact for the PR 👏

```

MattExact:
> Almost identical to mine (#1759). Missing the following though: when testing, you're sending `json` on get requests. Query args are send via the `query_string`.

Ah cheers, not sure why I went with using werkzeug to encode the url

Though I'm not sure if you're getting a bit confused - we don't want to remove support/testing for JSON

Will check that out - given me the bump I needed to finish off and land this PR!

basvandriel:
> Though I'm not sure if you're getting a bit confused - we don't want to remove support/testing for JSON

Ofcourse. I think for every POST request the `application/json` Content-Type should be used. So the `json` property. 
https://graphql.org/learn/serving-over-http/#post-request

patrick91:
@MattExact thanks for doing this to *all* the integrations :)

I was chatting with @basvandriel about adding a check for only allowing queries. I think we can add an `allowed_operations` param to `execute` and `execute_sync` and have it based on the request type (GET only allows QUERY, POST allows all), what do you think?

patrick91:
@MattExact I've pushed a commit that disables mutations and subscriptions in get requests, this was based on @basvandriel's work in https://github.com/strawberry-graphql/strawberry/pull/1759/

Hope that's ok! I'll try to finish this and add it to the other integrations tomorrow (and fix mypy)

we should probably also allow to disable queries via GET (if for some reason someone doesn't want them)

patrick91 (approved):
I've made some changes and disallowed mutations (and subscriptions) in all GET requests 😊 I've also added a flag to disable queries via GET.

I'll give another look at this tomorrow and then merge! thanks for the initial work (especially for trying all the integrations at once)

strawberry/flask/views.py
```diff
@@ -35,21 +38,24 @@ def process_result(self, result: ExecutionResult) -> GraphQLHTTPResponse:
...
 
     def dispatch_request(self):
-        if "text/html" in request.environ.get("HTTP_ACCEPT", ""):
-            if not self.graphiql:
-                abort(404)
-
-            template = render_graphiql_page()
-            return self.render_template(template=template)
+        method = request.method
+        content_type = request.content_type or ""
```
BryceBeagle, line 42: This function (and some above) are missing return annotations

strawberry/schema/schema.py
```diff
@@ -34,6 +35,13 @@
 from .execute import execute, execute_sync
 
 
+DEFAULT_ALLOWED_OPERATION_TYPES = (
+    OperationType.QUERY,
+    OperationType.MUTATION,
+    OperationType.SUBSCRIPTION,
+)
```
BryceBeagle, line 42:
```suggestion
DEFAULT_ALLOWED_OPERATION_TYPES = {
    OperationType.QUERY,
    OperationType.MUTATION,
    OperationType.SUBSCRIPTION,
}
```

strawberry/types/graphql.py
```diff
@@ -0,0 +1,24 @@
...
+    def from_http(method: str) -> Tuple[OperationType, ...]:
+        if method == "GET":
+            return (OperationType.QUERY,)
+
+        if method == "POST":
+            return (
+                OperationType.QUERY,
+                OperationType.MUTATION,
+                OperationType.SUBSCRIPTION,
+            )
```
BryceBeagle, line 22: This should just return a `set` because every usage of it immediately converts it to a set

strawberry/types/graphql.py
```diff
@@ -0,0 +1,24 @@
+from __future__ import annotations
+
+import enum
+from typing import Tuple
+
+
+class OperationType(enum.Enum):
+    QUERY = "query"
+    MUTATION = "mutation"
+    SUBSCRIPTION = "subscription"
```
BryceBeagle, line 10: Are these values necessary? Can we use `auto()`?

strawberry/types/graphql.py
```diff
@@ -0,0 +1,24 @@
...
+            return (OperationType.QUERY,)
+
+        if method == "POST":
+            return (
+                OperationType.QUERY,
+                OperationType.MUTATION,
+                OperationType.SUBSCRIPTION,
+            )
+
+        raise ValueError(f"Unsupported HTTP method: {method}")  # pragma: no cover
```
BryceBeagle, line 24: What is this `no cover` for?

tests/aiohttp/test_query_params.py
```diff
@@ -0,0 +1,71 @@
...
+    query = {
+        "query": """
+            query {
+                hello
+            }
+        """
+    }
+
+    response = await aiohttp_app_client_no_get.get("/graphql", params=query)
+    assert response.status == 400
```
BryceBeagle, line 69: Why does this return a `400` and not a `405`?

tests/asgi/test_query_params.py
```diff
@@ -0,0 +1,36 @@
...
+def test_post_fails_with_query_params(test_client):
+    response = test_client.post("/graphql", params={"query": "{ hello }"})
+
+    assert response.status_code == status.HTTP_415_UNSUPPORTED_MEDIA_TYPE
+
+
+def test_fails_mutation_using_get(test_client):
+    response = test_client.get("/graphql", params={"query": "mutation { hello }"})
+
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
```
BryceBeagle, line 28: Why does this return a 400 and not a 405?

BryceBeagle (changes requested):
Such a big diff -- was unfortunately too long to give the whole thing a super detailed view :disappointed:

docs/integrations/aiohttp.md
```diff
@@ -4,8 +4,8 @@ title: AIOHTTP
 
 # AIOHTTP
 
-Strawberry comes with a basic AIOHTTP integration. It provides a view that you can
```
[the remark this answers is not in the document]
patrick91: yes, I formatted the code (it's a habit). I was thinking of making sure we do that in pre-commit, but that's for another PR :)

strawberry/aiohttp/handlers/http_handler.py
```diff
@@ -35,23 +41,54 @@ async def handle(self) -> web.StreamResponse:
...
+                query=request_data.query,
+                root_value=root_value,
+                variable_values=request_data.variables,
+                context_value=context,
+                allowed_operation_types=allowed_operation_types,
+            )
+        except InvalidOperationTypeError as e:
+            raise web.HTTPBadRequest(
+                reason=e.as_http_error_reason(method=method)
+            ) from e
```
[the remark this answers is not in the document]
patrick91: the linter suggested me to do this, I'm not 100% sure it is the best idea, I'd need double check, it might be useful

strawberry/asgi/handlers/http_handler.py
```diff
@@ -117,25 +121,52 @@ async def get_http_response(
...
         return JSONResponse(response_data, status_code=status.HTTP_200_OK)
 
+    def should_render_graphiql(self, request: Request) -> bool:
+        if not self.graphiql:
+            return False
+
+        return any(
+            supported_header in request.headers.get("accept", "")
+            for supported_header in ("text/html", "*/*")
+        )
```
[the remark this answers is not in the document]
patrick91: I'd do this in a future task after we do this: https://github.com/strawberry-graphql/strawberry/issues/1809

strawberry/asgi/handlers/http_handler.py
```diff
@@ -117,25 +121,52 @@ async def get_http_response(
...
     def get_graphiql_response(self) -> HTMLResponse:
         html = get_graphiql_html()
 
         return HTMLResponse(html)
 
     async def execute(
-        self, query, variables=None, context=None, operation_name=None, root_value=None
+        self,
+        query: str,
+        variables: Dict[str, Any] = None,
```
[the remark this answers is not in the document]
patrick91: yes, I think right now there's an implicit optional, but better to make it explicit :)

strawberry/asgi/handlers/http_handler.py
```diff
@@ -117,25 +121,52 @@ async def get_http_response(
...
+        context: Any = None,
+        operation_name: Optional[str] = None,
+        root_value: Any = None,
```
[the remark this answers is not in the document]
patrick91: probably :D I still find it weird to use object as a type annotation

strawberry/chalice/views.py
```diff
@@ -33,44 +32,24 @@ def has_html_been_asked_for(headers: CaseInsensitiveMapping) -> bool:
...
-
-        return False
+        return any(
+            supported_header in request.headers.get("accept", "")
+            for supported_header in ("text/html", "*/*")
+        )
 
     @staticmethod
-    def invalid_query_response() -> Response:
+    def error_response(message, error_code, http_status_code, headers=None) -> Response:
```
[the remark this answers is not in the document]
patrick91: done

strawberry/chalice/views.py
```diff
@@ -81,41 +60,63 @@ def execute_request(self, request: Request) -> Response:
...
-                )
 
-        if not request.method == "POST":
-            return self.invalid_rest_verb_response()
-
-        try:
-            request_data = request.json_body
-        except BadRequestError:
-            return self.invalid_query_response()
+        if request.method not in ["POST", "GET"]:
```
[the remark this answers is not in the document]
patrick91: done

strawberry/asgi/handlers/http_handler.py
```diff
@@ -117,25 +121,52 @@ async def get_http_response(
...
 
         return JSONResponse(response_data, status_code=status.HTTP_200_OK)
 
+    def should_render_graphiql(self, request: Request) -> bool:
+        if not self.graphiql:
+            return False
+
+        return any(
+            supported_header in request.headers.get("accept", "")
+            for supported_header in ("text/html", "*/*")
```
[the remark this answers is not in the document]
patrick91: done, though I'm not sure how much this is useful :)

strawberry/types/graphql.py
patrick91: what's the benefit in using auto?

strawberry/types/graphql.py
patrick91: I didn't want to add a test for this case :)

tests/aiohttp/test_query_params.py
patrick91: because the method is supported, but the usage is wrong :)

strawberry/types/graphql.py
BryceBeagle: It indicates that the enum is only ever used as a symbol and never for its value. It can be pretty useful when exploring an API

patrick91: we do use it by value: `OperationType(definition.operation.value)` inside https://github.com/strawberry-graphql/strawberry/blob/867e3a4b0d60a79262264cda43f060f195352c5b/strawberry/utils/operation.py#L35

BryceBeagle: I'm a but confused why that's not just `return definition.operation`?

patrick91: that's a different enum (GraphQL-core's)

BryceBeagle: Oh because that's graphql-core's operation :facepalm:

strawberry/asgi/handlers/http_handler.py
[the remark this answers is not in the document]
patrick91: Let's leave this for another PR, we'd need to fix in a lot of places anyway, hopefully it will happen with the refactoring of the integrations

strawberry/aiohttp/handlers/http_handler.py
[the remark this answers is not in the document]
patrick91: image
doesn't look like it changes anything, but might be useful for logs, so I'll leave it there

PR #1701, opened 2022-03-06
Fix issue when using generics and interfaces together

### LINKED ISSUE #1654: Generic class implementing a `strawberry.interface` raises error
Congrats on this awesome package, I've been using it for the past 2 weeks and really enjoy it! In my present usage I've run into 
an issue with interfaces that are implemented by generics. This is demonstrated by the minimal example below:

````python
from typing import Any, Generic, TypeVar

import strawberry


@strawberry.interface
class ObjectType:

    obj: strawberry.Private[Any]

    @strawberry.field
    def repr(self) -> str:
        return str(self.obj)


T = TypeVar("T")


@strawberry.type
class GenericObject(ObjectType, Generic[T]):
    @strawberry.field
    def value(self) -> T:
        return self.obj


@strawberry.type
class Query:
    @strawberry.field
    def foo(self) -> GenericObject[str]:
        return GenericObject("foo")


query_result = schema.execute_sync(
    """
        query {
            foo {
                __typename
                value
                repr
            }
        }
    """
)
print(query_result)


````

Will raise the following error:

````bash
graphql.error.graphql_error.GraphQLError: Expected value of type 'StrGenericObject' but got: <GenericObject instance>.

GraphQL request:3:13
2 |         query {
3 |             foo {
  |             ^
4 |                 __typename
````

However, changing the use of `strawberry.interface` to `strawberry.type` as follows:
```python
@strawberry.type
class ObjectType:
    ...
```

Will return the expected result:

```bash
ExecutionResult(data={'foo': {'__typename': 'StrGenericObject', 'value': 'foo', 'repr': 'foo'}}, errors=None, extensions={})
```

patrick91:
This pr closes #1654 :)

### DIFF
--- a/strawberry/schema/schema_converter.py
+++ b/strawberry/schema/schema_converter.py
@@ -285,23 +285,30 @@ class GraphQLCoreConverter:
             assert isinstance(graphql_object_type, GraphQLObjectType)  # For mypy
             return graphql_object_type
 
-        is_type_of: Optional[Callable[[Any, GraphQLResolveInfo], bool]]
-        if object_type.is_type_of:
-            is_type_of = object_type.is_type_of
-        elif object_type.interfaces:
+        def _get_is_type_of() -> Optional[Callable[[Any, GraphQLResolveInfo], bool]]:
+            if object_type.is_type_of:
+                return object_type.is_type_of
+
+            if not object_type.interfaces:
+                return None
 
             def is_type_of(obj: Any, _info: GraphQLResolveInfo) -> bool:
+                if object_type.concrete_of and (
+                    hasattr(obj, "_type_definition")
+                    and obj._type_definition.origin is object_type.concrete_of.origin
+                ):
+                    return True
+
                 return isinstance(obj, object_type.origin)
 
-        else:
-            is_type_of = None
+            return is_type_of
 
         graphql_object_type = GraphQLObjectType(
             name=object_type_name,
             fields=lambda: self.get_graphql_fields(object_type),
             interfaces=list(map(self.from_interface, object_type.interfaces)),
             description=object_type.description,
-            is_type_of=is_type_of,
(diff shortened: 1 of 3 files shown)


strawberry/schema/schema_converter.py
```diff
@@ -480,3 +471,25 @@ def from_union(self, union: StrawberryUnion) -> GraphQLUnionType:
...
+            return object_type.is_type_of
+
+        if object_type.interfaces:
+
+            def is_type_of(obj: Any, _info: GraphQLResolveInfo) -> bool:
+                if object_type.concrete_of and (
+                    hasattr(obj, "_type_definition")
+                    and obj._type_definition.origin is object_type.concrete_of.origin
+                ):
+                    return True
```
patrick91, line 489: @BryceBeagle any suggestion for improving this? :)

tests/schema/test_generics.py
```diff
@@ -1,17 +1,17 @@
 import textwrap
-import typing
 from enum import Enum
```
patrick91, line 2: I changed the imports :)

strawberry/schema/schema_converter.py
BryceBeagle: Hmm... 

Since this is only used in one specific place, I would probably _not_ make this a method on the class, instead making it another nested function in `from_object`.

Other than that, I'm not sure of anything else that could be done.

tests/schema/test_generics.py
BryceBeagle: :clap:

botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

strawberry/schema/schema_converter.py
patrick91: moved :)

Issue #1707, opened 2022-03-07
automatic persisted queries like Apollo GraphQL


basvandriel:
[More info here](https://www.apollographql.com/docs/apollo-server/performance/apq/) and related to #960

indigoviolet:
@basvandriel Is this still being worked on?

PR #1708, opened 2022-03-08
automatic persisted queries

basvandriel:
## Description

Supporting [Apollo's automatic persisted queries](https://www.apollographql.com/docs/apollo-server/performance/apq/)
in [Strawberry GraphQL](https://github.com/strawberry-graphql/strawberry) 🍓

## Types of Changes

- [x] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* #1707 

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [tool.poetry]
 name = "strawberry-graphql"
 packages = [ { include = "strawberry" } ]
-version = "0.215.3"
+version = "0.216.0"
 description = "A library for creating GraphQL APIs"
 authors = ["Patrick Arminio <patrick.arminio@gmail.com>"]
 license = "MIT"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 CHANGELOG
 =========
 
+0.216.0 - 2023-12-06
+--------------------
+
+Override encode_json() method in Django BaseView to use DjangoJSONEncoder
+
+Contributed by [Noam Stolero](https://github.com/noamsto) via [PR #3273](https://github.com/strawberry-graphql/strawberry/pull/3273/)
+
+
 0.215.3 - 2023-12-06
 --------------------
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,0 @@
-Release type: minor
-
-Override encode_json() method in Django BaseView to use DjangoJSONEncoder


botberry:



---

Supporting [Apollo's automatic persisted queries](https://www.apollographql.com/docs/apollo-server/performance/apq/)
in [Strawberry GraphQL](https://github.com/strawberry-graphql/strawberry) 🍓

## Notes
- For persisting queries, a simple `TTLCache` is used. Will be changed later.
- Only flask views are supported for now.

## Benchmarks
All benchmarks are made using the `pytest-benchmark` package.

|                          | min (ms) | max (ms) | average (ms) |
|--------------------------|----------|----------|--------------|
| no persisting, default   | 1.9934   | 99.0644  | 2.5739       |
| persistiging, not cached | 3.8509   | 211.6294 | 5.0793       |
| persisting, cached       | 1.8495   | 21.6539  | 2.1182       |


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to @bvandriel for the PR 👏

```

strawberry/persistance/exceptions.py
```diff
@@ -0,0 +1,6 @@
+class PersistedQueryNotFoundError(Exception):
```
estyxx, line 1: Until now we have kept all the exceptions in `strawberry.exceptions (except for pydantic because it is still experimental). So I will move this exception with others.

tests/flask/persistance/app.py
```diff
@@ -0,0 +1,19 @@
+from flask import Flask
+from strawberry.flask.persistance.view import ApolloGraphQLPersistanceView
+
+from ..schema import Query, schema
+
+
+def create_app(**kwargs):
```
estyxx, line 7: Can we share the `persistence.app` in all the flask tests? or there is something particular we have to set for persistence's tests?

tests/flask/persistance/conftest.py
```diff
@@ -0,0 +1,23 @@
+from typing import Callable
+
+import pytest
+
+from .app import create_app
+
+
+@pytest.fixture
+def client():
```
estyxx, line 9: same above :)

tests/persistance/test_exceptions.py
```diff
@@ -0,0 +1,18 @@
...
+
+def test_notfound_exception_error_and_message():
+    with pytest.raises(Exception) as exc:
+        raise PersistedQueryNotFoundError()
+
+    assert exc.value.code == "PERSISTED_QUERY_NOT_FOUND"
+    assert exc.value.message == "PersistedQueryNotFound"
+
+
+def test_found_exception_instance():
```
estyxx, line 14: 🤔don't think we need this test...  would be better to have it check in an 'integration' test if it's possible :)

estyxx (commented):
Hey, thank you for this PR! looks good :) 
I just added some little comments :)

strawberry/flask/persistance/view.py
```diff
@@ -0,0 +1,126 @@
...
+from strawberry.persistance.funcs.create_notfound_response import (
+    create_notfound_response,
+)
+from strawberry.persistance.funcs.get_extensions import get_extensions
+from strawberry.persistance.funcs.get_query_hash import get_query_hash
+from strawberry.persistance.http import GraphQLRequestData
+
+
+# Let's see.
+cache: Cache = TTLCache(maxsize=1000, ttl=timedelta(hours=12), timer=datetime.now)
```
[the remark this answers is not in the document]
basvandriel: Absolutely. Used that for test purposes really, I'll fix it when having CI passed.

tests/flask/persistance/app.py
basvandriel: `ApolloGraphQLPersistanceView` is specific for using persistance on the Flask view. Since the `query` is not always in the request (when the hash is only provided for example), `parse_request_data` will throw an error. This is why I couldn't use extensions.

strawberry/flask/persistance/view.py
```diff
@@ -0,0 +1,132 @@
...
+    @property
+    def hash_in_cache(self) -> bool:
+        return self.query_hash in cache
+
+    def __init__(self, **kwargs) -> None:
+        """
+        Initialize the request data
+        """
+        data = self.resolve_request_data()
+        query: str = data.get("query") or ""
```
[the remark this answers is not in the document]
patrick91: unless data["query"] might be set as None

strawberry/persistance/funcs/create_notfound_response.py
```diff
@@ -0,0 +1,14 @@
+from strawberry.http import GraphQLHTTPResponse
```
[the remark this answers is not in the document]
patrick91: also I think this function can just be inside the view :)

strawberry/persistance/http.py
```diff
@@ -0,0 +1,32 @@
...
+
+class RequestPopulator(ABC):
+    def populate(self, data: Dict):
+        """Populates the Flask request with dictionary-data"""
+        raise NotImplementedError("Method should be implemented in subtypes")
+
+
+class RequestJsonPopulator(RequestPopulator):
+    def populate(self, data: Dict):
+        request._cached_json = (data, data)
```
patrick91, line 27: I'm not sure this is somethign we should do, using a private attribute to return JSON sounds wrong, is there a particular reason why we don't return JSON directly?

patrick91 (commented):
Great work so far! I've added some initial comments, let me know if you have any doubts :)

strawberry/persistance/__init__.py
```diff
@@ -0,0 +1,5 @@
+from typing import Final
+
+
+# The key where the query hash is in.
+APOLLO_PERSTISANCE_EXT_KEY: Final = "persistedQuery"
```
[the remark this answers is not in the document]
basvandriel: `persisted-query` would be invalid, this is the object key the Apollo Client sends

strawberry/persistance/http.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class RequestArgsPopulator(RequestPopulator):
+    def populate(self, data: Dict):
+        request.args = ImmutableMultiDict(data)
```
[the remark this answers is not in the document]
basvandriel: This is created for preparation to another issue. The default Flask view is only retrieving data from the `request.json` right now. Apollo (and other clients) also send data via the request.args when using GET requests. More on that later.

strawberry/persistance/http.py
basvandriel: Regarding my previous comment, the default Flask view initially checks for data in the `request.json` dict. In this case, this always gets called, unless when a custom view is defined. That's why re-populating the request json is nessacery in this case.

strawberry/exceptions.py
```diff
@@ -240,3 +240,11 @@ def __init__(self):
...
         )
         super().__init__(message)
+
+
+class PersistedQueryNotFoundError(Exception):
+    message: str = "PersistedQueryNotFound"
+    code: str = "PERSISTED_QUERY_NOT_FOUND"
+
+    def __init__(self):
+        super().__init__(self.code)
```
BryceBeagle, line 250: Should the `super()` call provide a message that's more like a sentence? What's the reasoning for using `code` as the exception message (not to be confused with the `.message` here)

strawberry/persistance/funcs/get_extensions.py
```diff
@@ -0,0 +1,7 @@
+from typing import Dict, Optional
+
+from strawberry.utils.get_dict_normalize_if_str import get_dict_normalize_if_str
+
+
+def get_extensions(data: Dict) -> Optional[Dict]:
```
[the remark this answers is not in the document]
BryceBeagle: Yeah these functions should live on a class and not be module-namespaced

BryceBeagle (changes requested):
Great addition. I have a couple smaller requests too

strawberry/flask/views.py
```diff
@@ -42,15 +42,19 @@ def dispatch_request(self):
             template = render_graphiql_page()
             return self.render_template(template=template)
 
-        if request.content_type.startswith("multipart/form-data"):
+        # GET requests doesn't have content types
```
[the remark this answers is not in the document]
basvandriel: Whoops.

pyproject.toml
```diff
@@ -46,6 +46,7 @@ sanic = {version = ">=20.12.2,<22.0.0", optional = true}
 aiohttp = {version = "^3.7.4.post0", optional = true}
 fastapi = {version = ">=0.65.2", optional = true}
 "backports.cached-property" = "^1.0.1"
+cachetools = "^5.0.0"
```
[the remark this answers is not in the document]
patrick91: actually, do you think we could make this pluggable? maybe we can provide a basic lru_cache (the python builtin) and then allow people to override it?

strawberry/exceptions.py
patrick91: we also aren't settings code and message anywhere, right?

patrick91 (commented):
Added some more comments :)

jkimbo:
Is there a reason this cannot be implemented as an extension? That way we can use this for all web frameworks.

PR #1714, opened 2022-03-10
Allow all types to `Schema(types=[...])`

paulo-raca:
## Description

I accidently found that Strawberry assumes all types specified in
Schema constructor are expected to be object types.

Because of this, setting `types=[InputType]` would cause the schema to contain `type InputType`
instead of `input InputType`, and everything crashed from there (Same for interfaces).

This change ensures all types are allowed :)

This is a minor fix and probably doesn't require a release

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/strawberry/schema/schema.py
+++ b/strawberry/schema/schema.py
@@ -4,6 +4,7 @@ from typing import Any, Dict, Optional, Sequence, Type, Union
 from graphql import (
     ExecutionContext as GraphQLExecutionContext,
     GraphQLNamedType,
+    GraphQLNonNull,
     GraphQLSchema,
     get_introspection_query,
     parse,
@@ -79,9 +80,13 @@ class Schema(BaseSchema):
 
         graphql_types = []
         for type_ in types:
-            graphql_type = self.schema_converter.from_type(type_)
-            if isinstance(graphql_type, GraphQLNamedType):
-                graphql_types.append(graphql_type)
+            graphql_type = self.schema_converter.from_maybe_optional(type_)
+            if isinstance(graphql_type, GraphQLNonNull):
+                graphql_type = graphql_type.of_type
+            assert isinstance(
+                graphql_type, GraphQLNamedType
+            ), f"{graphql_type} is not a named GraphQL Type"
+            graphql_types.append(graphql_type)
 
         self._schema = GraphQLSchema(
             query=query_type,
--- a/strawberry/schema/schema.py
+++ b/strawberry/schema/schema.py
@@ -3,6 +3,7 @@ from typing import Any, Dict, Optional, Sequence, Type, Union
 
 from graphql import (
     ExecutionContext as GraphQLExecutionContext,
+    GraphQLNamedType,
     GraphQLSchema,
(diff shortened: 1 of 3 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

strawberry/schema/schema.py
```diff
@@ -78,8 +79,9 @@ def __init__(
 
         graphql_types = []
         for type_ in types:
-            graphql_type = self.schema_converter.from_object(type_._type_definition)
-            graphql_types.append(graphql_type)
+            graphql_type = self.schema_converter.from_type(type_)
+            if isinstance(graphql_type, GraphQLNamedType):
+                graphql_types.append(graphql_type)
```
patrick91, line 84: why do we need this check?

paulo-raca: Good question!

I initially added it just to keep mypy happy :grimacing: 

but now that you brought it up, I've looked a bit more into it:

As far as I can tell, the only non-named type that could be returned here is a GraphQLList.

Adding `types=[List[MyType]]` would definitely be weird, but shouldn't be ignored. I'll replace it with an assertion instead

I'll replace this with an assertion instead

patrick91: yes, assertion would be better! thanks 😊

strawberry/schema/schema.py
```diff
@@ -78,7 +80,12 @@ def __init__(
 
         graphql_types = []
         for type_ in types:
-            graphql_type = self.schema_converter.from_object(type_._type_definition)
+            graphql_type = self.schema_converter.from_maybe_optional(type_)
+            if isinstance(graphql_type, GraphQLNonNull):
+                graphql_type = graphql_type.of_type
+            assert isinstance(
+                graphql_type, GraphQLNamedType
+            ), f"{graphql_type} is not a named GraphQL Type"
```
BryceBeagle, line 88: I see that @patrick91 requested this be an `assert` but I think this would just be better as a custom exception type. Asserts shouldn't typically be used to check logic the user can influence (except in tests).

BryceBeagle (changes requested):

strawberry/schema/schema.py
patrick91: yes, you're right :) a custom exception is better 😊 sorry @paulo-raca!

paulo-raca: Can it be a ValueError? :thinking:

patrick91: let's make our own custom exception :)

`InvalidTypeError` or something like this :)

paulo-raca: `SchemaConverter.from_type()` raises TypeError. Sounds close enough :)

patrick91 (approved):
Sorry I missed the update for this PR! Looks perfect to me! Thanks you! 😊

paulo-raca:
Thanks!

PR #1723, opened 2022-03-13
Improve schema directives printer

bellini666:
* Improve schema directives printer
* Print the schema definitions and any extra types defined by them as well
* Fix typing issues that the change to SchemaDirective caused

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/strawberry/directive.py
+++ b/strawberry/directive.py
@@ -4,7 +4,7 @@ import dataclasses
 import inspect
 import sys
 from itertools import islice
-from typing import Callable, List, Optional, TypeVar
+from typing import Any, Callable, List, Optional, TypeVar
 
 from graphql import DirectiveLocation
 
@@ -13,7 +13,7 @@ from strawberry.arguments import StrawberryArgument
 from strawberry.field import StrawberryField
 
 
-def directive_field(name: str) -> StrawberryField:
+def directive_field(name: str) -> Any:
     return StrawberryField(
         python_name=None,
         graphql_name=name,
--- a/strawberry/federation/field.py
+++ b/strawberry/federation/field.py
@@ -15,7 +15,6 @@ from typing_extensions import Literal
 
 from strawberry.field import _RESOLVER_TYPE, StrawberryField, field as base_field
 from strawberry.permission import BasePermission
-from strawberry.schema_directive import StrawberrySchemaDirective
 from strawberry.unset import UNSET
 
 
@@ -41,7 +40,7 @@ def field(
     deprecation_reason: Optional[str] = None,
     default: Any = UNSET,
     default_factory: Union[Callable, object] = UNSET,
-    directives: Sequence[StrawberrySchemaDirective] = (),
+    directives: Sequence[object] = (),
 ) -> T:
     ...
 
@@ -64,7 +63,7 @@ def field(
     deprecation_reason: Optional[str] = None,
     default: Any = UNSET,
     default_factory: Union[Callable, object] = UNSET,
-    directives: Sequence[StrawberrySchemaDirective] = (),
+    directives: Sequence[object] = (),
 ) -> Any:
     ...
 
@@ -87,7 +86,7 @@ def field(
     deprecation_reason: Optional[str] = None,
     default: Any = UNSET,
     default_factory: Union[Callable, object] = UNSET,
-    directives: Sequence[StrawberrySchemaDirective] = (),
+    directives: Sequence[object] = (),
 ) -> StrawberryField:
     ...
 
@@ -109,7 +108,7 @@ def field(
     deprecation_reason=None,
     default=UNSET,
     default_factory=UNSET,
-    directives: Sequence[StrawberrySchemaDirective] = (),
+    directives: Sequence[object] = (),
     # This init parameter is used by PyRight to determine whether this field
     # is added in the constructor or not. It is not used to change
     # any behavior at the moment.
@@ -128,25 +127,25 @@ def field(
     directives = list(directives)
 
     if provides:
-        directives.append(Provides(" ".join(provides)))  # type: ignore
+        directives.append(Provides(" ".join(provides)))
 
     if requires:
-        directives.append(Requires(" ".join(requires)))  # type: ignore
+        directives.append(Requires(" ".join(requires)))
 
     if external:
-        directives.append(External())  # type: ignore
+        directives.append(External())
 
     if shareable:
-        directives.append(Shareable())  # type: ignore
+        directives.append(Shareable())
 
     if tags:
-        directives.append(Tag(" ".join(tags)))  # type: ignore
+        directives.append(Tag(" ".join(tags)))
 
     if override:
-        directives.append(Override(override))  # type: ignore
+        directives.append(Override(override))
 
     if inaccessible:
-        directives.append(Inaccessible())  # type: ignore
+        directives.append(Inaccessible())
 
     return base_field(
         resolver=resolver,
--- a/strawberry/printer.py
+++ b/strawberry/printer.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
+import dataclasses
 from itertools import chain
 from typing import TYPE_CHECKING, Any, Dict, Optional, cast
 
+from graphql.language.printer import print_ast
 from graphql.type import (
     is_enum_type,
     is_input_type,
@@ -10,6 +12,8 @@ from graphql.type import (
     is_scalar_type,
     is_specified_directive,
 )
+from graphql.type.directives import GraphQLDirective
+from graphql.utilities.ast_from_value import ast_from_value
 from graphql.utilities.print_schema import (
     is_defined_type,
     print_args,
@@ -26,7 +30,7 @@ from graphql.utilities.print_schema import (
 )
 
 from strawberry.field import StrawberryField
-from strawberry.schema_directive import Location
+from strawberry.schema_directive import Location, StrawberrySchemaDirective
 from strawberry.types.types import TypeDefinition
 from strawberry.unset import UNSET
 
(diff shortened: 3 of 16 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

strawberry/printer.py
```diff
@@ -34,21 +40,55 @@
...
 def print_schema_directive(
     directive: StrawberrySchemaDirective, schema: BaseSchema
 ) -> str:
-    params = directive.instance.__dict__ if directive.instance else {}
+    from strawberry.schema.schema import Schema
+
+    if isinstance(schema, Schema):
+        schema_converter = schema.schema_converter
+    else:
+        schema_converter = GraphQLCoreConverter(schema.config, {})
```
patrick91, line 83: why do we need this check?

strawberry/schema/schema_converter.py
```diff
@@ -135,6 +140,46 @@ def from_directive(self, directive: StrawberryDirective) -> GraphQLDirective:
...
+                type_annotation=StrawberryAnnotation(
+                    annotation=field.type,
+                    namespace=module.__dict__,
+                ),
+                default=default,
+            )
+            args[self.config.name_converter.from_argument(arg)] = self.from_argument(
+                arg
+            )
+
```
patrick91, line 174: don't we do something like this somewhere else already?

tests/test_printer/test_schema_directives.py
```diff
@@ -79,8 +105,10 @@ def user(self, input: Input) -> User:
 
     type User @sensitiveData(reason: "GDPR") {
       firstName: String!
+      phone: String! @sensitiveData(reason: "PRIVATE", meta: [{key: "can_share_field", value: "phone_share_accepted"}])
+      phoneShareAccepted: Boolean!
     }
-    """
+    """  # noqa:E501
```
patrick91, line 112: 💯 

thanks so much for doing this, I kept postponing it :D

strawberry/printer.py
bellini666: Because `BaseSchema` doesn't have a `.schema_converter` in it, and those printers are expecting a `BaseSchema` instance.

Maybe `BaseSchema` should add a `schema_converter` to its protocol?

strawberry/schema/schema_converter.py
bellini666: Hrm, not that I found (but I'm still no expert in strawberry's code).

This is similar to the `from_directive` code.

strawberry/printer.py
patrick91: yes, let's do that for now, I think we should revisit Schema and BaseSchema, as BaseSchema doesn't really provide any useful thing at the moment, we are still far away from having a full abstraction on GraphQL core :)

bellini666: Done!

bellini666:
Hey @patrick91 , I refactored this PR to consider the schema directive refactor changes. I also added printing of the schema directives itself and any extra typing they define.

I also fixed some typing issues that arise from the schema directive refactor itself.

strawberry/directive.py
```diff
@@ -13,7 +13,7 @@
 from strawberry.field import StrawberryField
 
 
-def directive_field(name: str) -> StrawberryField:
```
BryceBeagle, line 16: Why is this getting changed? This always return `StrawberryField`?

strawberry/federation/field.py
```diff
@@ -87,7 +86,7 @@ def field(
     deprecation_reason: Optional[str] = None,
     default: Any = UNSET,
     default_factory: Union[Callable, object] = UNSET,
-    directives: Sequence[StrawberrySchemaDirective] = (),
```
BryceBeagle, line 90: I'm pretty confused by the loosening of the types everywhere

strawberry/printer.py
```diff
@@ -35,33 +40,77 @@
...
 
 
-def print_schema_directive_params(params: Dict) -> str:
+@dataclasses.dataclass
+class PrintExtras:
+    directives: Set[str] = dataclasses.field(default_factory=set)
+    types: Set[type] = dataclasses.field(default_factory=set)
+
+
+def _serialize_dataclasses(value: Any) -> Any:
```
BryceBeagle, line 49: Can we narrow the type definitions here?
```suggestion
def _serialize_dataclasses(value: object) -> Union[dict, list]:
```

strawberry/federation/field.py
bellini666: That's because the current schema directives are not `StrawberrySchemaDirectives` anymore, they are dataclasses that don't derive from anything.

Consider this example:

```python
import strawberry
from strawberry.schema_directive import Location


@strawberry.schema_directive(locations=[Location.OBJECT])
class SomeSchemaDirective:
    foo: str
    bar: int


@strawberry.type(
    directives=[SomeSchemaDirective(foo="foo", bar=0)],
)
class SomeType:
    bin: str
```

Without this change, it is impossible to use this code without having to `type:ignore` pyright/mypy/etc. For example, running pyright on that code:

```
/tmp/xxx/xxx.py
  /tmp/xxx/xxx.py:12:17 - error: Expected no arguments to "SomeSchemaDirective" constructor (reportGeneralTypeIssues)
  /tmp/xxx/xxx.py:12:16 - error: Argument of type "list[SomeSchemaDirective]" cannot be assigned to parameter "directives" of type "Sequence[StrawberrySchemaDirective] | None" in function "type"
    Type "list[SomeSchemaDirective]" cannot be assigned to type "Sequence[StrawberrySchemaDirective] | None"
      TypeVar "_T_co@Sequence" is covariant
        "SomeSchemaDirective" is incompatible with "StrawberrySchemaDirective"
      Type cannot be assigned to type "None" (reportGeneralTypeIssues)
2 errors, 0 warnings, 0 informations 
```

strawberry/directive.py
bellini666: Yes, but it conflicts with its use case. For example:

```python
import strawberry
from strawberry.schema_directive import Location


@strawberry.schema_directive(locations=[Location.OBJECT])
class SomeSchemaDirective:
    foo: str = strawberry.directive_field(name="foo")
    bar: int
```

If the typing says it returns a `StrawberryField`, that means I can't set it on a variable typed as anything else (i.e. in that example where the type expects `str`).

```
/tmp/xxx/xxx.py
  /tmp/xxx/xxx.py:7:16 - error: Expression of type "StrawberryField" cannot be assigned to declared type "str"
    "StrawberryField" is incompatible with "str" (reportGeneralTypeIssues)
1 error, 0 warnings, 0 informations
```

This is the same reason to why `dataclasses.field` is also typed as `Any` (it is actually overloaded, with the other possibility being the type of variable passed to `default` when it is given)

strawberry/printer.py
bellini666: The problem is that value can actually be anything (i.e. `str`, `int`, an object that is a dataclass, `list`, `dict`)

But I can improve it using overloads, which will match the actual typing closer

BryceBeagle: Right, that's why I think we want `object`. It ensures that anything done to `value` within the function has to be possible on _all_ `object`s. If you use `Any`, the type checker just gives up and lets you do whatever you want

bellini666: It's done!

bellini666:
@BryceBeagle just a ping to check if you can review this again!

@patrick91 do you want to take another look after the latest changes? You previously reviewed an older and incomplete version of this PR

strawberry/directive.py
patrick91: @bellini666 we should implement this for directives: https://peps.python.org/pep-0681/

strawberry/federation/field.py
patrick91: that's correct, unfortunately we can't do much to improve the typing for directives (and types) for the time being

strawberry/directive.py
patrick91: oh, looks like you did already!

bellini666: @patrick91 I already did :)

https://github.com/strawberry-graphql/strawberry/pull/1723/files#diff-f2d5b9421987cb0ccccf8d32cad8245d96c5aaef62893bdd9e94e6aa8fa5c253R39

tests/test_printer/test_schema_directives.py
```diff
@@ -66,14 +79,35 @@ class Input:
...
+            return User(
+                first_name=input.first_name,
+                phone="+551191551234",
+                phone_share_accepted=False,
+            )
 
     expected_type = """
+    directive @sensitiveData(reason: String!, meta: [SensitiveValue!]) on OBJECT | FIELD_DEFINITION
+
+    directive @sensitiveInput(reason: String!, meta: [SensitiveValue!]) on INPUT_OBJECT
```
patrick91, line 109: was there a particular reason for adding the schema directives to the schema? just curious 😊

bellini666: @patrick91 AFAIK (I might be wrong =P) the schema directives should be printed to the schema, shouldn't them?

But the main reason here is to allow those to be parsed by external libs, and when they find a field/type/etc "decorated" by a schema directive, they can look at the definition and understand what they mean exactly. In this example exactly, the field is defined as `phone: String! @sensitiveData(reason: "PRIVATE", meta: [{key: "can_share_field", value: "phone_share_accepted"}])`. Without knowing definition it would be impossible to know the real type for `key` and `value` (in this case they are actually `String`, but they could be an enum for example)

patrick91: yes, the reason I'm asking is that for example apollo federation doesn't need them to be printed so I thought it wouldn't be necessary for most use cases 😊

are you doing any work with schema directives and external tools?

bellini666: I'm actually already using it for a while now :)

Basically I created a custom [schema directive extension](https://github.com/blb-ventures/strawberry-django-plus/blob/main/strawberry_django_plus/directives.py#L103) to be able to run schema directives at runtime.

I use it mainly to [check permissions](https://github.com/blb-ventures/strawberry-django-plus/blob/main/strawberry_django_plus/permissions.py) (i.e. the user is logged in, the user is a superuser or the user has a given django permission globally or in the returned object). And since those directives get printed to the schema, my frontend partners parse it to know in runtime if the user can or cannot access some field, allowing them to dynamically generate queries/forms/etc. For that to work I had to monkey-patch the printerpy module in `strawberry-django-plus`, which is basically a simpler version of what I ended up suggesting here =P

Btw, that permissioning thing is a simple use case, I was inspired by the approach mentioned in places like:

* https://gqlgen.com/reference/directives/
* https://blog.logrocket.com/a-complete-guide-to-permissions-in-a-graphql-api/
* https://www.graphql-tools.com/docs/schema-directives#enforcing-access-permissions

botberry:
# Pre-release
:wave:

Releasing commit [a2286544e919029df22261c1aac24aff9c9790ff] to PyPi as pre-release! :package:

botberry:
# Pre-release
:wave:

Pre-release **0.114.0.dev.1653002317** [a2286544e919029df22261c1aac24aff9c9790ff] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.114.0.dev.1653002317
```

strawberry/schema/base.py
```diff
@@ -20,6 +21,7 @@
 
 class BaseSchema(Protocol):
     config: StrawberryConfig
+    schema_converter: GraphQLCoreConverter
```
patrick91, line 24: I don't think this change should be here, but for now it is fine, we'll probably refactor the BaseSchema class at some point :)

strawberry/printer.py
```diff
@@ -35,33 +52,95 @@
...
-def print_schema_directive(directive: Any, schema: BaseSchema) -> str:
-    name_converter = schema.config.name_converter
+def print_schema_directive(
+    directive: Any, schema: BaseSchema, *, extras: PrintExtras
+) -> str:
+    strawberry_directive = cast(
+        StrawberrySchemaDirective, directive.__class__.__strawberry_directive__
+    )
+    schema_converter = schema.schema_converter
+    gql_directive = schema_converter.from_schema_directive(directive)
```
patrick91, line 118: in theory the printer shouldn't access the schema converter (it shouldn't know about it)

do you think we change this code to to use the directive directly?

strawberry/schema/base.py
bellini666: This is due to this discussion we had in this comment: https://github.com/strawberry-graphql/strawberry/pull/1723#discussion_r825463393

I can remove it if you prefer though :). Or we can keep it for now and refactor `BaseSchema` in the future

strawberry/printer.py
bellini666: Sure! I didn't know if I should make a big change here so I went for this

We can probably extract and save the schema directives in the `Schema` itself when creating it, just like we do with types, client directives, etc and then retrieve it from there.

Will do that and send an update in this PR soon.

obs. this will also fix the other comment you've made since there will be no reason to access the `schema_converter` in here

strawberry/schema/base.py
patrick91: ah, sorry I didn't review properly last time!

patrick91 (approved):
I spoke with @bellini666 privately about this and improving how the printer works to not depend on the schema convert will take a bit of time, so in order to not keep this PR stuck for a long time I think we can merge it and make some issues for improving the code in future 😊

Issue #1731, opened 2022-03-15
graphql_transport_ws:  Calling `complete` for a single request operation results in a disconnect.


kristjanvalur:
For the client->server `complete` messate, the protocol spericifation states (boldface mine):
```
Client -> Server indicates that the client has stopped listening and wants to complete the subscription. No further events, relevant to the original subscription, should be sent through. Even if the client **completed** a **single result operation** before it resolved, the result should not be sent through once it does.
```
The problem is, that the message **id** of a single results operation is not placed in `self.subscriptions` and so any attempt at referring to it causes a `KeyError` and a closure.

Now, due to the potential race conditions in the bi-directional Websockets channel, it is impossible to guarantee that a `next` message is not sent after a client sends a `complete` message, so ensureing that a `complete` message from the client actually interrupts a single-request-operation is not of the highest importance.  However, it should not cause a disconnect.

the `complete` message is inherently *racy* because both parties can complete a subscription at any time.  Therefore, both `complete` messages to old subscriptions must be ignored, and a client must ignore `next` or `error` messages to subscriptions it has previously `completed`.

kristjanvalur:
please see https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md which explains how single response requests are just special cases of subscriptions.

DoctorJohn:
This issue has been resolved in #1792. Thanks for the report and the PR :)

PR #1732, opened 2022-03-15
Issue #1731 - Allow `complete` messages for single-result-operations

kristjanvalur:
Single result operations must also accept `complete` messages. See #1731

## Description
Since single-result-operations don't put their IDs into the `self.subscriptions` dict, we need to ignore `complete`
messages that don't have an ID in that dict.
Previously such messages would cause a disconnect, because  `KeyError` was caught by ann exception handler intended
to catch malformed messages.

The PR doesn't attempt to actually cancel the single-request-operation.  Although the protocol specification says that a cancellation should be possible, a `complete` message may always come too late for that due to the non-synchronous bi-directional nature of the Websocket: The client may issue a `complete` message while a `next` and `complete` messages are in-flight from the server, and still receive them.  As such, a `complete` message can never guarantee that nothing more is ever received with that ID.

Actually cancelling the single request operation is an optional enhancement, for some future PR.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #1732

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,4 +1,4 @@
 Release type: patch
 
-This release fixes issue #1732, allowing a `graphql-transport-ws` client
+This release fixes issue #1731, allowing a `graphql-transport-ws` client
 to send `complete` messages for a *single result request*
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,4 @@
+Release type: patch
+
+This release fixes issue #1732, allowing a `graphql-transport-ws` client
+to send `complete` messages for a *single result request*
--- a/tests/aiohttp/test_graphql_transport_ws.py
+++ b/tests/aiohttp/test_graphql_transport_ws.py
@@ -419,7 +419,7 @@ async def test_subscription_cancellation(aiohttp_client):
 
         response = await ws.receive_json()
         assert response == CompleteMessage(id="sub2").as_dict()
-        
+
         # Issue #1731
         # Check that a racing complete message to an already completed subscription
         # is ignored by server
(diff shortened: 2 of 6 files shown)


tests/aiohttp/test_graphql_transport_ws.py
```diff
@@ -420,6 +420,11 @@ async def test_subscription_cancellation(aiohttp_client):
         response = await ws.receive_json()
         assert response == CompleteMessage(id="sub2").as_dict()
 
+        # Issue #1731
```
estyxx, line 423: Did we really need to reference the issue in our code? 🤔

kristjanvalur: I don't know, In projects that I have worked on it is customary to link tests to issues if they are regression tests verifying that an issue is fixed.  CPython being one of them.

But I'll follow whatever guidelines set by this project, of course :)

BryceBeagle: I'm not opposed to having the regression test link to the GH issue, but I think it should be a separate unit test.

kristjanvalur: You mean, have a separate .py file for regression tests?  Any suggestions?

BryceBeagle: No, just separate `def test_X` functions. You modified an existing test to check for the regression. I was suggesting that they become distinct tests instead.

RELEASE.md
```diff
@@ -0,0 +1,4 @@
+Release type: patch
+
+This release fixes issue #1732, allowing a `graphql-transport-ws` client
```
```suggestion
This release fixes issue #1731, allowing a `graphql-transport-ws` client
```

BryceBeagle (changes requested):

tests/aiohttp/test_graphql_transport_ws.py
kristjanvalur: Agreed, that was sloppy.

kristjanvalur:
Closing this since #1792 was merged which incorporates this functionality.

Issue #1736, opened 2022-03-15
DataLoader.load/load_many include any exceptions types in return value


huonw:
Thanks for Strawberry!

The `DataLoader` class is neat. It seems like the expectation is that the list of loaded values returned from the `load_fn` includes any failures as `Exception`s directly, and they're propagated through the futures, to be re-thrown when `.load()` is awaited:

https://github.com/strawberry-graphql/strawberry/blob/67e6d749fe5652a7ec5788097aa8962770edf34d/strawberry/dataloader.py#L141-L142

For a fully typed code base, this means the load function return value needs to specify the exceptions, like `-> list[Union[MyRealType, Exception]]`. Currently the `load_fn` parameter is typed as:

https://github.com/strawberry-graphql/strawberry/blob/67e6d749fe5652a7ec5788097aa8962770edf34d/strawberry/dataloader.py#L52

so specifying exceptions in the return value like the above sets `T = Union[MyRealType, Exception]`, and this flows through to the return value of the `load` method. However, it's impossible for `load` to ever return an `Exception`, and so code calling `.load` is forced to do type checks/asserts/casts when it tries to do real things with `MyRealType` (e.g. access properties).

It'd be nice if the types were more accurate, so that `load` doesn't need to return the exceptions. One option might be to have `load_fn` have type something like `load_fn: Callable[[List[K]], Awaitable[List[Union[T, Exception]]]],`. The `List` may need to change to `Sequence` to be backwards compatible, because `List`'s type parameter is invariant.

### Code example

`pyproject.toml`:
```toml
[tool.poetry]
name = "strawberry-testing"
version = "0.1.0"
description = ""
authors = []

[tool.poetry.dependencies]
python = "^3.9"
strawberry-graphql = "^0.103.1"
mypy = "^0.941"

[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
```

<details><summary>Click for <code>poetry.lock</code></summary>

```toml
[[package]]
name = "backports.cached-property"
version = "1.0.1"
description = "cached_property() - computed once per instance, cached as attribute"
category = "main"
optional = false
python-versions = ">=3.6.0"

[[package]]
name = "click"
version = "8.0.4"
description = "Composable command line interface toolkit"
category = "main"
optional = false
python-versions = ">=3.6"

[package.dependencies]
colorama = {version = "*", markers = "platform_system == \"Windows\""}

[[package]]
name = "colorama"
[...]

PR #1737, opened 2022-03-16
Omit exceptions from DataLoader.load/load_many return type annotation

huonw:
## Description

Per the proposal in #1736, this adjusts the type hints on `DataLoader.__init__` so that the `T` type variable never includes exceptions that are re-raised by `load`/`load_many`. In particular, when the `load_fn` is explicitly annotated to return a type like `list[Union[SomeModel, ValueError]]`, the type variable is inferred to `T  = SomeModel`, and thus the return value of `load` is `SomeModel`, as desired.

This requires some somewhat subtle `@overload`s to ensure that using a completely unannotated `load_fn` is handled appropriately. Fortunately there's an existing mypy test covering that, and this PR adds mypy tests for `Union[]`s with exception, and without (`Optional`).

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* #1736

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,5 @@
+Release type: patch
+
+The return type annotation for `DataLoader.load` and `load_many` no longer
+includes any exceptions directly returned by the `load_fn`. The ability to
+handle errors by returning them as elements from `load_fn` is now documented too.
--- a/strawberry/dataloader.py
+++ b/strawberry/dataloader.py
@@ -12,7 +12,10 @@ from typing import (
     Iterable,
     List,
     Optional,
+    Sequence,
     TypeVar,
+    Union,
+    overload,
 )
 
 from .exceptions import WrongNumberOfResultsReturned
@@ -47,9 +50,31 @@ class DataLoader(Generic[K, T]):
     cache: bool = False
     cache_map: Dict[K, Future]
 
+    @overload
+    def __init__(
+        self: "DataLoader[K, T]",
+        # any BaseException is rethrown in 'load', so should be excluded from the T type
+        load_fn: Callable[[List[K]], Awaitable[Sequence[Union[T, BaseException]]]],
+        max_batch_size: Optional[int] = None,
+        cache: bool = True,
+        loop: AbstractEventLoop = None,
+    ) -> None:
+        ...
+
+    # fallback if load_fn is untyped and there's no other info for inference
+    @overload
+    def __init__(
+        self: "DataLoader[K, Any]",
+        load_fn: Callable[[List[K]], Awaitable[List[Any]]],
+        max_batch_size: Optional[int] = None,
+        cache: bool = True,
+        loop: AbstractEventLoop = None,
+    ) -> None:
+        ...
+
     def __init__(
         self,
-        load_fn: Callable[[List[K]], Awaitable[List[T]]],
+        load_fn: Callable[[List[K]], Awaitable[Sequence[Union[T, BaseException]]]],
         max_batch_size: Optional[int] = None,
         cache: bool = True,
         loop: AbstractEventLoop = None,
--- a/strawberry/dataloader.py
+++ b/strawberry/dataloader.py
@@ -52,7 +52,7 @@ class DataLoader(Generic[K, T]):
 
     @overload
     def __init__(
-        self: "DataLoader[K, T]",
+        self,
         # any BaseException is rethrown in 'load', so should be excluded from the T type
         load_fn: Callable[[List[K]], Awaitable[Sequence[Union[T, BaseException]]]],
         max_batch_size: Optional[int] = None,
--- a/tests/mypy/test_dataloaders.yml
+++ b/tests/mypy/test_dataloaders.yml
@@ -26,7 +26,7 @@
     async def load(keys: List[int]) -> List[str]:
         return [str(k) for k in keys]
 
-    loader = DataLoader[int, str](load)
+    loader = DataLoader(load)
 
     reveal_type(loader)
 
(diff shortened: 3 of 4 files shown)


botberry:



---

The return type annotation for `DataLoader.load` and `load_many` no longer
includes any exceptions directly returned by the `load_fn`. The ability to
handle errors by returning them as elements from `load_fn` is now documented too.


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to @huon_w for the PR 👏

```

tests/mypy/test_dataloaders.yml
```diff
@@ -37,3 +37,45 @@
...
+
+    reveal_type(loader)
+
+    async def run() -> None:
+      user = await loader.load(1)
+
+      reveal_type(user)
+  out: |
+    main:10: note: Revealed type is "strawberry.dataloader.DataLoader[builtins.int*, Union[builtins.str, None]]"
+    main:15: note: Revealed type is "Union[builtins.str, None]"
```
patrick91, line 81: nice!

strawberry/dataloader.py
```diff
@@ -47,9 +50,31 @@ class DataLoader(Generic[K, T]):
...
+    ) -> None:
+        ...
+
     def __init__(
         self,
-        load_fn: Callable[[List[K]], Awaitable[List[T]]],
+        load_fn: Callable[[List[K]], Awaitable[Sequence[Union[T, BaseException]]]],
         max_batch_size: Optional[int] = None,
         cache: bool = True,
         loop: AbstractEventLoop = None,
```
patrick91, line 80: I usually remove the types from the non overloaded method so it doesn't have to be in sync with the rest :)

huonw: Ah. I tried what I think you're meaning (let me know if I've misinterpreted) and it looks like the parameters end up typed as `Any` in each overload. Specifically, I switched these to:

```python
    @overload
    def __init__(
        self: "DataLoader[K, T]",
        # any BaseException is rethrown in 'load', so should be excluded from the T type
        load_fn: Callable[[List[K]], Awaitable[Sequence[Union[T, BaseException]]]],
        max_batch_size = None,
        cache = True,
        loop = None,
    ) -> None:
        ...

    # fallback if load_fn is untyped and there's no other info for inference
    @overload
    def __init__(
        self: "DataLoader[K, Any]",
        load_fn: Callable[[List[K]], Awaitable[List[Any]]],
        max_batch_size = None,
        cache = True,
        loop = None,
    ) -> None:
        ...
```

`reveal_type(DataLoader.__init__)` gives (reformatted for ease of reading):

```python
Overload(
  def (
    self: strawberry.dataloader.DataLoader[K`1, T`2],
    load_fn: def (builtins.list[K`1]) -> typing.Awaitable[typing.Sequence[Union[T`2, builtins.BaseException]]], 
    max_batch_size: Any =, 
    cache: Any =, 
    loop: Any =
  ), 
  def (
    self: strawberry.dataloader.DataLoader[K`1, Any],
    load_fn: def (builtins.list[K`1]) -> typing.Awaitable[builtins.list[Any]],
    max_batch_size: Any =, 
    cache: Any =,
    loop: Any =
  )
)
```

and a call like `loader = DataLoader(load, cache="this isn't a bool")` passes type checking.

strawberry/dataloader.py
```diff
@@ -47,9 +50,31 @@ class DataLoader(Generic[K, T]):
     cache: bool = False
     cache_map: Dict[K, Future]
 
+    @overload
+    def __init__(
+        self: "DataLoader[K, T]",
```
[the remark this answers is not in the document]
huonw: Nope, it's not: I just included it for consistency with the other overload. I'll remove it.

strawberry/dataloader.py
patrick91: I meant something like this: https://github.com/strawberry-graphql/strawberry/blob/c7322c0aefc009c021b29de28f1e2a2302bc75f1/strawberry/object_type.py#L152-L190

it's not a big deal anyway 😊

RELEASE.md
```diff
@@ -0,0 +1,5 @@
+Release type: patch
+
+The return type annotation for `DataLoader.load` and `load_many` no longer
+includes any exceptions directly returned by the `load_fn`. The ability to
+handle errors by returning them as elements from `load_fn` is now documented too.
```
patrick91, line 5: Sorry, one last thing, these release notes seem to be a bit misleading, because now we added BaseException to the return type of load_fn, right?

strawberry/dataloader.py
huonw: Ah, I see; thanks for explaining 👍

RELEASE.md
huonw: Could you be a bit more specific about what you find misleading and/or what you'd prefer to see?

I think the change to the type of the `load_fn` parameter is essentially an implementation detail of the change to `load`/`loda_many`, because sub-typing means it's backwards compatible and nearly invisible to users.

patrick91: > The return type annotation for `DataLoader.load` and `load_many` no longer includes any exceptions directly returned by the `load_fn`. 

correct me if I'm wrong or if I misunderstood this, but we didn't touch `DataLoader.load`, right?

huonw: Yeah, we didn't touch the code of `load`, but the visible impact this change has on users is the value of the type parameter `T`, meaning the (inferred) type for `DataLoader[K, T]` and the return type of `load`/`load_many` will change from including `Exception`s to not including them. I think describing it in terms of changing `__init__`/`load_fn` will hide value to users of this change.

patrick91: ah,right! thanks for explaining 😊

patrick91 (approved):

botberry:

huonw:
Woohoo, thank you for reviewing and merging so promptly 🎉

PR #1759, opened 2022-03-22
GET request data retrieval from request.args in Flask

basvandriel:
## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist


botberry:



---

GET-request's data gets no longer retrieved from request.json


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to @bvandriel for the PR 👏

```

patrick91 (changes requested):
Hey, sorry for taking so long to review this :)

So, overall it looks great! I didn't realise that the tests where using "GET" instead of "POST", your change makes sense, since we want to allow queries via "GET", but before merging this, can we make sure we don't allow mutations via "GET"?

strawberry/flask/views.py
```diff
@@ -42,20 +42,32 @@ def dispatch_request(self):
...
+        # Shorthand for using it
 
         try:
             request_data = parse_request_data(data)
         except MissingQueryError:
             return Response("No valid query was provided for the request", 400)
 
+        # Move to the @mutation annotation later?
+        if using_get and "mutation" in request_data.query:
+            return Response("GET requests don't support mutations", 400)
```
patrick91, line 69: not sure this it the best check for this, mutation might appear in a random field of the query :)

I think we might need to do something more clever here and change our `execute` and `execute_sync` functions on the schema to accept a list of operation type they can allow.

so then in dispatch_request we do something like this:


```
        result = self.schema.execute_sync(
            request_data.query,
            variable_values=request_data.variables,
            context_value=context,
            operation_name=request_data.operation_name,
            root_value=self.get_root_value(),
            allowed_operation_types=[QUERY] if get else [MUTATION, QUERY, SUBSCRIPTION]
        )
```

I'm open to other suggestions :) /cc @BryceBeagle for more ideas 😊

BryceBeagle: Yeah, as `request_data.query` is a `str`, I think using `in` is probably pretty unsafe. I don't have any other better solutions though

basvandriel: Currently made use of the `parse` function from `graphql-core` to retrieve the operation type. https://github.com/strawberry-graphql/strawberry/pull/1759/files#diff-88aa6fd17e4c6feac6e7152ebd3f2b8f972544c444a071b550e4d23061b97a3fR76

patrick91:
This was integrated in https://github.com/strawberry-graphql/strawberry/pull/1686 I'll close

PR #1760, opened 2022-03-23
Move console plugin to end

patrick91:
Not sure why I decided to put the console plugin as the first plugin, especially now that we can specify a new one it doesn't make a lot of sense.

Having it at the end allows other plugins to edit the content of the files (for example running black) before they are written on disk.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,5 @@
+Release type: patch
+
+This release moves the console plugin for the codegen command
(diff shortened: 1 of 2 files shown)


botberry:



---

This release moves the console plugin for the codegen command
to be last one, allowing to run code before writing files to
disk.


---




Here's the tweet text: 
```

```

PR #1766, opened 2022-03-25
Add link from GraphQL types to Strawberry types

paulo-raca:
## Description

I'm toying with strawberry internals, and I often need to find the links between `GraphQLType`s and their strawberry definitions.

For now I have been using `Schema.get_type_by_name(graphql_type.name)` and similar funcions, but it feels a bit clumsy.

Instead, this PR takes advantage of `GraphQLNamedType`s (and others) "extensions" to store links back to the originating strawberry type.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

## Checklist


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

strawberry/federation/schema.py
```diff
@@ -68,12 +68,10 @@ def _extend_query_type(self):
...
         fields.update(query_type.fields)
+        query_type_kwargs = query_type.to_kwargs()
+        query_type_kwargs["fields"] = fields
 
-        self._schema.query_type = GraphQLObjectType(
-            name=query_type.name,
-            description=query_type.description,
-            fields=fields,
-        )
+        self._schema.query_type = GraphQLObjectType(**query_type_kwargs)
```
BryceBeagle, line 74: Hmm, I'm not a fan of using `**kwargs`-style syntax unless absolutely necessary as it breaks type-checking. Is there a way to avoid this?

strawberry/schema/schema_converter.py
```diff
@@ -59,6 +59,10 @@
 from .types.concrete_type import ConcreteType
 
 
+# Extension key used to link a GraphQLType back into the Strawberry definition
+STRAWBERRY_DEFINITION = "strawberry-definition"
```
BryceBeagle, line 63: Can we give this a more descriptive name?

Also can we put this on some namespace other than the module? Do you think it would make sense to be on `GraphQLCoreConverter`?

What do you think of `GraphQLCoreConverter.DEFINITION_BACKREF`?

strawberry/federation/schema.py
paulo-raca: I usually agree with you: I could just have added the `extensions=query_type.extensions` argument and have the tests pass.

However in this kind of scenario where we want a "copy with tweaks",  it is **very** easy to leave optional stuff behind (e.g., the code is currently missing `interfaces`, `is_type_of`, `extensions`, `ast_node` and `extension_ast_nodes`. And maybe more stuff in the future). 

Which is why I think `**kwargs` is a good choice here.

However I'll change If you are strongly against it

BryceBeagle: Can we make use of [`copy`](https://docs.python.org/3/library/copy.html#module-copy) + [`GraphQLObjectType.__copy__`](https://github.com/graphql-python/graphql-core/blob/b9423b74ca9d22d009a1a4633b38b2bcd06c604b/src/graphql/type/definition.py#L728)?

paulo-raca: > Can we make use of [`copy`](https://docs.python.org/3/library/copy.html#module-copy) + [`GraphQLObjectType.__copy__`](https://github.com/graphql-python/graphql-core/blob/b9423b74ca9d22d009a1a4633b38b2bcd06c604b/src/graphql/type/definition.py#L728)?

This looks much better, thank you :)

BryceBeagle (approved):

paulo-raca:
Merged latest strawberry changes :)

patrick91 (approved):

patrick91:
@paulo-raca thanks for this PR! Me and @BryceBeagle spent a bit of time refactoring the tests and we also updated the release notes to make it easier to understand what this release adds.

paulo-raca:
Thank you for looking into those!

PR #1739, opened 2022-03-16
Fixes to single-response-requests on `graphql-transport-ws`

kristjanvalur:
Various problems with single-response-requests (as opposed ot streaming requests) are tackled

## Description

single-response-request are now handled as *tasks* and can be overlapped with other requests
on the same connection, can be cancelled, and must have unique requests IDs.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement/optimization
- [ ] Documentation

## Issues Fixed or Closed by This PR

* #1731
* #1733
* #1734
* #1738

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the CONTRIBUTING document.
- [x] I have added tests to cover my changes.
- [x] I have tested the changes and verified that they work and don't break anything (as well as I can manage).

Issue #1738, opened 2022-03-16
Single-result-operation cannot be overlapped or cancelled


kristjanvalur:
Single result operations are performed synchronously in response to a websocket message, meaning that other messages are not handled on the websocket until they complete.
This prevents single-result requests from taking a long time or being cancelled.

kristjanvalur:
The protocol spec at https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md has recently been updated to clarify how single-response-requests are simply special cases of subscriptions.

DoctorJohn:
This issue has been resolved in #1792. Thanks for the report and the PR :)

Issue #1734, opened 2022-03-15
Single-result-operations don't reserve the Subscription ID


kristjanvalur:
Single request operations should behave like regular streaming operations, in that they must have a
unique **Id** (for any active oprations).
The Id provided for a single result operation is not placed in `self.subscriptions` and therefore there is no validation on it.  Hence it would appear that it is possible to schedule two single result operations simultaneously with the same ID, or to use an ID that is currently in use for a regular streaming operation.

All Ids should probably be put into the same table (`self.subscriptions`) when encountered and removed from it once completed.

kristjanvalur:
Please see recently updated https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md

DoctorJohn:
This issue has been resolved in #1792. Thanks for the report and the PR :)

Issue #1733, opened 2022-03-15
graphql-transport-ws handler doesn't free up IDs


kristjanvalur:
when investingating #1731 , I came across a curious thing.
A client `complete` message will cause a `self.cleanup_operation()` for the supplied ID, thus releasing server resoures and freeing it up for re-use on the same Websocket connection.
But a server-side `complete` (i.e. the subscription drains or causes an error) **will not** cause the same kind of cleanup.
From looking at the source, it would appear that Id and resources are only released at the end of the WebsocketConnection, in 
`GraphQLTransportWSHandler.handler_request`.

This is not in accordance with protocol.  the `graphql-transport-ws` spec says:
```
If there is already an **active subscriber** for an operation matching the provided ID, regardless of the operation type, the server must close the socket immediately with the event 4409: Subscriber for <unique-operation-id> already exists.
```

According to this, the ID really should be freed up as soon as the operation is done.  This would also be in accordance with good housekeeping practices, to avoid having the state machinery fill up with old subscription metadata.

kristjanvalur:
Please see updated https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md which discusses how IDs should be forgotten once complete.

DoctorJohn:
This issue has been resolved in #1792. Thanks for the report and the PR :)

PR #1792, opened 2022-04-04
Unify single-response-requests with streaming requests on `graphql-transport-ws`

kristjanvalur:
(replaces closed #1739)

Various problems with single-response-requests existed as a result of their having a separate control
path from streaming requests.  Single-response-requests are really just special cases of streaming requests
that deliver __at most__ one `Next` message and need to behave as such.

## Description

single-response-request are now handled as *tasks* and can be overlapped with other requests
on the same connection, can be cancelled, and share the same unique requests ID namespace.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #1731
* #1733
* #1734
* #1738

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -45,7 +45,7 @@ repos:
       - id: check-toml
 
   - repo: https://github.com/psf/black
-    rev: 22.1.0
+    rev: 22.3.0
     hooks:
       - id: black
         exclude: ^tests/codegen/snapshots/python/
--- a/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
+++ b/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
@@ -2,7 +2,7 @@ import asyncio
 from abc import ABC, abstractmethod
 from contextlib import suppress
 from datetime import timedelta
-from typing import Any, AsyncGenerator, Callable, Dict, Optional
+from typing import Any, AsyncGenerator, Callable, Dict, List, Optional
 
 from graphql import (
     ExecutionResult as GraphQLExecutionResult,
@@ -43,8 +43,9 @@ class BaseGraphQLTransportWSHandler(ABC):
         self.connection_init_timeout_task: Optional[asyncio.Task] = None
         self.connection_init_received = False
         self.connection_acknowledged = False
-        self.subscriptions: Dict[str, AsyncGenerator] = {}
+        self.subscriptions: Dict[str, Optional[AsyncGenerator]] = {}
         self.tasks: Dict[str, asyncio.Task] = {}
+        self.completed_tasks: List[asyncio.Task] = []
 
     @abstractmethod
     async def get_context(self) -> Any:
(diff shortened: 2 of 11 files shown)


botberry:



---

This release fixes a number of problems with single-result-operations over
`graphql-transport-ws` protocol

- operation **IDs** now share the same namespace as streaming operations
  meaning that they cannot be reused while the others are in operation

- single-result-operations now run as *tasks* meaning that messages related
  to them can be overlapped with other messages on the websocket.

- single-result-operations can be cancelled with the `complete` message.

- IDs for single result and streaming result operations are now released
  once the operation is done, allowing them to be re-used later, as well as
  freeing up resources related to previous requests.


---




Here's the tweet text: 
```

```

.pre-commit-config.yaml
```diff
@@ -45,7 +45,7 @@ repos:
       - id: check-toml
 
   - repo: https://github.com/psf/black
-    rev: 22.1.0
+    rev: 22.3.0
```
estyxx, line 48: Maybe is better to update black in a separate PR and run it on all file as well :)

kristjanvalur: To clarify:  Without it, the commit checks would not run.

tests/aiohttp/test_graphql_transport_ws.py
```diff
@@ -285,6 +285,55 @@ async def test_duplicated_operation_ids(aiohttp_client):
         assert data.extra == "Subscriber for sub1 already exists"
 
 
+async def test_reused_operation_ids(aiohttp_client):
```
BryceBeagle, line 288: As these are fairly complicated tests, could you add some docstrings describing what functionality is being tested

BryceBeagle (changes requested):

tests/aiohttp/test_graphql_transport_ws.py
kristjanvalur: Added docstrings.
It is a bit annoying that we have three identical sets of tests, testing _internal_ functionality, that is in theory quite independent of the integration in use.   Would it make sense to separate "integration" tests from unit tests?

BryceBeagle: cc @patrick91, I think we were discussing ideas for doing this last week

patrick91: @kristjanvalur totally, I've opened up an issue for this: https://github.com/strawberry-graphql/strawberry/issues/1809

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -204,24 +170,67 @@ async def handle_streaming_operation(self, message: SubscribeMessage) -> None:
...
+            async def get_result_source():
+                yield await self.schema.execute(
+                    query=message.payload.query,
+                    variable_values=message.payload.variables,
+                    context_value=context,
+                    root_value=root_value,
+                    operation_name=message.payload.operationName,
+                )
+
+            result_source = get_result_source()
```
patrick91, line 193: maybe in a future PR we can move this to a classmethod to clean this method 😊

patrick91 (approved):
this looks great! thanks so much and sorry for the late review!

PR #1813, opened 2022-04-16
changed location of `UNSET` to `strawberry` and `strawberry.unset` & new documentation subsection explaining `UNSET` and `Optional`

### LINKED ISSUE #1805: Add better documentation for `UNSET`
Currently, the only documentation of `UNSET` is on our [pagination docs ](https://strawberry.rocks/docs/guides/pagination)(and in our unit tests). We should add first-class documentation for it

DominiqueGarmier:
## Description

changed the location of the instantiation of `UNSET` to `strawberry.unset`, `UNSET` can also be imported from `strawberry` directly. `is_unset` was also moved to `strawberry.unset`.

Adjusted all imports to import from `strawberry.unset` (and `from strawberry import UNSET` in some tests).

Old imports are still backwards compatible (should I add deprecation warnings when importing from `strawberry.arguments` ?)

Closes #1805

## Types of Changes

- [x] New feature
- [x] Documentation

## Checklist


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to @dominiqgarmier for the PR 👏

```

patrick91:
> Old imports are still backwards compatible (should I add deprecation warnings when importing from strawberry.arguments ?)

Yes please 😊

botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to @dominiqgarmier for the PR 👏

```

strawberry/unset.py
```diff
@@ -1,6 +1,22 @@
...
 
     def __bool__(self):
         return False
+
+
+UNSET: Any = _Unset()
+
+
+def is_unset(value: Any) -> bool:
+    return type(value) is _Unset
```
patrick91, line 16: do you think we can deprecate this function too and allow doing `something is UNSET`? :)

DominiqueGarmier: yes this should work since all instances of `_Unset` come from this single `UNSET`.

DominiqueGarmier: we should also use `Literal[UNSET]` in that case instead of `_Unset`

patrick91: where?

DominiqueGarmier:
I can with no problem implement warnings for `is_unset`, however I don't know If I can do something like that for `UNSET`. It would require some metaclass sorcery with id's and even then I'm not sure if its possilbe.

patrick91:
> > > Old imports are still backwards compatible (should I add deprecation warnings when importing from strawberry.arguments ?)
> > 
> > 
> > Yes please 😊
> 
> I can with no problem implement warnings for `is_unset`, however I don't know If I can do something like that for `UNSET`. It would require some metaclass sorcery with id's and even then I'm not sure if its possilbe.

I think we cause getattr on the module: https://peps.python.org/pep-0562/#rationale

strawberry/unset.py
DominiqueGarmier: https://github.com/strawberry-graphql/strawberry/blob/7d79bf9a7d7f2816bd1243f24216903c020758e3/strawberry/experimental/pydantic/utils.py#L81 and a few more times in tests

patrick91: ah yes! good idea 😊

DominiqueGarmier:
locally `tests/cli/test_server.py::test_automatic_reloading` is failing, it seems the reloading is not getting detected properly. I  think this is just something with my local setup. Hardly think this is due to my changes.

DominiqueGarmier:
> 

im not sure

> > > > Old imports are still backwards compatible (should I add deprecation warnings when importing from strawberry.arguments ?)
> > > 
> > > 
> > > Yes please 😊
> > 
> > 
> > I can with no problem implement warnings for `is_unset`, however I don't know If I can do something like that for `UNSET`. It would require some metaclass sorcery with id's and even then I'm not sure if its possilbe.
> 
> I think we cause getattr on the module: https://peps.python.org/pep-0562/#rationale

I went through all regex matches of "unset" and couldn't find anything even in strings.

patrick91:
> > 
> 
> im not sure
> 
> > > > > Old imports are still backwards compatible (should I add deprecation warnings when importing from strawberry.arguments ?)
> > > > 
> > > > 
> > > > Yes please 😊
> > > 
> > > 
> > > I can with no problem implement warnings for `is_unset`, however I don't know If I can do something like that for `UNSET`. It would require some metaclass sorcery with id's and even then I'm not sure if its possilbe.
> > 
> > 
> > I think we cause getattr on the module: https://peps.python.org/pep-0562/#rationale
> 
> I went through all regex matches of "unset" and couldn't find anything even in strings.

I meant adding a `__getattr__` inside strawberry.arguments, something like:

```python
from warnings import warn

deprecated_names = ["old_function", ...]

def _deprecated_old_function(arg, other):
    ...

def __getattr__(name):
    if name in deprecated_names:
        warn(f"{name} is deprecated", DeprecationWarning)
        return globals()[f"_deprecated_{name}"]
    raise AttributeError(f"module {__name__} has no attribute {name}")

```

strawberry/unset.py
DominiqueGarmier: `Literal[UNSET]` did not work, mypy was complaining about it, I believe it has to do with the fact that `UNSET` has type `Any` (I assume so that `arg: Optional[str] = UNSET` works). So I just left it the way it was.

DominiqueGarmier:
> ```python
> _deprecated_old_function
> ```

hmm, yes i just tried this, but the problem is `arguments.py` still uses `UNSET` so I think of a way this could work right now, unless we would be willing to rename `UNSET` inside the `arguments.py` file

patrick91:
> hmm, yes i just tried this, but the problem is `arguments.py` still uses `UNSET` so I cant think of a way this could work right now, unless we would be willing to rename `UNSET` inside the `arguments.py` file

Yes, we could temporarily rename 😊


> PS: i guess the confusion might be why we would even deprecate it in the first place then (the `UNSET`). I was manly talking about removing it from `__all__`. so that might be acceptable without a dep. warning?

I think we should that later :)

DominiqueGarmier:
> 
> Yes, we could temporarily rename 😊
done

DominiqueGarmier:
should i write tests for the deprecated `is_unset` and the deprecation warnings? coverage is complaining because of it.

strawberry/arguments.py
```diff
@@ -119,8 +124,8 @@ def convert_argument(
     if value is None:
         return None
 
-    if is_unset(value):
-        return value
+    if value is _deprecated_UNSET:
+        return _deprecated_UNSET
```
DominiqueGarmier, line 128: I tried to look at adding tests for this, but from what I could tell this already has tests. So I don't really know what to do

.devcontainer/Dockerfile
```diff
@@ -1,5 +0,0 @@
-ARG VARIANT=3.8
-FROM mcr.microsoft.com/vscode/devcontainers/python:${VARIANT}
-
-RUN pip3 install poetry pre-commit
-RUN poetry config virtualenvs.in-project true
```
patrick91, line 5: can you restore these files? 😊

strawberry/arguments.py
```diff
@@ -24,17 +25,19 @@
...
-
-def is_unset(value: Any) -> bool:
-    return type(value) is _Unset
+DEPRECATED_NAMES: Dict[str, str] = {
+    "UNSET": (
+        "importing `UNSET` from `strawberry.arguments` is deprecated,"
+        "import instead from `strawberry` or from `strawberry.unset`"
+    ),
+    "is_unset": "`is_unset` is deprecated use `value is UNSET` instead",
+}
```
patrick91, line 40: 👍

strawberry/arguments.py
patrick91: codecov is probably wrong :D

patrick91 (commented):
Looks great! Thank you so much :) Can you add a `RELEASE.md` file so we can get this merged and released? 😊

.devcontainer/Dockerfile
DominiqueGarmier: oh I though these were untracked, ofc

DominiqueGarmier: they should be back now, but believe i screwed up the blame now, the history is still intact though.

patrick91: that's fine 😊

RELEASE.md
```diff
@@ -0,0 +1,14 @@
+Release type: refactor
```
```suggestion
Release type: minor
```

DominiqueGarmier: oh that makes sense didn't realize it was for semver

patrick91 (approved):

patrick91:
Thank you!

Issue #1828, opened 2022-04-20
Mypy Plugin Issue and list unpacking


motherofcoconuts:
I am running into an issue with the strawberry plugin when using list unpacking to define a union type.

Command: `dmypy run -- --follow-imports=skip`
Results:
```
Daemon crashed!
Traceback (most recent call last):
  File "mypy/dmypy_server.py", line 229, in serve
  File "mypy/dmypy_server.py", line 272, in run_command
  File "mypy/dmypy_server.py", line 331, in cmd_run
  File "mypy/dmypy_server.py", line 391, in check
  File "mypy/dmypy_server.py", line 528, in fine_grained_increment
  File "mypy/server/update.py", line 245, in update
  File "mypy/server/update.py", line 328, in update_one
  File "mypy/server/update.py", line 393, in update_module
  File "mypy/server/update.py", line 602, in update_module_isolated
  File "mypy/semanal_main.py", line 78, in semantic_analysis_for_scc
  File "mypy/semanal_main.py", line 199, in process_top_levels
  File "mypy/semanal_main.py", line 326, in semantic_analyze_target
  File "mypy/semanal.py", line 414, in refresh_partial
  File "mypy/semanal.py", line 425, in refresh_top_level
  File "mypy/semanal.py", line 5222, in accept
  File "mypy/errors.py", line 902, in report_internal_error
  File "mypy/semanal.py", line 5220, in accept
  File "mypy/nodes.py", line 1155, in accept
  File "mypy/semanal.py", line 2091, in visit_assignment_stmt
  File "mypy/semanal.py", line 2377, in apply_dynamic_class_hook
  File "/Users/ryanhoulihan/Baton/anton-v2/.venv/lib/python3.10/site-packages/strawberry/ext/mypy_plugin.py", line 181, in union_hook
    tuple(_get_type_for_expr(x, ctx.api) for x in types.items)
  File "/Users/ryanhoulihan/Baton/anton-v2/.venv/lib/python3.10/site-packages/strawberry/ext/mypy_plugin.py", line 181, in <genexpr>
    tuple(_get_type_for_expr(x, ctx.api) for x in types.items)
  File "/Users/ryanhoulihan/Baton/anton-v2/.venv/lib/python3.10/site-packages/strawberry/ext/mypy_plugin.py", line 148, in _get_type_for_expr
    raise ValueError(f"Unsupported expression {type(expr)}")
[...]

PR #1831, opened 2022-04-21
Add extremely basic support for unpacking in union

patrick91:
Closes https://github.com/strawberry-graphql/strawberry/issues/1828

This PR add support for the following code in mypy:

```python
UserErrors = (
    Error,
    AnotherError,
)

Response = strawberry.union(types=(User, *UserErrors), name="Response")
```

unfortunately the support is not ideal as response now is typed as:

```
note: Revealed type is "Union[main.User, Any]"
```

which is not ideal, I'll spend a bit more time trying to see if I can find the value of the unpacked variable. If I can find a solution quickly enough I'll merge this since it prevents mypy from crashing at least.

### DIFF
--- a/strawberry/ext/mypy_plugin.py
+++ b/strawberry/ext/mypy_plugin.py
@@ -1,4 +1,5 @@
 from decimal import Decimal
+from itertools import chain
 from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union, cast
 
 from typing_extensions import Final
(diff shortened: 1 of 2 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

patrick91:
I'll close this since we'll change how union works in future

Issue #1830, opened 2022-04-20
mypy errors when using Pydantic integration


olivier-thatch:
Hello,

I'm trying to use the Pydantic integration and getting a few mypy errors.

Simplified version of the code:

```python
import strawberry
from pydantic import BaseModel, EmailStr


class UserBase(BaseModel):
    email: Optional[EmailStr] = None


class UserCreate(PatientBase):
    email: EmailStr


@strawberry.experimental.pydantic.input(model=UserCreate, all_fields=True)
class UserCreateInput:
    pass


@strawberry.mutation
def create_user(self, info: Info, input_data: UserCreateInput) -> PatientType:
    instance = input_data.to_pydantic()
    # ...
```

This causes 2 errors:

1. `@strawberry.experimental.pydantic.input(model=UserCreate, all_fields=True)` causes this error:
    ```
    error: model argument in decorator failed to be parsed
    ```

2. `instance = input_data.to_pydantic()` causes this error:
    ```
    error: "UserCreateInput" has no attribute "to_pydantic"
    ```

Package versions:
* Python 3.10.3
* strawberry-graphql 0.106.3
* pydantic 1.9.0

motherofcoconuts:
Secpnd this issue. Also get issues with `from_pydantic`. Ex: `error: "Type[User]" has no attribute "from_pydantic"`

thejaminator:
hey @motherofcoconuts  and @olivier-thatch are you using our mypy plugin? i fixed that
```
[mypy]
plugins = strawberry.ext.mypy_plugin
```

motherofcoconuts:
hey @thejaminator i am indeed:

thejaminator:
@motherofcoconuts uhoh. can you get your mypy version and strawberry version for me please

motherofcoconuts:
Pulled from my poetry.lock file:

```
[[package]]
name = "strawberry-graphql"
version = "0.97.0"
description = "A library for creating GraphQL APIs"
category = "main"
optional = false
python-versions = ">=3.7,<4.0"
```

```
[[package]]
name = "mypy"
version = "0.942"
description = "Optional static typing for Python"
category = "dev"
optional = false
python-versions = ">=3.6"
```

thejaminator:
@motherofcoconuts as a temporary workaround use the `is_input=True` parameter instead for `strawberry.experimental.pydantic.type` 

```
from typing import Optional
import strawberry
from pydantic import BaseModel, EmailStr


[...]

PR #1832, opened 2022-04-21
fix(mypy): Fix issue with input and interface pydantic decorators

thejaminator:
Resolves https://github.com/strawberry-graphql/strawberry/issues/1830

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

- [ x] I have added tests to cover my changes.
- [ x] I have tested the changes and verified that they work and don't break anything (as well as I can manage).

### DIFF
--- a/strawberry/experimental/pydantic/object_type.py
+++ b/strawberry/experimental/pydantic/object_type.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import builtins
 import dataclasses
 import warnings
-from functools import partial
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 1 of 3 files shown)


strawberry/ext/mypy_plugin.py
```diff
@@ -922,6 +922,7 @@ def _is_strawberry_pydantic_decorator(self, fullname: str) -> bool:
             for strawberry_decorator in {
                 "strawberry.experimental.pydantic.object_type.type",
                 "strawberry.experimental.pydantic.object_type.input",
+                "strawberry.experimental.pydantic.object_type.interface",
```
thejaminator, line 925: i think we forgot to support the interface decorator, so i just added it

strawberry/experimental/pydantic/object_type.py
```diff
@@ -301,6 +300,57 @@ def to_pydantic_default(self) -> PydanticModel:
...
+        model=model,
+        fields=fields,
+        name=name,
+        is_input=True,
+        is_interface=is_interface,
+        description=description,
+        directives=directives,
+        all_fields=all_fields,
+        use_pydantic_alias=use_pydantic_alias,
+    )
```
thejaminator, line 328: technically you could just leave in ```partial(type, is_input=True)``` and put this in the IF TYPE_CHECKING block. Anyways i made it like this so that its less confusing when users introspect. they may not be familar to IF TYPE_CHECKING

patrick91:
> > Makes sense, do you think we can also add this? https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/object_type.py#L152-L153
> > this should make it work slightly better with pyright
> 
> does that work out of the box for the pydantic conversion? not familar with how that works but i think the **init** for the dataclass wouldn't work well because not all the fields are in the class getting decorated.
> 
> e.g.
> 
> ```
> class UserCreate(BaseModel):
>     email: str
> 
> 
> @strawberry.experimental.pydantic.input(model=UserCreate, all_fields=True)
> class UserCreateInput:
>     pass # the fields aren't here
> ```
> 
> if it doesn't and gives false positives i'll settle it in another issue

it only works with fields that are defined, so yeah, let's do it in another issue :)

patrick91 (approved):

Issue #1821, opened 2022-04-18
0.107.0 GET graphiql (fastapi) always depends on context_getter


silas:
I'm not sure if we consider this a regression or not, but the 0.107.0 GET endpoint for fastapi now always depends on `context_getter` (regardless of the `allow_queries_via_get` setting), which means that for users that use a fastapi dependency to authenticate requests, they'll get an authentication error instead of the graphiql interface (assuming they're not authenticated).

Before:

`GET /graphql` would always return the graphiql interface (when enabled) regardless of what `context_getter` did

After:

`GET /graphql` will execute `context_getter` and possibly bail early.

Why:

I override the default graphiql interface to enable the `Request Headers` feature which allows users to enter their credentials into the graphiql interface, which allows them to authenticate after the graphiql page has loaded.

Possible solutions:

 * Work around 1: Just serve GraphiQL from a different path (perfectly fine just doing this, although it's nice to colocate the graphiql and graphql endpoints)
 * Work around 2: Special case the `GET /graphql` endpoint  in authentication dependency (not great, but it works)
 * Change to strawberry: If `allow_queries_via_get` is disabled, don't depend on `context_getter`
 * Change to strawberry: Allow users to provide an alternative `context_getter` for the get vs post handlers (I'd be happy to create a PR for this)

cc @MattExact

PR #1839, opened 2022-04-22
Support transport specific FastAPI `context_getters`

silas:
## Description

This updates the FastAPI integration to support defining transport specific `context_getter`s. This is intended to solve #1821, but would be generally useful to anyone that has transport specific dependencies.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation

* #1821

## Checklist

### DIFF
--- a/docs/integrations/fastapi.md
+++ b/docs/integrations/fastapi.md
@@ -47,9 +47,15 @@ The `GraphQLRouter` accepts the following options:
   interface.
 - `allow_queries_via_get`: optional, defaults to `True`, whether to enable
   queries via `GET` requests
-- `context_getter`: optional FastAPI dependency for providing custom context
-  value.
-- `root_value_getter`: optional FastAPI dependency for providing custom root
+- `context_getter`: optional, default FastAPI dependency for providing custom
+  context value.
+- `http_get_context_getter`: optional, override `context_getter` for queries
+  executed via HTTP GET.
+- `http_post_context_getter`: optional, override `context_getter` for queries
+  executed via HTTP POST.
+- `ws_context_getter`: optional, override `context_getter` for queries executed
+  via WebSockets.
+- `root_value_getter`: optional, FastAPI dependency for providing custom root
   value.
 
 ## context_getter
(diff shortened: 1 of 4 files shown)


botberry:



---

This release updates the FastAPI integration to support transport specific context getters.


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Silas Sewell for the PR 👏

```

docs/integrations/fastapi.md
```diff
@@ -47,9 +47,15 @@ The `GraphQLRouter` accepts the following options:
...
-  value.
-- `root_value_getter`: optional FastAPI dependency for providing custom root
+- `context_getter`: optional, default FastAPI dependency for providing custom
+  context value.
+- `http_get_context_getter`: optional, override `context_getter` for queries
+  executed via HTTP GET.
+- `http_post_context_getter`: optional, override `context_getter` for queries
+  executed via HTTP POST.
+- `ws_context_getter`: optional, override `context_getter` for queries executed
+  via WebSockets.
```
patrick91, line 57: wouldn't be nice to keep context getter and pass the request so one can make the decision of what to return based on that?

silas: So I'll start by saying there is a non-trivial chance I don't understand your suggestion.

What this PR is trying to do is allow you to inject a different set of dependencies based on the transport type.

In the current api `context_getter` is directly attached to the endpoint as a dependency. This means that the signature of that method defines what gets injected at runtime (basically it's a dag). If your `context_getter` doesn't depend on anything (or it's dependencies are common for all transports) than it's trivial to do an `if` statement on the request method and conditionally do something different. But if one of your dependencies does something specific for the transpor, than you can't use FastAPI to inject those dependencies because there is no way to tell the dependency injection system to only inject them for a given transport.

That said if we don't want to expand the API, the FastAPI integration is pretty simply and I'm happy to just fork it into my code.

Or if you have a more elegant way to expose this functionality I'm happy to update the PR.

silas:
I'll just fork the fastapi bits to solve my particular problem.

If you think there is a useful followup PR or different approach let me know and I came come up with something different.

PR #1853, opened 2022-04-26
Compiled resolvers proof of concept

skilkis:
Attempt to improves performance by compiling resolver functions to prevent
conditional look-up of arguments. In the future, this implementation could be
extended to also compile scalar converters, as their type is known ahead of
time. Therefore, dynamic look-up from the scalar registry can be avoided.


patrick91:
🏚️

Issue #159, opened 2019-09-25
Add SDL to strawberry conversion tool


patrick91:
It would be nice to have a command to convert an SDL to strawberry types, something like:

```gql
type User {
  name: String!
}

type Query {
  user: User
}
```

```
$ strawberry import-schema path.grapqhl
```

```
import typing
import strawberry

@strawberry.type
class User:
    name: str

@strawberry.type
class Query:
    user: typing.Optional[User]
```

This should support description and directives if possible :)

Jatinbalodhi96:
@patrick91 I want to work on this. How to get started on this?

patrick91:
@Jatinbalodhi96 I'll assign you to this issue :)

We have a small contribution guide here: https://github.com/strawberry-graphql/strawberry/blob/master/CONTRIBUTING.md#contributing-to-code

We also have a cli setup (which should be refactored I reckon): https://github.com/strawberry-graphql/strawberry/blob/master/strawberry/cli/__init__.py

So maybe you can start from that, what do you think?

Jatinbalodhi96:
thanks, @patrick91 I followed the guide but while running I faced two errors:

1.  `poetry run pytest tests` it gives me `ModuleNotFoundError: No module named 'termios'` error, after searching figure out that this library is for Linux only.
2. While `import strawberry` I got `no Module named 'graphql'` even I already ran poetry install which ran.

Also tried running on linux env git below error
![image](https://user-images.githubusercontent.com/10434201/65631828-f90ef780-dff5-11e9-88a0-890807cf0a4f.png)

patrick91:
@Jatinbalodhi96 uh, never seen that error, what os and python version are using?

Jatinbalodhi96:
I'm using python 3.7 on windows 10, Now I'm able to run
- I'm able run strawberry now but i installed graphql seperately using `pipenv install --pre "graphql-core>=3a"`

The error I sent earlier was because I was installing graphql using `poetry add graphql`, I'm not sure if it works same for all.

patrick91:
@Jatinbalodhi96  oh interest, I never tried strawberry or event GraphQL core on windows, I'll see if I can add tests for that, I don't have a windows machine at the moment, so it might take me a while to test this :)

Jatinbalodhi96:
@patrick91 I've never used poetry before but is this correct way to build the dev version.
[...]

PR #1868, opened 2022-05-02
Add SDL to strawberry command from awesome work of @srbiotik

Ambro17:
## Description

Given an SDL file, when one runs 
`strawberry import-schema <SDL_FILE_PATH> > schema.py`
One gets autogenerated code equivalent to the sdl dumped into `schema.py` file

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

#159 

## Checklist


botberry:



---

Adds import-schema command to cli import-schema
The command transpiles SDL from a path to strawberry types which are printed to the stdout to be used at will.
Call it by: `strawberry import-schema <PATH>`


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Nahuel Ambrosini for the PR 👏

```

strawberry/cli/commands/schema_importer/sdl_transpiler.py
```diff
@@ -0,0 +1,210 @@
...
+    """
+    Format and return a whole attribute string
+    consists of attribute name in snake case and field type
+    """
+    field_name = get_field_name(field.name.value)
+    field_type = get_field_type(field)
+    is_custom_type = not any(
+        builtin_type in field_type
+        for builtin_type in ("str", "int", "float", "ID", "bool")
+    )
```
BryceBeagle, line 121: Am I correct in understanding that `field_type` is a string? Any custom type with any of these as a _substring_ (e.g. if I had a type named "winter" or something), would make `is_custom_type == False` no?

Ambro17: You're correct. Given a string like `List[int]` it tries to detect if it uses a custom type with this logic. But it does have a problem you just found. Can you think of a better solution?

BryceBeagle: Sorry for the delayed response here. Could you do something like `field_type == "int" or "[int]" in field_type` (obviously expanded for all the types here)?

PR #1871, opened 2022-05-02
Add support for passing names to field in schema directives

patrick91:
This PR add support for passing names to field of strawberry directives, it also tried the `__strawberry_xxx__` approach, this will be probably refactored, but I'll probably merge this soon as it blocks other PRs 😊

This works now:

```python
@strawberry.schema_directive(locations=[Location.FIELD_DEFINITION])
class Sensitive:
    reason: str = directive_field(name="as")
    real_age_2: str = directive_field(name="real_age")

@strawberry.type
class Query:
    first_name: str = strawberry.field(
        directives=[Sensitive(reason="GDPR", real_age_2="42")]
    )

expected_type = """
type Query {
    firstName: String! @sensitive(as: "GDPR", real_age: "42")
}
```

This is done by using `strawberry.directive_field` that uses StrawberryField, we might introduce a new type, but for now this will work 😊

PR #1876, opened 2022-05-03
tests: harden test_using_different_names_for_directive_field

blueyed:
I think this helps to clarify what is being used where.

ref: https://github.com/strawberry-graphql/strawberry/pull/1871


patrick91 (approved):
Good catch! Thank you!

PR #1877, opened 2022-05-03
tests: rename test plugins to avoid PytestCollectionWarning

blueyed:
> …/tests/cli/test_codegen.py:11:
>   PytestCollectionWarning: cannot collect test class 'TestConsolePlugin'
>   because it has a __init__ constructor (from: tests/cli/test_codegen.py)

### DIFF
--- a/tests/cli/test_codegen.py
+++ b/tests/cli/test_codegen.py
@@ -8,14 +8,14 @@ from strawberry.codegen import CodegenFile, CodegenResult, QueryCodegenPlugin
 from strawberry.codegen.types import GraphQLOperation, GraphQLType
 
 
-class TestConsolePlugin(ConsolePlugin):
+class ConsoleTestPlugin(ConsolePlugin):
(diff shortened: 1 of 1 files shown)


tests/cli/test_codegen.py
```diff
@@ -8,14 +8,14 @@
...
-class TestConsolePlugin(ConsolePlugin):
+class ConsoleTestPlugin(ConsolePlugin):
     def on_end(self, result: CodegenResult):
         result.files[0].path = "renamed.py"
 
         return super().on_end(result)
 
 
-class TestPlugin(QueryCodegenPlugin):
+class QueryCodegenTestPlugin(QueryCodegenPlugin):
```
blueyed, line 18: Not really necessary here, but I think it is clearer, more consistent/explicit.

patrick91 (approved):
I've been meaning to do that for a while!

Thank you for reducing the number of warnings 😊

Issue #1899, opened 2022-05-10
Fields with resolvers raise Type Errors when using Generic Subclasses


invokermain:
## Describe the Bug

Following on from: https://github.com/strawberry-graphql/strawberry/pull/1891

The case where a field's resolver has a subclassed generic return type still raises a Type Error. The `isinstance` check is too weak.

See this gist for examples: https://gist.github.com/invokermain/31c783c76dd26e1da862cbe5218d2500
 
## System Information

 - Strawberry version (if applicable): 0.111.2

## Additional Context

There is a wider discussion about resolver return type hint behaviour here: https://github.com/strawberry-graphql/strawberry/issues/1896

invokermain:
Behavior table:


Let `T = TypeVar("T")` ... the extended table might like:


<html>
<body>


field type | resolver return type | current behaviour | expected behaviour | current type | expected type
-- | -- | -- | -- | -- | --
int | T | no error | no error [1] | int | int
int | GenericClass[T] | runtime error | no errors, no warnings | int | int
int | Optional[T] | runtime error | no errors, no warnings | int | int

</body>
</html>

[1] recently fixed here: https://github.com/strawberry-graphql/strawberry/pull/1891

patrick91:
Fixed in #1900

Issue #1896, opened 2022-05-09
Improve how we determine types for fields that have resolvers


patrick91:
This was prompted by this issue: https://github.com/strawberry-graphql/strawberry/issues/1887

Currently there are some cases where the behaviour we have might not be what one expects. And we could also improve how we handled cases where field type and resolver type don't match.

Based on this type:

```python
def resolver() -> another_type:
    ...

@strawberry.type
class Example:
     field: a_type
```

Here's a table of the current behaviours

| field type | resolver return type | current behaviour      | expected behaviour                   | current type | expected type |
|------------|----------------------|------------------------|--------------------------------------|--------------|---------------|
| int        | TypeVar("T")         | no error               | no error [1]                         | int          | int           |
| int        | int                  | no errors, no warnings | no errors, no warnings               | int          | int           |
| int        | missing              | no errors, no warning  | no error, maybe a (opt-out) warning? | int          | int           |
| int        | str                  | no errors, no warnings | error                                | str          | int           |

[1] recently fixed here: https://github.com/strawberry-graphql/strawberry/pull/1891

PR #1900, opened 2022-05-10
fix: TypeError when using generic types in resolver in more complex c…

invokermain:
…ases

## Description

Fixes a field's resolvers from raising a type error when using a subclassed generic return type hint by widening the check for generics in the `StrawberryField.type` property, [original code here](https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/field.py#L200).

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* https://github.com/strawberry-graphql/strawberry/issues/1899

## Checklist


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Tim OSullivan for the PR 👏

```

invokermain:
Holding off on adding a RELEASE.md file as would like some discussion around this first:

1. Is it a good idea to use `StrawberryType.is_generic` for this?
2. Is `tests/schema/test_resolvers.py` the best location for the tests? (slightly overwhelmed by the number of unit tests!)
3. Follow on for above, is there a more granular/small unit that could be used to tests these, my understanding of the internals of this library is too poor to know the answer to this. Maybe these should be unit tests on `GraphQLCoreConverter` instead?
4. Are there more edge cases we should test against? Maybe `Union[T, SomeThingElse]`, for example if someone was returning error responses as part of their schema or the generic type.
5. The big question (probably should have been asked before the previous pull request), if we are evading the `MissingTypesForGenericError` so much, I wonder what else uses it to make it useful? If the answer is not much, then I wonder what happens if we don't raise that error.

Ultimately I am guessing these PRs are a short term solution to allow you to type hint resolver factories correctly, and in the long term you would like to maybe check that say `Container[SolidType]` & `Container[GenericType]` are compatible. Although that sounds complicated! But that would be part of the warning/incorrect type checking logic hinted at in: https://github.com/strawberry-graphql/strawberry/issues/1896

patrick91:
> Holding off on adding a RELEASE.md file as would like some discussion around this first:
> 
> 1. Is it a good idea to use `StrawberryType.is_generic` for this?

Based on our current code structure, I'd say yes. I was a bit worried that doing that would break some code, but we do have quite a bit of tests for generics so we should be fine :)

> 2. Is `tests/schema/test_resolvers.py` the best location for the tests? (slightly overwhelmed by the number of unit tests!)

I think so, we don't have a great structure for tests at the moment, so I wouldn't worry too much 😊

> 3. Follow on for above, is there a more granular/small unit that could be used to tests these, my understanding of the internals of this library is too poor to know the answer to this. Maybe these should be unit tests on `GraphQLCoreConverter` instead?

The test you wrote is fine for me, writing a more granular test might be good, but we still want to make sure the resolvers works. Maybe we can add a test for the schema, to check that the types are what we expect :)

> 4. Are there more edge cases we should test against? Maybe `Union[T, SomeThingElse]`, for example if someone was returning error responses as part of their schema or the generic type.

Union[T, SomethingElse], Optional[T] and List[T], not sure if we should test more cases

> 5. The big question (probably should have been asked before the previous pull request), if we are evading the `MissingTypesForGenericError` so much, I wonder what else uses it to make it useful? If the answer is not much, then I wonder what happens if we don't raise that error.

I'm going to review all our expected errors at some point, so I wouldn't change this too much for now 😊

> Ultimately I am guessing these PRs are a short term solution to allow you to type hint resolver factories correctly, and in the long term you would like to maybe check that say `Container[SolidType]` & `Container[GenericType]` are compatible. Although that sounds complicated! But that would be part of the warning/incorrect type checking logic hinted at in: #1896

I'd worry about this in future, we can definitely add some basic checks which might cover most of the common errors 😊 we'll see 😊

invokermain:
I have added some extra unit tests covering the Union[T, SomethingElse], Optional[T] and List[T] cases.

> Maybe we can add a test for the schema, to check that the types are what we expect :)
With regards to this happy to add them with more pointers. Currently I am not 100% sure if its required/what is required as not sure if any of my code changes affect the schema generation?

I was under the impression that the Schema is generated from the top level type hint (e.g. [here ](https://gist.github.com/invokermain/31c783c76dd26e1da862cbe5218d2500#file-strawb_generic_example-py-L51) it would be defined by the `list[MyItemType]` type hint no?).

invokermain:
playground for example in RELEASE.md: https://play.strawberry.rocks/?gist=715c09744893b8662b8468f29579ee90

strawberry/field.py
```diff
@@ -47,6 +47,18 @@
...
+def _is_generic(resolver_type: Union[StrawberryType, type]) -> bool:
+    """Returns True if `resolver_type` is generic else False"""
+    if isinstance(resolver_type, StrawberryType):
+        return resolver_type.is_generic
+
+    # solves the Generic subclass case
+    if hasattr(resolver_type, "_type_definition"):
+        return resolver_type._type_definition.is_generic  # type: ignore
+
+    return False
```
patrick91, line 59: adding a note for the future (maybe), we could move this method on the resolver itself potentially

patrick91 (approved):
Apologies for the very late review!

I've merged main into this PR and reviewed the code, looks good! I've left a minor note for the future, but I'll merge this now when tests are green :)

invokermain:
> Apologies for the very late review!
> 
> I've merged main into this PR and reviewed the code, looks good! I've left a minor note for the future, but I'll merge this now when tests are green :)

No worries! It was a very minor issue, but just means I can uncomment some generic type hints now, thanks 👍

PR #1904, opened 2022-05-11
New Lazy to replace LazyType and fix typing issues

### LINKED ISSUE #388: Improve MyPy support for LazyTypes
We've introduced LazyType["TypeName", "module.path"] as a way to use types that depend on each other but are defined in different modules, ie:

```
# a.py

if TYPE_CHECKING:
    from module.b import B

@strawberry.type
class A:
    b: LazyType["B", "module.b"]
```

```
# b.py

if TYPE_CHECKING:
    from module.a import A

@strawberry.type
class B:
    a: LazyType["A", "module.a"]
```

There's an example of this here: https://github.com/strawberry-graphql/strawberry/tree/master/tests/test_cyclic

As you can see from the tests, we have to import the module where the type was defined for mypy to not complain. I wonder if we can improve our mypy plugin (https://github.com/strawberry-graphql/strawberry/blob/master/strawberry/ext/mypy_plugin.py) to tell mypy to ignore the secondo argument or at least remove the error.

### LINKED ISSUE #1177: LazyType on resolver type-hint collides with return type
## Description

While using LazyTypes on function-based resolvers, the IDE code inspector goes crazy with the return type and type-hint mismatch.

## Context 

I've been trying to define an extensible and modular project file structure. After some iterations, I came up with the following:
* schema.py
* types.py
* resolvers.py

This structure became pretty comfortable as query and type contain very compact method-less classes that define my schemas. And most business-related logic is contained in resolvers and other util modules.

But as I've progressed on my project, I noted that this project structure has some circular dependencies that difficult the develop. Expanding this idea, the resolvers functions need a reference to the types, as they will return an Object Type to populate the resolved field. And types will need a reference to the resolvers, as a way to indicate the schema how to resolve each field that has inputs and some complex resolving logic.

## Example code
The following is the project structure that avoids any circular import issues.

> schema.py
```python
@strawberry.type
class Query: 
    library: Library = strawberry.field(
        Resolvers.library
    )

```

> types.py
```python
@strawberry.type
class Book:
    title: str
    language: str

@strawberry.type
class Library:
    """For this example we assume a Library can only has Books of the library's language.""""
    name: str
    language: str
    books: Book = strawberry.field(
        Resolvers.books
    )
```

> resolvers.py
```python
if TYPE_CHECKING:
    from types import Book, Library

class Resolvers:
    @staticmethod
    def library(self, name: str, language: str) -> LazyType["Library", "types"]:
        from types import Library
        # the next line shows: Expected type 'LazyType[Library, Any]', got 'Library' instead 
        return Library(name=name, language=language)

    @staticmethod
    def books(self, title: str) -> List[LazyType["Book", "types"]]:
        from types import Book
         # the next line shows: Expected type 'List[LazyType[Book, Any]]', got 'List[Book]' instead 
        return [Book(title=title, language=self.language)]
```

## Questions

Generally speaking, I think the issues I'm having might be a code-smell pointing that my project's structure isn't healthy. 
So my first question is if there's any suggestion on an extensible and clean project structure with some separation of concerns?

My second question is if there are any updates on how are the good practices on using LazyTypes that help me avoid the warnings that are popping up in my IDE with this approach?

As I'm using PyCharm I think #796 might be slightly related to this issue.

bellini666:
An attempt to solve typing issues with `LazyType`.

Currently there's a major issue with `LazyType` that requires you to always have to `# type:ignore` because of the module path. It also makes the final attribute expects a `LazyType` instance instead of the real type that we want to use.

This exposes a single `Lazy` that should be used in place of `LazyType`. For type checkers (i.e. inside a `if TYPE_CHECKING`), it is aliased as an `Annotated` because it will make it behave exactly the way we want. At runtime it is aliased to the `LazyType` itself.

At first I actually defined `if TYPE_CHECKING: LazyType = Annotated` and `else: class LazyType(Generic[...])):`, but it messes with the typing of schema parsers and other places of the lib. Then I thought that exposing a simple `Lazy` would be better, but I'm not sure what you all think (cc @patrick91 @BryceBeagle )

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* Fix #388
* Fix #1177

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,6 @@
+Release type: minor
+
+New `Lazy` that will replace `LazyType` in the future. It is aliased to
+`LazyType` at runtime, but static type checkers (e.g. pyright) see them as
+`Annotated`. That means that using it will prevent typing issues and also
+make those same static type checkers to properly type them.
--- a/strawberry/__init__.py
+++ b/strawberry/__init__.py
@@ -5,7 +5,7 @@ from .custom_scalar import scalar
 from .directive import directive, directive_field
 from .enum import enum, enum_value
 from .field import field
-from .lazy_type import LazyType
+from .lazy_type import Lazy, LazyType
 from .mutation import mutation, subscription
 from .object_type import input, interface, type
 from .permission import BasePermission
@@ -22,6 +22,7 @@ __all__ = [
     "experimental",
     "ID",
     "UNSET",
+    "Lazy",
     "LazyType",
     "Private",
     "Schema",
--- a/strawberry/ext/mypy_plugin.py
+++ b/strawberry/ext/mypy_plugin.py
@@ -880,7 +880,10 @@ class StrawberryPlugin(Plugin):
         )
 
     def _is_strawberry_lazy_type(self, fullname: str) -> bool:
-        return fullname == "strawberry.lazy_type.LazyType"
+        return fullname in [
+            "strawberry.lazy_type.LazyType",
+            "strawberry.lazy_type.Lazy",
+        ]
 
     def _is_strawberry_decorator(self, fullname: str) -> bool:
         if any(
--- a/strawberry/lazy_type.py
+++ b/strawberry/lazy_type.py
@@ -41,7 +41,7 @@ class LazyType(Generic[TypeName, Module]):
         return None
 
 
-if TYPE_CHECKING:  # pragma: nocover
+if TYPE_CHECKING:  # pragma: no cover
     # Static types like pyright expects a type to be passed to generic classes, which
     # means that if we pass a string it will say the type is unknown, and if we import
     # the module they will say a module can't be used in place of a type.
--- a/strawberry/lazy_type.py
+++ b/strawberry/lazy_type.py
@@ -1,7 +1,9 @@
 import importlib
 import inspect
 from dataclasses import dataclass
-from typing import Generic, Optional, Type, TypeVar
+from typing import TYPE_CHECKING, Generic, Optional, Type, TypeVar
+
+from typing_extensions import Annotated
 
 
 TypeName = TypeVar("TypeName")
@@ -37,3 +39,14 @@ class LazyType(Generic[TypeName, Module]):
 
     def __call__(self):  # pragma: no cover
         return None
+
+
+if TYPE_CHECKING:  # pragma: nocover
+    # Static types like pyright expects a type to be passed to generic classes, which
+    # means that if we pass a string it will say the type is unknown, and if we import
+    # the module they will say a module can't be used in place of a type.
+    # This will tricky it into thinking that LazyType is Annotated, not only avoiding
+    # the issue but also statically typing it correctly.
+    Lazy = Annotated
+else:
+    Lazy = LazyType
--- a/tests/mypy/test_lazy_type.yml
+++ b/tests/mypy/test_lazy_type.yml
@@ -1,0 +1,21 @@
+- case: test_unset
+  main: |
+    from typing import TYPE_CHECKING
+    import strawberry
+    from strawberry.lazy_type import Lazy
+
+
+    @strawberry.type
+    class SomeType:
+        some_var: Lazy[str, "strawberry.scalars"]
+
+
+    obj = SomeType(
+        some_var="foobar",
+    )
+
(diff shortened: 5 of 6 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

patrick91 (commented):
I think this makes sense! Can we have a test for mypy too? :)

bellini666:
Hi @patrick91 ,

Hrm, so strange, I wrote the same code as for pyright and got this:

```
main:11: error: Module "strawberry.scalars" is not valid as a type  [valid-type] (diff)
main:11: error: Variable "strawberry.lazy_type.Lazy" is not valid as a type  [valid-type] (diff)
main:11: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases (diff)
main:18: note: Revealed type is "Lazy?[Any, strawberry.scalars?]" (diff)
main:19: note: Revealed type is "Lazy?[Any, strawberry.scalars?]" (diff)
```

It seems that mypy isn't understanding tha `Lazy` is an alias to `Annotated`, and thus it should work just like if we used `Annotated` directly. Probably a bug in it?

Either way, I noticed that even without this code it would also produce an error, it was just working fine before because of the plugin. I added `Lazy` to be considered for the plugin as well and it worked. Updating the PR...

RELEASE.md
```diff
@@ -0,0 +1,6 @@
+Release type: minor
+
+New `Lazy` that will replace `LazyType` in the future. It is aliased to
```
patrick91, line 3: sorry I just released we are renaming LazyType to Lazy, is there any reason for that? 

Could we keep the same name?

bellini666: You want to keep `LazyType` or just `Lazy`?

We can keep just one! =P. Since I created the PR as a proof of concepot and didn't want to make any radical changes if you guys didn't like it, I decided to create a new `Lazy` for that. But it is up to you :)

patrick91: we should keep one, I think the name `LazyType` might be misleading since it is similar to `typings.Type`, so maybe `Lazy` is better /cc @BryceBeagle what do you think?

botberry:
# Pre-release
:wave:

Releasing commit [34f9f11b964a1e30cc741b3b207c7e02431ed5d7] to PyPi as pre-release! :package:

botberry:
# Pre-release
:wave:

Pre-release **0.124.0.dev.1659635965** [34f9f11b964a1e30cc741b3b207c7e02431ed5d7] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.124.0.dev.1659635965
```

RELEASE.md
BryceBeagle: Hey sorry for the delayed response -- just seeing this now! I think the potential for confusion with `Type` could occur, but I'm not sure it would be that common. Another option could be `LazyVar` (similar to `ClassVar` and `TypeVar`) but I don't really have strong preferences

bellini666: For me it is up to you @patrick91 and @BryceBeagle to call it `Lazy`, `LazyVar`, or just keep the current `LazyType` and define it in the `else` part of a `if TYPE_CHECKING`.

patrick91: Let's go with `LazyVar`! 😊

bellini666:
@patrick91 @BryceBeagle just a FYI, when trying to update this PR to `LazyVar` as we discussed I hit a problem where newer versions of pyright are not working properly with the aliased `Annotation`.

The end result is correct, but they give an error at the module part because it tries to parse it as a type. Unfortunately it is something that pyright will not fix: https://github.com/microsoft/pyright/issues/3853

We can go 3 ways here:

1) Abandon the idea here. I don't think we should because currently this breaks the typing for things other than mypy that have a plugin, meaning that everyone trying to create a type that uses `LazyType` needs to `# type:ignore` the variable they are passing to that var.

2) Go ahead with the current solution. It works, the user just need to `# type:ignore` the `LazyType`/`LazyVar` definition, but when creating instantiating the class he will still be able to pass the proper typed variable to it.

3) Another solution that makes pyright happy and also allows us to do the lazy importing thing? I'm trying to think in a solution but so far nothing has come to my mind.

Issue #1983, opened 2022-07-02
Regression using mutable default values in arguments.


coady:
## Describe the Bug
Prior to v0.115
```python
@strawberry.input
class Options:
    flag: str = ''


@strawberry.type
class Query:
    @strawberry.field
    def field(self, x: list[str] = [], y: Options = {}) -> str:
        return f'{x} {y}'
```
would correctly resolve to
```graphql
type Query {
  field(x: [String!]! = [], y: Options! = {}): String!
}
```
As of v0.115 it raises
```python
  File "lib/python3.10/site-packages/strawberry/types/fields/resolver.py", line 87, in find
    resolver._resolved_annotations[parameter] = resolved_annotation
  File "/lib/python3.10/inspect.py", line 2740, in __hash__
    return hash((self.name, self.kind, self.annotation, self.default))
TypeError: unhashable type: 'list'

```

For lists, there is a workaround to use a tuple instead, but it's not ideal because GraphQL type coercion will correctly supply a list. For objects, there's no clean workaround; one would have to  use the equivalent of a `frozendict`.

## System Information
 - Strawberry version 0.115

skilkis:
Thanks for the bug report @coady, I'm looking for a fix now 😊

PR #1984, opened 2022-07-02
Mutable default values are frozen to support lists and objects.

coady:
## Description
Restores the ability to use mutable default values, freezing them to support lists and objects.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #1983 

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,8 @@
+Release type: patch
+
+Mutable default values are frozen to support lists and objects.
+```python
+@strawberry.field
+def search(self, arg: List[int] = [], input: MyInput = {}) -> int:
+    ...
+```
--- a/strawberry/types/fields/resolver.py
+++ b/strawberry/types/fields/resolver.py
@@ -1,7 +1,9 @@
 from __future__ import annotations as _
 
 import builtins
+import functools
 import inspect
+import operator
 import sys
 import warnings
 from inspect import isasyncgenfunction, iscoroutinefunction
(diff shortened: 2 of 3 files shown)


botberry:



---

Mutable default values are frozen to support lists and objects.
```python
@strawberry.field
def search(self, arg: List[int] = [], input: MyInput = {}) -> int:
    ...
```


---




Here's the tweet text: 
```

```

RELEASE.md
```diff
@@ -0,0 +1,8 @@
+Release type: patch
+
+Mutable default values are frozen to support lists and objects.
```
patrick91, line 3: shall we mention that this is only for arguments?

RELEASE.md
```diff
@@ -0,0 +1,8 @@
...
+```python
+@strawberry.field
+def search(self, arg: List[int] = [], input: MyInput = {}) -> int:
```
patrick91, line 6: `input: MyInput = {}`

Should this be `input: MyInput = MyInput()`?

skilkis:
@coady perhaps we can join forces on this PR? In haste to fix the issue, I didn't realize you had already proposed this fix until after creating #1985 🙈

I've glanced through your implementation and I'm worried that the `freeze_value` function will not return a hashable value in all cases, for example for a `set` object. Since we can't predict / control what default value the user will decide to use, the most robust way I could find to address this issue is to neglect the default value when computing the hash of a `Parameter` object. 

Please let me know how you feel about this solution, and how you'd like to proceed. You definitely deserve credit for this fix as you were the one to find out about the regression in the first place 😊

coady:
Closing in favor of #1985.

RELEASE.md
coady: That never worked historically, but could be a new feature.

I think literals should work because of the resemblance between GraphQL objects, JSON objects, and Python dicts.

Issue #1964, opened 2022-06-23
cannot set graphql_name with strawberry.field on pydantic.type


benzolium:
cannot set `graphql_name` with `strawberry.field(name='someName')` on `pydantic.type`

## Describe the Bug

It's better explained as a test for `tests/experimental/pydantic/test_basic.py`

```
def test_alias_fields():
    class User(pydantic.BaseModel):
        age: int

    @strawberry.experimental.pydantic.type(User)
    class UserType:
        age: strawberry.auto = strawberry.field(name="ageAlias")

    definition: TypeDefinition = UserType._type_definition
    assert definition.name == "UserType"

    field1 = definition.fields[0]

    assert field1.python_name == "age"
    assert field1.graphql_name == "ageAlias"  # <-- field1.graphql_name is None
    assert field1.type is int
```

benzolium:
I understand that I can use `pydantic.Field(alias="ageAlias")` on model, but the reason I use `pydantic` with strawberry is because I want to separate models and graphql types.

PR #1986, opened 2022-07-03
pydantic: use strawberry alias if any

benzolium:
Closes https://github.com/strawberry-graphql/strawberry/issues/1964:

> cannot set `graphql_name` with `strawberry.field(name='someName')` on `pydantic.type`

Feel free to edit my PR if such behaviour was not intended. 🙂

### DIFF
--- a/strawberry/experimental/pydantic/object_type.py
+++ b/strawberry/experimental/pydantic/object_type.py
@@ -71,11 +71,14 @@ def _build_dataclass_creation_fields(
     else:
         # otherwise we build an appropriate strawberry field that resolves it
         existing_field = existing_fields.get(field.name)
+        graphql_name = None
+        if existing_field and existing_field.graphql_name:
+            graphql_name = existing_field.graphql_name
+        elif field.has_alias:
+            graphql_name = field.alias
         strawberry_field = StrawberryField(
             python_name=field.name,
-            graphql_name=field.alias
-            if field.has_alias and use_pydantic_alias
-            else None,
+            graphql_name=graphql_name,
             # always unset because we use default_factory instead
             default=UNSET,
             default_factory=get_default_factory_for_field(field),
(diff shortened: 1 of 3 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Alex for the PR 👏

```

patrick91:
@Matt343 @thejaminator would you able to review this PR? 😊

strawberry/experimental/pydantic/object_type.py
```diff
@@ -71,11 +71,14 @@ def _build_dataclass_creation_fields(
     else:
         # otherwise we build an appropriate strawberry field that resolves it
         existing_field = existing_fields.get(field.name)
+        graphql_name = None
+        if existing_field and existing_field.graphql_name:
+            graphql_name = existing_field.graphql_name
+        elif field.has_alias and use_pydantic_alias:
```
thejaminator, line 77: what happens if there `existing_field.graphql_name` and the user passes use_pydantic_alias?

 i think we should error out?  or maybe we can leave it such that existing_field.graphql_name takes precedence (which is how it is now). we'll just need to enforce that behaviour with a test @Matt343 @patrick91

patrick91: maybe erroring out might be best, when would you use graphql_name and use_pydantic_alias together?

thejaminator: @patrick91 
```
    age: int
    state: str = Field(alias="_state")
    country: str = Field(alias="_country")

@strawberry.experimental.pydantic.type(User, use_pydantic_alias=True)
class UserType:
    age: strawberry.auto = strawberry.field(name="ageAlias") # graphql is ageAlias
    state: strawberry.auto = strawberry.field(name="state") # graphql is state
    country: strawberry.auto # graphql is _country
```

maybe because use_pydantic_alias is a blanket settings. and the person happens to have a model where there are two pydantic fields with a pydantic alias. but he only wants one field to use the pydantic alias (poor guy)

thejaminator: yep! we should write a test to enforce this behaviour too. i can take it in another MR, or @benzolium if he feels like it :D

patrick91 (approved):
Thank you! <3

botberry:

PR #1994, opened 2022-07-05
Add temporary fix for strawberry.auto and future annotations

### LINKED ISSUE #1819: Postponed annotation evaluation causes `Annotated` to break
When using postponed annotation evaluation, annotating resolver arguments no longer works:

```python

import random
from typing import Annotated

import strawberry


@strawberry.type
class Query:
    @strawberry.field
    def dice_roll(
        self,
        sides: Annotated[
            int,
            strawberry.argument(description="Number of sides the die should have."),
        ] = 6,
    ) -> int:
        return random.randint(1, sides)


strawberry.Schema(query=Query)
```

The example above raises this TypeError:

```
TypeError: Query fields cannot be resolved. Unexpected type 'typing.Annotated[int, <strawberry.arguments.StrawberryArgumentAnnotation object at 0x7fd12e130d00>]'
```

When the first line (`from __future__ import annotations`) is left out, everything works as intended. This will probably also break once Python 3.11 lands, since the behavior will become mandatory then. #1586 refers to a somewhat related issue.

patrick91:
This PR is a temporary fix for auto not working when using `from __future__ import annotations`.

We should find a better way to fix this, but it will take a bit of time to refactor some of the code. /cc @thejaminator maybe we can discuss how to best make this work :)

It might fix #1819, @eddysen I'll make a pre-release, could you check if that works for you? :)

### DIFF
--- a/strawberry/auto.py
+++ b/strawberry/auto.py
@@ -57,7 +57,7 @@ class StrawberryAutoMeta(type):
             if args[0] is Any:
                 return any(isinstance(arg, StrawberryAuto) for arg in args[1:])
 
-        return False
+        return instance == "strawberry.auto"
 
 
 class StrawberryAuto(metaclass=StrawberryAutoMeta):
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,4 @@
+Release type: patch
+
+This release adds an initial fix to make `strawberry.auto` work when using
+`from __future__ import annotations`.
--- a/tests/experimental/pydantic/schema/test_forward_reference.py
+++ b/tests/experimental/pydantic/schema/test_forward_reference.py
@@ -1,0 +1,50 @@
+from __future__ import annotations
+
+import textwrap
+from typing import Optional
+
+import pydantic
+
+import strawberry
+
+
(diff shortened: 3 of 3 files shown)


botberry:
# Pre-release
:wave:

Releasing commit [bff7b709b3727daa696663b9fb0b86c362673a95] to PyPi as pre-release! :package:

strawberry/auto.py
```diff
@@ -57,7 +57,7 @@ def __instancecheck__(
             if args[0] is Any:
                 return any(isinstance(arg, StrawberryAuto) for arg in args[1:])
 
-        return False
+        return instance == "strawberry.auto"
```
thejaminator, line 60: :D

thejaminator: yeah it complicated because we only resolve the annotations at a later point of time......

patrick91: I wonder if we can leverage what we do for strawberry.type, but probably not since the type resolution is deferred quite a bit 🤔

thejaminator: @patrick91 think we should merge this in the meantime XD

PR #2029, opened 2022-07-19
fix: edge case for nested generics in a union type

### LINKED ISSUE #2028: StrawberryUnion's type resolver does not prioritise types in the Union
## Describe the Bug

StrawberryUnion's type resolver iterates over all known types and returns the first type which successfully implements the returned object from the resolver.

Due to limitations of `TypeDefinition.is_implemented_by` when checking generic fields, this check might succeed prematurely for an object with the same top level structure.

If the wrong type is selected a `UnallowedReturnTypeForUnion` error is raised shortly afterwards.

Potentially it might be possible to improve `TypeDefinition.is_implemented_by` check for generic fields but that is beyond the scope of this issue.

This issue suggests a simple fix of first checking the types given in the union before falling back to checking all types. This might also be a small performance improvement if it is true that in most cases the correct type is listed in the Union.

Minimal reproduceable example: https://play.strawberry.rocks/?gist=f7d88898d127e65b12140fdd763f9ef2

## Existing Code
The below check is logically incorrect as `possible_type.is_implemented_by(root)` might be True for a similar object that is not the correct type.

```python
# Iterate over all of our known types and find the first concrete type that
# implements the type
for possible_concrete_type in type_map.values():
    possible_type = possible_concrete_type.definition
    if not isinstance(possible_type, TypeDefinition):
        continue
    if possible_type.is_implemented_by(root):
        return_type = possible_concrete_type.implementation
        break
else:
    return_type = None
```

## Example Fix
Here we check the types explicitly listed in the Union first, and then fallback to checking all types.

```python
# Iterate over the given types in the union and check if any provide a
# concrete type that implements the type.
for gql_object_type in type_.types:
    type_name = gql_object_type.name
    possible_concrete_type = type_map.get(type_name)
    if possible_concrete_type:
        possible_type = possible_concrete_type.definition
        if not isinstance(possible_type, TypeDefinition):
            continue
        if possible_type.is_implemented_by(root):
            return_type = possible_concrete_type.implementation
            break
else:
    return_type = None

# Otherwise, iterate over all of our known types and find the first concrete
# type that implements the type.
if return_type is None:
    for possible_concrete_type in type_map.values():
        possible_type = possible_concrete_type.definition
        if not isinstance(possible_type, TypeDefinition):
            continue
        if possible_type.is_implemented_by(root):
            return_type = possible_concrete_type.implementation
            break
```

## System Information

 - Operating system: N/A
 - Strawberry version (if applicable): 0.119.1

invokermain:
## Description

In `StrawberryUnion.get_type_resolver` prioritise testing against types that are explicitly in the Union type before checking all other known types. This fixes an edge case where StrawberryUnion would pick out an incorrect type that weakly matches (due to nested generics) the resolved object before testing the actual type in the union.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* Resolves #2028.

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,9 +1,8 @@
 Release type: patch
 
 Fixed edge case where `Union` types raised an `UnallowedReturnTypeForUnion` error when
-returning the correct type from the resolver due to only being able to partially match
-nested generic types. This is fixed by prioritising the types explicitly listed in the
-Union.
+returning the correct type from the resolver. This also improves performance of
+`StrawberryUnion.get_type_resolver()` from `O(n)` to `O(1)` in the majority of cases.
 
 For example the below ([link to playground](https://play.strawberry.rocks/?gist=f7d88898d127e65b12140fdd763f9ef2))
 would previously raise the error when querying `two` as `StrawberryUnion` would incorrectly
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,42 @@
+Release type: patch
+
+Fixed edge case where `Union` types raised an `UnallowedReturnTypeForUnion` error when
+returning the correct type from the resolver due to only being able to partially match
+nested generic types. This is fixed by prioritising the types explicitly listed in the
+Union.
+
+For example the below ([link to playground](https://play.strawberry.rocks/?gist=f7d88898d127e65b12140fdd763f9ef2))
+would previously raise the error when querying `two` as `StrawberryUnion` would incorrectly
+determine that the resolver returns `Container[TypeOne]`.
+
+```python
+import strawberry
+from typing import TypeVar, Generic, Union, List, Type
+
+T = TypeVar("T")
+
+@strawberry.type
+class Container(Generic[T]):
+    items: List[T]
+
+@strawberry.type
+class TypeOne:
+    attr: str
+
+@strawberry.type
+class TypeTwo:
+    attr: str
+
+def resolver_one():
+    return Container(items=[TypeOne("one")])
+    
+def resolver_two():
+    return Container(items=[TypeTwo("two")])
+
+@strawberry.type
+class Query:
+    one: Union[Container[TypeOne], TypeOne] = strawberry.field(resolver_one)
+    two: Union[Container[TypeTwo], TypeTwo] = strawberry.field(resolver_two)
+
+schema = strawberry.Schema(query=Query)
+```
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,12 +1,14 @@
 Release type: patch
 
-Fixed edge case where `Union` types raised an `UnallowedReturnTypeForUnion` error when
-returning the correct type from the resolver. This also improves performance of
-StrawberryUnion's `_resolve_union_type` from `O(n)` to `O(1)` in the majority of cases.
+Fixed edge case where `Union` types raised an `UnallowedReturnTypeForUnion`
+error when returning the correct type from the resolver. This also improves
+performance of StrawberryUnion's `_resolve_union_type` from `O(n)` to `O(1)` in
+the majority of cases where `n` is the number of types in the schema.
 
-For example the below ([link to playground](https://play.strawberry.rocks/?gist=f7d88898d127e65b12140fdd763f9ef2))
-would previously raise the error when querying `two` as `StrawberryUnion` would incorrectly
-determine that the resolver returns `Container[TypeOne]`.
+For
+[example the below](https://play.strawberry.rocks/?gist=f7d88898d127e65b12140fdd763f9ef2))
+would previously raise the error when querying `two` as `StrawberryUnion` would
+incorrectly determine that the resolver returns `Container[TypeOne]`.
 
 ```python
 import strawberry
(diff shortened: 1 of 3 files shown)


botberry:



---

Fixed edge case where `Union` types raised an `UnallowedReturnTypeForUnion` error when
returning the correct type from the resolver due to only being able to partially match
nested generic types. This is fixed by prioritising the types explicitly listed in the
Union.

For example the below ([link to playground](https://play.strawberry.rocks/?gist=f7d88898d127e65b12140fdd763f9ef2))
would previously raise the error when querying `two` as `StrawberryUnion` would incorrectly
determine that the resolver returns `Container[TypeOne]`.

```python
import strawberry
from typing import TypeVar, Generic, Union, List, Type

T = TypeVar("T")

@strawberry.type
class Container(Generic[T]):
    items: List[T]

@strawberry.type
class TypeOne:
    attr: str

@strawberry.type
class TypeTwo:
    attr: str

def resolver_one():
    return Container(items=[TypeOne("one")])
    
def resolver_two():
    return Container(items=[TypeTwo("two")])

@strawberry.type
class Query:
    one: Union[Container[TypeOne], TypeOne] = strawberry.field(resolver_one)
    two: Union[Container[TypeTwo], TypeTwo] = strawberry.field(resolver_two)

```


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Tim OSullivan for the PR 👏

```

botberry:



---

Fixed edge case where `Union` types raised an `UnallowedReturnTypeForUnion` error when
returning the correct type from the resolver due to only being able to partially match
nested generic types. This is fixed by prioritising the types explicitly listed in the
Union.

For example the below ([link to playground](https://play.strawberry.rocks/?gist=f7d88898d127e65b12140fdd763f9ef2))
would previously raise the error when querying `two` as `StrawberryUnion` would incorrectly
determine that the resolver returns `Container[TypeOne]`.

```python
import strawberry
from typing import TypeVar, Generic, Union, List, Type

T = TypeVar("T")

@strawberry.type
class Container(Generic[T]):
    items: List[T]

@strawberry.type
class TypeOne:
    attr: str

@strawberry.type
class TypeTwo:
    attr: str

def resolver_one():
    return Container(items=[TypeOne("one")])
    
def resolver_two():
    return Container(items=[TypeTwo("two")])

@strawberry.type
class Query:
    one: Union[Container[TypeOne], TypeOne] = strawberry.field(resolver_one)
    two: Union[Container[TypeTwo], TypeTwo] = strawberry.field(resolver_two)

```


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Tim OSullivan for the PR 👏

```

invokermain:
I refactored the solution for performance and simplicity.

I benchmarked performance as well, as expected in the best-case scenario (i.e. only the two types used in the Union are present in the schema) it introduces a slight overhead.

However in the worst case scenario (schema defines many types, and all of them appear before the actual resolver type in `type_map.values()`) performance is vastly improved. 

image

This graph tests worse case scenario (although 0 is best case scenario as schema only has the two types):
- The new logic is practically constant with number of types in the schema: `~O(1)`.
- The old logic scales linearly with number of types in the schema: `O(n)`.

patrick91 (approved):
Nice one! I've updated the release notes slightly and added a TODO comment, it doesn't look like we need to iterate over all the types anymore, right? In any case this can be deal in another PR :)

invokermain:
Thanks for merging this, was blocking a ticket at work for me! 

> Nice one! I've updated the release notes slightly and added a TODO comment, it doesn't look like we need to iterate over all the types anymore, right? In any case this can be deal in another PR :)

Yeah I did think that. At first glance you'd think the check:

```python
if return_type not in type_.types:
    raise UnallowedReturnTypeForUnion
```

means we don't need to iterate over the whole `type_map`, i.e. iterating over` type_.types` is enough. But also I don't understand the internals of Strawberry enough to confidently say there isn't an edge case, maybe once you bring generics into the mix. So I did it was just the safest way possible in terms of breaking changes. If I was a more confident person maybe I would change it and just trust the existing unit tests cover all the relevant cases 😄

patrick91:
> Thanks for merging this, was blocking a ticket at work for me!

ohh, next time ping me and I'll try to merge it quicker 😊

 
> > Nice one! I've updated the release notes slightly and added a TODO comment, it doesn't look like we need to iterate over all the types anymore, right? In any case this can be deal in another PR :)
> 
> Yeah I did think that. At first glance you'd think the check:
> 
> ```python
> if return_type not in type_.types:
>     raise UnallowedReturnTypeForUnion
> ```
> 
> means we don't need to iterate over the whole `type_map`, i.e. iterating over` type_.types` is enough. But also I don't understand the internals of Strawberry enough to confidently say there isn't an edge case, maybe once you bring generics into the mix. So I did it was just the safest way possible in terms of breaking changes. If I was a more confident person maybe I would change it and just trust the existing unit tests cover all the relevant cases 😄

Ahaha, same to be honest, we can definitely do this in another PR. I don't remember why I did iterate over all types :'D

PR #2047, opened 2022-07-27
Improve support for Apollo Federation 2.0

patrick91:
## Description

While looking at https://github.com/strawberry-graphql/strawberry/issues/2044 I found some issues with our support for apollo federation, this PR addresses that, it also improves Mypy and Pyright support

- [x] Split up tests into multiple files
- [x] Add a link directive
- [x] Add federation directives support for input and interface
- [x] Automatically add the link directive on the schema based on usage
- [x] Decide if we should keep printing the directives (we do that for v1)
- [x] Add opt-in into v2
- [x] check if we have all the directives and that they are defined correctly
- [x] docs
- [x] federation support for scalars, enums, enum values and unions

Closes #1827

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [x] Enhancement/optimization
- [ ] Documentation

PR #2048, opened 2022-07-28
Add support for printing dictionaries when using custom scalars

patrick91:
This PR add support for printing dictionaries as default values, for example:

```python
@strawberry.input
class MyInput:
    j: JSON = strawberry.field(default_factory=dict)
    j2: JSON = strawberry.field(default_factory=lambda: {"hello": "world"})

@strawberry.type
class Query:
    @strawberry.field
    def search(self, input: MyInput) -> JSON:
        return input.j
```

will print the following

```graphql
input MyInput {
    j: JSON! = {}
    j2: JSON! = {hello: "world"}
}

type Query {
    search(input: MyInput!): JSON!
}
```

Here's an example of this working in another server: https://codesandbox.io/s/apollo-json-default-value-4mkreh?file=/index.js:0-899

## TODO:

- [x] allow nested dictionaries
- [x] this, but also for arguments

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,25 @@
+Release type: minor
+
+This release adds support for printing default values for scalars like JSON.
+
+For example the following:
+
+```python
+import strawberry
+from strawberry.scalars import JSON
+
+
+@strawberry.input
+class MyInput:
+    j: JSON = strawberry.field(default_factory=dict)
+    j2: JSON = strawberry.field(default_factory=lambda: {"hello": "world"})
+```
+
+will print the following schema:
+
+```graphql
+input MyInput {
+    j: JSON! = {}
+    j2: JSON! = {hello: "world"}
+}
+```
--- a/strawberry/printer/ast_from_value.py
+++ b/strawberry/printer/ast_from_value.py
@@ -73,7 +73,7 @@ def ast_from_leaf_type(
             ]
         )
 
-    raise TypeError(f"Cannot convert value to AST: {inspect(serialized)}.")
+    raise TypeError(f"Cannot convert value to AST: {inspect(serialized)}.")   # pragma: no cover
 
 
 def ast_from_value(value: Any, type_: GraphQLInputType) -> Optional[ValueNode]:
--- a/strawberry/printer/ast_from_value.py
+++ b/strawberry/printer/ast_from_value.py
@@ -73,7 +73,9 @@ def ast_from_leaf_type(
             ]
         )
 
-    raise TypeError(f"Cannot convert value to AST: {inspect(serialized)}.")   # pragma: no cover
+    raise TypeError(
+        f"Cannot convert value to AST: {inspect(serialized)}."
+    )  # pragma: no cover
 
 
 def ast_from_value(value: Any, type_: GraphQLInputType) -> Optional[ValueNode]:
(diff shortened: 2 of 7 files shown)


botberry:



---

This release adds support for printing default values for scalars like JSON.

For example the following:

```python
import strawberry


@strawberry.input
class MyInput:
    j: JSON = strawberry.field(default_factory=dict)
    j2: JSON = strawberry.field(default_factory=lambda: {"hello": "world"})
```

will print the following schema:

```graphql
input MyInput {
    j: JSON! = {}
    j2: JSON! = {hello: "world"}
}
```



Here's the tweet text: 
```

```

strawberry/printer/ast_from_value.py
```diff
@@ -0,0 +1,145 @@
...
+    if isinstance(serialized, bool):
+        return BooleanValueNode(value=serialized)
+
+    # Python ints and floats correspond nicely to Int and Float values.
+    if isinstance(serialized, int):
+        return IntValueNode(value=str(serialized))
+    if isinstance(serialized, float) and isfinite(serialized):
+        value = str(serialized)
+        if value.endswith(".0"):
+            value = value[:-2]
```
BryceBeagle, line 52: why do you single out the `.0` case? Is there a common scenario that leads to this where you'd want it printed as an integer? I notice you don't have a test-case for this either

strawberry/printer/printer.py
```diff
@@ -161,6 +162,30 @@ def print_field_directives(
...
+    return (
+        "(\n"
+        + "\n".join(
+            print_description(arg, f"  {indentation}", not i)
+            + f"  {indentation}"
+            + print_input_value(name, arg)
+            for i, (name, arg) in enumerate(args.items())
+        )
+        + f"\n{indentation})"
+    )
```
BryceBeagle, line 186: Do you think it would be better to split this into normal `for` loops instead of a really dense one-liner?

RELEASE.md
```diff
@@ -0,0 +1,25 @@
...
+```python
+import strawberry
+from strawberry.scalars import JSON
+
+
+@strawberry.input
+class MyInput:
+    j: JSON = strawberry.field(default_factory=dict)
+    j2: JSON = strawberry.field(default_factory=lambda: {"hello": "world"})
+```
```
BryceBeagle, line 16: Does `strawberry.field(default=...)` already work? The description says "default values" but the implementation here seems to be only for "default factories"?

strawberry/printer/printer.py
patrick91: yes, but I also want to rewrite the whole printer to not use any GraphQL core code

RELEASE.md
patrick91: yes, it works, but you shouldn't use it for mutable values 😊

strawberry/printer/ast_from_value.py
patrick91: the test is here: https://github.com/strawberry-graphql/strawberry/blob/63f17292e18eadde6d30c71c0174369b665fa776/tests/test_printer/test_basic.py#L138

This was done here: https://github.com/graphql-python/graphql-core/pull/164

I think we should/could refine the logic at some point, but I also don't want to change the current printing behaviour 😊

patrick91:
@BryceBeagle I'll merge this since I need it for the federation 2.0 fixes (https://github.com/strawberry-graphql/strawberry/pull/2047)

I'll open an issue about rewriting the printer 😊

RELEASE.md
BryceBeagle: Ahh right. Of course

PR #2054, opened 2022-07-29
Add support for schema directives on schema

patrick91:
This PR adds support for printing schema directives on the schema itself. There's not a lot of use cases for this, but Apollo Federation (https://github.com/strawberry-graphql/strawberry/pull/2047) uses them.

I'm a bit annoyed at the fact we now have 2 parameters, `directives` and `schema_directives` and specifically the fact that `directives` is actually used for schema directives when used on the fields:

```
strawberry.Schema(
    Query, 
    directives=[...], # directives that can be used in GraphQL operations
    schema_directives=[...], # directives that are printed on the schema
)
```

We **could** rename things, but it is annoying, and needs to be done in multiple steps, for example:

1. rename `directives` to `client_directives` and deprecate passing `directives` to the Schema
2. remove `directives` after some time
3. rename `schema_directives` to `directives` and deprecate passing `schema_directives` to the Schema
4. remove `schema_directives`

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,31 @@
+Release type: minor
+
+This PR adds support for adding schema directives to the schema of
+your GraphQL API. For printing the following schema:
+
+```python
+@strawberry.schema_directive(locations=[Location.SCHEMA])
+class Tag:
+    name: str
+
+@strawberry.type
+class Query:
+    first_name: str = strawberry.field(directives=[Tag(name="team-1")])
+
+schema = strawberry.Schema(query=Query, schema_directives=[Tag(name="team-1")])
+```
+
+will print the following:
+
+```graphql
+directive @tag(name: String!) on SCHEMA
+
+schema @tag(name: "team-1") {
+    query: Query
+}
+
+type Query {
+    firstName: String!
+}
+"""
+```
--- a/strawberry/directive.py
+++ b/strawberry/directive.py
@@ -16,12 +16,14 @@ from strawberry.types.fields.resolver import (
     ReservedType,
     StrawberryResolver,
 )
+from strawberry.unset import UNSET
 
 
-def directive_field(name: str) -> Any:
+def directive_field(name: str, default: object = UNSET) -> Any:
     return StrawberryField(
         python_name=None,
         graphql_name=name,
+        default=default,
     )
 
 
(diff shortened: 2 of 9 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

BryceBeagle:
> I'm a bit annoyed at the fact we now have 2 parameters, `directives` and `schema_directives` and specifically the fact that `directives` is actually used for schema directives when used on the fields:
> 
> ```
> strawberry.Schema(
>     Query, 
>     directives=[...], # directives that can be used in GraphQL operations
>     schema_directives=[...], # directives that are printed on the schema
> )
> ```
> 
> We **could** rename things, but it is annoying, and needs to be done in multiple steps, for example:
> 
> 1. rename `directives` to `client_directives` and deprecate passing `directives` to the Schema
> 2. remove `directives` after some time
> 3. rename `schema_directives` to `directives` and deprecate passing `schema_directives` to the Schema
> 4. remove `schema_directives`

Would it make sense to just have a single argument used for both? Or is there use-case too different to make this a good idea?

```python
class Directive: ...
class SchemaDirective(Directive): ...
class ClientDirective(Directive): ...

class Schema:
    directives: Sequence[Directive]
```

Otherwise it might be a bit weird to have either [`directives`/`schema_directives`] or [`client_directives`/`directives`]. It might be too confusing to users, especially new ones, why one appears to be a superset of the other but isn't. Perhaps it would be best to have [`client_directives`/`schema_directives`]

patrick91:
> Would it make sense to just have a single argument used for both? Or is there use-case too different to make this a good idea?

I thought about that, but I think it might be confusing, schema directives are printed on the schema like so

```graphql
schema @tag(name: "team-1") {
    query: Query
}
```

client directives are printed like so:

```graphql
directive @uppercase on FIELD_DEFINITION
```

and used like so:

```graphql
query {
    user {
        name @uppercase
    }
}
```

> Otherwise it might be a bit weird to have either [`directives`/`schema_directives`] or [`client_directives`/`directives`]. It might be too confusing to users, especially new ones, why one appears to be a superset of the other but isn't. Perhaps it would be best to have [`client_directives`/`schema_directives`]

I think we can do that ([`client_directives`/`schema_directives`]) We won't have a lot of people using schema directives anyway, so having a verbose name should be fine.

And client directives are only allowed on the schema, so the longer name is fine :)

PR #2058, opened 2022-08-01
Add support for printing custom scalars in schema directives

patrick91:
This PRs add support for printing custom scalars in schema directive. Previously they wouldn't be printed if only the schema directive was using that scalar directive.


botberry:

```markdown
Release type: patch

```


Here's the tweet text: 
```

```

PR #2059, opened 2022-08-01
Add support for printing enums only used in schema directives

patrick91:
Exactly the same as https://github.com/strawberry-graphql/strawberry/pull/2058 but for enums 😊

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,36 @@
+Release type: patch
+
+This release adds support for priting custom enums used only on
+schema directives, for example the following schema:
+
+```python
+@strawberry.enum
+class Reason(str, Enum):
+    EXAMPLE = "example"
+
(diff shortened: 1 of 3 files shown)


botberry:



---

This release adds support for priting custom enums used only on
schema directives, for example the following schema:

```python
@strawberry.enum
class Reason(str, Enum):
    EXAMPLE = "example"

@strawberry.schema_directive(locations=[Location.FIELD_DEFINITION])
class Sensitive:
    reason: Reason

@strawberry.type
class Query:
    first_name: str = strawberry.field(
        directives=[Sensitive(reason=Reason.EXAMPLE)]
    )
```

prints the following:

```graphql
directive @sensitive(reason: Reason!) on FIELD_DEFINITION

type Query {
    firstName: String! @sensitive(reason: EXAMPLE)
}

enum Reason {
    EXAMPLE
}
```

while previously it would omit the definition of the enum.



Here's the tweet text: 
```

```

PR #2113, opened 2022-08-21
support return types from permissions

### LINKED ISSUE #2100: Return types from permission classes, instead of raising exceptions.
Return graphql  types from permission classes, instead of raising exceptions.

## Feature Request Type

- [ ] Core functionality
- [x] New behavior

## Description
Currently this is what you'll do for permissions:
```py
# copied from the docs. 

    # This method can also be async!
    def has_permission(self, source: typing.Any, info: Info, **kwargs) -> bool:
        return False

@strawberry.type
class Query:
    user: str = strawberry.field(permission_classes=[IsAuthenticated])
```

The problem is that it is not considered best practice in GraphQL to raise GraphQL exceptions 
for stuff that you really except.

### Therefore a proposal:
```py
@strawberry.type
class AuthOutput:
    success: bool
    code: int
    error: str = None

    def has_permission(self, source: typing.Any, info: Info, **kwargs) -> AuthOutput:
        return AuthOutput(success=True, code=0)

@strawberry.type
class Query:
    user: str = strawberry.field(permission_classes=[IsAuthenticated])
```

 There are 4 ways I can think of to make this work:
1. migrate the return type of the field to `Union[Value[str],  AuthOutput]` 

| Pros                                               | Cons                                       |
|----------------------------------------------|----------------------------------------|
| No need to override the original return type | Issues when the field returns a scalar |
| Easier to combine multiple permissions|
2. Create a new type with the fields from both the field and the permission class return type.

| Pros                                               | Cons                                       |
|----------------------------------------------|----------------------------------------|
| can handle scalars seamlessly  | Quite messy and  determines stuff to the user upfront |
3. Let the user handle it, It is up to the user to determine the return type for both the field and the permission class.  

| Pros                                               | Cons                                       |
|----------------------------------------------|----------------------------------------|
| Easier on the strawberry side  | More code for the user |

4. Let users extend a base Permission interface that we will create.
```py
# this would be something that strawberry provides.
@strawberry.interface
class PermissionInterface:
    success: bool

# user code:
    @strawberry.type
    class IsAuthenticatedPayload(PermissionInterface):
        success: bool

    def has_permission(self, source: typing.Any, info: Info, **kwargs) -> IsAuthenticatedPayload:
        return self.IsAuthenticatedPayload(success=False, message="Not Authorized")
```
This would generate an Union of the field that is wrapped with the permission and
the `IsAuthenticatedPayload` type.

 | Pros                                               | Cons                                       |
|----------------------------------------------|----------------------------------------|
| More elegant                                 | might be unwanted by some users,  |
| We can easily tell if `has_permission` was successful with the `success` field. | the `success` field might be useless in the schema |

nrbnlulu:
Support return strawberry types from permissions instead of raising exceptions.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

* resolves #2100

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -135,6 +135,19 @@ $ poetry run pre-commit install
 Your code must always be accompanied by corresponding tests. If tests are not
 present, your code will not be merged.
 
+#### Debugging
+During debugging, you might tackle errors like:
+```bash
+ModuleNotFoundError: No module named 'aiohttp.test_utils'
+```
+or
+```bash
+ImportError: cannot import name 'GenericAlias' from 'types'
+```
+
+these are caused because of namespace clashes, make sure that you are running the tests from
+strawberry root directory and not from inside `strawberry/tests` directory.
+
 #### Pull requests
 
 - Be sure that your pull request contains tests that cover the changed or added
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -29,16 +29,14 @@ from strawberry.types.info import Info
 from strawberry.union import StrawberryUnion
 from strawberry.unset import UNSET
 
-from .permission import BasePermission
 from .types.fields.resolver import StrawberryResolver
 
 
 if TYPE_CHECKING:
     from .object_type import TypeDefinition
-
+    from .permission import BasePermission
 
 _RESOLVER_TYPE = Union[StrawberryResolver, Callable, staticmethod, classmethod]
-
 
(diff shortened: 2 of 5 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Nir for the PR 👏

```

PR #2114, opened 2022-08-21
Only install `backports.cached-property` if Python < 3.8

### LINKED ISSUE #2108: `backports.cached-property` should only be required for python<3.8
Hi,

the change introduced in #1582 causes my pipeline to fail, I think due to the dependency name it cannot be found on our private repo server. As far as I know, this library is in any case only required for Python<3.8, so would it be possible to mark the dependency as such?

Suggested change to pyproject.toml would be this:

```
backports-cached-property = { version = "^1.0.1", python = ">=3.6,<3.8" }
```

If for some reason, you don't want to restrict to a specific Python version, could we rename the dependency to `backports-cached-property`?

Happy to make a PR for either case.

ljnsn:
Update `backports.cached-property` dependency to only install when Python < 3.8.

## Description

- Added the Python version restriction to pyproject.toml.
- Update poetry.lock
- Update code to use `functools.cached_property` when Python >= 3.8.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

Closes #2108 

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,7 +52,7 @@ sanic = {version = ">=20.12.2,<22.0.0", optional = true}
 aiohttp = {version = "^3.7.4.post0", optional = true}
 fastapi = {version = ">=0.65.2", optional = true}
 channels = {version = ">=3.0.5", optional = true}
-"backports.cached-property" = "^1.0.1"
+backports-cached-property = {version = "^1.0.2", python = "<3.8"}
 
 [tool.poetry.dev-dependencies]
 pytest = "^7.1"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,7 +52,7 @@ sanic = {version = ">=20.12.2,<22.0.0", optional = true}
 aiohttp = {version = "^3.7.4.post0", optional = true}
 fastapi = {version = ">=0.65.2", optional = true}
 channels = {version = ">=3.0.5", optional = true}
-backports-cached-property = {version = "^1.0.2", python = "<3.8"}
+"backports.cached-property" = {version = "^1.0.2", python = "<3.8"}
 
 [tool.poetry.dev-dependencies]
 pytest = "^7.1"
--- a/strawberry/directive.py
+++ b/strawberry/directive.py
@@ -4,7 +4,6 @@ import dataclasses
 import inspect
 from typing import Any, Callable, List, Optional, TypeVar
 
-from backports.cached_property import cached_property
 from typing_extensions import Annotated
 
 from graphql import DirectiveLocation
@@ -17,6 +16,7 @@ from strawberry.types.fields.resolver import (
     StrawberryResolver,
 )
 from strawberry.unset import UNSET
+from strawberry.utils.cached_property import cached_property
 
 
 def directive_field(name: str, default: object = UNSET) -> Any:
--- a/strawberry/utils/cached_property.py
+++ b/strawberry/utils/cached_property.py
@@ -1,0 +1,15 @@
+import sys
+from typing import Any, Callable, TypeVar
+
+
+if sys.version_info < (3, 8):
+    from backports.cached_property import cached_property as _cached_property
+else:
+    from functools import cached_property as _cached_property
+
+T = TypeVar("T")
+
+
+def cached_property(func: Callable[[Any], T]) -> _cached_property[T]:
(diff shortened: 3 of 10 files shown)


ljnsn:
Let me know if I need to update documentation or any additional steps.

botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to ljnsn for the PR 👏

```

pyproject.toml
```diff
@@ -52,7 +52,7 @@ sanic = {version = ">=20.12.2,<22.0.0", optional = true}
 aiohttp = {version = "^3.7.4.post0", optional = true}
 fastapi = {version = ">=0.65.2", optional = true}
 channels = {version = ">=3.0.5", optional = true}
-"backports.cached-property" = "^1.0.1"
+backports-cached-property = {version = "^1.0.2", python = ">=3.6,<3.8"}
```
patrick91, line 55: just curious about the change in the name, was it just cosmetic or there a reason for it?

also I think we can remove the lower bound limit, since we don't support python 3.6 anyway 😊

strawberry/directive.py
```diff
@@ -19,6 +19,12 @@
 from strawberry.unset import UNSET
 
 
+if sys.version_info < (3, 8):
+    from backports.cached_property import cached_property
+else:
+    from functools import cached_property
```
patrick91, line 25: could we make our own helper that does this? so we don't have to repeat in multiple places?

patrick91 (commented):
Thank you for working on this 😊 I've added a few minor comments

pyproject.toml
ljnsn: I agree about the lower bound, will remove that.

About the name, I've had a problem with the module not being able to be resolved on my work PC that runs Windows. I'm not sure if the issue was due to the private repo we're using there or Windows. On Linux it resolves fine and even if I do `poetry add backports-cached-property` it's added to pyproject.toml as `backports.cached-property`. However, I think under Windows it's added as `backports-cached-property`. I can check and confirm that tomorrow.

In any case, I should not have a problem at work anymore, since we use Python > 3.8, so I don't really mind either way, although I do think that `backports-cached-property` without the quotes is a bit nicer. Let me know what you prefer and I'll change it accordingly :)

strawberry/directive.py
ljnsn: Good idea, where would you suggest putting that?

patrick91: `utils/cached_property.py` sounds like a good enough place 😊

strawberry/utils/cached_property.py
```diff
@@ -0,0 +1,15 @@
...
+    from backports.cached_property import cached_property as _cached_property
+else:
+    from functools import cached_property as _cached_property
+
+T = TypeVar("T")
+
+
+def cached_property(func: Callable[[Any], T]) -> _cached_property[T]:
+    """Helper to return the cached_property from functools or backports."""
+    return _cached_property(func)
```
patrick91, line 15: do we need this function? can't we rely on the import? 😊

ljnsn: I tried with the import only, for some reason all the mypy tests then failed, all with the same error that instead of getting the expected error they got "Unnecessary # type: ignore comment" or something along the lines. For example for this line: https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/field.py#L139.

This seemed to be totally unrelated to cached_property, so I wasn't sure what to make of that. Using this helper function all the tests pass fine.

patrick91: interesting! I'll do a quick test and let you know! Otherwise this is fine to merge 😊

pyproject.toml
patrick91: I've reverted the name as it was causing a random failure; https://github.com/strawberry-graphql/strawberry/runs/7974876734

strawberry/utils/cached_property.py
ljnsn: Yes, looks good, thanks!

patrick91 (approved):
Thank you so much for this! I'm glad we reduced the number of dependencies for modern python, great stuff!

botberry:

PR #2117, opened 2022-08-22
[WIP] Support custom fields

nrbnlulu:
## Description
Disclaimer:  _This PR was started as a fix for #920 and expanded to refactoring of the strawberry type system._

### Motivation:
While walking threw the code I have found that being able to extend `StrawberryField` was pretty hard for these reasons:
 1. the field evaluation process relied mostly on functions (namely `_get_fields`).
 2. `StrawberryField` was inheriting `dataclasses.Field` and that IMO unnecessary burden, results namespace clashes, tights strawberry to dataclasses too much and finally it is not a dataclass like many other parts of strawberry.
 3. generic types was evaluated in a very complex way (`copy_with`) and it is not clear at a first glance when a type is already evaluated or not,
 4.  Integrations (like pydantic) are building lots of things on their own and some functionalities that strawberry provides or not allows might not be covered in the integration, as well as they have lots of redundant code that mimics what strawberry-core already does. 
 5. Usage of `@property` was quite extensive and that is a runtime burden and makes hard to deduce what will a value be at a given point.   
 6.  there are probably more things I forgot but the general idea was to padronize the way strawberry does things and rely more on classes than on functions to allow extending.
___
#### Generics:
- if a class inherits `Generic` it would be created as `class TemplateTypeDefinition(TypeDefinition)` This class can generate a whole new "`StrawberryObject`" based on `__args__`. the generator is `TemplateTypeDefinition.generate(<types>)`. it is lunched by the class `__getitem__` that defined in `StrawberryMeta`, so  i.e `Apple[int]` is equivalent to `Apple._type_definition.generate((int, ))`.
- I introduced new `validate` abstract method to `StrawberryType` it is used to determine generic unions.   
  *NOTE*:  With the new API for generating templates this might not even be necessary because users would be able to do this:
 ```py
@strawberry.type
class A(Generic[T]):
    a: T

@strawberry.type
class SomeType():
    @strawberry.field
    def union_generic(self) -> Union[A[int], A[str]]
        return A[int](1)
 ```
 though this would break existing code.
- generics are now cached by `__args__` signature and every type is created just once. This is tested using `timeit`. 

#### Scalars:
 I removed `ScallarWrapper` as it made issued for me with the new `validate` method of `StrawberryType`, every scalar now is just a `ScalarDefinition`, I would consider renaming this to `StrawberryScalar`.
 
#### strawberry.ID
`strawberry.ID` is no longer a plain `NewType` but a `ScalarDefinition` using implementation from graphql-core, I saw that is is a 'TODO' and also it did issues with `validate()` of generic unions.  

#### pydantic:
I have also refactored pydantic integration, and it is no longer building things on it's own, it connects to hooks from strawberry-core to append and modify fields.

#### StrawberryObject:
strawberry object is what will lay behind `@strawberry.type` and it is extending the `StrawberryMeta` class to evaluate generics as explained above.
This has the benifit of duck-typing `_type_definition` and allows integrations to connect to strawberry type creation system.
The workflow is like this:
1. `StrawberryObject` creates a new type from the decorated class that extends `StrawberryObject`.
2. `StrawberryObject` calls a customizable function to create the `TypeDefinition` (to support extending by integrations). 
3. `TypeDefinition` will find all the fields in the class including pre-evaluated dataclasses fields.
7. `TypeDefinition` will create a `StrawberryField` for each field (integrations can define their own `StrawberryField` extended and it will be created instead).
8. `TypeDefinition` will call `__pre_dataclass_creation__` to allow integrations to add  or modify fields.
9. dataclass is created and the rest is history.

As you can see this PR won't only allow custom fields but also custom `StrawberryObject` and `TypeDefinition` the pydantic integration can set an example for this.
___

Please contact me if You want more information, Thank you @patrick91, @BryceBeagle  for your valuable time and patience.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

* #920 

## Checklist

### tests
- [x] renamer field
- [x] type changer field
- [x] argument modification field
- [x] override `get_result` field
- [ ] batched fields.
- [ ] field "`__post_init__`" 
- [x] not supported resolver assertion
- [x] not supported evolvable 
- [ ] StrawberryField was initialized with origin but with no python name assertion.
- [ ] test validate generic union with `StrawberryObject`


botberry:



---

This release will try to address [#1100](https://github.com/strawberry-graphql/strawberry/pull/1100).

### New features:
- allow overriding `StrawberryField`'s
    - `get_result` - for overriding the execution of resolvers.


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Nir for the PR 👏

```

pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [tool.poetry]
 name = "strawberry-graphql"
 packages = [ { include = "strawberry" } ]
-version = "0.126.0"
+version = "0.126.1"
```
patrick91, line 4: this is done automatically when merging 😊 can you revert it?

strawberry/object_type.py
```diff
@@ -45,33 +41,15 @@ def _check_field_annotations(cls: Type):
...
-
-                cls_annotations[field_name] = field_.base_resolver.type_annotation
-
-            # TODO: Make sure the cls annotation agrees with the field's type
-            # >>> if cls_annotations[field_name] != field.base_resolver.type:
-            # >>>     # TODO: Proper error
-            # >>>    raise Exception
+        if isinstance(field_, StrawberryLazyField):
+            strawberry_field = field_.evaluate(cls, field_name)
+            setattr(cls, field_name, strawberry_field.to_dataclass_field)
```
patrick91, line 52: `to_dataclass_field` sounds like a function, shall we change it do be a regular function instead of a property?

tests/fields/test_basic.py
```diff
@@ -111,6 +108,3 @@ class Query:
     assert instance.the_list == []
     fields = Query._type_definition.fields
     assert [field.default_value for field in fields] == [3, []]
-
-    with pytest.raises(InvalidDefaultFactoryError):
-        strawberry.field(default_factory=round)
```
patrick91, line 116: why is this not raising anymore?

tests/fields/test_custom.py
```diff
@@ -0,0 +1,23 @@
...
+        def get_result(self, *args, **kwargs):
+            res = super().get_result(args, **kwargs)
+            return res + " world"
+
+    def resolve_hello(self) -> str:
+        return "hello"
+
+    @strawberry.type
+    class Query:
+        hello: str = WorldField.evolve_with_resolver(resolve_hello)
```
patrick91, line 19: I'm not sure `evolve_with_resolver` is easy to understand, could this follow the same api as `strawberry.field`, so:

`WorldField(resolver=resolve_hello)`?

strawberry/field.py
```diff
@@ -23,129 +23,289 @@
...
-            except TypeError as exc:
-                raise InvalidDefaultFactoryError() from exc
-
-        self.is_subscription = is_subscription
+@dataclasses.dataclass(frozen=True)
+class StrawberryField:
+    python_name: Optional[str]
+    """
+    name to be used by python, i.e in the generated __init__
+    """
```
patrick91, line 120: nit: I don't we need all these doc strings, I think they make the code less readable and don't provide a lot of value 😊

strawberry/field.py
```diff
@@ -23,129 +23,289 @@
...
+        is_subscription: bool = False,
+        description: Optional[str] = None,
+        deprecation_reason: Optional[str] = None,
+        permission_classes=None,  # type: ignore
+        directives: Sequence[object] = (),
+        dataclass_options: Optional[DataclassFieldSettings] = DEFAULT_FIELD_SETTINGS,
+        is_basic_field: bool = False,
+        # legacy compatibility, will be passed to dataclass_options:
+        default: object = UNSET,
+        default_factory: Union[Callable[[], Any], object] = UNSET,
```
patrick91, line 226: why is this legacy? this is (also) part of Strawberry's api 😊

strawberry/field.py
```diff
@@ -23,129 +23,289 @@
...
-        self.deprecation_reason = deprecation_reason
+            assert_message = (
+                "Field must have a name by the time the schema is generated"
+            )
+            assert self.python_name is not None, assert_message
 
-    def __call__(self, resolver: _RESOLVER_TYPE) -> "StrawberryField":
-        """Add a resolver to the field"""
+    @classmethod
+    def create(
```
patrick91, line 210: why a class method and not just a constructor? :)

strawberry/field.py
```diff
@@ -23,129 +23,289 @@
...
-        # Allow for StrawberryResolvers or bare functions to be provided
-        if not isinstance(resolver, StrawberryResolver):
-            resolver = StrawberryResolver(resolver)
+    @classmethod
+    def evolve_with_resolver(
+        cls,
+        instance: "StrawberryField",
+        base_resolver: StrawberryResolver,
+        origin: Union[Type, Callable, staticmethod, classmethod],
+    ) -> "StrawberryField":
```
patrick91, line 265: I'm unclear why we need this and can't use standard inheritance 🤔

patrick91 (commented):
Thanks so much for doing this, I've added a few comments 😊

pyproject.toml
nrbnlulu: sure

strawberry/object_type.py
nrbnlulu: it is a function :upside_down_face: thats a typo

tests/fields/test_custom.py
nrbnlulu: I'll have a try

tests/fields/test_basic.py
nrbnlulu: I thought that this is redundant since it is pure dataclasses logic and has nothing to do with strawberry am I wrong?

strawberry/field.py
nrbnlulu: I want to remove this and let users use 
`DataclassFieldSettings` instead, this will allow to modify the fields kwargs as well as provide better typing information.

also this is not related to `StrawberryField` I want to dataclass logic from strawberry logic as much as possible.

tests/fields/test_basic.py
patrick91: that's a good point! I'd keep it though, mostly so that we don't accidentally remove this dataclasses logic 😊

strawberry/field.py
nrbnlulu: lol I though this would allows to modify the `dataclass __init__` along the way didn't though to call `super().__init__(...)`

strawberry/field.py
nrbnlulu: This is not for overriding `StrawberryField` this is just me refactoring... instead of having the `StrawberryField` built up in so many places we build it only using several functions.

patrick91: mmh, what do you mean exactly?

nrbnlulu: for instance field.origin is determined in `strawberry.type` IIRC it is quite cumbersome IMO.

nrbnlulu: Here:
https://github.com/strawberry-graphql/strawberry/blob/1fc34fa708ebb0d21c6e1a34f86c6202fcfb4783/strawberry/types/type_resolver.py#L116

not only that it is weird that `_get_field` (that is a getter) is messing up with state, it is also unclear to me why wouldn't this logic would live inside `StrawberryField`.

strawberry/field.py
nrbnlulu: What do you have in mind? Should i drop the `DataclassFieldSettings`?

tests/fields/test_basic.py
nrbnlulu: This can make it easier to generate docs later on though..

strawberry/field.py
nrbnlulu: This can make it easier to generate docs later on though...

strawberry/field.py
nrbnlulu: actually I was right, you cannot assign to a frozen dataclass in __init__ see [this SO](https://stackoverflow.com/questions/57893902/how-can-i-set-an-attribute-in-a-frozen-dataclass-custom-init-method)

strawberry/field.py
patrick91: That sounds good, but I'm not sure why we need to use these methods and not use the constructor or have a single evaluate call for this 😊

strawberry/field.py
patrick91: yup 😊 we can declare all the params on our function, and maybe use this in future to make it cleaner; https://peps.python.org/pep-0692/

strawberry/field.py
nrbnlulu: > use the constructor
It is impossible with frozen dataclasses.
> have a single evaluate call for this.
there actually is only one evaluate call for this in `StrawberryLazyField` :upside_down_face:  , Wym?

patrick91: do we need to make it frozen?

when is `evolve_with_resolver` called?

nrbnlulu: > do we need to make it frozen?  

I like it this way, keeps things more robust.

> when is evolve_with_resolver called? 

currently [here](https://github.com/strawberry-graphql/strawberry/pull/2117/files#diff-00af9d43c9b59404ba170e72e470939f2a6dd50e2eae24d0e24ef105431e5414R454).Though  I am still trying to figure out where is the best place to call `StrawberryLazyField.evaluate`

nrbnlulu:
Hey, I have dropped the idea of a frozen styled class even though it makes thing more orgenized, it would be less dynamic.

Currently I am relying heavily on `__call__` of `StrawberryField`, this has several benefits:

- supports using the `StrawberryField` as a decorator.
- having only one public entry for field creation
- allows to batch fields like this:
```
class StrawberryField:
    def __init__(self, child=None)
        self.child = child

@strawberry.type
class Query:
    @UpperCaseField(AuthenticationRequiredField())
    def get_username(self, id: strawberry.ID) -> str:
        return "jkimbo"
```
this is not currently implemented but I think I'll be able to do this,

strawberry/types/types.py
```diff
@@ -23,6 +29,80 @@
...
+    class if one is not set by either using an explicit strawberry.field(name=...) or by
+    passing a named function (i.e. not an anonymous lambda) to strawberry.field
+    (typically as a decorator).
+    """
+    # Deferred import to avoid import cycles
+    from strawberry.field import StrawberryField
+
+    fields: Dict[str, StrawberryField] = {}
+
+    # TODO: What is this?
```
nrbnlulu, line 90: @patrick91 can you tell me if that is necessary? or what is that for? fields test are passing locally for me without it.

nrbnlulu:
Also, I have rename `_type_defenition` to `__strawberry_definition__` I hope you will forgive me not putting this in another PR, though `_type_defenition` is kinda cryptic :upside_down_face: ...

docs/types/exceptions.md
```diff
@@ -200,45 +200,84 @@ MyCustomScalar2 = strawberry.scalar(
...
-.../venv/lib/python3.9/site-packages/graphql/type/schema.py:208: in __init__
-    collect_referenced_types(query)
-.../venv/lib/python3.9/site-packages/graphql/type/schema.py:422: in collect_referenced_types
-    for field in named_type.fields.values():
-....9envet__
-    val = self.func(instance)
-.../venv/lib/python3.9/site-packages/graphql/type/definition.py:769: in fields
-    raise TypeError(f"{self.name} fields cannot be resolved. {error}")
-E   TypeError: Query fields cannot be resolved. Scalar `MyCustomScalar` has already been registered
+... / venv / lib / python3
```
BryceBeagle, line 212: I think the autoformatter got a bit eager here

tests/schema/test_lazy_types/type_a.py
```diff
@@ -20,3 +20,5 @@ def type_b(self) -> strawberry.LazyType["TypeB", ".type_b"]:  # noqa
         from .type_b import TypeB
 
         return TypeB()
+
+    done: bool = True
```
BryceBeagle, line 24: What is this change here for? I see that it's used in `test_cyclic` but I'm wondering why. Also what drove it to be added as a resolver-less field _below_ a field with a resolver -- is that necessary?

tests/schema/test_lazy_types/test_cyclic.py
```diff
@@ -30,5 +31,23 @@ class Query:
...
+    }
+    """,
+        root_value=Query(a=TypeA(), b=TypeB()),
+    )
+    assert res.errors is None
+    assert res.data["a"]["typeB"]["typeA"]["done"]
+    expected = textwrap.dedent(expected).strip()
+    schema = print_schema(schema).splitlines()
+    for line in expected.splitlines():
+        assert line in schema
```
BryceBeagle, line 53: Why this change? Are we not guaranteeing order?

tests/fields/test_resolvers.py
```diff
@@ -134,9 +132,11 @@ def adios(self):
...
     with pytest.raises(MissingArgumentsAnnotationsError) as e:
 
-        @strawberry.field
-        def hello(self, query) -> str:
-            return "I'm a resolver"
+        @strawberry.type
+        class ShouldRaiseHere:
+            @strawberry.field
+            def hello(self, query) -> str:
+                return "I'm a resolver"
```
BryceBeagle, line 139: Not sure about this change. Is it no longer a requirement that a field is "complete" until it's added to a type?

tests/fields/test_custom.py
```diff
@@ -0,0 +1,9 @@
+import strawberry
+
+
+strawberry.field
+
+
+def test_override_get_result():
+    # TODO: after patrick approves changes.
```
BryceBeagle, line 8: Let's make sure not to merge until this is resolved

tests/fields/test_basic.py
```diff
@@ -109,8 +124,19 @@ class Query:
...
+        class ShouldRaiseHere:
+            round_takes_arguments: int = strawberry.field(default_factory=round)
+
+
+def test_evolve_with_wrong_argument():
+    with pytest.raises(TypeError):
+
+        @strawberry.type
+        class Query:
+            the_field: int = strawberry.field(default=3)(2)
```
BryceBeagle, line 142: This test is the only usage of the term "evolve" anywhere in your change -- what does it mean? What does this test check for?

strawberry/utils/typing.py
```diff
@@ -117,3 +117,9 @@ def __dataclass_transform__(
     field_descriptors: Tuple[Union[type, Callable[..., Any]], ...] = (()),
 ) -> Callable[[_T], _T]:
     return lambda a: a
+
+
+def update_annotations(cls: Type, annots: dict) -> None:
```
BryceBeagle, line 122: Possible to narrow the `dict` type annotation? I believe the key will always be a `str`?

strawberry/exceptions.py
```diff
@@ -246,3 +250,12 @@ def __init__(self):
             "that inherits from BaseContext or a dictionary"
         )
         super().__init__(message)
+
+
+class UncallableResolverError(Exception):
+    def __init__(self, resolver: "StrawberryResolver"):
```
BryceBeagle, line 256: Quotes around `StrawberryResolver` are unnecessary because we're using `__future__` annotations

strawberry/annotation.py
```diff
@@ -269,12 +275,19 @@ def _strip_lazy_type(cls, annotation: LazyType) -> type:
...
+def _get_strawberry_definition(type_: Any) -> Optional[StrawberryDefinition]:
+    origin = type_
+    # generics store their class in __origin__
+    if origin_ := getattr(type_, "__origin__", False):
+        origin = origin_
+    res = getattr(origin, "__strawberry_definition__", None)
+    if isinstance(res, StrawberryDefinition):
+        return res
+    else:
+        return None
```
BryceBeagle, line 293: You implemented this function (which I think is great!), but there are a lot of places where we're still manually grabbing the `__strawberry_definition__` off the types. I think it should be:
1. be moved out of `annotations.py`
2. used in more places

strawberry/types/types.py
```diff
@@ -23,26 +33,257 @@
...
+            for field in __strawberry_definition__.fields:
+                if field.python_name in base.__annotations__:
+                    origins.setdefault(field.python_name, base)
+
+    return list(fields.values())
+
 
 @dataclasses.dataclass(eq=False)
-class TypeDefinition(StrawberryType):
+class StrawberryDefinition(StrawberryType):
```
BryceBeagle, line 115: 1. @patrick91, what do you think of this renaming. I wonder if calling this `StrawberryDefinition` is too broad and this should be `StrawberryObjectDefinition` or `StrawberryTypeDefinition`
2. @nrbnlulu, I hate to ask this because it's going to turn into a fair bit of work for you but can you please do this renaming in a separate diff? This turns a large diff into an absolutely massive diff making this fields change really hard to review.

    This renaming of `TypeDefinition` -> `StrawberryDefinition` is also a somewhat breaking change -- we want to separate it from the fields change to allow easier version upgrading by our users

BryceBeagle (changes requested):
I've done an introductory review of the change. I didn't have much time to review tonight so I did everything _except_ look at the important, logic-changing part of the change. That should also be easier if the `TypeDefinition` renaming is pulled out into a separate change

strawberry/types/types.py
BryceBeagle: > Also, I have rename `_type_defenition` to `__strawberry_definition__` I hope you will forgive me not putting this in another PR, though `_type_defenition` is kinda cryptic :upside_down_face: ...

Just saw this comment :grimacing:. I think splitting it out is pretty important, unfortunately

patrick91: doing it in another PR sounds best :) this will break strawberry django and other libs I think :)

for the name I'd go with StrawberryTypeDefinition for object types, StrawberryInputTypeDefinition for input types and StrawberryInterfaceDefinition for interfaces. But that makes me think of how we have StrawberryScalar, which is not too consistent with these names, maybe we can rename it to StrawberryScalarDefinition (something for another PR )

tests/schema/test_lazy_types/type_a.py
nrbnlulu: Due to how I evaluated forward refs it seems like the schema would not be generated in the same order, so I just had that test there to be sure.

tests/schema/test_lazy_types/test_cyclic.py
nrbnlulu: See above

nrbnlulu: Is order mandatory?

tests/fields/test_resolvers.py
nrbnlulu: There are some stuff that cannot be determined until the origin is determined, for basic fields we must kniw the python name and the origin so fields are not evaluated at all until they are added in a type. I cant see a use case where this might cause problems, do you?

tests/fields/test_custom.py
nrbnlulu: I'll add an assertion

strawberry/annotation.py
nrbnlulu: Where do you think it should live?

strawberry/types/types.py
nrbnlulu: O.K. although we can have a property with a warning that will return a pointer to the new name...

tests/fields/test_basic.py
nrbnlulu: When a field is called (similar to how you did this) it will be added with a resolver or something else and will be re-evaluated in __post_init__

nrbnlulu:
@patrick.py#0235 @ignormies#0001 I have done major refactoring here, there was some things that I just couldn't do with the previous way they were.
I hope that you will find this helpful as I dedicated most of my time recently just for this PR.

Some notes about major changes:
 - I didn't touched the public API
 - I introduced new `validate` abstract method to `StrawberryType` it is used to determine generic unions
 - I have removed `copy_with` and `type_params` because I couldn't get it to work with the new way `StrawberryField` is determined, also it was very complicated IMO. The alternative now is a new `TemplateTypeDefinition` class, if a class is inheriting a `Generic` it would be instanciated as a template by `@strawberry.type`. This class can generate a whole new "`StrawberryObject`" based on `__args__` with just two functions and it also has a cache map based on `__args__` tuple.

strawberry/field.py
```diff
@@ -9,292 +8,305 @@
...
+    # TODO: replace type with StrawberryObject
+    def _evaluate_from_base_resolver(self) -> None:
+        """
+        called by __post_init__ if there is a base resolver and an origin.
+        """
+        # validate base_resolver
+        if not isinstance(self.base_resolver, StrawberryResolver):
+            if resolver := resolveable(self.base_resolver):
+                self.base_resolver = StrawberryResolver(resolver)
+            #     TODO: Tests this fails
```
nrbnlulu, line 197: forgot this todo

strawberry/field.py
```diff
@@ -394,24 +443,29 @@ def field(
...
+    # if called like @strawberry.field or not as decorator but with field(resolver=...)
     if resolver:
         assert init is not True, "Can't set init as True when passing a resolver."
-        return field_(resolver)
-    return field_
+        field_(resolver)
+        return field_
+
+    # called like @strawberry.field(description="ABC") or not as decorator
+    # further on the decorator would call StrawberryLazyField.__call__(decorated)
```
nrbnlulu, line 466: no more `StrawberryLazyField` :smile:

strawberry/type.py
```diff
@@ -142,3 +126,17 @@ def __eq__(self, other) -> bool:
             return self.type_var == other
 
         return super().__eq__(other)
+
+
+class StrawberryGqlWrapper(StrawberryType):
```
nrbnlulu, line 131: I think it is not used anymore, to delete.

strawberry/types/types.py
```diff
@@ -23,6 +38,176 @@
...
+            if type_definition.is_interface:
+                interfaces.append(type_definition)
+
+        for inherited_interface in _get_interfaces(base):
+            interfaces.append(inherited_interface)
+
+    return interfaces
+
+
+def _get_fields(cls: Type) -> List["StrawberryField"]:
```
nrbnlulu, line 58: maybe the same as the new _eval_fields.

tests/fields/test_arguments.py
```diff
@@ -145,6 +145,8 @@ def name_resolver(
     class NameQuery:
         name: str = strawberry.field(resolver=name_resolver)
 
+    assert hasattr(NameQuery, "_type_definition")
```
nrbnlulu, line 148: replace with `get_type_definition`

strawberry/field.py
nrbnlulu: done.

nrbnlulu:
I tried to fix some pydantic tests though I don't really understand much there, it seems like they were relying on `copy_with` and it is needed to be migrated to `TemplateTypeDefinition.generate()`

strawberry/schema/name_converter.py
```diff
@@ -94,27 +92,42 @@ def from_union(self, union: StrawberryUnion) -> str:
...
-        for type_ in types:
-            name = self.get_from_type(type_)
-            names.append(name)
-
-        return "".join(names) + generic_type_name
+        """
+        generated types needed to be renamed, to avoid naming conflicts.
+        """
+        template = generic_type.concrete_of
+        assert template
```
nrbnlulu, line 110: Just assert  `generic_type.concrete_of`

tests/objects/generics/test_names.py
```diff
@@ -41,8 +41,8 @@ class TypeB:
         ([StrawberryUnion("Union", (TypeA, TypeB))], "UnionExample"),  # type: ignore
         ([TypeA], "TypeAExample"),
         ([CustomInt], "CustomIntExample"),
-        ([TypeA, TypeB], "TypeATypeBExample"),
-        ([TypeA, LazyType["TypeB", "test_names"]], "TypeATypeBExample"),  # type: ignore
+        ([TypeB], "TypeBExample"),
```
nrbnlulu, line 44: I am uncertain what were these params, this is invalid generation, this class takes only one generic we should probably provide an error if someone tries this.

tests/test_repr.py
```diff
@@ -12,8 +12,10 @@ class MyType:
         f: float
         id: strawberry.ID
 
+    instance = MyType("a", 1, True, 3.2, "123")
+    instance.__repr__()
```
nrbnlulu, line 16: revert this.

RELEASE.md
```diff
@@ -0,0 +1,7 @@
+Release type: patch
```
```suggestion
Release type: minor
```
I think this deserves a minor release 😊

strawberry/custom_scalar.py
```diff
@@ -31,42 +33,80 @@ def identity(x):
...
 
     # Optionally store the GraphQLScalarType instance so that we don't get
     # duplicates
     implementation: Optional[GraphQLScalarType] = None
 
-    def copy_with(
-        self, type_var_map: Mapping[TypeVar, Union[StrawberryType, type]]
-    ) -> Union[StrawberryType, type]:
-        return super().copy_with(type_var_map)
+    def to_graphql_core(self):
```
patrick91, line 52: can we keep this in the schema converter?

strawberry/arguments.py
```diff
@@ -152,9 +152,9 @@ def convert_argument(
     if isinstance(type_, LazyType):
         return convert_argument(value, type_.resolve_type(), scalar_registry, config)
 
-    if hasattr(type_, "_type_definition"):  # TODO: Replace with StrawberryInputObject
-        type_definition: TypeDefinition = type_._type_definition  # type: ignore
-
+    if type_definition := get_type_definition(
```
patrick91, line 155: we can't use python 3.8 syntax just yet :)

also can we make get_type_definition "private"? I don't we should expose it just yet to users 😊

strawberry/custom_scalar.py
```diff
@@ -31,42 +33,80 @@ def identity(x):
...
+            parsed = self.parse_value(value)
+            if self.serialize(parsed) == value:
+                return True
+        with contextlib.suppress(Exception):
+            serialized = self.serialize(value)
+            if self.parse_value(serialized) == value:
+                return True
+        return False
 
-class ScalarWrapper:
```
patrick91, line 58: why did you remove this? :)

strawberry/field.py
```diff
@@ -9,292 +8,305 @@
...
 UNRESOLVED = object()
 
 
-class StrawberryField(dataclasses.Field):
-    python_name: str
-    default_resolver: Callable[[Any, str], object] = getattr
-
-    def __init__(
+@dataclasses.dataclass()
+class StrawberryField(StrawberryType):
```
patrick91, line 52: I'm not sure this should extend StrawberryType /cc @BryceBeagle

patrick91 (commented):
thanks for putting so much work into this, I'm having troubles finding enough time to review this unfortunately, is there a way we can break it into multiple PRs? I've seen some changes that don't seem strictly related to the custom field work, it might be best to keep them separated

strawberry/custom_scalar.py
nrbnlulu: I'll have a try.

strawberry/arguments.py
nrbnlulu: Sure

strawberry/custom_scalar.py
nrbnlulu: check my comment: https://github.com/strawberry-graphql/strawberry/pull/2117#issuecomment-1231474766

> I removed ScallarWrapper as it made issued for me with the new validate method of StrawberryType, every scalar now is just a ScalarDefinition, I would consider renaming this to StrawberryScalar.

strawberry/field.py
nrbnlulu: This is also for `validate()` see https://github.com/strawberry-graphql/strawberry/pull/2117#issuecomment-1231474766

strawberry/custom_scalar.py
patrick91: is it needed for this PR?

nrbnlulu: Yes. it is how generic unions would be resolved here.

https://github.com/strawberry-graphql/strawberry/pull/2117/files#diff-111023ca6e4b982dfea8bea22a5eafe990b8639010fc30dbd7e515a96fc16fb2R120

patrick91:
@nrbnlulu ok, I'll reach out to you in the next few days 😊 thanks for making time to go through the code

tests/experimental/pydantic/test_conversion.py
```diff
@@ -844,52 +839,6 @@ class UserType:
     assert user.passwords == ["hunter2"]
 
 
-def test_sort_creation_fields():
```
nrbnlulu, line 847: I dropped the `DataclassCreationField` it is no longer needed with the new `StrawberryObejct` field creation hook.

tests/experimental/pydantic/test_conversion.py
```diff
@@ -946,32 +895,6 @@ class User(BaseModel):
             password: Optional[str] = Field(default=None, default_factory=lambda: None)
 
 
-def test_can_convert_pydantic_type_to_strawberry_with_additional_field_resolvers():
```
nrbnlulu, line 949: This test is not a valid strawberry behavior. strawberry doesn't create dataclasses fields for fields with resolvers. so I removed this for now. let me know what you think.

strawberry/experimental/pydantic/conversion.py
```diff
@@ -72,12 +72,9 @@ def convert_pydantic_model_to_strawberry_class(cls, *, model_instance=None, extr
             getattr(model_instance, python_name, None) if model_instance else None
         )
 
-        # only convert and add fields to kwargs if they are present in the `__init__`
-        # method of the class
-        if field.init:
```
nrbnlulu, line 77: test are not affected by this but this should probably be reverted

tests/schema/test_generics.py
```diff
@@ -135,6 +136,21 @@ class Query:
         def users(self) -> Connection[User]:
             return Connection(edge=Edge(node=User("Patrick")))
 
+    from strawberry.utils.typing import is_generic_subclass
+
+    class StrawberryGenericsGenerator:
```
nrbnlulu, line 141: (: I think thats a blooper

tests/schema/test_generics.py
```diff
@@ -492,6 +536,10 @@ class Edge(Generic[T]):
         cursor: strawberry.ID
         node: T
 
+    assert Edge[int]._type_definition.fields[1].type is int
```
nrbnlulu, line 539: can be removed.

tests/experimental/pydantic/test_conversion.py
nrbnlulu: @patrick91 do you approve this? also when #1187 is merged this is probably useless.

tests/experimental/pydantic/test_conversion.py
nrbnlulu: @patrick91 what do you think?

nrbnlulu:
@patrick91 Any updates on this one?

patrick91:
@nrbnlulu not yet sorry, I'm travelling for conferences so I don't have a lot of focused time 😊

PR #2128, opened 2022-08-25
Use dataclasses.MISSING for default value for default and default_factory

patrick91:
This PR changes the default for `default` and `default_factory` to `dataclasses.MISSING` to allow using `UNSET` as default when doing something like this:

```python
@strawberry.input
class FilterInput:
    in_: Optional[str] = strawberry.field(name="in", default=strawberry.UNSET)
```

I don't like the double checks we added for UNSERT and dataclasses.MISSING, maybe we can add a util function, or see if we can remove some of them.

I wanted to update the mypy plugin too, but it seem to trigger an unrelated error, so I'll avoid doing that for now :)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

strawberry/field.py
```diff
@@ -76,17 +75,11 @@ def __init__(
 
         # kw_only was added to python 3.10 and it is required
         if sys.version_info >= (3, 10):
-            kwargs["kw_only"] = dataclasses.MISSING
+            kwargs["kw_only"] = True
```
patrick91, line 78: @paulo-raca does this make sense to you? I think kw_only should be True by default, no?

strawberry/experimental/pydantic/object_type.py
```diff
@@ -216,13 +212,26 @@ def is_type_of(cls: Type, obj: Any, _info: GraphQLResolveInfo) -> bool:
...
             namespace["to_pydantic"] = cls.to_pydantic
 
+        kwargs: Dict[str, object] = {}
+
+        # Python 3.10 introduces the kw_only param. If we're on an older version
+        # then generate our own custom init function
+        if sys.version_info >= (3, 10):
+            kwargs["kw_only"] = True
+        else:
+            kwargs["init"] = False
```
patrick91, line 222: I'm not 100% sure we need this, I need to double check

strawberry/field.py
paulo-raca: Either works. MISSING` will inherit it its value from the @dataclass, which is also True

patrick91: ah, nice, maybe I'll should default to missing then

strawberry/experimental/pydantic/object_type.py
patrick91: looks like it is needed 😊 removing it will trigger this error: `TypeError: non-default argument 'other' follows default argument`

PR #2149, opened 2022-09-03
Dataloader's cache: support invalidation and importing

paulo-raca:
A big part of the power of Dataloaders comes from its internal cache.

This PR aims to take improve it with 2 new features:
- Data invalidation: After a mutations, it is important to remove old data from the cache so we can continue with up-to-date data.
- Data insertion: We can now insert data retrieved externally (Usually by a direct database call) into the cache, avoiding duplicate fetching between places that use the database directly and places that use the dataloader.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

## Checklist


botberry:



---

This release improves the dataloader class with new features:

- Explicitly cache invalidation, prevents old data from being fetched after a mutation
- Importing data into the cache, prevents unnecessary load calls if the data has already been fetched by other means.


---




Here's the tweet text: 
```

```

docs/guides/dataloaders.md
```diff
@@ -118,6 +118,62 @@ the list for each incorrect key. A call with `keys == [1, 3]` returns
...
+### Cache invalidation
+
+By default DataLoaders use an internal cache. It is great for performance, however it can cause problems when the data is modified
+(i.e., a mutation), as the cached data is no longer be valid! 😮
+
+To fix it, you can explicitly invalidate the data in the cache, using one of these ways:
+
+- Specifying a key with `loader.invalidate(id)`,
+- Specifying several keys with `loader.invalidate_many([id1, id2, id3, ...])`,
+- Invalidating the whole cache with `loader.invalidate_all()`
```
patrick91, line 130: should we add a note that this is the same a clear in dataloader js?

and same for preset being prime in dataloader js?

paulo-raca: Good point!

I really didn't think about keeping consistency between this and the JS api, but it sounds like a good idea

paulo-raca: I renamed the method to `clear_*` and `prime_*`

docs/guides/dataloaders.md
```diff
@@ -118,6 +118,62 @@ the list for each incorrect key. A call with `keys == [1, 3]` returns
 directly. If the `load_users` function raises an exception, even `load`s with an
 otherwise valid key, like `await loader.load(1)`, will raise that exception.
 
+### Cache invalidation
```
patrick91, line 121: doesn't look like the docs are updated with the new names 😊

patrick91: ah, it is missing docs about priming, I will add them

patrick91 (approved):
Nice one! Thank you 😊

paulo-raca:
Thanks for looking into it 👍

PR #2157, opened 2022-09-08
Add compatibility for dev builds of Mypy

skilkis:
## Description
Mypy dev builds are labeled as X.YYY+dev. Previously Decimal conversion was failing due to the added version suffix.

Note that `pkg_resources` is not used for version parsing as it is shipped together with `setuptools` and hence would not always be available.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/strawberry/ext/mypy_plugin.py
+++ b/strawberry/ext/mypy_plugin.py
@@ -1,4 +1,5 @@
 import re
+import warnings
 from decimal import Decimal
 from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union, cast
 
@@ -1016,5 +1017,8 @@ def plugin(version: str):
         MypyVersion.VERSION = Decimal(".".join(match.groups()))
     else:
         MypyVersion.VERSION = FALLBACK_VERSION
+        warnings.warn(
+            f"Mypy version {version} could not be parsed. Reverting to v0.800"
+        )
 
     return StrawberryPlugin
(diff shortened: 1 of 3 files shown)


strawberry/ext/mypy_plugin.py
```diff
@@ -1007,7 +1011,10 @@ def _is_strawberry_pydantic_decorator(self, fullname: str) -> bool:
 
 
 def plugin(version: str):
-    # Save the version to be used by the plugin.
-    MypyVersion.VERSION = Decimal(version)
+    match = VERSION_RE.match(version)
+    if match:
+        MypyVersion.VERSION = Decimal(".".join(match.groups()))
+    else:
+        MypyVersion.VERSION = FALLBACK_VERSION
```
patrick91, line 1018: is this safe to do? should we at least add a warning?

skilkis: Good suggestion, a warning would be nice, I'll add it!

Safety-wise, at worst-case certain features won't be correctly enabled for the required version, but `mypy` will still attempt checking the code, instead of directly failing 😊

patrick91: gotcha!

botberry:

```markdown
Release type: patch

```


Here's the tweet text: 
```

```

botberry:

```markdown
Release type: patch

```


Here's the tweet text: 
```

```

skilkis:
@patrick91 sorry for the barrage of pushes, it's been a long day so I made quite a few mistakes 🤣 PR is ready to review now!

patrick91 (approved):

PR #2158, opened 2022-09-09
Add new strawberry.lazy for better typing support for lazy types

patrick91:
This PR adds a new way to annotate types that can't be imported directly due to circular dependencies. This replaces `LazyType[Type, "module"]` and uses `Annotated` to make the code work with all type checkers, without any need for a plugin.

```python

import strawberry


if TYPE_CHECKING:
    from .type_b import TypeB


@strawberry.type
class TypeA:
    list_of_b: Annotated["TypeB", strawberry.lazy("path_to_module_b")]
```

Closes #1904

@bellini666 let me know if this makes sense to you 😊

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -148,7 +148,7 @@ markers = ["django", "starlette", "channels"]
 asyncio_mode = "auto"
 filterwarnings = [
     "ignore::DeprecationWarning:strawberry.*.resolver",
-    "ignore::DeprecationWarning:LazyType is deprecatedy",
+    "ignore:LazyType is deprecated:DeprecationWarning",
 ]
 
 [tool.autopub]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -147,7 +147,8 @@ testpaths = ["tests/"]
 markers = ["django", "starlette", "channels"]
 asyncio_mode = "auto"
 filterwarnings = [
-    "ignore::DeprecationWarning:strawberry.*.resolver"
+    "ignore::DeprecationWarning:strawberry.*.resolver",
+    "ignore::DeprecationWarning:LazyType is deprecatedy",
 ]
 
 [tool.autopub]
--- a/strawberry/annotation.py
+++ b/strawberry/annotation.py
@@ -8,7 +8,6 @@ from typing import (  # type: ignore[attr-defined]
     Dict,
     List,
     Optional,
-    Type,
     TypeVar,
     Union,
     _eval_type,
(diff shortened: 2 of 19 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

strawberry/annotation.py
```diff
@@ -63,14 +67,36 @@ def __eq__(self, other: object) -> bool:
...
+
+        if is_list(annotation):
+            return List[self._parse_annotated(get_args(annotation)[0])]
+
+        return annotation
+
     def resolve(self) -> Union[StrawberryType, type]:
-        annotation: object
+        annotation = self._parse_annotated(self.annotation)
+
```
patrick91, line 94: @BryceBeagle does this make sense to you? I'm recursively parsing the annotations that use `Annotated`

I'll try to add some unit tests later

pyproject.toml
```diff
@@ -147,7 +147,8 @@ testpaths = ["tests/"]
 markers = ["django", "starlette", "channels"]
 asyncio_mode = "auto"
 filterwarnings = [
-    "ignore::DeprecationWarning:strawberry.*.resolver"
+    "ignore::DeprecationWarning:strawberry.*.resolver",
+    "ignore::DeprecationWarning:LazyType is deprecatedy",
```
bellini666, line 151: Shouldn't it be `deprecated`?

strawberry/lazy_type.py
```diff
@@ -15,6 +16,15 @@ class LazyType(Generic[TypeName, Module]):
     package: Optional[str]
```
bellini666, line 16: I saw that you had to add `None` to a lot of places defining `LazyType`. Maybe `package` could be set to `None` by default to avoid that?

bellini666 (changes requested):
Awesome @patrick91 . Very nice solution to the problem we identified :)

Just 1 issue and 1 suggestion, the rest LGTM

pyproject.toml
patrick91: yes :D also I'm not sure it is actually working :D

strawberry/lazy_type.py
patrick91: yes, good suggestion!

bellini666 (approved):
LGTM :)

paulo-raca:
Nice solution!

However, I think `Annotated["User", strawberry.lazy(".users")]` looks more complicated than `LazyType("User", ".users").

Maybe we can modify `LazyType.__class_getitem__` to return `Annotated[type_name, StrawberryLazyReference(module)]` instead of deprecating it?

patrick91:
> Maybe we can modify `LazyType.__class_getitem__` to return `Annotated[type_name, StrawberryLazyReference(module)]` instead of deprecating it?

We could, but there's no way to make this supported in type checkers unfortunately :(

PR #2162, opened 2022-09-10
Replaced Text with str type

yasoob:
I wasn't sure whether Text was the right type as it doesn't exist in Python by default. Replaced that with `str` which might have been what the author meant. Please correct me if I am wrong :)


patrick91 (approved):
Good shout! I totally missed that!

PR #2174, opened 2022-09-15
fix enum handling of non-unique values or strawberry.enum_value

devkral:
We try to initialize an enum with the value of a value. This doesn't work (why it worked some time ago, I don't know).
This PR fixes the logic:
- first execute the checks of the underlying graphql enum
- if no error happened then return the value by using the proper enum access

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

Strange: no issue was opened? The code must have been broken for some time.
It fails for me for some versions.

Edit: see comments or heading.

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,2 @@
+Release type: patch
+Fix enum issues with strawberry.enum_value or non-unique values
--- a/strawberry/schema/schema_converter.py
+++ b/strawberry/schema/schema_converter.py
@@ -16,6 +16,7 @@ from typing import (
 )
 
 from graphql import (
+    EnumValueNode,
     GraphQLArgument,
     GraphQLDirective,
     GraphQLEnumType,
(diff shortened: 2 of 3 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

devkral:
there is some magic in use (at least with parse_literal). Please recheck my PR even the tests wok.

patrick91:
hi @devkral, can you add some tests for the broken behaviour? 

I'm unsure on how this was broken, I've used enums successfully with strawberry :)

devkral:
found the issue, the former code was partly correct:

enums allow the retrieval of values by wrapping them with enums.
This behavior works only correct if values are unique.
Also it fails if strawberry.enum_value is in use.

I modified the tests to demonstrate the issue.

devkral:
a nice sideeffect: by don't searching the values it should also improve the performance for bigger enums

devkral:
a nice sideeffect: it doesn't use the complicated code path of enums returning the value as a singleton. I think it will improve the speed at least a little bit.

devkral:
found a second issue:
EnumValueDefinition should not appear as a value

devkral:
sry, my test only works because of strawberry.enum_value. This is why the different keys with the same value can be differed. Actually you cannot differ keys with the same value.

the former code was correct, but the issue was at a different place: enums are not replaced

PR #2180, opened 2022-09-17
Fix exclude_unset issue when converting strawberry type to pydantic type

### LINKED ISSUE #2163: Pydantic "exclude_unset" dict option doesn't work on experimental strawberry type
When creating a Pydantic model you can create an instance of it without Optional arguments, and when exporting as a dictionary you can pass 'exclude_unset' to exclude optional fields from the return dict. https://pydantic-docs.helpmanual.io/usage/exporting_models/

```python
In [1]: from pydantic import BaseModel

In [2]: from typing import Optional
    class User(BaseModel):
        id: int
        name: str
        surname: Optional[str]

In [3]: user = User(id=1, name="Test")
In [4]: user.dict(exclude_unset=True)
Out[4]: {'id': 1, 'name': 'Test'}

```

## Feature Request Type

- [ ] Core functionality
- [x] New behavior

## Description
After creating a strawberry.experimental.pydantic.type which uses a pydantic model, you should be able to convert an instance of that type to_pydantic and export a dict the same way with 'exclude_unset' and return all fields that haven't been set but it returns all fields instead. 

Example:
```python
import strawberry
from typing import Optional


class User(BaseModel):
    id: int
    name: str
    surname: Optional[str]

@strawberry.experimental.pydantic.type(model=User, all_fields=True)
class UserType:
    pass

In [2]: user = UserType(id=1, name="Test")

```
Actual output:

```python
In [3]: user.to_pydantic().dict(exclude_unset=True)
Out[3]: {'id': 1, 'name': 'Test', 'surname': None}
```
Expected output:
```python
In [3]: user.to_pydantic().dict(exclude_unset=True)
Out[3]: {'id': 1, 'name': 'Test'}
```

na86421:
## Description
Fix exclude_unset issue when converting strawberry type to pydantic type

In the process of converting to pydantic type through `to_pydantic_default`, if Value is unset(== None) it is not put in `instance_kwargs`. 
So, make sure have the `__fields_set__` value properly.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* Fix #2163 

## Checklist

### DIFF
--- a/tests/experimental/pydantic/test_conversion.py
+++ b/tests/experimental/pydantic/test_conversion.py
@@ -1247,3 +1247,31 @@ def test_can_convert_optional_union_type_expression_fields_to_strawberry():
 
     assert test.optional_list == [1, 2, 3]
     assert test.optional_str is None
+
+
+def test_exclude_unset_parameter_when_convert_strawberry_type_to_pydantic_type():
+    class User(BaseModel):
+        id: int
+        name: str
+        surname: Optional[str]
+
+    @strawberry.experimental.pydantic.type(model=User, all_fields=True)
+    class UserType:
+        pass
+
+    user = User(id=1, name="leffe")
+    user_type = UserType(id=1, name="leffe")
+    user_type_pydantic = user_type.to_pydantic()
+
+    assert user.dict(exclude_unset=True) == user_type_pydantic.dict(exclude_unset=True)
+    assert user.json(exclude_unset=True) == user_type_pydantic.json(exclude_unset=True)
+
+    user_type_has_surname = UserType(id=1, name="leffe", surname="yoon")
+    user_type_has_surname_pydantic = user_type_has_surname.to_pydantic()
+
+    assert user.dict(exclude_unset=True) != user_type_has_surname_pydantic.dict(
+        exclude_unset=True
+    )
+    assert user.json(exclude_unset=True) != user_type_has_surname_pydantic.json(
+        exclude_unset=True
+    )
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,2 @@
+Release type: patch
+Fix exclude_unset issue when converting strawberry type to pydantic type
--- a/strawberry/experimental/pydantic/object_type.py
+++ b/strawberry/experimental/pydantic/object_type.py
@@ -251,6 +251,7 @@ def type(
                     getattr(self, f.name)
                 )
                 for f in dataclasses.fields(self)
+                if getattr(self, f.name) is not None
             }
             instance_kwargs.update(kwargs)
             return model(**instance_kwargs)


botberry:



---

Fix exclude_unset issue when converting strawberry type to pydantic type


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to JunKi Yoon for the PR 👏

```

tests/experimental/pydantic/test_conversion.py
```diff
@@ -1247,3 +1247,31 @@ class Test:
...
+        name: str
+        surname: Optional[str]
+
+    @strawberry.experimental.pydantic.type(model=User, all_fields=True)
+    class UserType:
+        pass
+
+    user = User(id=1, name="leffe")
+    user_type = UserType(id=1, name="leffe")
+    user_type_pydantic = user_type.to_pydantic()
```
thejaminator, line 1264: can we have a test where the `user_type = UserType(id=1, name="leffe", surname=None)`. And check that the None in the dict doesn't get removed?

thejaminator: and another one where the `surname: Optional[str] = "default"`. And don't set the surname, so it uses the default.

na86421: > can we have a test where the user_type = UserType(id=1, name="leffe", surname=None). And check that the None in the dict doesn't get removed?

You mean dict is `user_type.to_pydantic.dict()` ?  `not user_type.to_pydantic.dict(exclude_unset=True)` ?
If it's the former, It's testable and I'll add it.

thejaminator: > > can we have a test where the user_type = UserType(id=1, name="leffe", surname=None). And check that the None in the dict doesn't get removed?
> 
> You mean dict is `user_type.to_pydantic.dict()` ? not `user_type.to_pydantic.dict(exclude_unset=True)` ? If it's the former, It's testable and I'll add it.

no, this

```
user_type = UserType(id=1, name="leffe", surname=None)
user_type.to_pydantic.dict(exclude_unset=True)
```
Because now you are setting surname=None explicitly. So that is not unset. It is set. so the None should still be in the dict

na86421: Thank you, I understood.
In fact, It is a case that I was concerned about while patching, but the existing tests all worked well, so I did't worry about that.
But after seeing your comment, I think it should be considered. rigth?

thejaminator:
Yes, it has to be considered, i don't think the new tests  that i suggest
will pass. Because you are only checking for none :)

ughyg:
I took a look into this and came up with a proposal here PR https://github.com/strawberry-graphql/strawberry/pull/2631

PR #2185, opened 2022-09-20
WIP - Fix `enum_value`

patrick91:
This PR will fix an issue that doesn't allow you from using Enums that use `enum_value` for defining their value. The issue comes from `Enum(...)` trying to find the enum value using it's value, but when using `strawberry.enum_value` the value is the enum value definition.

We can fix this by changing the value or by adding an `__eq__` method, not sure which one is better. 

The PR has only the tests :)

### DIFF
--- a/tests/enums/test_enum.py
+++ b/tests/enums/test_enum.py
@@ -92,6 +92,10 @@ def test_can_deprecate_enum_values():
     assert definition.values[2].value == "chocolate"
     assert definition.values[2].deprecation_reason is None
 
+    assert IceCreamFlavour("strawberry") == IceCreamFlavour.STRAWBERRY
+    assert IceCreamFlavour("vanilla") == IceCreamFlavour.VANILLA
+    assert IceCreamFlavour("chocolate") == IceCreamFlavour.CHOCOLATE
+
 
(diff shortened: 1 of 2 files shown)


devkral:
I think there is a better solution: a subclass of an enum and using meta keyword arguments. Also with adapter function


would look like:

``` python
class Foo(strawberry.Enum, description="foo")
```
and for existing enum (metadata like description is added via function)


``` python
strawberry.Enum.adapt(enum, metadata_fn=foo)
```

devkral:
my solution would also wipe the EnumValueDefinition

devkral:
my idea is not working: creating a custom Enum requires much work as Enum subclassing is highly restricted

devkral:
I will upload a draft

devkral:
the limitation has been lifted and it seems that only some small issues remain

anyway: to remove the EnumValueDefinition you can define a an enum subclass by enum_value. this will be directly used.

Is there interest in my solution? Shall I continue working on it?

patrick91:
@devkral I'm really sorry I haven't answered you earlier! 

I think I found an easier solution in #2306, let me know what you think

PR #719, opened 2021-02-22
Add support for python 3.10

patrick91:
This PR add tests for python 3.10 and fixes a small issue when running Strawberry on python 3.10

PR #2189, opened 2022-09-22
Do not use `kw_only=False` when specifying strawberry.field()

paulo-raca:
The changes in #719 suggest that strawberry dataclasses should always use the `kw_only` mode. This is great, as it allows mixing the order of fields with/without default values, particularly when inheritance is involved.

However this falls apart when the fields without a default value is specified with `field: type = strawberry.field()`: Due to a minor mistake, this particular field is now marked with kw_only=False

Example:

```
# This Works
@strawberry.type
class FooBar1:
    foo: int = 1
    bar: int

# This fails with TypeError: non-default argument 'bar' follows default argument
@strawberry.type
class FooBar2:
    foo: int = strawberry.field(default=1)
    bar: int = strawberry.field()
```
## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist


botberry:



---

Bug fix: Do not force kw-only=False in fields specified with strawberry.field()


---




Here's the tweet text: 
```

```

patrick91 (approved):
Good find! Thank you!

PR #2194, opened 2022-09-22
Improve resolving performance by avoiding extra calls for basic fields

jkimbo:
## Description

This change improves performance of resolving a query by skipping `Info` creation and permission checking for fields that don't have a resolver or permission classes. In local benchmarks it improves performance of large results by ~14%.

Before:
CleanShot 2022-09-22 at 15 32 43@2x

After:
CleanShot 2022-09-22 at 15 32 23@2x

Benchmarked on a Macbook M1 Air.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/strawberry/schema/schema_converter.py
+++ b/strawberry/schema/schema_converter.py
@@ -411,6 +411,19 @@ class GraphQLCoreConverter:
     def from_resolver(
         self, field: StrawberryField
     ) -> Callable:  # TODO: Take StrawberryResolver
+        field.default_resolver = self.config.default_resolver  # type: ignore
+
+        if field.is_basic_field:
+
+            def _get_basic_result(_source: Any, *args, **kwargs):
+                # Call `get_result` without an info object or any args or
+                # kwargs because this is a basic field with no resolver.
+                return field.get_result(_source, info=None, args=[], kwargs={})
+
+            _get_basic_result._is_default = True  # type: ignore
+
+            return _get_basic_result
+
         def _get_arguments(
             source: Any,
             info: Info,
@@ -498,8 +511,6 @@ class GraphQLCoreConverter:
             await _check_permissions_async(_source, strawberry_info, kwargs)
 
             return await await_maybe(_get_result(_source, strawberry_info, **kwargs))
-
-        field.default_resolver = self.config.default_resolver  # type: ignore
 
         if field.is_async:
             _async_resolver._is_default = not field.base_resolver  # type: ignore
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,8 @@
+Release type: patch
+
+Improve resolving performance by avoiding extra calls for basic fields.
+
+This change improves performance of resolving a query by skipping `Info`
+creation and permission checking for fields that don't have a resolver
+or permission classes. In local benchmarks it improves performance of large
+results by ~14%.
--- a/strawberry/field.py
+++ b/strawberry/field.py
@@ -155,7 +155,7 @@ class StrawberryField(dataclasses.Field):
         return self
 
     def get_result(
-        self, source: Any, info: Info, args: List[Any], kwargs: Dict[str, Any]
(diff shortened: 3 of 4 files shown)


botberry:



---

Improve resolving performance by avoiding extra calls for simple fields.

This change improves performance of resolving a query by skipping `Info`
creation and permission checking for fields that don't have a resolver
or permission classes. In local benchmarks it improves performance of large
results by ~14%.


---




Here's the tweet text: 
```

```

strawberry/schema/schema_converter.py
```diff
@@ -411,6 +411,19 @@ def is_type_of(obj: Any, _info: GraphQLResolveInfo) -> bool:
     def from_resolver(
         self, field: StrawberryField
     ) -> Callable:  # TODO: Take StrawberryResolver
+        field.default_resolver = self.config.default_resolver  # type: ignore
```
patrick91, line 414: we were thinking of fetching the default resolver from `info.schema.config`, do you think it is possible to access the schema from info with this update?

patrick91:
thanks for this! do you have an example of the code you used for the benchmarking? I'll add it here: https://github.com/strawberry-graphql/benchmarks

botberry:
# Pre-release
:wave:

Releasing commit [0dae103c0b6ffa9b9f60ece70290e24d57bab179] to PyPi as pre-release! :package:

botberry:
# Pre-release
:wave:

Pre-release **0.131.2.dev.1663861889** [0dae103c0b6ffa9b9f60ece70290e24d57bab179] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.131.2.dev.1663861889
```

jkimbo:
> thanks for this! do you have an example of the code you used for the benchmarking? I'll add it here: https://github.com/strawberry-graphql/benchmarks

@patrick91 its the benchmarking code in the repo already: https://github.com/strawberry-graphql/strawberry/tree/main/tests/benchmarks

strawberry/schema/schema_converter.py
jkimbo: I don't understand, where are you trying to access `info.schema`?

patrick91: we aren't using info now, but I was thinking with @BryceBeagle to change `field.default_resolver` to be a property or something on field that gets the default resolver from `info.schema.config.default_resolver` so we don't have to set the default resolver here, which is a bit awkward

jkimbo: Right I see. I would just pass the default resolver to `get_result` on the field rather than accessing it through the `info` object. Creating the `info` object is just unnecessary for fields that don't use it.

jkimbo:
Alex doesn't seem to like `simple`. I could change it to `basic`?

strawberry/schema/schema_converter.py
patrick91: I think we can access it from the original info object, maybe I can attempt a PR after we merge this one 😊

strawberry/schema/schema_converter.py
```diff
@@ -411,6 +411,19 @@ def is_type_of(obj: Any, _info: GraphQLResolveInfo) -> bool:
...
+
+        if field.is_simple_field:
```
patrick91, line 416: what we did `if not field.has_resolver`?

otherwise we go with basic :)

jkimbo: It's not just having a resolver that determines this, also need to check if there are any permission classes (and potentially extensions in the future). 

I'll change it to basic.

patrick91: yes, my thought was that if you add a permission class the field would get a resolver automatically, but that's probably confusing ☺️

patrick91 (approved):
Thank you so much!

PR #2195, opened 2022-09-22
Support retrieving the current resolver's `Info` anywhere

paulo-raca:
Instead of adding the `Info` explicitly to the resolver's arguments, it is now possible to retrieve it anywhere with `strawberry.get_info()`.

While using the argument is fine if the info is used locally in the resolver, the new approach saves a lot of plumbing if the info is accessed in a separate function that is called everywhere.

This is particularly useful when acessing context information, like current user, database connection, etc: These are now available immediately anywhere in the code -- Which is why I also added a `get_context()` function

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

## Checklist


botberry:



---

Support retrieving the current resolver's `Info` without relying on arguments.

Instead of adding the `Info` explicitly to the resolver's arguments, it is now possible to retrieve it
anywhere with `strawberry.get_info()`.


---




Here's the tweet text: 
```

```

patrick91:
@paulo-raca sorry for the late reply on this 😊

I'll close since we settled for keeping info as parameter for resolvers, hope that's ok 😊

PR #1840, opened 2022-04-23
Create initial structure for unified HTTP tests

patrick91:
## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement/optimization
- [ ] Documentation

## Issues Fixed or Closed by This PR

*

## Checklist

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the CONTRIBUTING document.
- [ ] I have added tests to cover my changes.
- [ ] I have tested the changes and verified that they work and don't break anything (as well as I can manage).

PR #2213, opened 2022-09-29
Add json encoder settings to all integrations

patrick91:
This PR brings support for passing the following arguments to all our http integrations:

```
json_encoder: Type[json.JSONEncoder] = json.JSONEncoder,
json_dumps_params: Optional[JSONDumpsParams] = None,
```

so they are consistent between each other.

# TODO: 

- [x] Add docs

From #1840


botberry:



---

This release adds support for the following options in all our HTTP integrations:

- `json_encoder`: optional JSON encoder, defaults to `json.JSONEncoder`, will
  be used to serialize the data.
- `json_dumps_params`: optional dictionary of keyword arguments to pass to the
  `json.dumps` call used to generate the response. To get the most compact JSON
  representation, you should specify `{"separators": (",", ":")}`, defaults to
  `None`.


---




Here's the tweet text: 
```

```

patrick91:
I didn't add tests because they are in https://github.com/strawberry-graphql/strawberry/pull/1840 and I didn't want to duplicate work, hope that's ok :)

jkimbo (commented):
The code changes look fine but wouldn't it be cleaner to be able to pass a `response_serializer` function instead (which defaults to `json.dumps`)? That way you only need 1 extra argument. Or define a class method that could be overridden.

patrick91:
> The code changes look fine but wouldn't it be cleaner to be able to pass a `response_serializer` function instead (which defaults to `json.dumps`)? That way you only need 1 extra argument. Or define a class method that could be overridden.

Mmm, I don't remember why we went with passing the two arguments, passing a `response_serializer` might be a bit cumbersome if you want to pass custom params to the encoder (you'd need to make a custom function). Allowing to customise a method on the class could be nice though.

I'll try both options in this PR and see how it goes, thanks!

patrick91:
@jkimbo I might need a hand here, I like the idea of having one function `response_serializer` and/or a method on the class views, but I'm not 100% sure on how the should work.

The Django view uses `process_result` to get the result dictionary with data, errors and extensions, then it passes it to `_create_response` which creates a JSON response, with headers, status code and the serialised response.

Most GraphQL API will return JSON data, if you return something else you might also want to change the Content Type of the response, so maybe `response_serializer` might be not be the right choice since its name mean we could return anything, right? In this case I think having two options for the JSON encoder might be better, as it makes things more explicit. 

But I do think we should also have the method that allows to change the whole response, but maybe we should do that later, I'll have a PR that adds support for changing headers and status codes in all integrations and I think that will help (or at least change) how we implement this. 

What do you think?

jkimbo:
It could be called `json_serializer` then? I'm also not against passing these extra params, it just seems neater to pass a custom function with the signature `(data) -> str` to handle the whole json serialization step rather than a function and arguments separately.

patrick91:
> It could be called `json_serializer` then? I'm also not against passing these extra params, it just seems neater to pass a custom function with the signature `(data) -> str` to handle the whole json serialization step rather than a function and arguments separately.

mmh, maybe, I'll think about it a bit more 😊 also /cc @BryceBeagle for more opinions :)

I just checked where `json_dumps_params` comes from and it is from django's JSONResponse class

patrick91:
@jkimbo, I spoke with @marcoacierno and @BryceBeagle and they also agree with your suggestion, now I wonder how we can do implement this without breaking many users, we currently have the following arguments in django and sanic views:

```
json_encoder: Type[json.JSONEncoder] = json.JSONEncoder,
json_dumps_params: Optional[JSONDumpsParams] = None,
```

and we want to introduce a new argument that accepts a callable and returns a string, I like the name `json_encoder` but it is already used. Should we deprecate passing a class to it and only allow a callable?

Or should we find another name? `json_encoder` is also very similar to `json.JSONEncoder` so the name might be a bit misleading 😊

jkimbo:
> Or should we find another name? json_encoder is also very similar to json.JSONEncoder so the name might be a bit misleading 😊

@patrick91 `json_serializer`?

patrick91:
> > Or should we find another name? json_encoder is also very similar to json.JSONEncoder so the name might be a bit misleading 😊
> 
> @patrick91 `json_serializer`?

Just realised that anything sounding like a noun will always sound wrong:

```
json_serializer(data)
```

we can call it `serialize_json` or `encode_json`

```python
serialize_json(data)
```

I'll update the other PR. Thanks for the suggestion!

Issue #39, opened 2019-04-07
Investigate caching options and Apollo Cache Control


patrick91:
https://github.com/apollographql/apollo-cache-control [deprecated]
https://www.apollographql.com/docs/apollo-server/performance/caching/

Caching in GraphQL is not super easy, we might add some utilities for it :)

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/strawberry-graphql) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/strawberry-graphql/strawberry/issues/39">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/39/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/39/pledge.svg">
</picture>
</a>

vamshiaruru-virgodesigns:
Hi @patrick91 , are there any plans to implement this? If there are no plans for implementation, do you know of any strategies for this?
I imagine we have to set the cache-control headers before sending response back. Apollo-server does this by reading the @cacheControl directive. Is there a way for us to do the same using strawberry plugins? Maybe on the `on_request_end` hook? Can we read directives from this hook in `execution_context`?
I might not be able to get a clean way of doing this, but if I can get it done even in a rough way I'll be happy to share it :)

patrick91:
@vamshiaruru-virgodesigns adding cache control directives is easy, doing the caching in an extension is more difficult at the moment. 

We do have an example of validation cache here: https://strawberry.rocks/docs/extensions/validation-cache#validationcache this is not what you want, but it could be a start for an implementation :)

estyxx:
Hey I started digging into this...
As a first step maybe we can provide the CacheControl directive? 

I was thinking about an `apollo` namespace... but maybe there are better places :) 
https://github.com/strawberry-graphql/strawberry/compare/add-cache-control-directive?expand=1

patrick91:
I think apollo as a namespace might be fine, but I'm not sure if adding the directive is enough, as it can only be used when using federation (or any gateway that knows about this directive)

[...]

PR #2232, opened 2022-10-06
Add CacheControl directive

estyxx:
## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

#39 

## Checklist


tests/apollo/test_printer.py
```diff
@@ -0,0 +1,22 @@
...
+
+import strawberry
+from strawberry.apollo.schema_directives import CacheControl
+from strawberry.printer import print_schema
+
+
+def test_print_simple_cache_control():
+    @strawberry.type
+    class Query:
+        first_name: str = strawberry.field(directives=[CacheControl(max_age=10)])
```
estyxx, line 11: don't know if this test is necessary... Directives are tested somewhere else I think 🤔

strawberry/extensions/apollo_cache_control.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def on_request_start(self):
+        self.fields_caches: Dict[str, CachePolicy] = {}
+        self.overall_cache_policy = CachePolicy()
+
+    def on_executing_end(self):
+        if not self.calculate_http_headers:
+            return
+
+        self.execution_context.context["response"].headers[
```
estyxx, line 91: This should be tested for each integration... That means I should add ApolloCacheControlExtension to each schema... will mess up the tests I think, but I am not sure how to test this __in isolation__ otherwise...

botberry:



---

This release adds `ApolloCacheControl` extension that can be used to define cache control settings (`max_age` and `scope`) for each field in your schema.

```python
import strawberry
from strawberry.apollo.schema_directives import CacheControl, CacheControlScope
from strawberry.extensions import ApolloCacheControl

@strawberry.type
class Query:
    @strawberry.field(directives=[CacheControl(max_age=60, scope=CacheControlScope.PUBLIC)])
    def hello(
        self,
    ) -> str:
        return "🍓"

schema = strawberry.Schema(
    query=Query,
    extensions=[
        ApolloCacheControl(
            calculate_http_headers=False, default_max_age=10
        )
    ],
)
```


---




Here's the tweet text: 
```

```

omarzouk:
@estyxx Hello! I think the point of the cache directive is to tell the *api user* for how long they should be allowed to cache something. So this directive would inform apollo clients in the apps/web, you shouldn't cache this for longer than 30 minutes or some such.

So even if the server doesn't cache anything on its end, it is still valid to tell the clients, "you don't need to ask me again until 5 hours from now because my data will most likely not change" or "this data is valid for only x minutes".

I think this feature would be super useful to our organization

keithhackbarth:
@estyxx @patrick91 - Would it be possible to please re-evaluate this ticket?

See @omarzouk's comments above. There is a misunderstanding about what this PR does and how it can be used. With the Apollo client, the CacheControl directive tells the _client_ how often the field should be refreshed. Contrary to the explanation in the "reason for closing", it has nothing to do with a mechanism in Strawberry itself or anything server-side. This is only a directive for the client.

Apollo client is the leading client out there and this is a table-stakes feature in their recommended implementation starting with version 3.1 of the client. We appreciate all the maintainers. And we would appreciate a second look!

patrick91:
> See @omarzouk's comments above. There is a misunderstanding about what this PR does and how it can be used. With the Apollo client, the CacheControl directive tells the client how often the field should be refreshed. 

where did you read this?

PR #2239, opened 2022-10-10
Add documentation for testing with an API Client

estyxx:
I just relealised I never update the documentation for testing with an API Client. 
I think I was waiting to add the missing API Clients, I have to check if they require customizations... 
I was also waiting for [#1840](https://github.com/strawberry-graphql/strawberry/pull/1840) before adding more test on 
http... 
Anyway even if we don't have a custom `GraphQLClient` for all the integration, think is time to add it to the
documentation... and even if they can always use the `BaseGraphQLClient`... 

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/docs/operations/testing.md
+++ b/docs/operations/testing.md
@@ -8,7 +8,8 @@ The GraphiQL playground integrated with Strawberry available at
 [http://localhost:8000/graphql](http://localhost:8000/graphql) (if you run the schema
 with `strawberry server`) can be a good place to start testing your queries and
 mutations. However, at some point, while you are developing your application (or even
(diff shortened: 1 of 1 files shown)


patrick91:
thanks for this, I'm a bit afraid of merging it since there's a possibility of us changing the test client after I merge that PR. I'll see if I can work on it in the next few days and then then get back to this PR

patrick91:
Closing as it's old and we're thinking about extracting the test clients from the tests and making them first class

PR #2245, opened 2022-10-11
Fix support for arguments where `arg.type=LazyType["EnumType"]`

paulo-raca:
## Description

`convert_argument` is usually called with `type=EnumDefinition(...)`, which works fine.

However if the type is a lazy type, it is called initially with `type=LazyType["EnumType"]`, and then with `type=EnumType`.

Code was missing to convert the EnumType into an EnumDefinition

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/tests/schema/test_enum.py
+++ b/tests/schema/test_enum.py
@@ -5,7 +5,10 @@ from typing import List, Optional
 
 import pytest
 
+from typing_extensions import Annotated
+
 import strawberry
+from strawberry.lazy_type import lazy
 
 
 def test_enum_resolver():
@@ -98,6 +101,25 @@ def test_enum_arguments():
 
     assert not result.errors
     assert result.data["eatCone"] is True
+
+
+def test_lazy_enum_arguments():
+    LazyEnum = Annotated[
+        "LazyEnum", lazy("tests.schema.test_lazy_types.test_lazy_enums")
+    ]
+
+    @strawberry.type
+    class Query:
+        @strawberry.field
+        def something(self, enum: LazyEnum) -> LazyEnum:
+            return enum
+
+    schema = strawberry.Schema(query=Query)
+
+    query = "{ something(enum: BREAD) }"
+    result = schema.execute_sync(query)
+    assert not result.errors
+    assert result.data["something"] == "BREAD"
 
 
 def test_enum_falsy_values():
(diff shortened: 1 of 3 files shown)


botberry:



---

Fix support for arguments where `arg.type=LazyType["EnumType"]`


---




Here's the tweet text: 
```

```

tests/schema/test_enum.py
```diff
@@ -100,6 +110,26 @@ def eat_cone(self, input: ConeInput) -> bool:
     assert result.data["eatCone"] is True
 
 
+def test_lazy_enum_arguments():
+    @strawberry.type
+    class Query:
+        @strawberry.field
+        def something(self, enum: LazySharedEnum) -> LazySharedEnum:
```
patrick91, line 117: can you add a test for `Annotated[SharedEnum, strawberry.lazy()`?

we deprecated LazyType in favour of the new syntax https://strawberry.rocks/docs/types/lazy#lazy-types

paulo-raca: Actually that triggers a new bug :disappointed: 

This seems unrelated to the original bug and I'll leave it to a separate PR.

In summary:

When using `LazySharedEnum = Annotated["SharedEnum", lazy(__name__)]`:
- The argument type in the function signature is `Annotated[ForwardRef("SharedEnum"), lazy(__name__)]`
- `ReservedType.find()` calls `_eval_type()` and converts it into `Annotated[SharedEnum, lazy(__name__)]`
- `StrawberryArgument._parse_annotated()` calls `StrawberryLazyReference.resolve_forward_ref( SharedEnum )`. It crashes because it is not a ForwardRef

Also, I'm not sure what the proper fix would be:

- Should we ignore the current module and load it from the `lazy()` hint?
- Should it ignore the lazy() hint and load from the current module()?
- Should it be an error to apply `lazy()` annotation if the class can be resolved on the current module?
- Should it be a no-op to define `Annotated["ClassName", lazy("current_module")]` ?

paulo-raca: It works fine if the enum is actually defined in a different file, I've updated the test to use it

patrick91 (approved):
Thank you so much!

PR #1594, opened 2022-01-25
Fix compatibility with starlette 0.18.0

TimPansino:
## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement/optimization
- [ ] Documentation

## Issues Fixed or Closed by This PR

*

## Checklist

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the CONTRIBUTING document.
- [ ] I have added tests to cover my changes.
- [ ] I have tested the changes and verified that they work and don't break anything (as well as I can manage).

PR #2247, opened 2022-10-12
Use Tox for Matrix Testing

### LINKED ISSUE #1917: Isolate Integration Testing Environments
This issue is a follow up / postmortem on PR #1594

Considering this PR has been unmerged and blocked by dependency hell since January, and no one has been able to properly use strawberry with recent versions of Starlette, I think you should reconsider your current CI tooling. It can be quite frustrating for contributors and users alike when issues are unfixed because of workflow issues, not because no one has tried.

I would highly suggest you use independent environments for integration testing with external libraries. It would allow you to much more easily test these without having to worry about conflicting dependencies blocking features and patches.

The other issue I see is that you now have 2 distinct versions of starlette that you could continue to support, but without regression testing older versions are more likely to break and not be noticed. “Most” package maintainers try to do matrix based testing where they test against an array of versions of other packages, and against an array of python versions as well.

To that end, I would suggest you use [tox](https://tox.wiki/en/latest/) as it seems to be the tool of choice for package maintainers. It’s very simple to separate your tests into suites by folders, and have each test suite run in an individual environment with isolated dependencies. (Similar tools exist that I don’t have much experience with, see nox and thx)

As an example, we maintain a lengthy and complex tox.ini for [New Relic](https://github.com/newrelic/newrelic-python-agent) with a long comment explaining our naming conventions at the top. There are much simpler examples in the tox docs that are enough to try the tool. 

If you decide to use tox I’m happy to contribute or review designs, suggestions, and PRs.

TimPansino:
## Description

* Refactor testing to use `tox` for dispatching tests.
  * Added `tox` as a dev dependency in multiple places.
  * Convert existing matrix testing for `starlette` and `django` to `tox` matrixes.
* Reorganized `tests/` subfolders by dependencies required, using those as delineators for `tox` environments.
* Added pytest hook to set markers for each `tests/` subfolder (AKA `tox` environment), removing existing ones that were just for `starlette` and `django`.

The above changes SHOULD allow one to continue to use `poetry` based virtual environments for ease of local testing, but also allows the use of `tox` locally to test against multiple versions of frameworks or python quickly and easily. I highly recommend using [pyenv](https://github.com/pyenv/pyenv) to set up and maintain an array of python versions for use with tox if you want to test with it locally.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* Closes #1917

## Checklist


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Timothy Pansino for the PR 👏

```

TimPansino:
@patrick91 

On my fork I was able to run and pass tests, but I suspect there's still a code coverage issue. Likely each successive run replaces the existing coverage file, but as I don't have codecov setup on my end I was unable to confirm. There are solutions to that but I didn't want to implement without being able to debug by looking at codecov.

patrick91:
Hi @TimPansino thanks for much for starting this!

I have two questions:

1. would you be interested in trying out nox? That's the tool I wanted to try personally
2. can we keep the tests in the same folders they are now? I have a PR that removes a lot of them and I think it will be a lot of work to rebase that PR :)

patrick91:
Hi @TimPansino I went on and created #2275, I really wanted to check nox, plus I think I might need to a new a matrix for testing sanic (seems like there have been some breaking changes there)

patrick91:
I might close this PR if I'm happy with nox. Hope that's ok 😊

TimPansino:
If you'd rather use nox for creating venvs that's fine by me. I'm not personally a fan of nox, I like the more declarative style of tox and like to let it do the entire job of managing environments rather than mixing tools with things like poetry being run by nox. Not using a lock file based installer for testing is convenient for making sure you're testing with the most up to date versions of packages at all times to spot issues fast, and can easily pin older versions if you're blocked by a package update.

I would say you should definitely consider separating every major dependency you test against into a separate environment to ensure things are tested in isolation. To go with that, making every environment you want to matrix test into a separate folder in your tests folder lets you quickly and easily swap between testing with pytest/poetry by hand and using tox/nox for matrix testing. It's incredibly handy to have them logically separated.

If you'd prefer to use nox, might I suggest you swap the tox file for a nox file in this PR and keep the separated test folders? If you still have another PR that removes some tests that you'd like this rebased off of I can rebase it fairly easily after that's merged.

patrick91:
> If you'd rather use nox for creating venvs that's fine by me. I'm not personally a fan of nox, I like the more declarative style of tox and like to let it do the entire job of managing environments rather than mixing tools with things like poetry being run by nox. Not using a lock file based installer for testing is convenient for making sure you're testing with the most up to date versions of packages at all times to spot issues fast, and can easily pin older versions if you're blocked by a package update.

> I would say you should definitely consider separating every major dependency you test against into a separate environment to ensure things are tested in isolation. To go with that, making every environment you want to matrix test into a separate folder in your tests folder lets you quickly and easily swap between testing with pytest/poetry by hand and using tox/nox for matrix testing. It's incredibly handy to have them logically separated.

I'm not against this, but I have a PR that's going to refactor how the tests for the integration run, there will be only one test folder for all the integrations. I think it might be easier to isolate tests using markers for this specific case. What do you think?

> If you'd prefer to use nox, might I suggest you swap the tox file for a nox file in this PR and keep the separated test folders? If you still have another PR that removes some tests that you'd like this rebased off of I can rebase it fairly easily after that's merged.

I don't have a strong preference, I'm not familiar with either, I did like being able to use python for the configuration but other than that I don't know 😊 
Does tox allow to parallelise runs? At the moment the tests for django, starlette and other run in parallel since they are different actions, I'd like to keep that if possible

TimPansino:
> Does tox allow to parallelise runs? At the moment the tests for django, starlette and other run in parallel since they are different actions, I'd like to keep that if possible

Yes, the `-p auto` flag sets it to run with as many threads in parallel as possible based on your CPU count. (https://tox.wiki/en/latest/example/basic.html#parallel-mode)

Would recommend setting `TOX_PARALLEL_NO_SPINNER=1` for CI as it has a spinner by default that will spam your logs.

TimPansino:
I'm curious now if nox does parallelization, I didn't see anything in their docs. Do you happen to know?

patrick91:
> I'm curious now if nox does parallelization, I didn't see anything in their docs. Do you happen to know?

it does not :(

tox.ini
```diff
@@ -0,0 +1,100 @@
...
+    starlette: httpx
+    starlette: pygments
+    starlette: python-multipart
+    starlette: requests
+    starlette: uvicorn
+    starlette-starlettelatest: starlette
+    starlette-starlette001701: starlette==0.17.1
+    starlette-starlette001800: starlette==0.18.0
+    starlette-starlette001901: starlette==0.19.1
+    starlette-starlette002004: starlette==0.20.4
```
patrick91, line 81: this feels a lot more complicate than what it should be if I'm honest. 

I do get the benefits of not using a lock file, but I don't really like the idea of having to duplicate the dependencies here

TimPansino: You CAN use poetry somehow if you really want to manage these, but the point of using a test dispatcher like tox is to fully separate your concerns. It's much cleaner to specify exactly what dependencies are needed for testing with different suites by using a separate service.

If you have poetry install every dependency as a base and then just have tox manage the version matrixed dependencies that's not terrible, but it also sort of misses the point that having other dependencies can causes crashes and dependency hell when you try to reinstall with an older version of some package, if it shares a dependency with another package that's not under test but was included in the base poetry config.

A good example of problematic side effects would be FastAPI. They're frequently trailing starlette releases by a good margin. If you install everything with poetry first, FastAPI will likely cause starlette tests to be run with older versions of starlette and not the latest, whether or not you want it to be. That silent and unintended side effect would prevent you from testing on newer releases of starlette until FastAPI updates themselves. For libraries that are trying to guarantee compatibility with up to date versions of all their integrations that can easily hide critical failures for months. To get the most consistent and trustworthy results you want to test with only the dependencies you absolutely need for each integration.

Personally I don't think poetry is up to the task at all of managing any test dependencies of libraries. It's really good as a tool for managing the environment of applications, and for getting libraries published with minimal config. But if you're building a library and not an application, I don't think it's a very good tool for managing tests.

TimPansino: And actually thinking back, that was the exact problem with your testing environment that stalled #1594 in the first place prompting me to open an issue about this. If you use poetry to continue managing your base environment and include all your other dependencies in there like starlette and FastAPI you'll be only marginally more resilient and miss the original root cause.

patrick91: > You CAN use poetry somehow if you really want to manage these, but the point of using a test dispatcher like tox is to fully separate your concerns. It's much cleaner to specify exactly what dependencies are needed for testing with different suites by using a separate service.

would recommend removing the dev dependencies and always use tox then?

> Personally I don't think poetry is up to the task at all of managing any test dependencies of libraries. It's really good as a tool for managing the environment of applications, and for getting libraries published with minimal config. But if you're building a library and not an application, I don't think it's a very good tool for managing tests.

Ok, I think I'm getting convinced 😊 is it ok if I revert the test for the time being and use markers for dividing them? I don't know when I'll merge that other PR, and I want to keep the tests where they are for now (there's also quite a few PRs in flight)

TimPansino: That's fine, I would definitely recommend eventually going the 1 test suite per folder route but if it's easier you can always make that change later. Feel free to edit this PR in place and use it. I don't really have much more time to invest in actually writing code for at least a week.

TimPansino: > would recommend removing the dev dependencies and always use tox then?

I personally just use tox but it certainly takes a good 20-30 seconds sometimes to set up a venv before each test run. I have a bash alias for sourcing the tox env's created virtual environment after 1 run and editable installing the package into it. That being said, it's mostly because of how obnoxiously many tox environments we have (I think 500+). 

That being said, it might be easier for development workflows to have all those dev dependencies and to just use pytest directly. You can rely on your CI to handle the matrix testing and ensure nothing slips by, and focus on development with just any singular version of the framework you're testing against.

So really it just comes down to personal preference for local development. I can see the value either way.

PR #2248, opened 2022-10-12
Add MaskErrors extension

### LINKED ISSUE #1221: Allow to configure exception catching for schema execution

jkimbo:
## Description

Add a new `MaskErrors` extension that can be used to hide error messages from the client to prevent exposing sensitive details.

It also moves the `Schema.process_errors` call to happen before the `on_request_end` extension hook so that the unmasked errors are logged.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* Fixes #1221 

## Checklist

### DIFF
--- a/strawberry/extensions/mask_errors.py
+++ b/strawberry/extensions/mask_errors.py
@@ -1,20 +1,25 @@
-from typing import Sequence, Type
+from collections.abc import Callable
 
 from graphql.error import GraphQLError
 
 from strawberry.extensions import Extension
 
 
+def default_should_mask_error(_) -> bool:
+    # Mask all errors
+    return True
+
+
 class MaskErrors(Extension):
-    visible_errors: Sequence[Type[Exception]]
+    should_mask_error: Callable[[GraphQLError], bool]
     error_message: str
 
     def __init__(
         self,
-        visible_errors: Sequence[Type[Exception]],
+        should_mask_error: Callable[[GraphQLError], bool] = default_should_mask_error,
         error_message: str = "Unexpected error.",
     ):
-        self.visible_errors = visible_errors
+        self.should_mask_error = should_mask_error
         self.error_message = error_message
 
     def __call__(self, execution_context):
@@ -36,19 +41,9 @@ class MaskErrors(Extension):
         if result and result.errors:
             processed_errors = []
             for error in result.errors:
-                if not error.original_error:
-                    processed_errors.append(error)
+                if self.should_mask_error(error):
+                    processed_errors.append(self.anonymise_error(error))
                 else:
-                    original_error = error.original_error
-
-                    if not self.visible_errors:
-                        # anonymise all errors
-                        processed_errors.append(self.anonymise_error(error))
-                    else:
-                        for visible_error_cls in self.visible_errors:
-                            if not isinstance(original_error, visible_error_cls):
-                                processed_errors.append(self.anonymise_error(error))
-                            else:
-                                processed_errors.append(error)
+                    processed_errors.append(error)
 
             result.errors = processed_errors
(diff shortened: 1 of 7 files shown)


strawberry/extensions/mask_errors.py
```diff
@@ -0,0 +1,54 @@
...
+from strawberry.extensions import Extension
+
+
+class MaskErrors(Extension):
+    visible_errors: Sequence[Type[Exception]]
+    error_message: str
+
+    def __init__(
+        self,
+        visible_errors: Sequence[Type[Exception]],
```
estyxx, line 14: What about an `exclude_errors` ? 
Sometimes you want to exclude only a few errors. Would be easier if we just pass an "exclude" instead of mentioning all the other possible errors that can happen... 
what do you think?

jkimbo: I think passing an allow list is better than passing an exclude list because your exclude list is potentially unknowable and constantly expanding because any number of exceptions can be thrown. This way a new exception won't be shown to the client which I think is the right default.

I could change this API to accept a `show_show_error` predicate function instead? That way the developer could implement an allow list or an exclude list. It's a more complex API but it's more flexible.

jkimbo: @estyxx I've changed the API to use a `should_mask_error` function instead. I think thats more flexible. Let me know what you think.

estyxx: @patrick91 can we do a pre-release of this, please?

botberry:
# Pre-release
:wave:

Releasing commit [c1ad38d53339e50e5afa4e9b7463240cc0390256] to PyPi as pre-release! :package:

botberry:
# Pre-release
:wave:

Pre-release **0.134.0.dev.1665768624** [c1ad38d53339e50e5afa4e9b7463240cc0390256] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.134.0.dev.1665768624
```

estyxx (approved):
Yes think with the new `should_mask_error` it's more flexible! 
Just remember to update the `RELEASE.md` file ;)

strawberry/extensions/mask_errors.py
```diff
@@ -0,0 +1,49 @@
...
+        self,
+        should_mask_error: Callable[[GraphQLError], bool] = default_should_mask_error,
+        error_message: str = "Unexpected error.",
+    ):
+        self.should_mask_error = should_mask_error
+        self.error_message = error_message
+
+    def __call__(self, execution_context):
+        self.execution_context = execution_context
+        return self
```
patrick91, line 27: is this needed?

jkimbo: Nope, removed

patrick91 (approved):
Thank you! 💯

jerber:
@jkimbo what do you think about adding a status_code option too? So you can add a custom `error_message` and `status_code`, presumably defaulting to 500?

jerber:
> @jkimbo what do you think about adding a status_code option too? So you can add a custom `error_message` and `status_code`, presumably defaulting to 500?

I created an [issue](https://github.com/strawberry-graphql/strawberry/issues/2276) and a [pull request](https://github.com/strawberry-graphql/strawberry/pull/2277) for this feature.

Aponace:
This should definitely be added to the deployments docs in the "security extensions" section (https://strawberry.rocks/docs/operations/deployment).

Issue #1921, opened 2022-05-21
Generic types with scalars use the python type name.


coady:
## Describe the Bug
Generics types follow a naming convention of prepending the variable name to the type name. But when the variable is a builtin scalar, it's using the name of the python class, instead of the corresponding GraphQL scalar.

```python
@strawberry.type
class Output(Generic[T]):
    data: T

Output[str] # resolves to StrOutput instead of StringOutput
```
## System Information

0.113

Issue #2131, opened 2022-08-27
Rename generic types on the fly


benzolium:
It would be nice to rename generic types by annotating them.

E.g. with `typing.Annotated` like for arguments.

My use case is to annotate generic types inside the union:
```python
import strawberry
import typing as tp

GenericType = tp.TypeVar('GenericType')

@strawberry.type(name='SomeGeneric')
class SomeGenericType(
    tp.Generic[GenericType]
):
    value: GenericType

SomeType = strawberry.union(
    name='SomeTypeName',
    types=(
        tp.Annotated[SomeGenericType[int], strawberry.type(name='SomeGenericTypeInt')],
        tp.Annotated[SomeGenericType[float], strawberry.type(name='SomeGenericTypeFloat')],
    )
)

@strawberry.type
class Query:
    @strawberry.field
    def user(self) -> tp.Optional[SomeType]:
        return None

schema = strawberry.Schema(query=Query)
```

nrbnlulu:
May be resolved by #2117 Where generics are evaluated by `__getitem__(annotations)`
users would be able to do:
```py
SomeGenericTypeInt = SomeGenericType[int]
SomeGenericTypeInt._type_definition.graphql_name = "SomeGenericTypeInt"
```
or we would provide some helper function for renaming a type that will do this.

PR #2265, opened 2022-10-20
Generic scalars use the GraphQL scalar name instead of the Python class name.

coady:
## Description
Generic scalars prepend the GraphQL scalar name, instead of the Python class name, to the base type name.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #1921 

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,9 @@
+Release type: patch
+
+Generic scalars use the GraphQL scalar name instead of the Python class name.
+ ```python
+@strawberry.type
+class Output(Generic[T]):
+    data: T
+```
+`Output[str]` would be named `StringOutput` in the schema.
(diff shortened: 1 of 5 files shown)


botberry:



---

Generic scalars use the GraphQL scalar name instead of the Python class name.
 ```python
@strawberry.type
class Output(Generic[T]):
    data: T
```
`Output[str]` would be named `StringOutput` in the schema.


---




Here's the tweet text: 
```

```

patrick91:
mmh, I think this is a breaking change as it will change the schema. I wonder if we can find a way to make it easier to migrate (maybe a temporary configuration option?)

@coady what's the reasoning behind this?

coady:
> @coady what's the reasoning behind this?

Maybe it can be superseded by #2131: allowing a custom name. The reasoning is just that the GraphQL name is used in all scenarios - except a builtin scalar. I think it hasn't been perceived as a bug just because the names are similar:
* int -> Int
* float -> Float
* str -> Str _ing_
* bool -> Bool _ean_

coady:
#2291 has an alternate proposal.

PR #2266, opened 2022-10-20
Refactor chalice view

patrick91:
From: https://github.com/strawberry-graphql/strawberry/pull/1840

This PR does the following changes to the view for Chalice:

- deprecates render_graphiql
- adds graphiql as param
- adds allow_queries_via_get
- changes return codes to 404 (same as other integrations, also leaking less info)
- allows to change the statust_code via `info.context["response"].status_code = 418`

~it's draft because it needs tests, which are in #1840, but I think it is worth having them here too~


botberry:



---

This release refactors the chalice integration in order to keep it consistent with
the other integrations.

## Deprecation

Passing `render_graphiql` is now deprecated, please use `graphiql` instead.

## New features

- You can now return a custom status by using `info.context["response"].status_code = 418`
- You can enabled/disable queries via get using `allow_queries_via_get` (defaults to `True`)

## Changes

Trying to access /graphql via a browser and with `graphiql` set to `False` will return a 404.


---




Here's the tweet text: 
```

```

PR #2272, opened 2022-10-21
Change json_encoder to be a function, add it to all integrations

patrick91:
This PR supersedes #2213 by adding a json_encoder param to the HTTP views. I've implemented yet for the other integrations, because I want to get feedback on the Django integration first.

Currently due to how django class views work, we need to mark the default encoder function as a static method. Not sure if this is nice, but it allows to make this function a classmethod too.

@jkimbo, @marcoacierno, @BryceBeagle what do you think?

Closes #2213

### DIFF
--- a/strawberry/django/views.py
+++ b/strawberry/django/views.py
@@ -1,6 +1,7 @@
 import asyncio
 import json
-from typing import Any, Dict, Optional, Type
+import warnings
+from typing import Any, Dict, Optional
 
 from django.core.exceptions import BadRequest, SuspiciousOperation
 from django.core.serializers.json import DjangoJSONEncoder
@@ -23,6 +24,7 @@ from strawberry.http import (
     parse_request_data,
     process_result,
 )
+from strawberry.http.json_encoder import JsonEncoder, default_json_encoder
 from strawberry.schema.exceptions import InvalidOperationTypeError
 from strawberry.types import ExecutionResult
 from strawberry.types.graphql import OperationType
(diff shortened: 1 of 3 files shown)


strawberry/django/views.py
```diff
@@ -54,23 +56,60 @@ class BaseView(View):
     graphiql = True
     allow_queries_via_get = True
     schema: Optional[BaseSchema] = None
-    json_encoder: Type[json.JSONEncoder] = DjangoJSONEncoder
+    json_encoder: JsonEncoder = staticmethod(default_json_encoder)
```
patrick91, line 59: this is where we mark the function as a static method (1)

tests/django/test_views.py
```diff
@@ -437,22 +438,50 @@ def test_json_encoder():
...
     )
 
+    def fake_encoder(data: GraphQLHTTPResponse) -> str:
+        return "fake_encoder"
+
+    response1 = GraphQLView.as_view(schema=schema, json_encoder=fake_encoder)(request)
+    assert response1.content.decode() == "fake_encoder"
+
+    class CustomGraphQLView(GraphQLView):
+        json_encoder = staticmethod(fake_encoder)
```
patrick91, line 448: this is where we mark the function as a static method (2)

strawberry/django/views.py
```diff
@@ -23,6 +24,7 @@
     parse_request_data,
     process_result,
 )
+from strawberry.http.json_encoder import JsonEncoder, default_json_encoder
```
patrick91, line 27: I should probably change `JsonEncoder` to `JSONEncoder`, but note the name being the same as python's json.JSONEncoder (and DjangoJSONEncoder)

strawberry/http/json_encoder.py
```diff
@@ -0,0 +1,11 @@
+import json
+from typing import Callable
+
+from . import GraphQLHTTPResponse
+
+
+JsonEncoder = Callable[[GraphQLHTTPResponse], str]
```
jkimbo, line 7: How about prefixing this with `T` to indicate that it's a type annotation? `TJsonEncoder` (or `TJsonSerializer` if we change the name)

patrick91:
@jkimbo I'm thinking this should just become a method on the view class instead of being passed. It's not uncommon to subclass the view anyway. What do you think?

jkimbo:
@patrick91 that makes sense to me!

patrick91:
@jkimbo I finally updated this, let me know what you think before I do the rest of the integrations 😊

jkimbo (approved):
Looks good @patrick91

patrick91:
I've updated this PR to add `encode_json` to all the integrations

botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

botberry:



---

This release updates all the HTTP integration and adds a new method called
`encode_json` that allows to customize the encoding of the JSON response. By
default we use `json.dumps` but you can override this method to use a different
encoder.

It also deprecates `json_encoder` and `json_dumps_params` in the Django and
Sanic views, `encode_json` should be used instead.


---




Here's the tweet text: 
```

```

noamsto:
HI @patrick91,
I ran int an issue with serializing `datetime` objects in my Strawberry-Django project, and saw you have commits which address this issue, why didn't your merge them?

patrick91:
@noamsto can you show me the commit you're talking about? 😊

I think everything from here was merged, if you have an issue with datetimes can you open a new issue? I'll be happy to help 😊

noamsto:
Hi @patrick91,
I see that it was all merged,
and then reworked in this PR: #2681 
And the default `DjangoJSONEncoder` was removed in favor of letting the user override the `encode_json()` method.

Since it's Django related in-anycase, wouldn't it be best to override `encode_json()` for the Django views at least with the `DjangoJSONEncoder`?

Since datetime is widely used it means almost every project will need to overwrite this method, and  since it's Django project it might handle other Django related encodings which I can think of right now :)

It shouldn't impact anyone who overerode this function already.

What do you think?
Should I open an issue for that? 
I can create a PR as well for that

PR #2273, opened 2022-10-22
Refactor sanic view

patrick91:
This PR updates sanic's view to be consistent with the other views we have. From #1840 

This PR does adds the following **breaking changes**:

- `get_context` now receives a response parameter
- `process_result` is now **async** and receives a `request` parameter

Deprecation:

- `get_context` return type is now a dictionary, accessing its values via the dotted notation is deprecated[1][2]

New:

- Removed the upper bound version of sanic
- Errors are consistent with other integrations
- You can use the response value of context to change the status code of the response

Things to do:

- [x] Add release notes
- [ ] Add tests for v21 and v22 if it makes sense
- [x] Add deprecation on getattribute
- [x] Document the changes

[1] TODO: I should check if we allow to return a different dictionary (without mypy complaining)
[2] we should do this for all the integrations, will open an issue for this later


botberry:



---

This release updates the Sanic integration and includes some breaking changes.
You might need to update your code if you are customizing `get_context` or
`process_result`

`get_context` now receives the request as the first argument and the response as
the second argument.

`process_result` is now async and receives the request and the GraphQL execution
result.

This change is needed to align all the HTTP integrations and reduce the amount
of code needed to maintain. It also makes the errors consistent with other
integrations.

It also brings a new feature and it allows to customize the HTTP status code by
using `info.context["response"].status_code = YOUR_CODE`.

It also removes the upper bound on the Sanic version, so you can use the latest
version of Sanic with Strawberry.


---




Here's the tweet text: 
```

```

Issue #2264, opened 2022-10-19
Improve SelectedFields API


estyxx:
## Feature Request Type

- [ ] Core functionality
- [x] Alteration (enhancement/optimization) of existing feature(s)
- [ ] New behavior

## Description

Atm `Info.selected_fields` expose only the graphQL CamelCase of the fields, would be nice to expose the `python_name` in snake_case as well. 

Also `InlineFragment` doesn't have the `name` attribute like `SelectedField` and `FragmentSpread`. This makes a little difficult to iter: 
```python
        for field in info.selected_fields:
            if isinstance(field, SelectedField) or isinstance(field, FragmentSpread):
                print(field.name, camel_to_snake(field.name))
```

P. S. the [docs](https://strawberry.rocks/docs/types/resolvers#api) is not correct: it says only `SelectedField` instead is a union [`Union["SelectedField", "FragmentSpread", "InlineFragment"]`](https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/types/nodes.py#L32)...

See [Discord](https://discord.com/channels/689806334337482765/1032282710028328970/1032282710028328970) thread.

PR #2274, opened 2022-10-22
Add python_name to SelectedField (#2264)

estyxx:
## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/strawberry/types/nodes.py
+++ b/strawberry/types/nodes.py
@@ -25,6 +25,7 @@ from graphql.language import (
     ValueNode as GQLValueNode,
     VariableNode as GQLVariableNode,
 )
+from graphql.pyutils import camel_to_snake
 
 
 Arguments = Dict[str, Any]
@@ -84,6 +85,7 @@ class FragmentSpread:
     """Wrapper for a FragmentSpreadNode."""
 
     name: str
+    python_name: str
     type_condition: str
     directives: Directives
     selections: List[Selection]
@@ -95,6 +97,7 @@ class FragmentSpread:
         fragment = info.fragments[name]
         return cls(
             name=name,
+            python_name=camel_to_snake(node.name.value),
             directives=convert_directives(info, node.directives),
             type_condition=fragment.type_condition.name.value,
             selections=convert_selections(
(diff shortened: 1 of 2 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

strawberry/types/nodes.py
```diff
@@ -95,6 +97,7 @@ def from_node(cls, info: GraphQLResolveInfo, node: GQLFragmentSpreadNode):
         fragment = info.fragments[name]
         return cls(
             name=name,
+            python_name=camel_to_snake(node.name.value),
```
patrick91, line 100: I think this might not be the right way to go, since it doesn't reflect the configuration options.

We can have auto camel case off, which might not be an issue for this case, but we can also have a custom name converter.

Do you think we can find a way to get the strawberry type here and access the python name from there?

estyxx: Yeah, I was about to write something about... 
In this way is more a guess: everything is lowercase when for example the Child types can be Title Case... etc. 

So I'm digging in the Node object if there is something that can help us, but I didn't find much... 
I was thinking find the field from the schema:

```
schema: Schema = info.schema._strawberry_schema  # type: ignore
        field: StrawberryField = schema.get_field_for_type(  # type: ignore
            field_name=node.name.value,
            type_name=info.parent_type.name,
        )
```
But I'm not sure I have the right parent_type... 🤔

estyxx: yeah, I just missing the `parent_type`, if I had that I could get the field from the schema... but I don't know where to find it unfortunately :(

patrick91: `info.parent_type.extensions['strawberry-definition']` this should be what you need :)

estyxx: The thing is info is always the same, and is not changing between the nodes so even in `info.parent_type.extensions['strawberry-definition']` there is always the `Query` root type and one field... I would like to find the other fields... every node should have its field somewhere...  🤔 
right?

patrick91: ah, that's annoying, I'll take a look at it tomorrow and see how we can do this

patrick91:
Closing this as it has been a while, and it looks like no one needed this 😊

Issue #2276, opened 2022-10-23
Ability to dynamically set `status_code` on Response from `MaskErrors` extension


jerber:
## Feature Request Type

- [ ] Core functionality
- [ ] Alteration (enhancement/optimization) of existing feature(s)
- [x] New behavior

## Description

`MaskErrors` is great but I also want to be able to set the Response `status_code` dynamically based on the error.

For example, internal errors that I am now masking should return `500` status codes whereas visible errors should return `400` status codes. I would also like the ability to further customize the `status_code` on a per error basis, and it seems like adding a function hook to `MaskErrors` would be the easiest way to achieve this.

Ideally, `MaskErrors` could take in an optional function hook that reads in the error and can set the `status_code` according to custom logic.

jkimbo:
Going to close this issue for the reasons outlined in the corresponding PR

PR #2277, opened 2022-10-23
Added ability to dynamically update the Response `status_code` in the `MaskErrors` extension

jerber:
## Description

Add a new optional field `status_code_hook` to the `MaskErrors` extension to allow dynamically setting the Response `status_code`.

`status_code_hook` takes in an optional function of `Callable[[GraphQLError, ExecutionContext], None]` so you can update the Response, which you get from the `execution_context`, dynamically per error.

I'm using this in our app like so:

```

def should_mask_error(error: GraphQLError) -> bool:
    original_error = error.original_error
    if original_error and isinstance(
        original_error, (DisplayException, PermissionError)
    ):
        return False
    return True


def status_code_hook(error: GraphQLError, execution_context: ExecutionContext) -> None:
    if should_mask_error(error):
        execution_context.context.response.status_code = (
            settings.internal_server_error_status_code
        )
    else:
        status_code = settings.display_error_status_code
        original_error = error.original_error
        if original_error and isinstance(original_error, DisplayException):
            status_code = (
                original_error.status_code or settings.display_error_status_code
            )
        execution_context.context.response.status_code = status_code

MaskErrors(
        error_message="Internal Server Error",
        should_mask_error=should_mask_error,
        status_code_hook=status_code_hook,
)
```

I've also updated the docs to reflect this change.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* Fixes https://github.com/strawberry-graphql/strawberry/issues/2276

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,3 +147,5 @@ pre-commit install
 
 The code in this project is licensed under MIT license. See [LICENSE](./LICENSE)
 for more information.
+
+![Recent Activity](https://images.repography.com/0/strawberry-graphql/strawberry/recent-activity/d751713988987e9331980363e24189ce.svg)
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -12,11 +12,14 @@ class VisibleError(Exception):
     pass
 
 def status_code_hook(error: GraphQLError, execution_context: ExecutionContext) -> None:
+    response = execution_context.context['response']
+    existing_status_code = response.status_code
     if error.original_error and isinstance(error.orginal_error, VisibleError):
-        execution_context.context['response'].status_code = 400
+        new_status_code = 400
     else:
-        execution_context.context['response'].status_code = 500
-
+        new_status_code = 500
+    if new_status_code > existing_status_code:
+        response.status_code = new_status_code
 
 schema = strawberry.Schema(
     Query,
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,27 @@
+Release type: minor
+
+This release adds a new optional parameter `status_code_hook` to the `MaskErrors` extension that can be used to dynamically set the Response `status_code` based on the errors raised.
+
+```python
+import strawberry
+from strawberry.extensions import MaskErrors
+from graphql.error import GraphQLError
+from strawberry.types import ExecutionContext
+
+class VisibleError(Exception):
+    pass
+
+def status_code_hook(error: GraphQLError, execution_context: ExecutionContext) -> None:
+    if error.original_error and isinstance(error.orginal_error, VisibleError):
+        execution_context.context['response'].status_code = 400
+    else:
+        execution_context.context['response'].status_code = 500
+
+
+schema = strawberry.Schema(
+    Query,
+    extensions=[
+        MaskErrors(status_code_hook=status_code_hook),
+    ]
+)
+```
--- a/docs/extensions/mask-errors.md
+++ b/docs/extensions/mask-errors.md
@@ -111,9 +111,6 @@ from strawberry.extensions import MaskErrors
 from graphql.error import GraphQLError
 from strawberry.types import ExecutionContext
 
-VISIBLE_ERROR_STATUS_CODE = 400
-HIDDEN_ERROR_STATUS_CODE = 500
-
 class VisibleError(Exception):
     pass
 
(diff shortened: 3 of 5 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Jeremy Berman for the PR 👏

```

AndrewIngram:
How would conflicting status codes be managed here?

Say the response document contains multiple errors, and the first error handled sets a 500, and a later one sets something else, say a 412. Right now, the latter would win, even if the 500 is arguably more correct -- intuitively i'd expect the most "serious" error to win.

jerber:
> How would conflicting status codes be managed here?

Good point @AndrewIngram. I wanted to make this a function so that you could add this logic if you wanted. For example:

```python
import strawberry
from strawberry.extensions import MaskErrors
from graphql.error import GraphQLError

class VisibleError(Exception):
    pass

def status_code_hook(error: GraphQLError, execution_context: ExecutionContext) -> None:
    response = execution_context.context['response']
    existing_status_code = response.status_code
    if error.original_error and isinstance(error.orginal_error, VisibleError):
        new_status_code = 400
    else:
        new_status_code = 500
    if new_status_code > existing_status_code:
        response.status_code = new_status_code

schema = strawberry.Schema(
    Query,
    extensions=[
        MaskErrors(status_code_hook=status_code_hook),
    ]
)
```

I just updated the RELEASE.md and example in docs to reflect this.

patrick91:
> @jerber I'm not sure it makes sense to put this logic in the `MaskErrors` extension, I think it probably needs its own extension.
> 
> Also I would be careful about modifying the status code of the response based on any errors because it might contravene the GraphQL over HTTP spec: https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#status-codes . Because a GraphQL server can return partial data it should always return a 200 unless it cannot generate a well formed response.

+1 for this

@jerber what are the use cases for updating the status code?

jerber:
@jkimbo @patrick91 I agree, it should be in a different extension. Although maybe I'll just open source an example of how I'm changing the status codes since the whole thing is really just one function.

My main use case for updating status codes is for webhooks that use the status code to signal success or failure. For example, Stripe uses response status codes to know whether to retry a webhook.

jerber:
I created this extension which takes in the function `status_code_from_error` that returns an optional status code given the error. @patrick91 let me know if you think it's worth me creating a pull request for this, I can see how it can be used against the best practices of the GQL spec.

```
import typing as T
from strawberry.extensions import Extension
from graphql.error import GraphQLError


class UpdateStatusCode(Extension):
    """This must go before MaskErrors so that you get the `error.original_error`"""

    status_code_from_error: T.Callable[[GraphQLError], T.Optional[int]]

    def __init__(
        self, status_code_from_error: T.Callable[[GraphQLError], T.Optional[int]]
    ):
        self.status_code_from_error = status_code_from_error

    def on_request_end(self) -> None:
        result = self.execution_context.result
        if result and result.errors:
            response = self.execution_context.context.response
            max_status_code = response.status_code
            for error in result.errors:
                if error.original_error:
                    error_status_code = self.status_code_from_error(error)
                    if error_status_code is not None and error_status_code > (
                        max_status_code or 0
                    ):
                        max_status_code = error_status_code
            if max_status_code is not None:
                response.status_code = max_status_code
```

jkimbo:
> My main use case for updating status codes is for webhooks that use the status code to signal success or failure. For example, Stripe uses response status codes to know whether to retry a webhook.

@jerber I don't understand this use case. Are you saying that you've setup Stripe to send a GraphQL query to your API server?

jerber:
> Are you saying that you've setup Stripe to send a GraphQL query to your API server?

@jkimbo right. For example, Stripe sends a call to our server when a user adds a bank account. Stripe will retry this call if they don't receive a 200 response, so if there's some sort of internal error when processing this info, I want to return a non 200 so Stripe knows to retry later. I have this pattern for a few other services as well.

I can always just make Stripe hit a REST endpoint but I wanted to keep these calls in GraphQL land, so I found setting the status_code for certain exceptions was easiest.

AndrewIngram:
> I can always just make Stripe hit a REST endpoint but I wanted to keep these calls in GraphQL land, so I found setting the status_code for certain exceptions was easiest.

Personally i'd bypass GraphQL for this kind of thing, there's no real value in it here;  and as you've already seen, it presents avoidable challenges. There's no reason why GraphQL has to be the only interface between your system and the internet.

jerber:
> @jerber but how do you know that it's stripe that is calling your GraphQL endpoint? Presumably stripe is not making a GraphQL request it's just making a POST call with some data. 


Stripe sends a header that is decoded by a secret key the server has.

jerber:
@AndrewIngram yeah I ended up creating REST endpoints for this.

jkimbo:
Ok I'm going to close this issue now because it sounds like you've solved this issue in another way.

PR #2278, opened 2022-10-24
Make Websocket context_getter dependency async

kristjanvalur:
## Description

A `GrapQLRouter` has a `context_getter` dependency.  This is evaluated internally by another hidden dependency
to add default values onto the context object (or dict).  However, this internal depdenency wasn't defined as `async`
causing FastAPI to unnecessarily evaluate it on a worker thread, increasing overhead.  This change makes it async, causing it to be evaluated normally.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/release.md
+++ b/release.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+Internal GrapQLRouter FastAPI dependency is now async
--- a/release.md
+++ b/release.md
@@ -1,3 +1,0 @@
-Release type: patch
-
-Internal GrapQLRouter FastAPI dependency is now async
--- a/strawberry/fastapi/router.py
+++ b/strawberry/fastapi/router.py
@@ -54,7 +54,7 @@ class GraphQLRouter(APIRouter):
     def __get_context_getter(
         custom_getter: Callable[..., Optional[CustomContext]]
     ) -> Callable[..., CustomContext]:
-        def dependency(
+        async def dependency(
             custom_context: Optional[CustomContext],
             background_tasks: BackgroundTasks,
             request: Request = None,
--- a/strawberry/fastapi/router.py
+++ b/strawberry/fastapi/router.py
@@ -1,11 +1,21 @@
 import json
 from datetime import timedelta
 from inspect import signature
-from typing import Any, Awaitable, Callable, Dict, Iterable, Optional, Sequence, Union
+from typing import (
+    Any,
+    Awaitable,
+    Callable,
+    Dict,
+    Iterable,
+    Optional,
+    Sequence,
+    Union,
+    cast,
+)
 
 from starlette import status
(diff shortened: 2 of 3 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

patrick91:
@kristjanvalur thanks for this, can you update the type annotations?

Does this affect existing user code?

kristjanvalur:
No, it doesn't affect anything.  It only makes sure that the creation of the final `info.context` object happens without invoking a worker thread to do it.  All `FastAPI` dependencies really should be `async` functions **unless** they perform blocking IO.  I don't think FastAPI docs really explain this.  If they aren't `async`, FastAPI will shcedule them to run on a worker thread, which adds unnecessary latency.

release.md
```diff
@@ -0,0 +1,3 @@
+Release type: patch
+
+Internal GrapQLRouter FastAPI dependency is now async
```
patrick91, line 3:
```suggestion
Release type: patch

This release changes an internal implementation for FastAPI's
GraphQL router. This should reduce overhead when using the context,
and it shouldn't affect your code.
```

can you rename this file to RELEASE.md? our check is case sensitive :D

strawberry/fastapi/router.py
```diff
@@ -52,26 +62,28 @@ async def __get_root_value():
...
         ) -> MergedContext:
-            default_context = {
-                "request": request or ws,
-                "background_tasks": background_tasks,
-                "response": response,
-            }
+            request = cast(Union[Request, WebSocket], connection)
             if isinstance(custom_context, BaseContext):
-                custom_context.request = request or ws
+                custom_context.request = request
```
patrick91, line 77: mmh, any reason why we changed this?

kristjanvalur: Yes, see above.  There is a dependency, HTTPConnection, which provides one or the other, avoiding the overhead of having two separate dependencies and choose between them (FastAPI dependencies aren't completely free)

kristjanvalur: Besides, the parameter spec: `request: Request = None` would fail type checking..  Unfortunately, FastApi does not allow `Optional` or `Union` in its dependency type spec, but that is fine, because the only place you'd want to use that is precicely in this case, when getting either a `Request` or a `WebSocket`.

release.md
kristjanvalur: done.

kristjanvalur:
fyi, I have a PR open at FastAPI which aims to dramatically improve performace of dependency injection: https://github.com/tiangolo/fastapi/pull/5554

patrick91 (approved):
Thank you!

Issue #2292, opened 2022-10-30
Support new `Self` type as an annotation.


coady:
## Feature Request Type

- [ ] Core functionality
- [x] Alteration (enhancement/optimization) of existing feature(s)
- [ ] New behavior

## Description
Python 3.11 introduced [the `Self` type](https://docs.python.org/3/library/typing.html#typing.Self), allowing methods to explicitly annotate that they return the same type as the instance. It would especially help when an implementation has no need to override an interface field, other than to annotate it.

```python
from typing import TypeVar
import strawberry

@strawberry.interface
class INode:
    @strawberry.field
    def field(self) -> 'INode':
        ...
        return self

@strawberry.type
class Node(INode):
    @strawberry.field
    def field(self) -> 'Node':
        return super().field()

schema = strawberry.Schema(Node)
print(strawberry.printer.print_schema(schema))
```

```GraphQL
schema {
  query: Node
}

interface INode {
  field: INode!
}

type Node implements INode {
  field: Node!
}
```
`Self` would allow the implementation to inherit fields without further annotation.
```python
try:
    from typing import Self
except ImportError:
    Self = TypeVar('Self')

@strawberry.interface
class INode:
    @strawberry.field
    def field(self) -> Self:
        ...
        return self

@strawberry.type
class Node(INode):
    ...

schema = strawberry.Schema(Node)
print(strawberry.printer.print_schema(schema))
```

PR #2295, opened 2022-10-31
Support new Self type as an annotation.

coady:
## Description
`typing.Self` is a supported annotation, for types and interfaces.

```python
from strawberry.types import Self

@strawberry.type
class Node:
    @strawberry.field
    def field(self) -> Self:
        return self
```

## Types of Changes
- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* #2292

## Checklist

### DIFF
--- a/strawberry/types/types.py
+++ b/strawberry/types/types.py
@@ -14,6 +14,8 @@ from typing import (
     Union,
 )
 
+from typing_extensions import Self
+
 from strawberry.type import StrawberryType, StrawberryTypeVar
 from strawberry.utils.typing import is_generic as is_type_generic
 
@@ -43,6 +45,12 @@ class TypeDefinition(StrawberryType):
     type_var_map: Mapping[TypeVar, Union[StrawberryType, type]] = dataclasses.field(
         default_factory=dict
     )
+
+    def __post_init__(self):
+        # resolve `Self` annotation with the origin type
+        for index, field in enumerate(self.fields):
+            if isinstance(field.type, StrawberryType) and field.type.has_generic(Self):
+                self.fields[index] = field.copy_with({Self: self.origin})  # type: ignore
 
     # TODO: remove wrapped cls when we "merge" this with `StrawberryObject`
     def resolve_generic(self, wrapped_cls: type) -> type:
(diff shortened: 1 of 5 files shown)


botberry:

```markdown
Release type: patch

```


Here's the tweet text: 
```

```

strawberry/types/types.py
```diff
@@ -23,6 +23,11 @@
 
     from strawberry.field import StrawberryField
 
+try:
+    from typing import Self  # type: ignore
+except ImportError:
+    Self = TypeVar("Self")
```
patrick91, line 29: what about importing it from typing extensions?

strawberry/types/types.py
```diff
@@ -44,6 +46,11 @@ class TypeDefinition(StrawberryType):
         default_factory=dict
     )
 
+    def __post_init__(self):
+        for index, field in enumerate(self.fields):
+            if isinstance(field.type, StrawberryType) and field.type.has_generic(Self):
+                self.fields[index] = field.copy_with({Self: self.origin})  # type: ignore
```
patrick91, line 52: can you add a comment explaining why this is needed? 😊

tests/schema/test_generics.py
```diff
@@ -815,6 +817,37 @@ def books(self) -> List[Book]:
...
+      fields: [INode!]!
+    }
+
+    type Node implements INode {
+      field: Node!
+      fields: [Node!]!
+    }
+    """
+
+    assert str(schema) == textwrap.dedent(expected_schema).strip()
```
patrick91, line 848: can we also add a test making sure we can fetch fields? :)

patrick91 (commented):
This looks neat! 

What are your use cases for using Self?

coady:
> What are your use cases for using Self?

I think the general use case is that using `Self` will become standard in annotated Python going forward, so the expectation that it works with Strawberry will increase.

But my use case is pretty much the test example:
* an interface `Foo`, with field(s) `bar` which returns `Foo`
* implementations `*Foo` override `bar` to return their compatible `*Foo`
* the `Foo` resolver uses `type(self)`, so the implementations could just inherit and reuse its resolver
* except... the resolvers needs to wrapped or duplicated anyway because they have to register `*Foo` somehow

patrick91 (approved):
Nice work!

PR #2306, opened 2022-11-02
Fix issue with enum values

### LINKED ISSUE #2305: strawberry.enum_value breaks input parsing
## Describe the Bug

If `strawberry.enum_value` is used for an enum variant, parsing this variant as input of a query fails.

Consider the following example:
```python
from enum import Enum
import traceback

import strawberry

@strawberry.enum
class TestEnum(Enum):
    A = strawberry.enum_value("A")
    B = "B"

@strawberry.type
class Query:
    @strawberry.field
    def receive_enum(self, test: TestEnum) -> int:
        return 0

query = """
    {
        receiveEnum(test: A)
    }
""" # works if A is replaced by B
result = schema.execute_sync(query)

if result.errors:
    [error] = result.errors
    orig = error.original_error
    print(error)
    traceback.print_exception(type(orig), orig, orig.__traceback__)
```

Output:
```
Expected value of type 'TestEnum!', found A; 'A' is not a valid TestEnum

GraphQL request:3:27
2 |     {
3 |         receiveEnum(test: A)
  |                           ^
4 |     }
  File "~/.local/lib/python3.9/site-packages/graphql/validation/rules/values_of_correct_type.py", line 141, in is_valid_value_node
    parse_result = type_.parse_literal(node)
  File "~/.local/lib/python3.9/site-packages/strawberry/schema/schema_converter.py", line 89, in parse_literal
    return self.wrapped_cls(super().parse_literal(value_node, _variables))
  File "/usr/lib/python3.9/enum.py", line 360, in __call__
    return cls.__new__(cls, value)
  File "/usr/lib/python3.9/enum.py", line 677, in __new__
    raise ve_exc
ValueError: 'A' is not a valid TestEnum
```

## System Information

 - Strawberry version: 0.138.1
 - Python version: 3.9.2
 - Operating system: Debian 11

patrick91:
I'm not certain how we haven't noticed this earlier, but using `strawberry.enum_values` was broken for inputs.

This PR fixes it by adding the enum value **manually** to `cls._value2member_map_`, as all enum values defined using `strawberry.enum_value` were missing from that map. 

Not 100% sure this is the best solution, but it seems to work fine. @BryceBeagle @skilkis what do you think?

Closes #2305
Closes #2203
Closes #2185
Closes #2315


botberry:
# Pre-release
:wave:

Releasing commit [9e976d5570eafc146c1481a4233636c54eefabd0] to PyPi as pre-release! :package:

botberry:
# Pre-release
:wave:

Pre-release **0.138.2.dev.1667430260** [9e976d5570eafc146c1481a4233636c54eefabd0] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.138.2.dev.1667430260
```

botberry:



---

This release fixes an issue that prevented using enums that
were using strawberry.enum_value, like the following example:

```python
from enum import Enum
import strawberry

@strawberry.enum
class TestEnum(Enum):
    A = strawberry.enum_value("A")
    B = "B"

@strawberry.type
class Query:
    @strawberry.field
    def receive_enum(self, test: TestEnum) -> int:
        return 0

```


---




Here's the tweet text: 
```

```

strawberry/enum.py
```diff
@@ -102,6 +102,12 @@ def _process_enum(
             deprecation_reason = item_value.deprecation_reason
             item_value = item_value.value
 
+            # TODO: this check should always be true, but if it's not we should
+            # raise an error (???)
+
+            if item.name not in cls._value2member_map_:
```
patrick91, line 108: this might not be true when doing something like:

@strawberry.enum
class TestEnum(Enum):
    A = "A"
    B = strawberry.enum_value("B")
    D = strawberry.enum_value("B")

strawberry/schema/schema_converter.py
```diff
@@ -86,7 +86,9 @@ def parse_value(self, input_value: str) -> Any:
     def parse_literal(
         self, value_node: ValueNode, _variables: Optional[Dict[str, Any]] = None
     ) -> Any:
-        return self.wrapped_cls(super().parse_literal(value_node, _variables))
+        parsed_value = super().parse_literal(value_node, _variables)
+
+        return self.wrapped_cls(parsed_value)
```
patrick91, line 91: I'll revert this change

strawberry/enum.py
patrick91: CleanShot 2022-11-02 at 23 14 31@2x

devkral:
This solution has a major drawback. You cannot access the value by using: .value
You have to access the value by entering .value.value and the other attributes via:
.value.description.
It may would be better to extract the value and drop the EnumValueDefinition as the attributes are only used for the _enum_definition

patrick91:
@devkral can you show me a full snippet of the issue?

I'll see if I can find a solution that works for that case too :)

devkral:
you can simplify you solution this way:

``` python
@dataclasses.dataclass
class EnumValueDefinition:
    value: Any
    deprecation_reason: Optional[str] = None
    directives: Iterable[object] = ()
    description: Optional[str] = None

    def __eq__(self, __o: object) -> bool:
        if isinstance(__o, EnumValueDefinition):
            return self.value == __o.value
        else:
            return self.value == __o

    def __hash__(self) -> int:
        return hash(self.value)
```
But this has the mentioned issue.
@patrick91  I write a simple reproduction for you

devkral:
``` python

from enum import Enum
import strawberry

@strawberry.enum
class TestEnum(Enum):
    A = "A"
    B = strawberry.enum_value("B")
    C = strawberry.enum_value("Coconut", deprecation_reason="We ran out")

# should be True
TestEnum.B.value == "B"

# we create a value list, oops there are EnumValueDefinition in it
list(
    map(lambda x: x.value, TestEnum.__members__.values())
)


```

devkral:
sidenote: If you are the devil then add to my simplification this line:
``` python
    def __repr__(self) -> str:
        return repr(self.value)
```
People will swear, I promise  ;)

strawberry/enum.py
```diff
@@ -102,6 +102,12 @@ def _process_enum(
...
+                cls._value2member_map_[item_value] = item
```
patrick91, line 109: @devkral I think here we can use item.value, and that should fix your use case as well, will try it soon 😊

devkral: this should work. But it is still unclean (you access internals).

What speaks against my solution?

It is much simpler and doesn't changes the internals of Enum.


Hint: you can change the item value by changing the item._value_ attribute not by changing the item.value

patrick91: > this should work. But it is still unclean (you access internals) 

I think this is fine, I wanted to treat `strawberry.enum_value` similar to how enum.auto is treated.

> and you still have somewhere the references to the former Enum item (memory) floating around.

We need that for the type generation, no? :)

> What speaks against my solution?
> It is much simpler and doesn't changes the internals of Enum.

Your solution is good too, but I think I prefer making use of internals to completely hide the fact that we are using strawberry.enum_value. The enum should be a python enum without any additional things. Similar to what happens when we use strawberry.field :)

> Hint: you can change the item value by changing the item._value_ attribute not by changing the item.value

Oh man! Thanks so much for this, I didn't manage to find a proper solution so far! :D

jkimbo (approved):
I'm a bit uncomfortable with modifying internals like this but I don't have a better suggestion tbh.

patrick91:
> I'm a bit uncomfortable with modifying internals like this but I don't have a better suggestion tbh.

yeah, the other option I had in mind was to create a new enum, but seems a bit like a waste :) 

there's also @devkral solution: https://github.com/strawberry-graphql/strawberry/pull/2315/files#diff-36dede1973d46cbb72786377dc32e6fc8c90926ad1256ad06ffffcd8afa004c9

bit I think I prefer this one, even if it touches enum's internal more

jkimbo:
Oh actually I think I like #2315 more. You also don't need this line: https://github.com/strawberry-graphql/strawberry/pull/2315/files#diff-36dede1973d46cbb72786377dc32e6fc8c90926ad1256ad06ffffcd8afa004c9R113 so you don't need to touch enum internals at all. Why is this solution better than that?

patrick91:
> Why is this solution better than that?

Because we aren't exposing `EnumDefinitionValue` to the end user or on the enum itself

jkimbo:
> Because we aren't exposing EnumDefinitionValue to the end user or on the enum itself

Ok that makes sense 👍 . Happy to ship this change then.

patrick91:
@devkral I'll ship this change later today, thanks for much for your input and patience! Are you on discord, if yes, send me a message 😊

PR #2312, opened 2022-11-04
Prefer class annotation over resolver annotation when determining field type

jkimbo:
## Description

This PR changes the type resolution priority to prefer the field annotation over the resolver return type.

```python
def my_resolver() -> str:
    return "1.33"

@strawberry.type
class Query:
    a: float = strawberry.field(resolver=my_resolver)

# Before:
str(schema) == """
type Query {
  a: String!
}
"""

# After:
str(schema) == """
type Query {
  a: Float!
}
"""
```

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,28 @@
+Release type: minor
+
+This release changes the type resolution priority to prefer the field annotation over the resolver return type.
+
+```python
+def my_resolver() -> str:
+    return "1.33"
+
+@strawberry.type
+class Query:
+    a: float = strawberry.field(resolver=my_resolver)
+
+schema = strawberry.Schema(Query)
+
+# Before:
+str(schema) == """
+type Query {
+  a: String!
+}
+"""
+
+# After:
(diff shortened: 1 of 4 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

patrick91 (approved):
I wonder if we should add warnings in future when types don't match 🤔 it might be annoying when using duck typing.

Anyway, the PR looks good!

jkimbo:
> I wonder if we should add warnings in future when types don't match 🤔 it might be annoying when using duck typing.

@patrick91 I thought about it (and there is a comment to that effect here: https://github.com/strawberry-graphql/strawberry/blob/846f060a63cb568b3cdc0deb26c308a8d0718190/strawberry/object_type.py#L84-L87) but it's not that simple when you consider lazy types.

PR #2313, opened 2022-11-04
Add graphql_type parameter to `strawberry.field`

### LINKED ISSUE #2308: Duck typing / being allow to override the field types 🦆
I'm opening this issue to document how we can make it easier to override return types.

## Problem description

When using a framework like Django, you tend to directly return models in resolvers, for example:

```python
@strawberry.type
class Query:
    @strawberry.field
    def all_users(self) -> list[UserType]:
         return models.User.objects.all()
```

Assuming that `UserType` has a subset of models.User fields, this will work at runtime as we don't do a type check of the returned type[1].

Unfortunately this break type checking (mypy, pyright, etc) as the return type of `models.User.objects.all()` is `Queryset[models.User]` and the return type of `all_users` is `list[UserType]`.

## Ideas 

One way to work around this is the following:

```python
@strawberry.type
class Query:
    @strawberry.field
    def all_users(self) -> list[UserType]:
         return cast(list[UserType], models.User.objects.all())
```

---

@jkimbo suggested to allow doing something like this:

```python
@strawberry.type
class Query:
    @strawberry.field(field_type=list[UserType])
    def all_users(self) -> Queryset[models.User]:
        return models.User.objects.all()
```

and 

```python
@strawberry.type
class MyType:
  my_field: float = strawberry.field(field_type=str)

MyType(my_field=1.33) # this wouldn’t be a problem
```

[1] graphql-core has some type checks, but it is for scalars and nullability.

jkimbo:
## Description

This PR adds a new `graphql_type` parameter to `strawberry.field` that allows you to explicitly set the field type. This parameter will take preference over the resolver return type and the class field type.

For example:

```python
@strawberry.type
class Query:
    a: float = strawberry.field(graphql_type=str)
    b = strawberry.field(graphql_type=int)

    @strawberry.field(graphql_type=float)
    def c(self):
        return "3.4"

str(schema) == """
  type Query {
    a: String!
    b: Int!
    c: Float!
  }
"""
```

### Notes:

~~I'm open to alternative names to `field_type`. It doesn't look that great to have the duplication on `strawberry.field(field_type` but I didn't want to use just `type` since it's a built in function in Python.~~
Renamed to `graphql_type`.

### Implementation notes:

I noticed that we almost have this feature with the `type_annotation` parameter on `StrawberryField` so I just renamed it to `graphql_type` and exposed it in the decorator.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* Fixes #2308

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,28 @@
+Release type: minor
+
+This PR adds a new `graphql_type` parameter to strawberry.field that allows you
+to explicitly set the field type. This parameter will take preference over the
+resolver return type and the class field type.
+
+For example:
+
+```python
+@strawberry.type
+class Query:
+    a: float = strawberry.field(graphql_type=str)
+    b = strawberry.field(graphql_type=int)
+
+    @strawberry.field(graphql_type=float)
+    def c(self) -> str:
+        return "3.4"
+
+schema = strawberry.Schema(Query)
+
+str(schema) == """
+  type Query {
+    a: String!
+    b: Int!
+    c: Float!
+  }
+"""
+```
--- a/strawberry/experimental/pydantic/object_type.py
+++ b/strawberry/experimental/pydantic/object_type.py
@@ -57,7 +57,7 @@ def _build_dataclass_creation_fields(
     auto_fields_set: Set[str],
     use_pydantic_alias: bool,
 ) -> DataclassCreationFields:
-    graphql_type = (
+    field_type = (
         get_type_for_field(field, is_input)
         if field.name in auto_fields_set
         else existing_fields[field.name].type
@@ -77,7 +77,6 @@ def _build_dataclass_creation_fields(
             graphql_name = existing_field.graphql_name
         elif field.has_alias and use_pydantic_alias:
             graphql_name = field.alias
-
         strawberry_field = StrawberryField(
             python_name=field.name,
             graphql_name=graphql_name,
@@ -97,7 +96,7 @@ def _build_dataclass_creation_fields(
 
     return DataclassCreationFields(
         name=field.name,
-        graphql_type=graphql_type,
+        field_type=field_type,
         field=strawberry_field,
     )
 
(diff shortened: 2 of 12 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

patrick91:
I haven't seen the code yet, but I wonder about the name of the parameter, it might lead to some confusion, or not be clear that it is the GraphQL type. 

Maybe we should call it graphql_type? :)

jkimbo:
> Maybe we should call it graphql_type? :)

Yep, like that name more.

jkimbo:
@patrick91 updated. Also it would be helpful to get a pre-release of this so that I can check it against the Django integration.

jkimbo:
This should land first: https://github.com/strawberry-graphql/strawberry/pull/2319

patrick91:
Looks good! Do we need to update Strawberry Django when merging this?

patrick91 (approved):

jkimbo:
> Looks good! Do we need to update Strawberry Django when merging this?

Based on my track record probably! Can you create a pre release?

patrick91:
@jkimbo I forgot, we need a release file for the release bot to work 😊

jkimbo:
> I forgot, we need a release file for the release bot to work 😊

@patrick91 that makes sense! I've added a release file now. Should probably update the docs as well.

botberry:
# Pre-release
:wave:

Releasing commit [4305e718efb12e3e412c66e9bcc08f57b758c4b7] to PyPi as pre-release! :package:

botberry:
# Pre-release
:wave:

Pre-release **0.141.0.dev.1668163238** [4305e718efb12e3e412c66e9bcc08f57b758c4b7] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.141.0.dev.1668163238
```

jkimbo:
Looks like I haven't broken anything in strawberry-django 🥳

Issue #2285, opened 2022-10-26
Generic types are not supported as inputs.


coady:
## Describe the Bug
```python
from typing import Generic, TypeVar
import strawberry

T = TypeVar('T')

@strawberry.type
class Node(Generic[T]):
    @strawberry.field
    def data(self, arg: T) -> T:  # `arg` is also generic
        return arg

@strawberry.type
class Query:
    node: Node[int]

schema = strawberry.Schema(Query)  # raises TypeError
print(strawberry.printer.print_schema(schema))
```

## System Information

N/A

## Additional Context
If `arg: T` is replaced with a regular type, the generic output type works.

patrick91:
ah, interesting, I never thought of this use case! I wonder how long it could take to implement it though 🤔

skilkis:
@patrick91 I believe it's an easy fix, I had to recently implement this functionality for a use-case in my thesis. We would have to modify `TypeDefinition.copy_with`. I've currently implemented this in my use-case as a pre-processing step for root-types such as Query, Mutation, Subscription types before schema conversion. However, it makes more sense to implement this re-usably for any type in the schema by modifying `TypeDefinition.copy_with`. 

I'm happy to work on this in December. If anyone needs this urgently, the modification to `TypeDefinition.copy_with` would need one to iterate over the generic arguments in each field and resolve them individually.

```python
fields = []
for field in self.fields:
    ...
    # Resolve generic arguments
    for argument in field.arguments:
        if isinstance(argument.type, StrawberryType):
            if argument.type.is_generic:
                annotation = argument.type_annotation
                argument.type_annotation = StrawberryAnnotation(
                    annotation=argument.type.copy_with(type_var_map),
                    namespace=annotation.namespace,
                )
```

patrick91:
ah neat! thanks for the pointer 😊 feel free to work on it when you have time (and after resting from your thesis)

coady:
#2293 isn't working right, because it modifies the argument. The test should include 2 subscripts.

skilkis:
@coady you're indeed right, the snippet I posted above wasn't meant to be production ready yet. `StrawberryArgument` can implement a `copy_with` method to return a new argument to prevent side-effecting 😊

PR #2316, opened 2022-11-06
Fix for generic types to be used as argument annotations.

coady:
## Description
Field arguments are a property of resolver arguments, which are themselves a cached property. Moves the `copy_with` logic from the type definition to the resolver to avoid side-effects.

## Types of Changes
- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #2285

## Checklist

### DIFF
--- a/strawberry/type.py
+++ b/strawberry/type.py
@@ -82,7 +82,7 @@ class StrawberryContainer(StrawberryType):
     def copy_with(
         self, type_var_map: Mapping[TypeVar, Union[StrawberryType, type]]
     ) -> StrawberryType:
-        of_type_copy: Union[StrawberryType, type]
+        of_type_copy: Union[StrawberryType, type] = self.of_type
 
         # TODO: Obsolete with StrawberryObject
         if hasattr(self.of_type, "_type_definition"):
@@ -93,8 +93,6 @@ class StrawberryContainer(StrawberryType):
 
         elif isinstance(self.of_type, StrawberryType) and self.of_type.is_generic:
             of_type_copy = self.of_type.copy_with(type_var_map)
-
-        assert of_type_copy
 
         return type(self)(of_type_copy)
 
--- a/strawberry/types/fields/resolver.py
+++ b/strawberry/types/fields/resolver.py
@@ -294,16 +294,24 @@ class StrawberryResolver(Generic[T]):
         if self.type:
             if isinstance(self.type, StrawberryType):
                 type_override = self.type.copy_with(type_var_map)
-            else:
-                type_override = self.type._type_definition.copy_with(  # type: ignore
+            elif hasattr(self.type, "_type_definition"):
+                type_override = self.type._type_definition.copy_with(
                     type_var_map,
                 )
 
-        return type(self)(
+        other = type(self)(
             func=self.wrapped_func,
             description=self._description,
             type_override=type_override,
         )
+        # Resolve generic arguments
+        for argument in other.arguments:
+            if isinstance(argument.type, StrawberryType) and argument.type.is_generic:
+                argument.type_annotation = StrawberryAnnotation(
+                    annotation=argument.type.copy_with(type_var_map),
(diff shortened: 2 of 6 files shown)


botberry:



---

This release fixes support for generics in arguments, see the following example:

 ```python
 T = TypeVar('T')

 @strawberry.type
 class Node(Generic[T]):
    @strawberry.field
    def data(self, arg: T) -> T:  # `arg` is also generic
        return arg
 ```


---




Here's the tweet text: 
```

```

strawberry/type.py
```diff
@@ -91,8 +91,6 @@ def copy_with(
         elif isinstance(self.of_type, StrawberryType) and self.of_type.is_generic:
             of_type_copy = self.of_type.copy_with(type_var_map)
 
-        assert of_type_copy
-
```
patrick91, line 95: I think we can keep this, no?

strawberry/types/fields/resolver.py
```diff
@@ -294,16 +294,24 @@ def copy_with(
         if self.type:
             if isinstance(self.type, StrawberryType):
                 type_override = self.type.copy_with(type_var_map)
-            else:
-                type_override = self.type._type_definition.copy_with(  # type: ignore
+            elif hasattr(self.type, "_type_definition"):
+                type_override = self.type._type_definition.copy_with(
```
patrick91, line 298: do you think we can add coverage for this line? :)

strawberry/types/types.py
```diff
@@ -65,38 +65,11 @@ def resolve_generic(self, wrapped_cls: type) -> type:
...
-                argument.type_annotation = StrawberryAnnotation(
-                    annotation=argument.type.copy_with(type_var_map),
-                    namespace=argument.type_annotation.namespace,
-                )
-
-            fields.append(field)
+        # TODO: Logic unnecessary with StrawberryObject
+        # TODO: All types should end up being StrawberryTypes
+        #       The first check is here as a symptom of strawberry.ID being a
+        #       Scalar, but not a StrawberryType
```
patrick91, line 71: here these comments up to date now?

tests/schema/test_generics.py
```diff
@@ -751,24 +751,30 @@ def test_generic_argument():
...
-        def by_id(self, ids: List[T]) -> List[T]:
-            return []
+        def edge(self, arg: T) -> bool:
+            return bool(arg)
 
     @strawberry.type
     class Query:
-        user: Collection[int]
+        iNode: Node[int]
+        bNode: Node[bool]
```
patrick91, line 762:
```suggestion
        i_node: Node[int]
        b_node: Node[bool]
```

tests/schema/test_generics.py
```diff
@@ -751,24 +751,30 @@ def test_generic_argument():
     T = TypeVar("T")
 
     @strawberry.type
-    class Collection(Generic[T]):
+    class Node(Generic[T]):
         @strawberry.field
-        def by_id(self, ids: List[T]) -> List[T]:
-            return []
+        def edge(self, arg: T) -> bool:
+            return bool(arg)
```
patrick91, line 757: we should also add a test for List[T]

patrick91 (commented):
Added some comments!

strawberry/type.py
coady: We can, but it doesn't do much now that it's initialized. Oddly, the original version would raise an `UnboundLocalError`, which would happen on the next line anyway.

strawberry/types/fields/resolver.py
coady: It's not covered now, which make me wonder if it's necessary.

strawberry/types/types.py
coady: Looks like only the first one is still relevant.

patrick91 (approved):
Thanks @coady!

By the way, feel free to ping me when I take too long to review a PR!

PR #2321, opened 2022-11-07
docs(authentication): access authenticated user in resolvers

Meemaw:
## Description

Add documentation on best practices to access authenticated user information in a type safe way through context.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation

*

## Checklist


docs/guides/authentication.md
```diff
@@ -48,3 +48,69 @@ class Mutation:
...
+
+class User:
+    pass
+
+
+class ProgrammingError(Exception):
+    pass
+
+
+class Context(BaseContext):
```
Meemaw, line 77: Wondering if it makes sense to extend the `BaseContext` here, or just inline the fields, so they are not optional for the usage later.

docs/guides/authentication.md
```diff
@@ -48,3 +48,69 @@ class Mutation:
...
+from strawberry.types import Info as _Info
+from strawberry.types.info import RootValueType
+
+
+class User:
+    pass
+
+
+class ProgrammingError(Exception):
+    pass
```
jkimbo, line 74: Is this required? It doesn't seem to be used anywhere.

docs/guides/authentication.md
```diff
@@ -48,3 +48,69 @@ class Mutation:
 
         return LoginSuccess(user=User(username=username))
 ```
+
+### Access authenticated user in resolver
+
+Its fairly common to require user information within a resolver. We can do that in a type safe way with a custom context dataclass.
+
+For example, in fastapi this might look like this:
```
jkimbo, line 56:
```suggestion
For example, in FastAPI this might look like this:
```

docs/guides/authentication.md
```diff
@@ -48,3 +48,69 @@ class Mutation:
...
+
+import strawberry
+from fastapi import FastAPI, Request
+from strawberry.fastapi import BaseContext, GraphQLRouter
+from strawberry.types import Info as _Info
+from strawberry.types.info import RootValueType
+
+
+class User:
+    pass
```
jkimbo, line 70: Can you add a comment explaining that this is just a stub for an actual user object?

docs/guides/authentication.md
jkimbo: @Meemaw I would remove the `_request` property and just check for `.request` inside the `user` property function. I think that makes the example a bit simpler.

docs/guides/authentication.md
```diff
@@ -48,3 +48,69 @@ class Mutation:
...
+        return authorization_service.authorize(authorization)
+
+
+Info = _Info[Context, RootValueType]
+
+
+@strawberry.type
+class Query:
+    @strawberry.field
+    def get_authenticated_user(self, info: Info, id: strawberry.ID) -> User:
```
jkimbo, line 97: `User` needs to be a `strawberry` type here.

jkimbo (changes requested):
Thanks for the contribution @Meemaw . I've added some suggestions on how to improve the example code.

docs/guides/authentication.md
Meemaw: Its optional, so they we need to handle that there.

jkimbo: I get that but it in the context of a simple example to explain the concept it adds unnecessary redirection IMO.

I think the `user` property should look something like this:

```python
@property
def user(self) -> Optional[User]:
    if not self.request:
        return None

    authorization_key = self._request.headers.get("Authorization", None)
    return authorization_service.authorize(authorization_key)
```

By making the User return optional it handles the case where the request is unauthenticated as well.

Oh and it should **not** be a `cached_property` because that cache will persist across requests.

Meemaw: You might call this a lot of times in resolvers (especially within connections). You don't want to do the authorization logic each time. AFAIK `cached_property` is scoped to the current class instance, in this case `Context`  which gets recreated with every request.

jkimbo: > AFAIK cached_property is scoped to the current class instance, in this case Context  which gets recreated with every request.

My bad, you're right. cached_property makes sense here then.

docs/guides/authentication.md
```diff
@@ -48,3 +48,62 @@ class Mutation:
...
+Its fairly common to require user information within a resolver. We can do that in a type safe way with a custom context dataclass.
+
+For example, in FastAPI this might look like this:
+
+```python
+from functools import cached_property
+from typing import cast
+
+import strawberry
+from fastapi import FastAPI, Request
```
jkimbo, line 63:
```suggestion
from fastapi import FastAPI
```

jkimbo (changes requested):
Some final things to update 🙂 Thanks for taking on board my suggestions.

jkimbo (approved):

patrick91:
The invitation bot didn't work, so I'll post the message 😊

PR #2332, opened 2022-11-09
Auto generate a `resolve_reference` method for federation types

patrick91:
This PR introduces a new feature that should make it easier to use Apollo Federation with Strawberry.

I was writing an API and I found myself writing the reference resolver[1] manually all the time, Usually a reference resolver (in Strawberry) gets a dictionary and converts it to the Strawberry type, pretty much what we do for arguments. 

This PR adds automatic conversion when the reference resolver is not defined 😊

TODO:

- [x] Add a proper error message when the automatic conversion fail

I also think we can promote convert_arguments to a better place (and a better name, since it also supports this use case), but maybe we can do that in another PR :)

[1] https://www.apollographql.com/docs/federation/entities/#2-define-a-reference-resolver

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,4 +1,4 @@
-Release type: patch
+Release type: minor
 
 This release adds support for an implicit `resolve_reference` method
 on Federation type. This method will automatically create a Strawberry
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,25 @@
+Release type: patch
+
+This release adds support for an implicit `resolve_reference` method
+on Federation type. This method will automatically create a Strawberry
+instance for a federation type based on the input data received, for
+example, the following:
+
+```python
+@strawberry.federation.type(keys=["id"])
+class Something:
+    id: str
+
+@strawberry.federation.type(keys=["upc"])
+class Product:
+    upc: str
+    something: Something
+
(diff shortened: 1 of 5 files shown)


botberry:



---

This release adds support for an implicit `resolve_reference` method
on Federation type. This method will automatically create a Strawberry
instance for a federation type based on the input data received, for
example, the following:

```python
@strawberry.federation.type(keys=["id"])
class Something:
    id: str

@strawberry.federation.type(keys=["upc"])
class Product:
    upc: str
    something: Something

    @staticmethod
    def resolve_reference(**data):
        return Product(
            upc=data["upc"], something=Something(id=data["something_id"])
        )
```

doesn't need the resolve_reference method anymore.


---




Here's the tweet text: 
```

```

botberry:
# Pre-release
:wave:

Releasing commit [af9d97fb8d19c763f1fb81813949e606cd59144f] to PyPi as pre-release! :package:

botberry:
# Pre-release
:wave:

Pre-release **0.140.4.dev.1668092284** [af9d97fb8d19c763f1fb81813949e606cd59144f] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.140.4.dev.1668092284
```

PR #2335, opened 2022-11-11
Support documenting permissions on the schema

paulo-raca:
I've been using Strawberry's permission checks and I'm quite satisfied with it, however the permission information needs to be manually documented, which is error-prone.

This PR adds 2 different ways to expose this in the GraphQL schema: Using schema directives and using descriptions.

The resulting schemas look like this:

```
type Query {
  protectedField: String! @requiresPermissions(permissions: ["MyPermission"])
}
```

or

```
type Query {
  """
  Required permissions:
   - *MyPermission*: Doesn't have my permission
  """
  protectedField: String!
}
``'

Using directives seem like the best correct solution, but it is likely to be ignored by documentation tools


## Types of Changes

<!--- What types of changes does your pull request introduce? Put an `x` in all the boxes that apply. -->
- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

## Checklist

<!--- Go over all the following points, and put an `x` in all the boxes that apply. -->
<!--- If you're unsure about any of these, don't hesitate to ask. We're here to help! -->

### CODE REVIEW

**Comment by botberry** [comment]

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

DoctorJohn:
Thanks for the PR @paulo-raca, I appreciate the well written and tested code!

I'm a user of strawberry's permission checks myself and generally agree that automatic generation is less error prone than manually keeping things in sync. Both your approaches are probably the best ones, at least nothing else comes into my mind at the moment.

That being said I think the concept of generating descriptions hasn't been used anywhere else in strawberry yet, so I'm curious to hear what the rest of @strawberry-graphql/core thinks about it. Personally, I think it could be a nice feature (but a part of me wants to over-engineer description generation so we can use it in a consistent way from other parts of strawberry).

We also need to think about the proposed directive. Directives are nice, but I'm wondering how useful our opinionated directive (mainly referring to it's fixed name) would be 🤔

Don't get me wrong, I really like your changes. We just need to discuss them a bit since they introduce new "concepts" :)

PR #2341, opened 2022-11-14
Add test for closing ws twice

bellini666:
When an error happens inside the handler I see a lot of `RuntimeError: Unexpected ASGI message 'websocket.close', after sending 'websocket.close'.` being raised by uvicorn: https://sentry.2u.app.br/share/issue/83c133f3792c4305b305f0ce31c8ad0c/

This usually happens because the init timeout handler calls `close` when another `close` has already being called (due to an error)

Are there any bad side effects for doing this?


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

tests/websockets/test_graphql_transport_ws.py
```diff
@@ -156,6 +161,49 @@ async def test_connection_init_timeout_cancellation(
         assert ws.closed
 
 
+async def test_close_twice(
+    mocker: MockerFixture, request, http_client_class: Type[HttpClient]
+):
+    if sys.version_info < (3, 8):
+        pytest.skip("Task name was introduced in 3.8 and we need it for this test")
```
patrick91, line 168: can we move this to a marker on the test? 

```python
@pytest.mark.skipif(...)
async def test_close_twice(
```

tests/websockets/test_graphql_transport_ws.py
```diff
@@ -156,6 +161,49 @@ async def test_connection_init_timeout_cancellation(
...
+
+    if http_client_class == AioHttpClient:
+        pytest.skip(
+            "Closing a AIOHTTP WebSocket from a task currently doesnt work as expected"
```
patrick91, line 172: how come, what happens there?

tests/websockets/test_graphql_transport_ws.py
bellini666: Sure! Will do that

tests/websockets/test_graphql_transport_ws.py
bellini666: Not really sure, I copied that skip from other tests (https://github.com/strawberry-graphql/strawberry/blob/main/tests/websockets/test_graphql_transport_ws.py#L103) that also test the timeout assuming that my test would have the same issue

I'll try without it though

bellini666: I removed and it worked fine =P

bellini666:
@patrick91 it seems that the timeout PR also fixed this issue. I changed it to instead of doing what I was doing to instead just add a test to make sure it will not break in the future.

I also removed the `RELEASE.md` because I don't think a test is enough to justify a new release

kristjanvalur:
Yes, the timeout PR specifically made sure that there were no race conditions in closing the websocket connection.

PR #2349, opened 2022-11-16
add example of extend error response to document

nnashiki:
## Description

An example of error response expansion was added to the guide.
I formateed example codes with black.
I confirmed that the example works as expected with the latest version.
スクリーンショット 2022-11-16 23 11 30

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation

* Nothing

## Checklist


RELEASE.md
```diff
@@ -0,0 +1,4 @@
+Release type: patch
+
+This release updates the documentation.
+An example of error response expansion was added to the guide.
```
patrick91, line 4: I think we could skip the release for this documentation change, unless you like my idea below :D

docs/guides/custom-extensions.md
```diff
@@ -42,6 +42,60 @@ class MyExtension(Extension):
         print('GraphQL request end')
 ```
 
+<details>
+  <summary>Extend error response format</summary>
+
+```python
+import strawberry
+from graphql import GraphQLError
```
patrick91, line 50: I think it might be worth having our own `GraphQLError` extension, which can extend from `graphql.GraphQLError`

I think it might be a better experience to expose GraphQLError from the strawberry namespace and it will also be useful for the future (in case we migrate to something else)

Would you be willing of adding this exception? No worries if you don't have time :)

docs/guides/custom-extensions.md
```diff
@@ -42,6 +42,60 @@ class MyExtension(Extension):
...
+
+@strawberry.type
+class Query:
+    @strawberry.field
+    def my_field(self) -> str:
+        raise ExampleError("This error format has 'additional_key' in extensions key")
+
+    @strawberry.field
+    def my_second_field(self) -> str:
+        raise Exception("This error has normal format")
```
patrick91, line 91: Do you think we can also add an example of raising GraphQLError directly? I think that should work too

patrick91 (commented):
Thank you so much for doing this, I've added a few comments 😊

docs/guides/custom-extensions.md
nnashiki: @patrick91 thanks for Great product and the review. I will give it a try.

PR #2355, opened 2022-11-17
Define PartialType metaclass

cikay:
To prevent defining the same field for different types on the same model I defined a metaclass called PartialType that makes fields optional. 

Currently, we define the same fields in multiple types for the same model since the field might be optional for one type but not for the other.

```py
from typing import Optional


@strawberry.type
class UserCreate:
    firstname: str
    lastname: str

@strawberry.type
class UserUpdate:
    firstname: Optional[str] = strawberry.UNSET
    lastname: Optional[str] = strawberry.UNSET

@strawberry.type
class UserQuery(UserUpdate):
    id: Optional[strawberry.ID] = strawberry.UNSET

```
Imagine that `UserCreate`  has a bunch of fields. One has to define all of them in the `UserUpdate` type. With `PartialType` metaclass reusing is possible. 

```py
from strawberry.tools import PartialType


@strawberry.type
class UserCreate:
    firstname: str
    lastname: str

@strawberry.type
class UserUpdate(UserCreate, metaclass=PartialType):
    pass

@strawberry.type
class UserQuery(UserCreate, metaclass=PartialType):
     id: Optional[strawberry.ID] = strawberry.UNSET

```

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

*

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,17 @@
+Release type: minor
+
+Add `PartialType` metaclass to make fields optional dynamically.
+
+```py
+from strawberry.tools import PartialType
+
+@strawberry.type
+class UserCreate:
+    firstname: str
+    lastname: str
+
+@strawberry.type
+class UserUpdate(UserCreate, metaclass=PartialType):
+    pass
+
+```
--- a/docs/guides/tools.md
+++ b/docs/guides/tools.md
@@ -76,3 +76,29 @@ type ComboQuery {
   performA: String!
 }
 ```
+
+### `PartialType`
+
+`PartialType` metaclass is used to extend your type but makes its all field optional. Consider
+you have different types for each operation on the same model such as `UserCreate`, `UserUpdate`
+and `UserQuery`. `UserQuery` should have id field but the other does not. All fields of
+`UserQuery` and `UserUpdate` might be optional. In this case instead of defining the same field for
+each type one can define in a single type and extend it.
+
+```py
+from strawberry.tools import PartialType
+
+@strawberry.type
+class UserCreate:
+    firstname: str
+    lastname: str
+
+@strawberry.type
+class UserUpdate(UserCreate, metaclass=PartialType):
+    pass
+
+@strawberry.type
+class UserQuery(UserCreate, metaclass=PartialType):
+    id: Optional[strawberry.ID]
+
+```
--- a/strawberry/tools/__init__.py
+++ b/strawberry/tools/__init__.py
@@ -1,8 +1,10 @@
 from .create_type import create_type
 from .merge_types import merge_types
+from .partialtype import PartialType
 
 
 __all__ = [
     "create_type",
     "merge_types",
+    "PartialType",
 ]
--- a/strawberry/tools/partialtype.py
+++ b/strawberry/tools/partialtype.py
@@ -1,0 +1,30 @@
+from abc import ABCMeta
+from typing import Optional
+
+import strawberry
+
+
+class PartialType(ABCMeta):
+    def __new__(cls, name: str, bases: tuple, namespaces: dict, **kwargs):
+        mro = super().__new__(cls, name, bases, namespaces, **kwargs).mro()
+        annotations = namespaces.get("__annotations__", {})
+        fields: list[str] = []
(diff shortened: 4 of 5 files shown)


botberry:



---

Adds `AllOptional` metaclass to make fields optional dynamically.

```py
from strawberry.alloptional import AllOptional


@strawberry.type
class UserCreate:
    firstname: str
    lastname: str

@strawberry.type
class UserUpdate(UserCreate, metaclass=AllOptional):
    pass

```


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Muzaffer Cikay for the PR 👏

```

DoctorJohn (commented):
Thanks for the PR @cikay and especially for including tests! I first wasn't sure whether this change would be specific enough for strawberry to include it into the library. But looking at the metaclass' implemention, I noticed it kinda is. I would suggest moving the metaclass into `tools` module. Also curious what the rest of @strawberry-graphql/core thinks about this :)

strawberry/alloptional.py
```diff
@@ -0,0 +1,23 @@
...
+
+import strawberry
+
+
+class AllOptional(ABCMeta):
+    def __new__(cls, name: str, bases: tuple, namespaces: dict, **kwargs):
+        annotations = namespaces.get("__annotations__", {})
+        fields: list[str] = []
+        for base in bases:
+            annotations.update(base.__annotations__)
```
bellini666, line 12: I think this would need to use the MRO so that all the classes in the inheritance hierarchy are considered, and in the correct order.

The problem here is that, since the class has not been created yet, you can't access `klass.__mro__` yet. Maybe there's a way to construct that MRO given the `bases`?

strawberry/alloptional.py
```diff
@@ -0,0 +1,23 @@
...
+            fields.extend(field.name for field in base._type_definition._fields)
+        for field in annotations:
+            if not field.startswith("_"):
+                annotations[field] = Optional[annotations[field]]
+        namespaces["__annotations__"] = annotations
+
+        klass = super().__new__(cls, name, bases, namespaces, **kwargs)
+        for field in fields:
+            setattr(klass, field, strawberry.UNSET)
```
bellini666, line 21: Probably this needs to first check if the klass already has defined a default value for the field

bellini666 (changes requested):
Just like @DoctorJohn I also like this. It is similar to something strawberry-django does with [partial types](https://strawberry-graphql.github.io/strawberry-graphql-django/references/types/#input-types). With this change it could probably deprecate its use.

One thing that I'm worried is static typing support. Is there a way to indicate for static typers (e.g. pyright) that the attributes inside that class are now optional? For `mypy` we can probably do that in the plugin, but for pyright/pyre/etc that's not possible

strawberry/alloptional.py
bellini666: Exactly! Just remember to iterate over them in the reverse order, or else lower priority bases (the ones that come later in the MRO) will override the annotations from the higher priority ones

cikay:
| I would suggest moving the metaclass into tools module.

Done. @DoctorJohn

DoctorJohn (approved):
Looks good to me now, thanks for collaborating! Let's see what @bellini666 thinks :)

bellini666 (approved):
The code looks correct to me and I like the approach.

I would still like to have a way to workaround the fact that type checkers, such as pyright, will not understand that the variable is now optional. I'm not sure this is doable though :(

cikay:
@bellini666 The previous implementation overrides the created class annotations. I have fixed and added its tests.

cikay:
@bellini666 The previous implementation overrides the created class annotations. I have fixed and added its tests.

patrick91:
@jkimbo do you have opinions on this metaclass? :)

I think in future we might have a more type safe approach when there's the ability of doing things like this in Python's typing system. I believe it will look quite different from this (maybe it will look like [Typescript's Partial type](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype))

Also, based on that, perhaps we can rename the meta to PartialType?

cikay:
> Also, based on that, perhaps we can rename the meta to PartialType?
Done.

Issue #404, opened 2020-08-01
Raise an exception if creating a type with the same name


patrick91:
See https://github.com/strawberry-graphql/strawberry/pull/403/files#r463982907 for context.

Right now when creating a type we check if it is already in our type schema, if it is there we use it, but we don't check for duplicates

Kommandat:
Hey @patrick91, I'd like to start contributing to Strawberry and this looks like a good first issue to tackle. Any advice on how to get started and where to look?

My understanding is we would want to raise an exception if someone has duplicate types. A test could look like:

```
def test_raises_exception_with_duplicate_types():
    @strawberry.type
    class Query:
        title: str
        age: int

    @strawberry.type
    class Query:
        title: str
        age: int

    with pytest.raises(
        TypeError, match="Schema contains duplicate types named 'Book'"
    ):
        strawberry.Schema(query=Query)
```

BryceBeagle:
Hi @Kommandat! Thanks for your interest. There's a big PR refactor coming down the pipeline (#906) that's going to completely change how the internal typing system is handled. It's been a long time coming, but I'm aiming to have it done really soon. 

I'd hold off on this issue for now. There's a potential that PR ends up resolving this issue.

Kommandat:
Thanks @BryceBeagle! I'll hold off then. Let me know if there are any other issues you know of where I can lend a hand.

coady:
#2356 has caused a regression with generic types. Subscripted types are being counted as "duplicates".
```python
T = TypeVar("T")

@strawberry.type
class A(Generic[T]):
    a: T

@strawberry.type
class Query:
    first: A[int]
    second: A[int]

strawberry.Schema(Query)  # raises TypeError... Duplicated type name: IntA
```

patrick91:
@coady thanks! I'll take a look :)

PR #2027, opened 2022-07-18
Initial implementation of friendlier errors

patrick91:
I'll add a proper description later, this PR is just for testing the CI steps.

Related: #2107

TODO:

- [x] assert that documentation exists (via pytest)
- [x] make all of this optional (flag and checking for imports)
- [x] Don't fail bad if ast parsing fails (for example with dynamic code)
- [x] don't add libcst as main dependency
- [x] print standard exception if we can't find the source (?)
- [x] redirect from errors.strawberry.rocks
- [x] add group to allow installing everything (strawberry-graphql[cli] should ok)

Future release:
- [ ] Errors in wasm (if libcst works)

PR #2356, opened 2022-11-19
Added validation for type name uniqueness in a schema

bartekbp:
## Description
Today, it's possible to unintentionally define a schema that uses the same type name twice. You can have two modules with the following:
```python
# query.py
@strawberry.type
class User:
    name: str


@strawberry.type
class Query:
    viewer: User
    management: Management
```

```python
# management.py
@strawberry.type
class User:
    firstname: str
    lastname: str
    created_at: int

@strawberry.type
class Management:
    users: list[User]
```

The resulting schema will contain only one type to represent a user, thus leading to a conflict - one type will shadow the other one.

This pr introduces a validation ensuring that if type with the same name is duplicated, its definition must be the same.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* #404 

## Checklist


botberry:



---

Added extra validation that types used in a schema are unique.
Strawberry starts to throw an exception `DuplicatedTypeName` when two types defined in a schema have the same name.


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Bartosz Polnik for the PR 👏

```

strawberry/schema/schema_converter.py
```diff
@@ -653,3 +664,11 @@ def is_type_of(obj: Any, _info: GraphQLResolveInfo) -> bool:
             return is_type_of
 
         return None
+
+    def validate_same_type_definition(
+        self, name: str, type_definition: StrawberryType, cached_type: ConcreteType
+    ) -> None:
+        if cached_type.definition != type_definition:
+            print(cached_type.definition)
+            print(type_definition)
```
DoctorJohn, line 673: We probably don't want to keep those prints. If the information they're printing is relevant they should be part of the exception :) What do you think?

```suggestion
```

strawberry/exceptions.py
```diff
@@ -246,3 +246,14 @@ def __init__(self):
             "that inherits from BaseContext or a dictionary"
         )
         super().__init__(message)
+
+
+class DuplicatedTypeName(Exception):
```
DoctorJohn, line 251: @strawberry-graphql/core This is not specific to this PR, but I'm wondering whether we eventually want all our errors to extend a common base exception, so that strawberry errors are easier identifiable and catch-able? (or is this even considered an anti-pattern, since users would be tempted to catch the base strawberry exception instead of more specific ones?)

DoctorJohn (changes requested):
Thanks for the PR @bartekbp! Overall I like these changes, makes strawberry even more polished. I added one suggestion and annoyed the whole core team by inviting them to discuss something not even specific to this PR lol

strawberry/exceptions.py
patrick91: yes, this is the plan, I think we can leave this PR as it is, and I'll update #2027 when this one is merged :)

DoctorJohn: Ah, I see. Thanks Patrick!

strawberry/schema/schema_converter.py
bartekbp: Indeed, this was just for debugging. I'll remove it:)

DoctorJohn (approved):
LGTM :)
(except for the repeating pattern, but there's probably not much we can do about it)

patrick91 (approved):
This is great! Thank so much! <3

patrick91:
@bartekbp thanks for contributing to Strawberry! I'll merge this PR once it's green 🎉 Also, I've invited you to join the Strawberry GraphQL organisation 😊

PR #2378, opened 2022-12-02
Fix issues with duplicated generics

### LINKED ISSUE #404: Raise an exception if creating a type with the same name

patrick91:
Thanks @coady for the [bug report](https://github.com/strawberry-graphql/strawberry/issues/404#issuecomment-1335094562)! 

I wonder if we can change the check in future to be done inside type definition, but I'll leave that for the future :)

Closes #404

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,4 @@
+Release type: patch
+
+This release fixes a regression that prevented Generic types
+from being used multiple types.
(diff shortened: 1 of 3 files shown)


botberry:



---

This release fixes a regression that prevented Generic types
from being used multiple types.


---




Here's the tweet text: 
```

```

Issue #2351, opened 2022-11-16
context on Subscription Info object does not bring custom headers from GraphiQL


krauss:
## Describe the Bug

Custom headers sent through GraphiQL interface don't come up within the Info object (`info.context['request'].headers.items()`) of a Subscription. However, they do come up when you run a Query. Am I missing something here or is this really a bug?

## System Information

 - Operating system: Fedora Linux 36
 - Strawberry version (if applicable): 0.142.2
 - FastAPI version: 0.87.0
 - Starlette version: 0.21.0
 - Uvicorn version: 0.19.0

## Additional Context

The chunk of code below can be used to reproduce the scenario.

```python
# std lib imports
import asyncio
from typing import AsyncGenerator

# 3rd imports
import strawberry
from fastapi import FastAPI
from strawberry.fastapi import GraphQLRouter
from strawberry.schema.config import StrawberryConfig
from strawberry.types import Info

#-----------------------------------------------------------------------------
@strawberry.type
class Subscription:

    @strawberry.subscription
    async def count(self, info: Info, target: int = 100) -> AsyncGenerator[str, None]:
        for i in range(target):
            yield f"Counting {i} - headers: {info.context['request'].headers.items()}"
            await asyncio.sleep(0.5)


@strawberry.type
class Query:

    @strawberry.field
    async def say_hi(self, info: Info, name: str) -> str:
        return f"Hi {name}, here are your custom headers: {info.context['request'].headers.items()}"

#-----------------------------------------------------------------------------
schema = strawberry.Schema(
    query=Query, 
    subscription=Subscription, 
    config=StrawberryConfig(auto_camel_case=False))
graphql_api = GraphQLRouter(schema)

app = FastAPI(title='UpData - GraphQL API')
app.include_router(graphql_api, prefix='/graphql')
app.add_websocket_route("/graphql", graphql_api)
```

krauss:
> Unrelated to the post, but does subscriptions work for you at all now?

Yes it does. By the way, I am using python 3.11  as well. There is one dependency you must install in order for Subscriptions to work. It is the websockets lib.

`pip install websockets`

DoctorJohn:
After a quick investigation I feel like this could be specific to the fastapi integration. @tsmith023 do you have an idea what's happening? Asking because it looks like you worked on the related code a lot and I'm slightly overwhelmed by the `__get_context_getter` method :)

tsmith023:
[...]

Issue #1703, opened 2022-03-07
No standard way to intercept the "connect" websockets message


kristjanvalur:
### Discussed in https://github.com/strawberry-graphql/strawberry/discussions/1640

<div type='discussions-op-text'>

<sup>Originally posted by **kristjanvalur** February 13, 2022</sup>
...
Alternatively, is there a way to intercept and handler the `connection_init` message and either accept or reject the conneciton based on the request headers (from the original connection) or the optional `payload` and attach custom data to the resulting `connection` object?

Basically, I'm searching for the canonical way to do request authentication and authorization when working with requests that arrive via the `graphql-transport-ws` transport, be they subscriptions, mutations or queries.

Cheers!</div>

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/strawberry-graphql) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/strawberry-graphql/strawberry/issues/1703">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/1703/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/1703/pledge.svg">
</picture>
</a>

kristjanvalur:
see #1652

PR #2212, opened 2022-09-29
Implement union using Annotated

patrick91:
This is still WIP, the goal is to implement union using the following syntax:

```python
Result = Annotated[Union[A, B], strawberry.union(name="Result")]
```

Issue #2091, opened 2022-08-12
Support for SSE in Subscriptions


csuriano23:
Is there a way to implement subscriptions as SSE instead of WebSockets or do you plan to add some support in the near future?

patrick91:
we don't have support for SSE just yet, but I'd love to have it :)

elitepc:
Hey, just checking in if someone has found an easy way to achieve this already.

patrick91:
@elitepc not yet, this would need us to implement the [SSE protocol](https://github.com/enisdenjo/graphql-sse/blob/master/PROTOCOL.md), and we haven't started working on in it just yet

devennavani:
Wanted to +1 interest in this. A lot of folks are using websockets as the transport protocol for GraphQL subscriptions that stream chat responses (like from OpenAI's chat API) to their UIs, when really SSE should be sufficient :)

patrick91:
@devennavani I've been working on implementing multipart subscriptions, would that work for you?

omarzouk:
Wanted to +1 interest in this as well. We would really like to take advantage of SSE since its simpler and more lightweight for our usecases

patrick91:
@omarzouk hopefully that's easy once I land #3076 :D

axiomofjoy:
Would be interested in this as well.

XChikuX:
Are we at a comfortable stage to look into this? @patrick91

helloworld:
We would be interested in opening a bounty for this! Would be very interested in support for SSE for subscriptions

patrick91:
@helloworld (nice username!) have you considered using multipart subscriptions? Just wanted to understand why SSE instead of that

helloworld:
I didn't even realize that was an option! Is there any recommended links you suggest for learning about the tradeoffs for using multipart subscriptions? We were only debating between SSE + websockets

marcammann:
For what it's worth, we needed SSE to support Hive Gateway in front of our APIs (which doesn't want to talk to multipart subscriptions). I added [this somewhat crude middleware implementation](https://gist.github.com/marcammann/32111868f26a45a3813122801975e6ea) (FastAPI) to translate SSE to Multipart and back.

Maybe it helps someone else to get started if necessary. (And if you find something that's broken, please do let me know)

XChikuX:
Hey @marcammann,

That would also be because the multipart subscriptions is supposed to use http2 sessions to work.

[...]

PR #2380, opened 2022-12-03
General support for connectionParams during websocket connection initialisation when operating Subscriptions

tsmith023:
This PR makes changes that allow for the definition of [`connectionParams`](https://www.apollographql.com/docs/react/data/subscriptions/#5-authenticate-over-websocket-optional) thereby facilitating token-based authentication over Subscription requests through websocket connections.

Following the standard set by the [Apollo-GraphQL ecosystem](https://www.apollographql.com/docs/apollo-server/data/subscriptions/#operation-context), all Strawberry WS protocols now read the payload of the initial connection message as an internal `connection_params` field. This `connection_params` field is then injected into the context object with a general implementation for dictionary-based contexts at the top-level and specific implementations for each integration that define a more complex class-based context object, _i.e._ FastAPI and Channels.

## Description

When the `handle_connection_init` hook is invoked by an instance of either `BaseGraphQLTransportWSHandler` or `BaseGraphQLWSHandler`, depending on the chosen protocol, the payload of the message is assigned to the `self.connection_params` instance attribute of either instance. This payload can be anything sent by the user, hence the loose typing `Dict[str, Any]`.

Subsequently, when the Subscription is formally begun, by `handle_subscribe` in the case of `BaseGraphQLTransportWSHandler` and `handle_start` in the case of `BaseGraphQLWSHandler`, the `self.connection_params` instance attribute is injected into the context object as returned by `self.get_context()` on the condition that `self.get_context()` has returned a dictionary as its result.

The single reason for this condition surrounds the fact that some integrations, _i.e._ FastAPI and Channels, define class-based contexts rather than dict-based contexts. In these cases, specific functionality is added within their integration implementations to inject the `self.connection_params` variable into the class-based context within the hierarchy of Handler inheritances.

When an HTTP request is made, e.g. a Query or a Mutation, then dict-based contexts will lack the `connection_params` key while class-based contexts will have the `connection_params` attribute equal to `None`. Only when a Subscription request is made will this variable be populated with the payload of the initial connection message.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

- Closes https://github.com/strawberry-graphql/strawberry/issues/2351

~~https://github.com/strawberry-graphql/strawberry/issues/1703, potentially also. @kristjanvalur, let me know if you feel this PR alleviates this issue and whether `connection_params` achieves a similar goal to [your PR](https://github.com/strawberry-graphql/strawberry/pull/1652).~~

## Checklist


kristjanvalur:
Thanks for tagging me.  good stuff!

I'd like to remind you that queries and mutations also can happen over WS connections.  This is often the best way to do them in case you've already established a ws connection to deal with subscriptions.

in #1703 I was trying to find a way to reject a connection based on initially provided data.  Does the Appollo-GraphQL ecosystem have a way to do that?

tsmith023:
@kristjanvalur, no problem at all!

You raise a good point, the `connection_params` field is populated whenever a websocket connection is initialised. If this is done using the `GRAPHQL_TRANSPORT_WS_PROTOCOL` then it will also be populated for Queries and Mutations since these will be transmitted over the websocket connection.

There is a key point here, however, which is that headers sent on Queries and Mutations will not work if the underlying protocol is `GRAPHQL_TRANSPORT_WS_PROTOCOL` due to the potential pitfalls as outlined in https://github.com/strawberry-graphql/strawberry/issues/2351 with headers over websockets. The client will need to be aware of this, do you agree?

With regards to your second point, I cannot find anything easily about rejecting subscriptions but my intuition would be that it is done using `connection_params`. For example, some credentials are sent on the websocket connection initialisation that is injected into the `info.context` object as `connection_params`. You can then check these credentials within your Subscription resolvers similarly to how one might check headers within the `request` object of `info.context` when working with Queries and Mutations over HTTP.

This may not be exactly what you're looking for, however, since this doesn't outright reject the connection but only rejects subsequent messages following the initial connection.

kristjanvalur:
Great, just wanted to mention Queries and Mutations because of the way the PR description is worded.
Indeed, a client **must** be aware that it is making requests over WS, and that there can be no headers sent with individual requests in such a case.  (it physically _cannot_, there is no place to _put_ headers when sending a WS message)

This brings up a third point:  If `GRAPHQL_TRANSPORT_WS_PROTOCOL` is enabled, then your resolvers need to work with both schemes.  This means that they cannot rely on headers to provide data to individual queries/mutations.  Headers and connection payload can only really be used for authentication and to provide constant options, and defaults for options that can be overridden by parameters.  An API needs to be designed with this in mind.

On a slightly related note, see issue #1754 illustrating how the get_context() FastAPI _dependency_ cannot be assumed to have a single-query scope (and thus cannot be used to provide, e.g. database sessions)

botberry:



---

This change allows clients to define connectionParams when making Subscription requests similar to the way [Apollo-Server](https://www.apollographql.com/docs/apollo-server/data/subscriptions/#operation-context) does it.

With [Apollo-Client (React)](https://www.apollographql.com/docs/react/data/subscriptions/#5-authenticate-over-websocket-optional) as an example, define a Websocket Link as so:
```
import { GraphQLWsLink } from '@apollo/client/link/subscriptions';
import { createClient } from 'graphql-ws';

const wsLink = new GraphQLWsLink(createClient({
  url: 'ws://localhost:4000/subscriptions',
  connectionParams: {
    authToken: user.authToken,
  },
}));
```
and the JSON passed to `connectionParams` here will appear within Strawberry's context as the `connection_params` attribute when accessing `info.context` within a Subscription resolver.


---




Here's the tweet text: 
```

```

tsmith023:
I agree with your comments, there's certainly some trickery involved with going between Queries, Mutations, and Subscriptions due to the difference between the underlying HTTP and WS protocols. Perhaps these problems might be alleviated if we were to support SSE instead of WS as requested here: https://github.com/strawberry-graphql/strawberry/issues/2091.

> On a slightly related note, see issue #1754 illustrating how the get_context() FastAPI _dependency_ cannot be assumed to have a single-query scope (and thus cannot be used to provide, e.g. database sessions)

This is especially true within the current context, so thanks for linking it!

kristjanvalur:
I just want to add an example, to try to hammer home a point as part of this general discussion (not in the context of this PR):

We are using Strawberry in the context of a game project.  A client will establish a WS connection to listen to events.  Since it already has this connection, it will use it too to issue queries and mutations.   A separate, simpler, command line utility also uses the same GraphQL endpoint.  It doesn't required Subscriptions and uses HTTP to issue Queries and mutations.

GraphQL is specified independently of the transport.  And the resolvers in GraphQL cannot assume that they are being called with this transport or that.  This means that we cannot _rely_ on per-request headers to provide metadata.  Similarly, any `connection_params` cannot be assumed to exist (except for Subscriptions, and only assuming that they only remain possible via WS).

For our use-case, it means that we restrict ourselves to HTTP Headers, and use those only to perform Authentication.

Because GraphQL is transport-agnostic, transport-specific data is tricky to use.

patrick91:
@tsmith023 thanks for this PR 😊

I haven't checked it properly yet because I was hoping @DoctorJohn would get to it (he's the most knowledgable person on subscriptions in Strawberry :D). But I'll try to review it over the  next few days :)

strawberry/channels/context.py
```diff
@@ -12,6 +12,7 @@ class StrawberryChannelsContext:
     """
 
     request: "ChannelsConsumer"
+    connection_params: Optional[Any] = None
```
DoctorJohn, line 15: We actually have a guarantee that it'll be a `Optional[Dict[str, Any]]`, since that's [specified as part of the underlying protocol](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#connectioninit) :)

strawberry/channels/handlers/base.py
```diff
@@ -85,8 +85,11 @@ async def get_root_value(self, request: Optional["ChannelsConsumer"] = None) ->
     async def get_context(
         self,
         request: Optional["ChannelsConsumer"] = None,
+        connection_params: Optional[Any] = None,
```
DoctorJohn, line 88: Same here :)

```suggestion
        connection_params: Optional[Dict[str, Any]] = None,
```

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -42,6 +42,7 @@ def __init__(
         self.subscriptions: Dict[str, AsyncGenerator] = {}
         self.tasks: Dict[str, asyncio.Task] = {}
         self.completed_tasks: List[asyncio.Task] = []
+        self.connection_params: Dict[str, Any] = {}
```
DoctorJohn, line 45: Here it's correct already :D

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -117,6 +118,8 @@ async def handle_message(self, message: dict):
         await self.reap_completed_tasks()
 
     async def handle_connection_init(self, message: ConnectionInitMessage) -> None:
+        self.connection_params = message.payload or {}
```
DoctorJohn, line 121: Do we actually want to default to an empty dict? The spec allows the connectionParams to be `null`/`None` and your typings in other places suggest that you wanted to allow `None` :eyes:

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -164,6 +167,8 @@ async def handle_subscribe(self, message: SubscribeMessage) -> None:
             )
 
         context = await self.get_context()
+        if isinstance(context, dict):
```
DoctorJohn, line 170: I'm still thinking about this part. It means the base class would only add the connection params to the context if the context is a dict. In case for at least one integration (fastapi) that would not be the case. That's why you handled it seperately there. I'm wondering whether we should let all the integrations handle it by themselves for consistency or whether we have a way to make it work with any context (tricky). Still not sure. What do you think? :thinking:

strawberry/subscriptions/protocols/graphql_ws/handlers.py
```diff
@@ -82,6 +84,9 @@ async def handle_message(
             await self.handle_stop(message)
 
     async def handle_connection_init(self, message: OperationMessage) -> None:
+        payload = cast(ConnectionInitPayload, message.get("payload"))
+        self.connection_params = payload or dict()
```
DoctorJohn, line 88: Same as with the other protocol. The payload/connectionParams are allowed to be `null`/`None`. I think it would be best to preserve that and don't make it an empty dict instead.

strawberry/subscriptions/protocols/graphql_ws/handlers.py
```diff
@@ -100,6 +105,8 @@ async def handle_start(self, message: OperationMessage) -> None:
         variables = payload.get("variables")
 
         context = await self.get_context()
+        if isinstance(context, dict):
```
DoctorJohn, line 108: Also wondering about this check, just like with the other protocol :thinking:

tests/aiohttp/test_graphql_transport_ws.py
```diff
@@ -955,3 +955,38 @@ async def test_single_result_duplicate_ids_query(aiohttp_client):
...
+
+
+async def test_injects_connection_params(aiohttp_client):
+    app = create_app()
+    aiohttp_app_client = await aiohttp_client(app)
+
+    async with aiohttp_app_client.ws_connect(
+        "/graphql", protocols=[GRAPHQL_TRANSPORT_WS_PROTOCOL]
+    ) as ws:
+        await ws.send_json(ConnectionInitMessage(payload="echo").as_dict())
```
DoctorJohn, line 967: I assume this test is passing just fine, but technically the payload is required (by both ws protocols) to be a JS object / Python dict or null/None. We should adjust the test accordingly even tho just having a string works :) I'm confused why mypy is not complaining since `ConnectionInitMessage` is typed accordingly.

tests/aiohttp/test_graphql_ws.py
```diff
@@ -586,3 +586,43 @@ def get_result_handler_tasks():
...
+
+async def test_injects_connection_params(aiohttp_client):
+    app = create_app(keep_alive=False)
+    aiohttp_app_client = await aiohttp_client(app)
+
+    async with aiohttp_app_client.ws_connect(
+        "/graphql", protocols=[GRAPHQL_WS_PROTOCOL]
+    ) as ws:
+        await ws.send_json(
+            {"type": GQL_CONNECTION_INIT, "id": "demo", "payload": "echo"}
```
DoctorJohn, line 599: Same as above. Payload should be a `dict` :) I'm linking the test tho <3

tests/asgi/test_graphql_transport_ws.py
```diff
@@ -792,3 +792,32 @@ def test_single_result_duplicate_ids_query(test_client):
         data = ws.receive()
         assert data["type"] == "websocket.close"
         assert data["code"] == 4409
+
+
+def test_injects_connection_params(test_client):
+    with test_client.websocket_connect("/", [GRAPHQL_TRANSPORT_WS_PROTOCOL]) as ws:
+        ws.send_json(ConnectionInitMessage(payload="echo").as_dict())
```
DoctorJohn, line 799: Marking it again, so we won't forget about it. (payload should be a dict).

tests/asgi/test_graphql_ws.py
```diff
@@ -514,3 +514,42 @@ def test_task_cancellation_separation(test_client):
...
         assert response["id"] == "debug1"
+
+
+def test_injects_connection_params(test_client):
+    with test_client.websocket_connect("/", [GRAPHQL_WS_PROTOCOL]) as ws:
+        ws.send_json(
+            {
+                "type": GQL_CONNECTION_INIT,
+                "id": "demo",
+                "payload": "echo",
```
DoctorJohn, line 525: (payload should be a dict).

tests/channels/test_graphql_transport_ws.py
```diff
@@ -422,3 +422,27 @@ async def test_subscription_exceptions(ws):
     assert response["id"] == "sub1"
     assert len(response["payload"]) == 1
     assert response["payload"][0] == {"message": "TEST EXC"}
+
+
+async def test_inject_connection_params(ws):
+    await ws.send_json_to(ConnectionInitMessage(payload="echo").as_dict())
```
DoctorJohn, line 428: (payload should be a dict).

tests/channels/test_graphql_ws.py
```diff
@@ -583,3 +583,37 @@ async def test_task_cancellation_separation():
     response = await ws1.receive_json_from()
     assert response["type"] == GQL_COMPLETE
     assert response["id"] == "debug1"
+
+
+async def test_injects_connection_params(ws):
+    await ws.send_json_to(
+        {"type": GQL_CONNECTION_INIT, "id": "demo", "payload": "echo"}
```
DoctorJohn, line 590: (payload should be a dict). I feel bad for commenting it over an over again lol

tests/fastapi/test_context.py
```diff
@@ -125,3 +177,117 @@ def get_context(value: str = Depends(custom_context_dependency)) -> str:
...
+    app = FastAPI()
+    schema = strawberry.Schema(query=Query, subscription=Subscription)
+    graphql_app = GraphQLRouter(schema=schema, context_getter=get_context)
+    app.include_router(graphql_app, prefix="/graphql")
+    test_client = TestClient(app)
+
+    with test_client.websocket_connect(
+        "/graphql", [GRAPHQL_TRANSPORT_WS_PROTOCOL]
+    ) as ws:
+        ws.send_json(ConnectionInitMessage(payload="echo").as_dict())
```
DoctorJohn, line 213: (payload should be a dict). loving all the tests btw

tests/fastapi/test_graphql_transport_ws.py
```diff
@@ -828,3 +828,34 @@ def test_single_result_duplicate_ids_query(test_client):
         data = ws.receive()
         assert data["type"] == "websocket.close"
         assert data["code"] == 4409
+
+
+def test_injects_connection_params(test_client):
+    with test_client.websocket_connect(
+        "/graphql", [GRAPHQL_TRANSPORT_WS_PROTOCOL]
+    ) as ws:
+        ws.send_json(ConnectionInitMessage(payload="echo").as_dict())
```
DoctorJohn, line 837: (payload should be a dict). damn we have so many integrations

tests/fastapi/test_graphql_ws.py
```diff
@@ -516,3 +516,36 @@ def test_task_cancellation_separation(test_client):
         response = ws1.receive_json()
         assert response["type"] == GQL_COMPLETE
         assert response["id"] == "debug1"
+
+
+def test_injects_connection_params(test_client):
+    with test_client.websocket_connect("/graphql", [GRAPHQL_WS_PROTOCOL]) as ws:
+        ws.send_json({"type": GQL_CONNECTION_INIT, "id": "demo", "payload": "echo"})
```
DoctorJohn, line 523: (payload should be a dict). <3

DoctorJohn (changes requested):
Hey @tsmith023, loving those changes and especialy the extensive testing. Mostly looks good to me. I left a bunch of comments tho, it's mainly the same one over and over again since we have some many integrations. Thanks for the PR. Curious about your thoughts on some of the comments.

strawberry/channels/context.py
tsmith023: Great news with regards to the established spec, I intended to keep it as general as possible to allow `Any` everywhere but that clearly misaligns with what I wrote elsewhere. I will converge on having `Optional[Dict[str, Any]]` as that is how you describe it here as part of the spec and in other comments.

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
tsmith023: To achieve consistency based on other comments, this will change to
```python
self.connection_params: Optional[Dict[str, Any]] = None`
```

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
tsmith023: I agree that it would be best to align closely to the established spec, so I will align all other instances of this to use `Optional[Dict[str, Any]]` and change empty dicts to `None`s :grin:

tests/aiohttp/test_graphql_transport_ws.py
tsmith023: I agree that tests should assume that the provided `connection_params` is passed as a dict. As a result of this assumption, should we enforce that `connection_params` be a dict, _i.e._ during the `handle_connection_init` test the typing of the object and reject it if it isn't dict-like?

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
tsmith023: My thoughts around this were to achieve as much generality as possible, while still maintaining specificity. I think that Strawberry should be able to handle subscriptions in the absence of an integration, e.g. by using the CLI, so I think `connection_params` must be injected here.

Alternatively, we could attempt to implement a general Strawberry-wide class-based context object, similar to the FastAPI one, and then have these two options be the standard implementations that integrations should sub-class from. Indeed, I remember @jkimbo opening such an issue here: https://github.com/strawberry-graphql/strawberry/issues/1258.

tsmith023:
@patrick91, the type checkers failed on `tests/pyright/test_union.py`  because of this:
```
At index 0 diff: Result(type='error', message='Invalid expression form for type alias definition (reportGeneralTypeIssues)', line=14, column=26) != Result(type='information', message='Type of "UserOrError" is "Type[User] | Type[Error]"', line=16, column=13)
```
which is that the `type` attributes of each `Result` class are different here. Any idea why this could be?

tsmith023:
@DoctorJohn, the type checkers test fail on the, seemingly unrelated, test: `tests/pyright/test_union.py`. It fails locally too!
Do you have an idea why this could be?

patrick91:
@tsmith023 this might be due to a pyright update, feel free to ignore it, I'll see if I can get #2212 in at some point this week :)

tests/aiohttp/test_graphql_transport_ws.py
tsmith023: I have implemented a check on the payload variable to enforce it as a dict or None, let me know if you think it's welcome :grin:

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
DoctorJohn: > achieve as much generality as possible

Good point, I was only concerned about maintainability. But I think it's okay :)

> e.g. by using the CLI

The CLI uses the ASGI integration under the hood.

> Alternatively, we could attempt to implement a general Strawberry-wide class-based context object

I'm very much in favour of that. But that would be another PR of course :)

tests/aiohttp/test_graphql_transport_ws.py
DoctorJohn: > should we enforce that connection_params be a dict

I think we should, but for all `payload` objects. The [graphql-ws-transport spec](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#invalid-message) specifies what should happen when a message is invalid. The [graphql-ws protocol](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md) doesn't appear to be to explicit about it. (note that the links are correct, the protocol repository names do not match their protocols name).

tsmith023:
@DoctorJohn, what's the status of this PR? Are there any unresolved issues or outstanding changes that should be addressed?

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -117,6 +119,12 @@ async def handle_message(self, message: dict):
         await self.reap_completed_tasks()
 
     async def handle_connection_init(self, message: ConnectionInitMessage) -> None:
+        if message.payload is not UNSET and not isinstance(message.payload, dict):
```
DoctorJohn, line 122: The `message.payload is not UNSET` case is currently only implicitly tested by the pre-existing tests. Personally, I would add explicit tests for that as well :)

DoctorJohn (changes requested):
Hey @tsmith023, sorry for the late reply I'm currently overseas. This PR looks good to go and I'm tempted to just approve and merge it. To make it perfect I suggested another testcase. Lemme know what you think :)

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
tsmith023: Good idea! I've added some tests that I believe handle this case, which are ready to review again now :D

DoctorJohn (approved):
Thanks for the collaboration @tsmith023, this PR is good to go.

Issue #2379, opened 2022-12-03
Generic types are duplicated when using lazy types


lonelyteapot:
## Code sample

```python
# main.py (module name used later in lazy annotation)
from typing import Annotated, Generic, TypeVar
import strawberry


NodeType = TypeVar("NodeType")


@strawberry.type
class Edge(Generic[NodeType]):
    node: NodeType


@strawberry.type
class Foo:
    pass


@strawberry.type
class Query:
    regular: Edge[Foo]
    lazy: Edge[Annotated["Foo", strawberry.lazy(".main")]]


schema = strawberry.Schema(query=Query)
```

## Exception

`
Query fields cannot be resolved. Types need to have unique names in the entire schema. Duplicated type name: FooEdge
`

## Expected behavior

Regular Strawberry types and their lazy annotations should be treated as equal and result in a single SDL type.

## Additional context

Here's the line that I suspect causes the issue (`{~NodeType: Foo}` != `{~NodeType: LazyType(Foo)}`):

https://github.com/strawberry-graphql/strawberry/blob/d5e62ac9ad38f061bcbd45c6d5ac420697f0e1c8/strawberry/schema/schema_converter.py#L703

lonelyteapot:
If someone can confirm that the solution is as easy as expanding the mentioned line to check for lazy types, I can submit a PR

patrick91:
@lonelyteapot yes, that’s where the fix should go! 

thanks for the report ☺️

PR #2381, opened 2022-12-03
Fix duplication of generics with lazy types

lonelyteapot:
## Description

This PR makes lazy types treated as equal to their original types when comparing TypeVars inside generics.

Lazy types are resolved for the comparison.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #2379 

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,15 @@
+Release type: patch
+
+This release fixes an issue with type duplication of generics.
+
+You can now use a lazy type with a generic even if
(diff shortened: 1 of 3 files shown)


patrick91 (approved):
Looks good! Thank you!

I think we can probably improve the check in future by adding methods (especially magic ones) to TypeDefinition and/or LazyTypes, so we have to less checks here, but let's do that in future :)

PR #2411, opened 2022-12-20
Fix enums inside generics

### LINKED ISSUE #2410: App breaking when using Enum as field for Generic
When using an Enum as a field on a Generic, the app breaks, throwing a `NotImplementedError`.

## Describe the Bug
The below code is an example of the error.
```python
from enum import Enum
from typing import Generic, Optional, TypeVar

import strawberry

T = TypeVar("T")


@strawberry.enum
class EstimatedValueEnum(Enum):
    test = "test"
    testtest = "testtest"


@strawberry.type
class EstimatedValue(Generic[T]):
    value: T
    type: EstimatedValueEnum


@strawberry.type
class Query:
    id_translations: Optional[EstimatedValue[int]]


```
Are we doing something wrong and this is intended or is this a bug?

## System Information

 - Operating system: Docker
 - Strawberry version (if applicable): Since [0.149.2](https://github.com/strawberry-graphql/strawberry/blob/main/CHANGELOG.md#01492---2022-12-09)

patrick91:
This PR implements `copy_with` inside `EnumDefinition`, previously it would just raise a `NotImplementedError` :)

Closes #2410

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,24 @@
+Release type: patch
+
+This release fixes an issue that prevented using generic
+that had a field of type enum. The following works now:
+
+```python
+@strawberry.enum
+class EstimatedValueEnum(Enum):
+    test = "test"
+    testtest = "testtest"
+
+
+@strawberry.type
+class EstimatedValue(Generic[T]):
+    value: T
+    type: EstimatedValueEnum
+
+
+@strawberry.type
+class Query:
+    @strawberry.field
+    def estimated_value(self) -> Optional[EstimatedValue[int]]:
+        return EstimatedValue(value=1, type=EstimatedValueEnum.test)
+```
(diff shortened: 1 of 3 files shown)


botberry:



---

This release fixes an issue that prevented using generic
that had a field of type enum. The following works now:

```python
@strawberry.enum
class EstimatedValueEnum(Enum):
    test = "test"
    testtest = "testtest"


@strawberry.type
class EstimatedValue(Generic[T]):
    value: T
    type: EstimatedValueEnum


@strawberry.type
class Query:
    @strawberry.field
    def estimated_value(self) -> Optional[EstimatedValue[int]]:
        return EstimatedValue(value=1, type=EstimatedValueEnum.test)
```


---




Here's the tweet text: 
```

```

PR #2440, opened 2023-01-09
Add support for FastAPI 0.89.0

patrick91:
Not pleased with this fix, but it was the fastest one :D

Latest FastAPI adds support for response models (based on Pydantic) so it tries to validate the response types, we did type our responses, but using a generic Response class which makes FastAPI complain.

I haven't spent time checking the docs to see if there's a way to opt-out from this behaviour, we also don't want to run validation on all the responses

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+This release adds support for FastAPI 0.89.0
--- a/poetry.lock
+++ b/poetry.lock
@@ -67,7 +67,7 @@ trio = ["trio (>=0.16,<0.22)"]
 
 [[package]]
 name = "asgiref"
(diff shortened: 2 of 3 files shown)


botberry:



---

This release adds support for FastAPI 0.89.0


---




Here's the tweet text: 
```

```

Kludex (commented):
`response_model=None` can be added on the method decorator, if you prefer.

Both are fine. This is an issue, I've already talked to Sebastian about it.

patrick91:
@Kludex thanks! I was going to use response_models, but I guess we need to support older version of FastAPI, so I'll merge this fix for now and I'll readd the types once there's a fix on FastAPI :D

Thank you! <3

PR #2457, opened 2023-01-13
Update documentation: offset-based pagination overview

kerenduchan:
This PR is an update to the documentation only.
Only the **Offset-Based Pagination** section in the **Pagination Overview** doc is affected.

**Changes:**
- Metadata is not mandatory in the result of offset-based pagination.
- offset+limit vs page_number+page_size, without mixing the two.
- Other clarifications / spelling stuff I stumbled upon.

I hope I have not overstepped with my edits.


patrick91 (approved):
Nice one! Thank you very much 😊

Issue #2397, opened 2022-12-13
Duplicate Auto-Generated Types


gravy-jones-locker:
## Describe the Bug

Having updated to 0.150.1 I started experiencing issues with Strawberry's auto-generated types.

I am using a `PaginatedConnection` type (similar to the Strawberry guide) to return paginated results for a particular object. Previously the relevant output annotation `PaginatedConnect[OutputType]` could be returned from multiple fields.

Since the update this raises `strawberry.exceptions.duplicated_type_name.DuplicatedTypeName`. The duplicate is the subsidiary OutputType which is exposed as an edge field with a dynamically generated name `OutputTypeEdge`.

This seems like a bug not a feature as it is important for multiple fields to be able to return the same type - even when those types are auto-generated by Strawberry.

## Additional Context

It seems this release is probably relevant 

![image](https://user-images.githubusercontent.com/67203163/207377980-b9d65e76-578d-46d6-8067-849a9286c0a3.png)

patrick91:
@gravy-jones-locker this is definitely a bug, do you have a small reproduction? 

We had this bug and we fixed it, so maybe there was a regression somewhere

gauthamkamath:
I am also facing same issue. First time strawberry user.

strawberry-graphql version == 0.152.0

Exception: **strawberry.exceptions.duplicated_type_name.DuplicatedTypeName: Type PlayerOptionalConnection is defined multiple times in the schema**

Reproduction:

```python
# straw/schemas/pagination.py
# Same as in documentation
from typing import List, Generic, TypeVar, Optional
import strawberry
from strawberry import UNSET

GenericType = TypeVar("GenericType")

@strawberry.type
class Connection(Generic[GenericType]):
    page_info: "PageInfo"
    edges: List["Edge[GenericType]"]

@strawberry.type
class PageInfo:
    has_next_page: bool
    has_previous_page: bool
    start_cursor: Optional[str]
    end_cursor: Optional[str]

@strawberry.type
class Edge(Generic[GenericType]):
    node: GenericType
    cursor: str

```

```python
# straw/schemas/team.py
import strawberry
from typing import TYPE_CHECKING, Annotated
from straw.schemas.pagination import Connection

if TYPE_CHECKING:
    from straw.schemas.player import Player

@strawberry.type
class Team:
    id: strawberry.ID
    name: str

    @strawberry.field
[...]

PR #2463, opened 2023-01-17
Fix duplicated type name on generics

skilkis:
This commit addresses #2379 by implementing additional logic to resolve
`strawberry.lazy` annotated types. Previously, if the same return type
was declared in one location using `strawberry.lazy` and in another
location without, then a `DuplicatedTypeName` exception would be raised.

Summary of Changes:
- Added logic to handle `StrawberryOptional` in `schema_converter.py`
- Added check to determine if `strawberry.lazy` refers to the
    `__main__` module to prevent an import leading to instantiation
    of new module members.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #2397 

## Checklist

### DIFF
--- a/strawberry/lazy_type.py
+++ b/strawberry/lazy_type.py
@@ -47,12 +47,13 @@ class LazyType(Generic[TypeName, Module]):
             # Refer to: https://github.com/strawberry-graphql/strawberry/issues/2397
             if main_module.__spec__ and main_module.__spec__.name == self.module:
                 module = main_module
-            elif (
-                main_module.__file__
-                and module.__file__
-                and Path(main_module.__file__).samefile(module.__file__)
-            ):
-                module = main_module
+            elif hasattr(main_module, "__file__") and hasattr(module, "__file__"):
+                if (
+                    main_module.__file__
+                    and module.__file__
+                    and Path(main_module.__file__).samefile(module.__file__)
+                ):
+                    module = main_module
         return module.__dict__[self.type_name]
 
     # this empty call method allows LazyTypes to be used in generic types
--- a/strawberry/lazy_type.py
+++ b/strawberry/lazy_type.py
@@ -1,7 +1,9 @@
 import importlib
 import inspect
+import sys
 import warnings
 from dataclasses import dataclass
+from pathlib import Path
 from typing import ForwardRef, Generic, Optional, Type, TypeVar, cast
 
 TypeName = TypeVar("TypeName")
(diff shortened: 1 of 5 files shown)


botberry:



---

Fix `DuplicatedTypeName` exception being raised on generics declared using
`strawberry.lazy`. Previously the following would raise:

```python
# issue_2397.py
from typing import Annotated, Generic, TypeVar

import strawberry

T = TypeVar("T")


@strawberry.type
class Item:
    name: str


@strawberry.type
class Edge(Generic[T]):
    node: T


@strawberry.type
class Query:
    edges_normal: Edge[Item]
    edges_lazy: Edge[Annotated["Item", strawberry.lazy("issue_2397")]]


if __name__ == "__main__":
```


---




Here's the tweet text: 
```

```

strawberry/lazy_type.py
```diff
@@ -38,7 +40,20 @@ def __class_getitem__(cls, params):
...
+            # Refer to: https://github.com/strawberry-graphql/strawberry/issues/2397
+            if main_module.__spec__ and main_module.__spec__.name == self.module:
+                module = main_module
+            elif hasattr(main_module, "__file__") and hasattr(module, "__file__"):
+                if (
+                    main_module.__file__
+                    and module.__file__
+                    and Path(main_module.__file__).samefile(module.__file__)
+                ):
+                    module = main_module
```
patrick91, line 56: I didn't know about `samefile` :D

is this only for code that imports things from `__main__`?

strawberry/schema/schema_converter.py
```diff
@@ -721,8 +721,18 @@ def validate_same_type_definition(
...
+                    type1.of_type, LazyType
+                ):
+                    type1.of_type = type1.of_type.resolve_type()
+
                 if isinstance(type2, LazyType):
                     type2 = type2.resolve_type()
+                elif isinstance(type2, StrawberryOptional) and isinstance(
+                    type2.of_type, LazyType
+                ):
+                    type2.of_type = type2.of_type.resolve_type()
```
patrick91, line 734: do we need to do this for lists as well?

strawberry/lazy_type.py
skilkis: @patrick91 yes unfortunately it is an annoying edge-case where you end up with two distinct class objects that are actually from the same module. One of them will have a `__module__ = "__main__"` and the other something like `__module__` = "foo.bar". I guess this situation with `strawberry.lazy` isn't very likely, but I did run into it in my minimal repro

strawberry/schema/schema_converter.py
skilkis: Yeah 🙁 I believe also for `StrawberryUnion`. I think the right way to go is to define `__eq__` on `TypeDefinition` so then in this method we can just compare equality on the types directly. The derived type classes can then implement their own behavior of how equality should be checked. For now I think this bit of code will get out of hand if we implement all cases. It's meant more of as a quick solution for @gauthamkamath issue. If you think it should be done in this PR I can work on implementing `__eq__` tonight. Otherwise I'll do it in a follow-up PR

patrick91: let's do it in another PR!

strawberry/lazy_type.py
patrick91: ok, good to know! thanks!

patrick91 (approved):

PR #2481, opened 2023-01-20
Improve the performance of conversions from NextMessage to a dict

### LINKED ISSUE #2479: Increased CPU usage when subscribing with the graphql-transport-ws protocol
## Describe the Bug
We have a Strawberry GraphQL server that we have been stress testing and running CPU performance tests on. We have found that there is a noticeable and consistent increase in the CPU usage of our server application when our client subscribes using the _graphql-transport-ws_ protocol compared to using the _graphql-ws_ protocol. 

I have done a bit of investigating and further profiling using py-spy and discovered that the Strawberry code is creating a `NextMessage` object ([here](https://github.com/strawberry-graphql/strawberry/blob/db9c22a53205cd82330a9c84d44ac1ee2731eafb/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py#L261)) for each message, which it then converts to a dictionary ([here](https://github.com/strawberry-graphql/strawberry/blob/db9c22a53205cd82330a9c84d44ac1ee2731eafb/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py#L283)) using the `dataclasses` `asdict() `method ([here](https://github.com/strawberry-graphql/strawberry/blob/db9c22a53205cd82330a9c84d44ac1ee2731eafb/strawberry/subscriptions/protocols/graphql_transport_ws/types.py#L12)). Some internet research shows that this `asdict()` method is doing a `deepcopy` of everything within the class. I ran a few timing tests and the `asdict()` method takes an order of magnitude longer than doing a simple `.__dict__` on the object. This is only done in the _graphql-transport-ws_ implementation and not the _graphql-ws_ implementation which explains why there is a difference in CPU usage between the 2 protocols.

I do not believe that we need to be doing a deepcopy when turning the class into a dictionary. What's more, I wonder whether we need to even be creating the `NextMessage` object because as far as I can see, we create it and pass it to a function that immediately turns it into a dictionary. So why don't we just create it as a dictionary and send it instead. This would bypass having to do any sort of conversion costing time. 

I.e. instead of line 261 and 262 ([here](https://github.com/strawberry-graphql/strawberry/blob/db9c22a53205cd82330a9c84d44ac1ee2731eafb/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py#L261)) which do:
```
    next_message = NextMessage(id=operation_id, payload=next_payload)
    await self.send_message(next_message)`
```
we could do something like:
```
    next_message = {"id":operation_id, "payload": next_payload, "type": "next"}
    await self.send_json(next_message)
```

When I ran the performance tests with the above change the CPU usage dropped and was consistent with the _graphql-ws_ protocol performance.

## System Information

 - Operating system: Centos 7
 - Strawberry version (if applicable): 0.154.1

## Additional Context
I have created a simple demo Strawberry GraphQL server and Python client on GitHub, available at: https://github.com/rjwills28/strawberry_cpu_demo/tree/master.
Instructions on how to install and run are in the readme. It simulates the tests that we were running where we have a server providing subscription updates at 10Hz and a client that creates 100 different subscriptions. Follow the example in the readme to first run with the _graphql-ws_ protocol (command line argument (`-p 1`) and then with the _graphql-transport-ws_ protocol (`-p 2`). Run both a few times and you should see that the average CPU usage is on the whole higher for the latter protocol. Please let me know if you have any problems running this.

rjwills28:
## Description
This PR relates to the issue described in issue #2479 where we noticed an increase in the CPU usage when running subscriptions using the graphql-transport-ws protocol compared to cases where we used the graphql-ws protocol. We think this is due to the conversion of the `NextMessage()` class to a dictionary, which uses the dataclasses method `asdict()`. Underneath the hood this method is performing a deepcopy of everything, which is taking sufficiently long when running lots of subscriptions to impact the performance. 

This deepcopy of the objects is not necessary and it is easier and quicker just to create a dictionary ourselves. This fix maintains compatibility with the rest of the code and tests with minimal changes required. Essentially we just override the base class `.as_dict()` method where we just create a dictionary from the class variables and return this. We can do this as we know exactly what keys we contain in the class.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* Closes #2479 

## Checklist


botberry:



---

This release fixes a bug in subscriptions using the graphql-transport-ws protocol
where the conversion of the NextMessage object to a dictionary took an unnecessary
amount of time leading to an increase in CPU usage.


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to rjwills28 for the PR 👏

```

patrick91 (approved):
Thank you so much!

PR #2482, opened 2023-01-21
resolve_references for federation schema

hrbolek:
This patch proposal allows to speed up resolving of multiple entities (same type) for queries on federation member.

## Description

In federation schema is method `entities_resolver` which calls appropriate `resolve_reference` on federation types one by one.
Proposed code allows to extend federation type by method `resolve_references` (notice "s" - plural form),
which can serve as multiple ids resolver. In case when a database query is needed, this could speed up 
the resolving significantly, especially when large query goes thorough multiple federation members.

Implementation of  `resolve_references` is optionally, when it is missing, `resolve_reference` is used.
`resolve_reference` remains mandatory if it needs more attributes than `info` and `id` (or other unique identificator).

`resolve_references` must use the singular form of parameter (aka `id` not ~~`ids`~~) passing unique identificators.

Only the method `entities_resolver` in /strawberry/federation/schema.py has been changed. This implementation
take into account both sync and async possible definitions of `resolve_reference` and `resolve_references`.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,11 @@
+Release type: patch
+
+For federation schema types has been added a posibility to implement `resolve_references` method.
+Such implementation can have `info` parameter and also a parameter which hols a list of unique identificators.
+Usually those identificators are ids primary keys. This parameter must have same name as it has in
+implementation of method `resolve_reference`.
+
+Such possibility allows to query for entities in database with one query. This can speed up response
+for large federation queries.
+
+`resolve_references` method is not mandatory, if it is not implemented, `resolve_reference` method is used.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,8 +1,8 @@
 Release type: patch
 
 For federation schema types has been added a posibility to implement `resolve_references` method.
-Such implementation can have `info` parameter and also a parameter which hols a list of unique identificators.
-Usually those identificators are ids primary keys. This parameter must have same name as it has in
+Such implementation can have `info` parameter and also a parameter which holds a list of unique identificators.
+Usually those identificators are ids (primary keys). This parameter must have same name as it has in
 implementation of method `resolve_reference`.
 
 Such possibility allows to query for entities in database with one query. This can speed up response
--- a/docs/guides/federation-v1.md
+++ b/docs/guides/federation-v1.md
@@ -154,13 +154,12 @@ the requested `id` and a fixed number of reviews.
 <Note>
 
 If the class implements `resolve_references` (focus on plural form), this can be used.
-The parameter ideeentifing entities is considered as a list of identificators. Such 
-implementation allows resolve multiple entities in a single call. This speeds up 
-resolving, especially if communication with a database server is needed as this can 
+The parameter ideeentifing entities is considered as a list of identificators. Such
+implementation allows resolve multiple entities in a single call. This speeds up
+resolving, especially if communication with a database server is needed as this can
 be done with single database query.
 
 </Note>
-
 
 If we were to add more fields to `Book` that were stored in a database, this
 would be where we could perform queries for these fields' values.
--- a/docs/guides/federation-v1.md
+++ b/docs/guides/federation-v1.md
@@ -151,6 +151,17 @@ by the books service. Recall that above we defined the `id` field as the `key`
 for the `Book` type. In this example we are creating an instance of `Book` with
 the requested `id` and a fixed number of reviews.
 
+<Note>
+
+If the class implements `resolve_references` (focus on plural form), this can be used.
+The parameter ideeentifing entities is considered as a list of identificators. Such 
+implementation allows resolve multiple entities in a single call. This speeds up 
+resolving, especially if communication with a database server is needed as this can 
+be done with single database query.
+
+</Note>
+
+
 If we were to add more fields to `Book` that were stored in a database, this
 would be where we could perform queries for these fields' values.
 
--- a/docs/guides/federation-v1.md
+++ b/docs/guides/federation-v1.md
@@ -153,8 +153,8 @@ the requested `id` and a fixed number of reviews.
 
 <Note>
 
-If the class implements `resolve_references` (focus on plural form), this can be used.
-The parameter ideeentifing entities is considered as a list of identificators. Such
+If the class implements `resolve_references` (focus on plural form), it can be used.
+The parameter identifying entities is considered as a list of identificators. Such
 implementation allows resolve multiple entities in a single call. This speeds up
 resolving, especially if communication with a database server is needed as this can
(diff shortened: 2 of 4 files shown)


botberry:



---

For federation schema types has been added a posibility to implement `resolve_references` method.
Such implementation can have `info` parameter and also a parameter which holds a list of unique identificators.
Usually those identificators are ids (primary keys). This parameter must have same name as it has in
implementation of method `resolve_reference`.

Such possibility allows to query for entities in database with one query. This can speed up response
for large federation queries.

`resolve_references` method is not mandatory, if it is not implemented, `resolve_reference` method is used.


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to hrbolek for the PR 👏

```

botberry:



---

For federation schema types has been added a posibility to implement `resolve_references` method.
Such implementation can have `info` parameter and also a parameter which holds a list of unique identificators.
Usually those identificators are ids (primary keys). This parameter must have same name as it has in
implementation of method `resolve_reference`.

Such possibility allows to query for entities in database with one query. This can speed up response
for large federation queries.

`resolve_references` method is not mandatory, if it is not implemented, `resolve_reference` method is used.


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to hrbolek for the PR 👏

```

patrick91:
Hi @hrbolek, thanks for taking time to submit this PR. I wonder if using data loaders will remove the need of this? I think that might be a more pragmatic approach. Was there any reason you didn't use dataloader for this? :)

hrbolek:
Hi,

Dataloaders can improve performance while resolving entities from their ids. 

However, even if dataloaders are used, the `resolve_reference` are called one by one and if entity must be taken from database, such call means one query to database. If the federated question is a bit larger, this means tens (usually SQL) queries to DB server.

Moreover, if lifespan of cache is short and Dataloaders are not implemented well (Redis can really help there), the communication with DB grows.

The proposed implementation looks like and overkill for those who do not need large queries spreading over multiplte federation members. At start I had idea to implement extra class with overided `entities_resolver` (in class `strawberry.federation.Schema`) and I am using this in my projects, but I propose to implement it for all cases.

patrick91:
> Hi,
> 
> Dataloaders can improve performance while resolving entities from their ids.
> 
> However, even if dataloaders are used, the `resolve_reference` are called one by one and if entity must be taken from database, such call means one query to database per single call. If a federated question is a bit larger, this means tens (usually SQL) queries to DB server.
If this is the case, then there's a bug, dataloaders are meant to be used to reduce the db calls :)


> Moreover, if lifespan of cache is short and Dataloaders are not implemented well (Redis can really help there), the communication with DB grows.
Isn't this the same with using `resolve_references`? :) 

> The proposed implementation looks like an overkill for those who do not need large queries spreading over multiplte federation members. At start I had idea to implement extra class with overided `entities_resolver` (in class `strawberry.federation.Schema`) as I am using this in my projects, but I propose to implement it for all cases.

sorry if it looks like I'm pushing but I want to understand why existing solution don't work 😊

hrbolek:
> > Hi,
> > Dataloaders can improve performance while resolving entities from their ids.
> > However, even if dataloaders are used, the `resolve_reference` are called one by one and if entity must be taken from database, such call means one query to database per single call. If a federated question is a bit larger, this means tens (usually SQL) queries to DB server.
> 
> If this is the case, then there's a bug, dataloaders are meant to be used to reduce the db calls :)
> 
> > Moreover, if lifespan of cache is short and Dataloaders are not implemented well (Redis can really help there), the communication with DB grows.
> 
> Isn't this the same with using `resolve_references`? :)
> 
I have a project, where are many federation members (strawberry based applications). To describe my findings I reduce this to two federation members. 

The master query goes to federation member B to get "master" entities (in my case `Events`). This particular request retrieves entities from database with single query. But request (incoming GraphQL query) define to get for each event a participants (`Users`) and some their attributes. As the responsible federation member for this entity (`User`) is federation member A, the subqueries are going there. Apollo federation transforms this into single query which goes to Federation member A. Federation member A must firstly reconstruct entities (`Users`) to get their attributes. And there goes in play `resolve_reference`. Unfortunately in this case a single entity (all `Users`) are resolved one by one by individual calls of `resolve_reference` method of `User` class. 

The subquery (synthetized by apollo) looks like

```
 query ($representations: [_Any!]!) {
      _entities(representations: $representations) {
          ... on User {
              email
              name
              surname
          }
      }
  }
```

As I know this subquery goes to `strawberry.federation.Schema`, method `entities_resolver`. This method creates entities (`Users`) but for each initialization it uses single call of `resolve_reference` (`User` class method). If the needed entites are not cached, what depends on cache strategy (and its implementation), we need many db queries. For, let say, 15 participants at avarage for 20 events, this could end up to 300 db queries (I suppose that all participants are unique).

I have tried to solve this with Dataloaders, but Dataloader is called as many times as apollo has asked. I am pretty sure that this is related only to GQL federation.

My proposal allows to gather all those repeated single calls into one call of `response_references` if it is implemented in appropriate type (in `User` class for this example).

As I noticed, the subquery (synthetized by apollo) can contain multiple asks for representations related to single primary key value. Apollo can ask multiple times for single user in one query. This is just my fininding which has no impact on proposal motivation.

> > The proposed implementation looks like an overkill for those who do not need large queries spreading over multiplte federation members. At start I had idea to implement extra class with overided `entities_resolver` (in class `strawberry.federation.Schema`) as I am using this in my projects, but I propose to implement it for all cases.
> 
> sorry if it looks like I'm pushing but I want to understand why existing solution don't work 😊

It is ok, I understand this need.

patrick91:
@hrbolek are you able to find a minimal reproduction for the dataloaders not working, I just did a demo and it works for me:

CleanShot 2023-01-24 at 11 02 01


code here: https://github.com/patrick91/strawberry-dataloader-federation/blob/main/api/schema.py#L11-L27

hrbolek:
Thanks for your code, I have spent several hours to recall or find the experiments I have done. I cannot find the experiment, which has confused me that function wrapped by dataloader is called multiple times. The only thing I have found is experiment in Jupyter environment. This experiment has different behaviour as I found now.
I am really sorry for forcing you to reinvent the wheel :).

patrick91:
@hrbolek so the code has a different behaviour when running under jupyter?
Can you share some code for that? if that's the case? Feel free to open a new issue, or join our [discord](https://strawberry.rocks/discord)

Also thanks for the PR! I think I'll close it for now since dataloaders should work. Feel free to request a sticker here: https://forms.gle/dmnfQUPoY5gZbVT67 if you want one 😊

PR #2483, opened 2023-01-22
Use sys.__excepthook__ and threading.__excepthook__

patrick91:
As suggested by @BryceBeagle , python provides `sys.__excepthook__`
and ~`threading.__excepthook__`~ (this is only in python 3.10), so there's no need to store them 
in variables.

We can merge this without a release 😊

### DIFF
--- a/strawberry/exceptions/handler.py
+++ b/strawberry/exceptions/handler.py
@@ -6,6 +6,12 @@ from typing import Any, Callable, Optional, Tuple, Type, cast
 
 from .exception import StrawberryException, UnableToFindExceptionSource
 
+if sys.version_info >= (3, 8):
+    original_threading_exception_hook = threading.excepthook
+else:
+    original_threading_exception_hook = None
+
+
 ExceptionHandler = Callable[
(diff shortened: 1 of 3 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

strawberry/exceptions/handler.py
```diff
@@ -76,18 +68,18 @@ def strawberry_threading_exception_handler(
...
     _get_handler(exception_type)(exception_type, exception, traceback)
 
 
 def reset_exception_handler():
-    sys.excepthook = original_exception_hook
+    sys.excepthook = sys.__excepthook__
 
     if sys.version_info >= (3, 8):
-        threading.excepthook = original_threading_exception_hook
+        threading.excepthook = threading.__excepthook__
```
BryceBeagle, line 82: `threading.__excepthook__` is new in 3.10 not 3.8

patrick91: thanks! I'll revert it then!

Issue #2477, opened 2023-01-19
Enable Ruff annotations checks


patrick91:
In #2476 we enabled a bunch of additional checks on Ruff, but I haven't enabled some that are probably more important for us, namely the ones for annotations. We'd like our library to be fully annotated 😊

We can enable all these checks one at the time:

- [ ] ANN001
- [ ] ANN002
- [x] ANN003
- [ ] ~ANN102~
- [x] ANN201
- [x] ANN202
- [x] ANN204
- [x] ANN205
- [x] ANN206
- [ ] ANN401

We can skip these checks on tests, it's nice to have tests type too, but it might take a lot of time 😊

Feel free to ask me any question for this, some of the annotations might be easy to add, some might not 😊

patrick91:
Annotations are disabled here: https://github.com/strawberry-graphql/strawberry/blob/main/pyproject.toml#L178-L187

To work on any of them, remove the line of the annotation and then run `ruff .` and see what you get 😊 Then you can work from there to add annotations :)

MadhuMPandurangi:
Hi @patrick91, I'm interested in contributing to this issue. Can you please assign this to me
Thank You.

patrick91:
> Hi @patrick91, I'm interested in contributing to this issue. Can you please assign this to me
> 
> Thank You. 

hi @MadhuMPandurangi, I think it's fine to start a PR directly ☺️ 

Please do only one check at the time, enabling all of them at once will require a lot of work ☺️

MadhuMPandurangi:
Thanks for the reply. If this requires a lot of work then can you suggest any other issue which I can work for starters.

alexauritt:
OK, diving into ANN001. @patrick91 no expectations that you'll want to check a WIP but if you want to take a peak and verify I'm on the right path, might save us some time...

https://github.com/strawberry-graphql/strawberry/commit/c81d989a1d9f9a0fba5f93cc9df79263bf39e0cd

patrick91:
@alexauritt yes, that looks good! Feel free to open a draft PR too if you want more feedback on specific lines 😊

patrick91:
@alexauritt it is also fine to have # noqa: ANN001 where it is quite complex to add annotations btw 😊 we can always do this in multiple PRs :)

JadHADDAD92:
I wanna work on ANN201, maybe it's better to make it with multiple PRs as there are 1387 errors, what do you think @patrick91

[...]

PR #2496, opened 2023-01-28
Update Ruff

alexauritt:
This PR updates ruff and also updates the code base to comply with some new checks, most importantly the check that asks to move imports inside a `if TYPE_DEFINITION` block.

Related: #2477


strawberry/aiohttp/handlers/graphql_ws_handler.py
```diff
@@ -1,11 +1,15 @@
 from contextlib import suppress
-from typing import Any, Optional
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Any, Optional
```
patrick91, line 5: do you if there's a way to skip having these imports behind `if TYPE_CHECKING`?

feels a bit annoying to have to do that all the time for things inside typing (we still need to import TYPE_CHECKING anyway) :D

patrick91:
By the failure on the tests:

```
  File "/home/runner/work/strawberry/strawberry/strawberry/utils/operation.py", line 14, in <module>
    graphql_document: DocumentNode,
NameError: name 'DocumentNode' is not defined
```

I think we need to add `from __future__ import annotations` at the top of the files that use `if TYPE_CHECKING`

patrick91:
can you also push the updated `pyproject.toml`? :)

strawberry/aiohttp/handlers/graphql_ws_handler.py
alexauritt: agree, quite annoying. I'm not sure if there's a way. I'll pause the work and do some digging. Thanks for flagging.

alexauritt: https://github.com/charliermarsh/ruff/issues/2267

alexauritt: looks like this is addressed in latest release of `ruff` - https://github.com/charliermarsh/ruff/releases/tag/v0.0.237

alexauritt: I'll rollback those changes, thanks again for flagging before I went too deep

alexauritt:
> can you also push the updated `pyproject.toml`? :)

added but rolled it back for now to keep the branch green. I'll re-update the toml once ruff is satisfied

alexauritt:
@patrick91 -  I think this PR is ready to come out of draft mode. I should note that while my original intention was just to enable `ANN001` checks (as part of https://github.com/strawberry-graphql/strawberry/issues/2477), this PR doesn't do that at all! I got confused + distracted by the TCH errors I was seeing, and I only just realized that this was because I was running a later version of ruff locally. 

So... I guess this branch has turned into an 'upgrade ruff' PR...

Is that helpful or welcome? Should I open a new issue?

I'll be happy to follow up with work on the ANN001

Thanks!

alexauritt:
Also - would be happy to keep plugging away and include the ANN001 checks on this branch if you prefer, but thought I should check in given the size of this PR

patrick91:
hi @alexauritt, sorry for the late reply! I'll check this in the next couple of days!
thanks so much for all the work you're putting in this PR 😊

alexauritt:
@patrick91 - I should acknowledge that I'm still seeing [that same error](https://github.com/strawberry-graphql/strawberry/actions/runs/4086912711/jobs/7134627292) surfacing when running tests. A bit stumped by that.

patrick91:
@alexauritt I'll probably review this on saturday during [our sprints](https://beta.guild.host/events/strawberry-graphql-sprints-wzz8u1) :D

alexauritt:
If you've any hints as to how I could help tackle the test failures let me know.

alexauritt:
Hoping to join the sprint, virtually, for a of couple hours at least

patrick91:
Hi @alexauritt, I took the liberty of doing some commits on the PR!

One issue was that the changes were also touching code that was needed at runtime (I forgot what it was now)

Another issue was that we needed `from __future__ import annotations` in a lot of places, since the types are behind the `if TYPE_CHECKING` flag (using future annotations makes python treat annotations as string, not code)

When using `from __future__ import annotations` `casts` that reference an object imported inside the `if TYPE_CHECKING` flag need to use quotes around them like so:

```python
if TYPE_CHECKING:
   from a import A

x = cast("A", 123)
```

on top of this `from __future__ import annotations` does break some of our code (in the federation/schema.py file). I haven't investigated why just yet 😊

And fastapi dependencies need to be always imported (so they can't be in the `if TYPE_CHECKING` flag)

I think that was all of it 😊

strawberry/ext/mypy_plugin.py
```diff
@@ -337,7 +354,7 @@ def add_static_method_to_class(
             cls.defs.body.remove(sym.node)
 
     # For compat with mypy < 0.93
-    if MypyVersion.VERSION < Decimal("0.93"):
+    if Decimal("0.93") > MypyVersion.VERSION:
```
patrick91, line 357: I'll revert this change, I don't think I like it :D

botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Alex Auritt for the PR 👏

```

patrick91 (approved):
Thanks @alexauritt for doing this! I think this should good to merge!

botberry:

Issue #1042, opened 2021-06-30
Migration from Graphene to Strawberry


Sergei-Rudenkov:
I use graphene==3.0.0b6 (it uses graphql-core v3 under the hood). Could you, please, provide some migration guide?

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/strawberry-graphql) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/strawberry-graphql/strawberry/issues/1042">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/1042/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/1042/pledge.svg">
</picture>
</a>

jkimbo:
@Sergei-Rudenkov at the moment unfortunately we don't have a migration guide other than rewriting your application 😢 

I am however experimenting with something that might allow you migrate from Graphene incrementally. If you're willing to give it a try follow the instructions here: https://github.com/jkimbo/strawberry-graphene

Let me know how it goes! I haven't actually tried this on a large schema so it's quite likely that things won't work but I'll try and fix them as soon as possible.

AlecRosenbaum:
We tried to use https://github.com/jkimbo/strawberry-graphene, but found that it only worked up until version `strawberry-graphql==0.68.0`.

We made a few modifications and have it mostly working on `strawberry-graphql==0.85.1`. Most of our schema is still `graphene`-based (and we're still running into issues around enums and interfaces), but our root `Schema` object is strawberry-based. Here it is in gist-form: https://gist.github.com/AlecRosenbaum/d23541cbcd35290f675a207d0b7f95a3

jkimbo:
@AlecRosenbaum this is awesome thanks for sharing! Do you mind if I update the repo with your changes?

Also what issues are you finding with enums and interfaces? Would be happy to help you debug them.

AlecRosenbaum:
Credit for that goes to @tsx 🙂 I just stumbled across this issue and posted it.

> Do you mind if I update the repo with your changes?

Sure!

> Also what issues are you finding with enums and interfaces?

[...]

PR #2511, opened 2023-02-04
feat: Initial relay implementation

### LINKED ISSUE #1573: Relay implementation proposal
Hey guys,

I saw that `relay` support is something planned here and I have created a relay implementation that I would like to know if you would be interested as a contribution.

You might have seen that I posted on discord at django channel about this lib I wrote: https://github.com/blb-ventures/strawberry-django-plus. In it I created a generic relay implementation:

- https://github.com/blb-ventures/strawberry-django-plus/blob/master/strawberry_django_plus/relay.py

I made it really generic and no attached to django at all because, not only I would like to have it be useful to non-django related types, but also because my intention was to either contribute it back to you or create a separated project for it (strawberry-relay maybe?). It is also in a single file for that purpose.

The way I did it, you just implement the `Node` interface by inheriting at the type you are creating and define at least 2 specific methods, one to resolve a single node given its it, and another one to resolve a list of nodes (optionally given a list of ids or none at all, in which the implementation is supposed to return an iterator of all nodes). There are other resolvers that can be overridden by the implementer, but they have a default implementation already.

With that, one can create a connection with `some_connection: Connection[NodeType] = relay.connection()`. A resolver can also be passed to that (by argument or by decorating a method) which is supposed to return an iterator of the `NodeType`, and that will be paginated by the field itself. Any arguments added to that resolver will be included together with the `first`, `last`, `before` and `after` arguments.

There's is a `some_node: NodeType = relay.node()` which creates a field that expects a global id and returns the node.

And lastly, there's an input mutation field that basically converts all of the arguments passed from the resolver function to an input `Input` that is automatically created for the mutation, and them converts them back to arguments when calling the function, so for the user it still defines the resolver the same way as a common mutation, without having to create a type for each one and specify it as the single input argument. The only thing I did not do here was to create the `Payload` type because I did not know how to do that the best way...

A complete example would be:

```python
fruits = [
    {
        "id": 1,
        "name": "Banana",
        "description": "Lorem ipsum",
    },
    {
        "id": 2,
        "name": "Apple",
        "description": None,
    },
    {
        "id": 3,
        "name": "Orange",
        "description": "Lorem ipsum",
    },
]


@strawberry.type
class Fruit(Node):
    name: str
    description: Optional[str]

    @classmethod
    def resolve_node(cls, node_id: str, *, info: Optional[Info] = None, required: bool = False):
        for fruit in fruits:
            if str(fruit["id"]) == node_id:
                return Fruit(**fruit)

        if required:
            raise ValueError(f"Fruit by id {node_id} not found.")

        return None

    @classmethod
    def resolve_nodes(
        cls, *, info: Optional[Info] = None, node_ids: Optional[Iterable[str]] = None
    ):
        node_ids = node_ids and set(node_ids)

        for fruit in fruits:
            if node_ids is not None and str(fruit["id"]) not in node_ids:
                continue

            yield Fruit(**fruit)


@strawberry.type
class Query:
    fruit: Fruit = relay.node()
    fruits_conn: relay.Connection[Fruit] = relay.connection()

    @relay.connection
    def fruits_conn_with_filter(self, name_startswith: str) -> Iterable[Fruit]:
        for fruit in fruits:
            if fruit["name"].startswith(name_startswith):
                yield Fruit(**fruit)


@strawberry.type
class Mutation:
    @relay.input_mutation
    def create_fruit(self, name: str, description: Optional[str]) -> Fruit:
        fruit_data = {
            "id": max(f["id"] for f in fruits) + 1,
            "name": name,
            "description": description,
        }
        fruits.append(fruit_data)
        return Fruit(**fruit_data)
```

The generated schema would be:

```gql
interface Node {
  id: GlobalID!
}

type Fruit implements Node {
  id: GlobalID!
  name: String!
  description: String
}

type FruitEdge {
  cursor: String!
  node: Fruit
}

type FruitConnection {
  edges: [ShipEdge!]!
  pageInfo: PageInfo!
}

type PageInfo {
  hasNextPage: Boolean!
  hasPreviousPage: Boolean!
  startCursor: String
  endCursor: String
}

type Query {
  fruit(id: GlobalID!): Fruit
  fruits_conn(
    before: String,
    after: String,
    first: Int,
    last: Int
  ): FruitConnection!
  fruits_conn_with_filter(
      before: String,
      after: String,
      first: Int,
      last: Int,
      nameStartswith: String!
  ): FruitConnection!
}

input CreateFruitInput {
  name: String!
  description: String
}

type Mutation {
  createFruit(input: CreateFruitInput!): Fruit
}
```

ORM integrations can implement those `resolve_node` and `resolve_nodes` to automatically provide them. I've done that with my lib for the django ORM. I also did override the `resolve_id` and `resolve_connection` to make their implementation async-safe, since django orm is not.

bellini666:
This is my relay implementation from [strawberry-django-plus](https://github.com/blb-ventures/strawberry-django-plus), which I mentioned in [this issue](https://github.com/strawberry-graphql/strawberry/issues/1573)

There are some things that might need some discussion:

1. I used `GlobalID` as the name of the scalar to avoid conflicts with the existing `ID`. Although it doesn't cause issues for the majority of people, [there are some cases](https://github.com/blb-ventures/strawberry-django-plus/issues/166) where the client expects the scalar to be named `ID`. It can be worked around by overriding the `ID` scalar to use the `GlobalID`, but maybe there's a better way? Note that the `GlobalID` object is a dataclass with some custom attributes and methods.
2. Find a way to allow the object to define an `id`, which is the non global version one, while still keeping our `Node.id` resolver working. You will notice in my tests that I defined the `Fruit` id as `_id` for that reason. Maybe use a metaclass approach to solve this?

TODO (already done):
- [x] Add tests for the `input_mutation`
- [x] Write a documentation for `input_mutation`

Fix #1573


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "strawberry-graphql"
-packages = [ { include = "strawberry" } ]
+packages = [{ include = "strawberry" }]
```
kaos, line 3: I discourage formatting changes mixed with other changes. Particularly for a PR already of this magnitude. It makes it so much more difficult to see what's actually changed.

strawberry/relay/types.py
```diff
@@ -0,0 +1,706 @@
...
+from strawberry.types.types import TypeDefinition
+from strawberry.utils.await_maybe import AwaitableOrValue
+
+from .utils import from_base64, to_base64
+
+_T = TypeVar("_T")
+_R = TypeVar("_R")
+connection_typename = "arrayconnection"
+
+NodeType = TypeVar("NodeType", bound="Node")
```
kaos, line 42: This looks like a class name to me, and reading the docs example for custom connections had me really confused (wrgt how it was implemented, not how to use it).

I'm not sure how to improve this really, though. Perhaps `TNode` does a better job at hinting that this is a generic type var?

however, looking at some more places this is used, I think `NodeType` works really well, so perhaps the doc mentioned is the only confusing part.. I'll ponder on this a bit more.

tests/relay/schema.py
```diff
@@ -0,0 +1,185 @@
...
+    strawberry.relay.Connection[strawberry.relay.NodeType]
+):
+    @strawberry.field
+    def something(self) -> str:
+        return "foobar"
+
+    @classmethod
+    def from_nodes(
+        cls,
+        nodes: Iterable[Fruit],
```
kaos, line 90: This should be `Iterable[NodeType]` given that the type is not specialized here, but still a `Generic[NodeType]`, which is wrong as we want it to be specialized now.

I think the correct use here is:
```python
@strawberry.type
class CustomPaginationConnection(
    strawberry.relay.Connection[Fruit]
):
    ...


@strawberry.type
class Query:
    ...
    fruits_custom_pagination: CustomPaginationConnection

```

This would explain my confusion I had with this, and also it would then make sense to perhaps call it `CustomFruitPaginationConnection` ;)

kaos (approved):
This is a massive PR, but I hope we can get through this as it looks really well worked through. 🚀 

And it will be easier to circle back and fix any remaining rough edges after we've got some experience using it.

pyproject.toml
bellini666: Oh, totally! I'm also the one who comments that on PRs I review. My autoformatter formatted this when I modified this file and I didn't even notice =P

tests/relay/schema.py
bellini666: Yeah, I agree. I actually tried to do that when I originally wrote this test on `strawberry-django-plus`, but strawberry was not able to generate the type that way.

Maybe now that this code is being merged here I can try to modify the Generic handling to support that code.

strawberry/relay/types.py
bellini666: I'm open to change it to `TNode` or anything else that makes more sense. I have to admit that I'm not good at naming things (`strawberry-django-plus` is a good example of that hahaha)

kaos: Naming stuff is among the top 3 most difficult programming jobs! :P 

I think I landed in being in favour of keeping it as `NodeType`, left the comment regardless to be transparent about my review process.

tests/relay/schema.py
bellini666: I investigated and discovered the issue here. It was not specific to the relay implementation, but with `Generic` type trying to create a specialized type in that way.

Basically, when creating the fields, the `type_var_map` was not being set for them. I fixed the issue and added tests for it in `test_generics`: https://github.com/strawberry-graphql/strawberry/pull/2511/commits/6bc2f4f0b4f311125144c64f1aa07cff8ec9fa06

docs/guides/relay.md
```diff
@@ -263,9 +261,18 @@ class CustomPaginationConnection(
 
 @strawberry.type
 class Query:
-    fruits: CustomPaginationConnection[Fruit]
+    fruits: CustomPaginationConnection
 ```
 
+<Note>
+
+In the example above we specialized the `FruitCustomPaginationConnection` by
```
kaos, line 269:
```suggestion
In the example above we specialized the `CustomPaginationConnection` by
```

docs/guides/relay.md
```diff
@@ -263,9 +261,18 @@ class CustomPaginationConnection(
...
+inheriting it from `relay.Connection[Fruit]`. We could still keep it generic by
+inheriting it from `relay.Connection[relay.NodeType]` and then specialize it
+in when defining the field.
```
kaos, line 272:
```suggestion
when defining the field.
```

kaos (approved):
Nice!

bellini666:
Just added a doc section for the input mutation and some tests for it. I think the PR is basically done, not considering any suggestions/improvements that might arrive from code reviews, of course :)

bellini666:
Note: I separated the specialized generic support to another PR (https://github.com/strawberry-graphql/strawberry/pull/2517). As soon as it gets merged I'll rebase this PR to make the changes here be only about relay itself

bellini666:
> Note: I separated the specialized generic support to another PR (#2517). As soon as it gets merged I'll rebase this PR to make the changes here be only about relay itself

It was merged! Just merged main in here, the diff should now only contain relay related stuff

botberry:
# Pre-release
:wave:

Releasing commit [5d582dbf43e2084603f3b4ec3c401ba1b5991c7f] to PyPi as pre-release! :package:

botberry:
# Pre-release
:wave:

Pre-release **0.157.0.dev.1675982393** [5d582dbf43e2084603f3b4ec3c401ba1b5991c7f] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.157.0.dev.1675982393
```

strawberry/types/type_resolver.py
```diff
@@ -50,6 +51,15 @@ class if one is not set by either using an explicit strawberry.field(name=...) o
...
     from strawberry.field import StrawberryField
 
+    # Relay uses @interface/@type while creating its types, which calls this function
+    try:
+        from strawberry.relay import Connection, ConnectionField, Node, NodeField
+    except ImportError:
+        Connection = None  # type: ignore[misc,assignment]
+        ConnectionField = None  # type: ignore[misc,assignment]
+        Node = None  # type: ignore[misc,assignment]
+        NodeField = None  # type: ignore[misc,assignment]
```
patrick91, line 61: why would we get an import error here?

strawberry/relay/types.py
```diff
@@ -0,0 +1,706 @@
...
+from strawberry.type import StrawberryContainer
+from strawberry.types.info import Info
+from strawberry.types.types import TypeDefinition
+from strawberry.utils.await_maybe import AwaitableOrValue
+
+from .utils import from_base64, to_base64
+
+_T = TypeVar("_T")
+_R = TypeVar("_R")
+connection_typename = "arrayconnection"
```
patrick91, line 40: could we avoid to use this somehow? I think we can always use the type name from strawberry, no? :)

strawberry/relay/mutations.py
```diff
@@ -0,0 +1,232 @@
...
+    deprecation_reason: Optional[str] = None,
+    default: Any = dataclasses.MISSING,
+    default_factory: Union[Callable[..., object], object] = dataclasses.MISSING,
+    metadata: Optional[Mapping[Any, Any]] = None,
+    directives: Optional[Sequence[object]] = (),
+    # This init parameter is used by pyright to determine whether this field
+    # is added in the constructor or not. It is not used to change
+    # any behavior at the moment.
+    init: Literal[True, False, None] = None,
+) -> Any:
```
patrick91, line 180: let's move this into another PR and decide if we should keep it or not 😊

strawberry/relay/fields.py
```diff
@@ -0,0 +1,553 @@
...
+        # but mypy is not accepting it
+        retval = cast(RelayField, super().copy_with(type_var_map))
+        retval.default_args = self.default_args
+        return retval
+
+
+class NodeField(RelayField):
+    """Relay Node field.
+
+    Do not instantiate this directly. Instead, use `@relay.node`
```
patrick91, line 94: is there any example of this?

strawberry/relay/fields.py
```diff
@@ -0,0 +1,553 @@
...
+        default_factory=default_factory,
+        metadata=metadata,
+        directives=directives or (),
+    )
+
+
+@overload
+def connection(
+    *,
+    resolver: Callable[[], _T],
```
patrick91, line 410: should this be resolver type?

strawberry/relay/fields.py
```diff
@@ -0,0 +1,553 @@
...
+    is_subscription: bool = False,
+    description: Optional[str] = None,
+    permission_classes: Optional[List[Type[BasePermission]]] = None,
+    deprecation_reason: Optional[str] = None,
+    default: Any = dataclasses.MISSING,
+    default_factory: Union[Callable[..., object], object] = dataclasses.MISSING,
+    metadata: Optional[Mapping[Any, Any]] = None,
+    directives: Optional[Sequence[object]] = (),
+    graphql_type: Optional[Any] = None,
+) -> ConnectionField:
```
patrick91, line 458: this might be annoying, and I'm happy to do it for, but would be possible to add some pyright tests?

I think there's some small issues with typing when using strict mode:

CleanShot 2023-02-09 at 23 23 35@2x

docs/guides/relay.md
```diff
@@ -0,0 +1,381 @@
...
+        return list(fruits.values())
+
+    @classmethod
+    def resolve_node(
+        cls,
+        node_id: str,
+        *,
+        info: Optional[Info] = None,
+        required: bool = False,
+    ):
```
patrick91, line 65: are these methods always required?

tests/relay/schema.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+@strawberry.type
+class Fruit(strawberry.relay.Node):
+    _id: strawberry.Private[int]
+    name: str
+    color: str
+
+    @classmethod
+    def resolve_id(cls, root: Self, *, info: Optional[Info] = None):
```
patrick91, line 16: is there a reason why info is optional?

tests/relay/schema.py
```diff
@@ -0,0 +1,209 @@
...
+        return root._id
+
+    @classmethod
+    def resolve_nodes(
+        cls,
+        *,
+        info: Optional[Info] = None,
+        node_ids: Optional[Iterable[str]] = None,
```
patrick91, line 24: when would get get no `node_ids`?

strawberry/relay/types.py
bellini666: This actually is the name of the cursor (maybe it is poorly named? =P)

It is used in here: https://github.com/strawberry-graphql/strawberry/pull/2511/files#diff-87ff2a77ed8a7dc2eb7fc8a320e55d457b77160ab7932c185b8becbf8662767bR526 . I used the same name that [graphql-relay-py](https://github.com/graphql-python/graphql-relay-py/blob/97ba15e573eabcefd13e249d05fc5557f0296804/src/graphql_relay/connection/array_connection.py#L158) uses.

I probably wanted to call this `cursor_typename` but wrote `connection_typename` by mistake. Do you have any suggestions for a better name?

strawberry/relay/mutations.py
bellini666: Sure! Will do that

strawberry/relay/fields.py
bellini666: Oh, didn't even notice that the typing is bad in here. WIll fix it

strawberry/relay/fields.py
bellini666: Sure, I can totally do that! :)

Does that only happen in strict mode? I use pyright in basic mode (with some extra checks enabled) and did not see that before. Will write some tests and fix any typing issues that strict mode shows :)

docs/guides/relay.md
bellini666: Yes, basically. I even defined then as `abstractmethods` in the `Node` interface.

The idea here is that any type implementing `Node` should define 3 methods:

1) `resolve_id`, which should return the real id of the object to be used to generate the global id. I would love to allow the type to use an `id` directly, but then there's no way to have it use the resolver (at least not without some refactoring to allow that)

2) `resolve_nodes`, which is expected to return an iterable of available nodes so that the connection can paginate over it

3) `resolve_node`, which should return the node given its id. During the migration I though about removing this one, since we can basically call `resolve_nodes` and pass `node_ids=[node_id]`. What do you think?

tests/relay/schema.py
bellini666: I thought that someone might want to call this somewhere they don't have access to an `info` object, and objects usually don't require it to resolve their IDs. But it is still there for objects that might need them

I can remove the optional if you prefer and make it required

kaos: +1 on making it non-optional. It's easy enough to simply call another method from this one if you need to share the implementation with another callee that doesn't have `Info`.

In my use-case, Info is required, and marking it Optional only adds to my implementation to ensure it's not None to satisfy type checking.

tests/relay/schema.py
bellini666: The main situation is: The `NodeField` have `id` or `ids` depending on whether it is a list or not. For example:

```python
@strawberry.type
class Query:
    node: Node
    nodes: List[Node]
```

This will generate:

```graphql
type Query {
  node(id: GlobalID!): Node!
  nodes(ids: [GlobalID!]!): [Node!]!
}
```

When querying `nodes` the list of global IDs will be passed in there

But it can be useful as a utility function for end users too.

docs/guides/relay.md
bellini666: Also note that integrations can define those automatically.

For example, in strawberry-django-plus, when a type is created and it implements `Node`, [I inject those resolvers into the type with async-safe resolvers](https://github.com/blb-ventures/strawberry-django-plus/blob/main/strawberry_django_plus/type.py#L293) that queries the django model, and also applying the query optimizer, etc. So anyone using the django integration there can just inherit from `Node` and not worry about those methods (unless he wants to override them for some reason)

strawberry/types/type_resolver.py
bellini666: Inside `relay/types.py` the types defined there use `@strawberry.type`, which imports this file to resolve their fields and causes a circular import error.

Probably my comment above is not enough to explain the issue here =P. But when creating the types there, `Node`, `Edge` and `Connection`, this will raise an `ImportError`. After that the import should work fine

strawberry/relay/fields.py
bellini666: You got me! hahaha

This is a very dated comment inherited from my code in strawberry-django-plus. Originally this field had the possibility to receive a resolver, but soon after I did that I noticed that it doesn't make sense to have one, because it is just a matter of creating a `@strawberry.field` with a single argument of `id: GlobalID` in it and removed the functionality.

I'll remove this part of the doc

tests/relay/schema.py
bellini666: Oh, actually you question is when we would "not" get those.

When resolving a connection for example. The connection asks the `Node` for his nodes and paginate them. In this situation there isn't a list of nodes to be filtered, it just wants an iterable of "everything" (not actually everything, usually will be a lazy object that can be sliced) to slice.

strawberry/types/type_resolver.py
patrick91: this looks like something we need to refactor, either in this PR or another one :D

strawberry/relay/types.py
patrick91: oh I mean `arrayconnection`, wouldn't be best to use the name of the types we are going to return? we do have access to those, no?

strawberry/relay/fields.py
patrick91: yup!

docs/guides/relay.md
patrick91: > resolve_id, which should return the real id of the object to be used to generate the global id. I would love to allow the type to use an id directly, but then there's no way to have it use the resolver (at least not without some refactoring to allow that)

I think we should find a way to make this work 😊

For `resolve_nodes` and `resolve_node`, I think it would be nice to have only one, but at the same time I wonder if `resolve_nodes` is needed all the time, for example if you do something like this:

```python
@strawberry.type
class Website:
    id: strawberry.ID
    domain: str

    @classmethod
    def from_model(cls, model: models.Website) -> Self:
        return cls(id=model.pk, domain=model.domain)

    @strawberry.relay.connection
    async def feedbacks(
        self, info: Info[Context, None]
    ) -> strawberry.relay.Connection[Feedback]:
        feedbacks = await find_feedbacks_for_site(self.domain)

        return strawberry.relay.Connection(
            edges=[
                strawberry.relay.Edge.from_node(Feedback.from_model(feedback))
                for feedback in feedbacks
            ],
            page_info=strawberry.relay.PageInfo(
                has_next_page=False,
                has_previous_page=False,
                start_cursor="",
                end_cursor="",
            ),
        )
```

I'm already passing the nodes to the connection, right? happy to chat about this on discord too :)

strawberry/relay/types.py
bellini666: They are actually not used in type names, but in the cursor. For example:

```json
{
  "edges" [
    {
      "cursor": "arrayconnection:0",  # this
      "node": {
        "__typename": "Fruit",
        "name": "Banana"
      },
    },
    {
      "cursor": "arrayconnection:1",  # this
      "node": {
        "__typename": "Fruit",
        "name": "Apple"
      },
    },
  ]
}
```

It is used in those places, but it is exposed as `base64` (I put them raw in the example above to show where they are used).

The cursor has the pagination information, which is not necessarily related to the type itself. For the base implementation, which uses limit/offset, it will be the row number of the result. So, if your connection has a total number objects of `1000`, you would have cursors from `arrayconnection:0` to `arrayconnection:999`. Then when asking for `first: 10` and `after: arrayconnection:50` (but in base64) for example your results would have edges from `arrayconnection:51` to `arrayconnection:61`

So the first cursor in a connection of 2 different types will both be `arrayconnection:0` because they are both the rownumber 0 in that connection.

Regarding the name, it is also used in this example (https://relay.dev/docs/guides/graphql-server-specification/). If you decode the base64 inside the cursor, you can see that it also uses `arrayconnection` for it

docs/guides/relay.md
bellini666: > I think we should find a way to make this work :blush:

download (3)

> I'm already passing the nodes to the connection, right?

In that specific case I would argue that you could use `@strawberry.field` directly, because you are not using any of the "magic" behind `@strawberry.relay.connection`

The main use cases for the connection field are those:

1) Define the field as `feedbacks: Connection[Feedback]`. The field would be created with the `first`, `last`, `before` and `after` arguments, call `resolve_nodes` and use those arguments to properly paginate it.

2) Use `@strawberry.relay.connection` to return an `Iterable[Feedback]`, which is basically your way of giving the paginator your own nodes so that it doesn't have to call `resolve_nodes`. In the resolver you could even define some extra args for filtering/ordering the results, but forget about the pagination stuff (which is the `Connection`'s responsibility). Just like this example in the tests: https://github.com/strawberry-graphql/strawberry/pull/2511/files#diff-089578b7231144ec508d0023bb9562b896c8ef0e24ed10eabadfe7409028886dR161

A real example for this that I use: I have a django type that can be soft deleted. Since `resolve_nodes` would return the queryset with all objects, even those that were "deleted", i pre-filter the deleted ones in the resolver. For example:

```python
@strawberry.relay.connection
def fruits(self, include_removed: bool = False) -> Iterable[FruitType]:
    return Fruit.objects.filter(is_removed=not include_removed).order_by("name")
```

This would still generate this in the schema:

```graphql
type Query {
  fruits (
    first: Int
    last: Int
    before: String
    after: String
    includeRemoved: Bool! = false
  ): FruitTypeConnection!
}
```

So paginating over that connection would retrieve only results that are not removed (unless passing `includeRemoved: true`) and ordered by name

But having said all of that, I can totally not make them "mandatory", specially because the user can always use just the resolver approach, meaning `resolve_nodes` would not be used in this case. I can probably check when creating the schema that, when using the approach number 1 the `Node` implemented type has a `resolve_nodes` and give a warning to the user. What do you think?

> happy to chat about this on discord too :)

Sure! Always open to chat about anything on discord :)

docs/guides/relay.md
```diff
@@ -0,0 +1,381 @@
...
+                has_next_page=(
+                    last_edge is not None and bool(edges) and edges[-1] != last_edge
+                ),
+            ),
+        )
+
+
+@strawberry.type
+class Query:
+    fruits: FruitCustomPaginationConnection
```
patrick91, line 264: this is neat, but it will prevent to instantiate the type (query in this case, but can be another object type) without passing `fruits` :)

a workaround I found is this, but it's not nice:

```python
@strawberry.type
class MyObjectType:
    fruits: FruitCustomPaginationConnection = None
```

patrick91: I don't know if there's a solution for this other than maybe allowing:

```python
@strawberry.type
class MyObjectType:
    fruits: FruitCustomPaginationConnection = strawberry.relay.connection()
```

tests/relay/schema.gql
```diff
@@ -0,0 +1,189 @@
...
+"""A connection to a list of items."""
+type FruitAsyncConnection {
+  """Pagination data for this connection"""
+  pageInfo: PageInfo!
+
+  """Contains the nodes in this connection"""
+  edges: [FruitAsyncEdge!]!
+
+  """Total quantity of existing nodes"""
+  totalCount: Int
```
patrick91, line 34: can we remove this field by default? I don't think it's in the spec 😊
Maybe we can have specialised connections for this?

docs/guides/relay.md
```diff
@@ -0,0 +1,381 @@
...
+    ):
+        # Note that this is a showcase implementation and is far from
+        # being optimal performance wise
+        edges_mapping = {
+            to_base64("fruit_name", n.name): strawberry.relay.Edge(
+                node=n,
+                cursor=to_base64("fruit_name", n.name),
+            )
+            for n in sorted(nodes, key=lambda f: f.name)
+        }
```
patrick91, line 228: this is the best place for allowing to customise edges?

bellini666: Currently, yes

strawberry/types/type_resolver.py
bellini666: I refactored it to make the code better to understand

strawberry/relay/mutations.py
bellini666: Done! Will open another PR soon

Should it go to `strawberry/mutation.py` as an alternative mutation approach?

docs/guides/relay.md
bellini666: All resolved

bellini666:
Just pushed a lot of fixes/refactors. The major highlights are:

* The `id` of the type implementing `Node` can now be `id: <something>`. For example an int id can be defined as `id: relay.NodeID[int]` and the implementation will know that this field is the id. This means we can also do `pk: relay.NodeID[int]` for example without having to override `resolve_id`. Also is explained in the documentation

* The nodefield can be defined as `node: Node`, `node: Node | None`, `node: list[Node]` or `node: list[Node | None]` and it will resolve them accordingly. More details in the documentation that I wrote for this

* The only "required" method to define in a type inheriting from `Node` now is `resolve_nodes`. It is not actually mandatory because if someone just uses it for connections with custom resolvers and also doesn't have any node fields, then this method will not be used. It is mandatory for anyone defining a generic `Node` (any of the 4 ways above) or a `Connection` field without a custom resolver

* Improved typing for the `connection` decorator/dataclass descriptor to fix pyright issues

* Removed `InputMutation`. I'll open another PR with it tomorrow. I'm thinking about making it generic inside `strawberry.mutation.py`, as an alternative approach to the default mutation. What do you think?

cc @patrick91

bellini666:
More changes:

* Add support for node resolvers return async generators
* Removal of `totalCount`
* Fixes for the limit/offset pagination logic
* New `node_converter` argument that can be passed to a `ConnectionField` to convert the node returned by the nodes resolver into its correct type when needed

kaos:
Are there a timeline/plan for this and/or is there a pre-release with the latest changes? If so I could try them out to help smoke out any surprise issues..

patrick91:
@kaos we might want to merge this pr #2567 before ^^

I'll make a pre-release now, so you can test things 😊

kaos:
> 

awesome, will try it out this weekend.

docs/guides/relay.md
```diff
@@ -0,0 +1,394 @@
...
+        required: bool = False,
+    ):
+        if node_ids is not None:
+            return [fruits[nid] if required else fruits.get(nid) for nid in node_ids]
+
+        return fruits.values()
+
+
+# Assume we have a dict mapping the fruits code to the Fruit object itself
+fruits: Dict[int, Fruit]
```
jthorniley, line 60: I think the type annotation here should be `Dict[str, Fruit]` - because the node ID field in the example is str.

bellini666: `int` is actually fine because the globalid will do `str(id)` on it. Most ids are usually ints/uuids

bellini666: > So I think what I mean is either the dictionary should be str -> Fruit or the resolve_nodes class should have an int(nid) cast before using it as a key.

Indeed if the id was set to `str` it would work out of the box. But we can't just do `int(nid)` when passing the ids to `resolve_nodes` because we can't actually know the type of the ids when we resolve them.

e.g. You might have `FooType:1` and that `1` might be an integer or really a string containing the character `"1"`.

I don't know if there's a better way to workaround this, but I decided to go the easy way in this initial implementation and let the implementer cast the value. In this case, since you know that the id is an integer, you can do `int(nid)` inside the `resolve_nodes` function, like: `[fruits[int(nid)] if required else fruits.get(int(nid)) for nid in node_ids]`

But I'm open to suggestions

jthorniley: I think the way you have it in the interface is fine. I don't think relay specifies any particular format for the ID so as you say you can't know.  Its fine to leave it up to the implementer.

I just mean in the example in the docs it would make it clearer to either have the `int(nid)` cast in the code, or make the dict keys `str` - as it is type checkers will error on the lookup. I've just noticed in the very similar code in the tests - `schema.py` line 122, the ids are cast to str in the lookup dictionary which makes sense.

bellini666: Oh, regarding the docs you are totally right, will fix it! Thanks for pointing it out

jthorniley:
Hi, first off thanks for creating this it looks like a really nice addition! I've been looking into using it for our project based off the branch and it should be really useful.

I just wanted to comment on the `GlobalID`/`ID` issue.

I'm trying to migrate our project from graphene to strawberry and we have *many* Node types which I can't migrate all at once. I'm using the graphene/strawberry schema code linked from [this issue](https://github.com/strawberry-graphql/strawberry/issues/1042), which generally allows us to have both graphene and strawberry types in the schema.

So having the node interface change from the type being called `ID` to `GlobalID` is potentially a bit of an issue (assuming we get past any issues with compiling JS types), as our graphene types annotate the id field as `ID` not `GlobalID` (schema validation fails because the two kinds of "Node" think they are talking about different types, and generally I'd like to avoid schema changes as much as possible).

I've experimented with just setting the annotations to `ID` instead of `GlobalID` and doing some trivial manipulations to parse the types when needed. This seems to work for my case so I'm wondering if its a viable approach? The `GlobalID` dataclass can still be used as a utility to do the conversion between encoded string and type/id pair. 

My changes are [here](https://github.com/jthorniley/strawberry/compare/589d5ac3949a368387c73072399fe78e668353ab...jthorniley:strawberry:relay) - but it currently fails a lot of pytests fairly unsurprisingly, I"m just wondering if this seems like a good approach or are there other issues? (I haven't investigated how this affects connections at all yet).

docs/guides/relay.md
bellini666: Fixed!

bellini666:
@patrick91 I did a major refactor of this code and added almost 100% coverage to it. If you can make a new pre-release from it to test (and also allow other people following this PR to do the same)

Note: After the refactoring and tests I tried to convert the `NodeField`/`ConnectionField` implementation to use `FieldExtension`. I made it work nicely (it is in [this commit](https://github.com/bellini666/strawberry/commit/7c662fd0fe78e860bfa18905da45ff6c15895777)), but there is a small issue that I wanted to discuss with you before moving forward:

In the commit I mentioned I ported `NodeField` to use `FieldExtension`. The implementation there is complete with the exception of [node_conveter](https://github.com/strawberry-graphql/strawberry/pull/2511/files#diff-ee83873f7394d1df47f446d9e9fe7a2e270a1c88ab3aa4478118d47e2f936af3R328). How can we pass `node_converter` to the node field without having to define a `node` function?

I was thinking in two possibilities:

1) Move the `node_converter` functionality to the `StrawberryField` itself as `result_converter` (or some other better name). It would allow the converter to be used for other fields as well, which could be actually good since it could resolve the issue with integrations where users usually have to rely on `cast` to return their objects.

2) Expose an `config: Dict[str, Any]` field (or use `**kwargs) on `StrawberryField` to be able to pass extra configurations to the field which the field extension can later read from. The big disadvantage here is that typing would not work very well.

Personally I like 1) a lot, but don't know what do you think about it.

cc @erikwrede

kaos (commented):
I still have an issue with using the simple
```python
@strawberry.type
class Query:
    node: relay.Node
    types: relay.Connection[TargetType]
```

which gives me a schema of
```gql
type Query {
  node: Node!
  types: TargetTypeConnection!
}
```

While using this relay connection works perfectly fine!:
```python
    @relay.connection
    def relay_tgt_types(self, info: Info, query: Optional[TargetTypesQuery] = None) -> Iterator[TargetType]:
        request_state = GraphQLContext.request_state_from_info(info)
        return TargetTypesQuery.filter(
            query,
            (
                TargetType.from_help(info)
                for info in request_state.all_help_info.name_to_target_type_info.values()
            ),
        )
```

I'll see if I can debug this further to isolate the root cause for this..

docs/guides/relay.md
```diff
@@ -0,0 +1,396 @@
...
+@strawberry.type
+class Fruit:
+    name: str
+    weight: str
+```
+
+We want it to have a globally unique ID, a way to retrieve a paginated results
+list of it and a way to refetch if if necessary. For that, we need to inherit it
+from the `Node` interface and implement its abstract methods: `resolve_id`,
+and `resolve_nodes`.
```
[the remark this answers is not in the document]
bellini666: Nice catch!

docs/guides/relay.md
```diff
@@ -0,0 +1,396 @@
...
+
+```python
+from strawberry import relay
+
+
+@strawberry.type
+class Fruit(relay.Node):
+    code: relay.NodeID[int]
```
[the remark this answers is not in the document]
bellini666: Sure, I think it is nice to explain how `NodeID` is translated here.

But regarding the issue, I could not reproduce it. The tests in this PR are all using the annotation without the `= strawberry.field(...)`. Can you maybe give me a minimal reproduction example to try here?

bellini666:
> I still have an issue with using the simple

When using just the annotation without a resolver, it will use the type's `resolve_nodes` to get the nodes, and `nodes_id` will be `None` because it is asking for everything

docs/guides/relay.md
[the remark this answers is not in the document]
kaos: I'll see if can reproduce this in a simpler setup.

mayteio:
What a rollercoaster ride reading the [issue started over a year ago](https://github.com/strawberry-graphql/strawberry/issues/1573) to now! The more the dates got more recent, the more excited I got! I appreciate your commitment to seeing this through. Pumped to see this released 🚀

jkimbo:
Btw @bellini666 and I had a discussion on discord about this PR (https://discord.com/channels/689806334337482765/1093176956524167258) and I've put forward a slightly tweaked API which you can see here: https://gist.github.com/jkimbo/3118b4bde65e67540d6cd45b9677cab4

Feedback welcome

bellini666:
The PR is updated with the following changes:

1) It is now using field extensions instead of a custom field implementation

2) Lots of changes to the API based on [this feedback](https://gist.github.com/jkimbo/3118b4bde65e67540d6cd45b9677cab4) from @jkimbo . Regarding those:

- The `Connection` type needs to be explicit defined for each connection field together with a resolver. The connection type defines the pagination that will be used and the resolver the nodes to be paginated. For example:

```python
@strawberry.type
class Query:
    # It can be defined like this
    fruits: relay.ListConnection[Fruit] = relay.connection(resolver=get_fruits)

    # Or with a resolver
    @relay.connection(relay.ListConnection[Fruit])
    def fruits_with_resolver(self) -> Iterable[Fruit]:
        ...
```

- As you might have noticed above, I used `relay.ListConnection` instead of `relay.Connection`. The `Connection` is now an abstract implementation that cannot be used directly, but instead their subclasses should implement the pagination logic. `ListConnection` is one implementation provided by this integration that paginates nodes using a limit/offset approach.

- @jkimbo suggested to only have a `resolve_node` in the `Node` type to be defined by subclasses. I still kept the `resolve_nodes` as well because the integration supports not only `node: relay.Node` but also `nodes: List[relay.Node]`. When using a list you can ask for a list of node ids to be resolved (they don't even need to be of the same type) and their corresponding nodes will be returned.

I changed the `nodes_id` to not be optional anymore, which was the case of the previous integration. All you need to do in the `Node` subclass is to define a `resolve_nodes` classmethod for it to work. You can also override `resolve_node` for a custom, more performatic implementation. By default it calls `resolve_nodes` passing a list with a single element (the node id) to it.

3) I rewrote a good amount of the docs to reflect the changes in the API. I hope I covered most of the use cases. As always, I'm not the best at writing docs, so feel free to suggest whatever changes you find necessary :)

Hopefully we can get this merged soon! :)

cc @patrick91 @jkimbo @erikwrede

docs/guides/pagination/overview.md
```diff
@@ -117,6 +117,14 @@ need a reliable and consistent way to handle pagination.
 
 ### Cursor based pagination
 
+<Note>
+
+Strawberry provides a cursor based pagination implementing the
+[relay spec](https://relay.dev/docs/guides/graphql-server-specification/).
+You can read more about it in the [relay](./input-types) page.
+
+<Note/>
```
patrick91, line 126:
```suggestion
</Note>
```

nrbnlulu:
Thanks @bellini666 !!!

kaos:
amazing!!

woojing:
Great job! Thanks you @bellini666 and whole reviewers :)

patrick91:
@lostclus can you open an issue for this? 

we were discussing some ideas around it on discord, but I think it is worth tracking it here on GitHub too :)

PR #2517, opened 2023-02-07
feat: support specialized generic types

### LINKED ISSUE #1195: TypeVar locking in types inherited from generics
Currently it's impossible to inherit from generic types and lock/fix their type variables
According to python docs it should be possible to do so:
```py
T = TypeVar('T')

class MyDict(Mapping[str, T]):
    ...
```

I created a repository with a reproducable example:
https://github.com/ThirVondukr/strawberry-generics-issue
```py
from typing import Generic, TypeVar

import strawberry

Node = TypeVar("Node")
Cursor = TypeVar("Cursor")


@strawberry.interface
class Edge(Generic[Node, Cursor]):
    node: Node
    cursor: Cursor


@strawberry.type
class SpecificEdge(Edge[str, str]):
    pass


@strawberry.type
class Root:
    edge: Edge[int, int]
    specific_edge: SpecificEdge


schema = strawberry.Schema(query=Root)
```
Expected behavior: `specificEdge` on schema having `{node: String!, cursor: String!}` type,
currently error occurs.

bellini666:
Fix #1195


RELEASE.md
```diff
@@ -0,0 +1,49 @@
+Release type: minor
+
+This release adds support for specialized generic types.
+Before, the following code would give an error, saying that \_T was not
```
```suggestion
Before, the following code would give an error, saying that `T` was not
```

strawberry/schema/name_converter.py
```diff
@@ -53,7 +53,7 @@ def from_argument(self, argument: StrawberryArgument) -> str:
         return self.get_graphql_name(argument)
 
     def from_object(self, object_type: TypeDefinition) -> str:
-        if object_type.concrete_of:
+        if object_type.concrete_of and not object_type.is_generic_specialized:
```
patrick91, line 56: could we do this check inside `from_generic`?

strawberry/types/type_resolver.py
```diff
@@ -138,6 +139,19 @@ class if one is not set by either using an explicit strawberry.field(name=...) o
             origin = origins.get(field.name, cls)
             module = sys.modules[origin.__module__]
 
+            if isinstance(field_type, TypeVar):
+                specialized_type_var_map = get_specialized_type_var_map(cls)
+                # If field_type is specialized and a TypeVar, replace it with its
```
patrick91, line 144: `replace it with its` I think a word is missing here :D

strawberry/types/types.py
```diff
@@ -113,6 +114,16 @@ def fields(self) -> List[StrawberryField]:
     def is_generic(self) -> bool:
         return is_type_generic(self.origin)
 
+    @property
+    def is_generic_specialized(self) -> bool:
```
patrick91, line 118:
```suggestion
    def is_specialized_generic(self) -> bool:
```

I think this sounds better 😊

botberry:



---

This release adds support for specialized generic types.
Before, the following code would give an error, saying that \_T was not
provided to the generic type:

```python
@strawberry.type
class Foo(Generic[T]):
    some_var: T

@strawberry.type
class IntFoo(Foo[int]):
    ....

@strawberry.type
class Query:
    int_foo: IntFoo
```

Also, because the type is already specialized, `Int` won't get inserted to its name,
meaning it will be exported to the schema with a type name of `IntFoo` and not
`IntIntFoo`.

For example, this query:

```python
@strawberry.type
class Query:
    int_foo: IntFoo
    str_foo: Foo[str]
```

Will generate a schema like this:

```graphql
type IntFoo {
  someVar: Int!
}

type StrFoo {
  someVar: String!
}

type Query {
  intFoo: IntFoo!
  strfoo: StrFoo!
}
```



Here's the tweet text: 
```

```

strawberry/types/types.py
```diff
@@ -125,7 +136,7 @@ def is_implemented_by(self, root: Union[type, dict]) -> bool:
         # TODO: Accept StrawberryObject instead
         # TODO: Support dicts
         if isinstance(root, dict):
-            raise NotImplementedError()
+            raise NotImplementedError
```
patrick91, line 139: 👀

strawberry/utils/inspect.py
```diff
@@ -14,3 +15,88 @@ def get_func_args(func: Callable[[Any], Any]):
...
+) -> Optional[Dict[TypeVar, Union[TypeVar, type]]]:
+    ...
+
+
+@overload
+def get_specialized_type_var_map(
+    cls: type,
+    *,
+    # NOTE: Union[bool, Literal[False]] is to make both mypy/pyright happy
+    include_type_vars: Union[bool, Literal[False]] = ...,
```
patrick91, line 34: what error where they throwing? :O

patrick91 (commented):
This is awesome work! I'm sure @ThirVondukr will love it!

RELEASE.md
```diff
@@ -0,0 +1,49 @@
...
+    ....
+
+@strawberry.type
+class Query:
+    int_foo: IntFoo
+```
+
+Also, because the type is already specialized, `Int` won't get inserted to its name,
+meaning it will be exported to the schema with a type name of `IntFoo` and not
+`IntIntFoo`.
```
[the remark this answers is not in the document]
bellini666: Sure, will do that!

strawberry/types/types.py
bellini666: I think my `ruff` autoformatter did this =P. It is a flake8 plugin rule that it might have implemented in a recent version

patrick91: ah, nice :D

strawberry/utils/inspect.py
bellini666: If I just use `Literal[False]`, the tests will have some pyright errors because I have a parametrize value of type `bool` and it says that `bool` cannot be assigned to `Literal[True] | `Literal[False]`

If I try to make pyright happy and define it like this:

```python
@overload
def get_specialized_type_var_map(
    cls: type,
    *,
    include_type_vars: Literal[True],
) -> Optional[Dict[TypeVar, Union[TypeVar, type]]]:
    ...


@overload
def get_specialized_type_var_map(
    cls: type,
    *,
    include_type_vars: Literal[False],
) -> Optional[Dict[TypeVar, type]]:
    ...


@overload
def get_specialized_type_var_map(
    cls: type,
    *,
    include_type_vars: bool = ...,
) -> Optional[Dict[TypeVar, Union[TypeVar, type]]]:
    ...
```

Mypy gives this warning:

```
E   pytest_mypy_plugins.utils.TypecheckAssertionError: Invalid output: 
E   Actual:
E     ../../home/bellini/dev/strawberry/strawberry/utils/inspect:30: error: Overloaded function signatures 2 and 3 overlap with incompatible return types  [misc] (diff)
E     main:9: note: Revealed type is "Union[builtins.str, None]" (diff)
E   Expected:
E     main:9: note: Revealed type is "Union[builtins.str, None]" (diff)
E   Alignment of first line difference:
E     E: main:9: note: Revealed type is "Union[builtins.str, None]"...
E     A: ../../home/bellini/dev/strawberry/strawberry/utils/inspect:30: error: Ov...
E        ^
```

I just noticed that when I added the generic clause, I mistakenly removed one of the overloads. I'm looking at it right now to try to solve it properly because when adding the overload again, even though pyright was fine with it, mypy wasn't

bellini666: Ok, found the issue. I assumed `Dict[TypeVar, type]` and `Dict[TypeVar, Union[TypeVar, Type]]` would be compatible, but they are not. The solution was to make a union of them both as the return

```python
@overload
def get_specialized_type_var_map(
    cls: type,
    *,
    include_type_vars: Literal[True],
) -> Optional[Dict[TypeVar, Union[TypeVar, type]]]:
    ...


@overload
def get_specialized_type_var_map(
    cls: type,
    *,
    include_type_vars: Literal[False] = ...,
) -> Optional[Dict[TypeVar, type]]:
    ...


@overload
def get_specialized_type_var_map(
    cls: type,
    *,
    include_type_vars: bool,
) -> Optional[
    Union[Optional[Dict[TypeVar, type]], Dict[TypeVar, Union[TypeVar, type]]]
]:
    ...
```

I'll send the fix together with some other comments soon

strawberry/types/type_resolver.py
bellini666: lol

strawberry/schema/name_converter.py
bellini666: Sure. Just noticed that this change was not required. Removing it

patrick91 (approved):
Awesome work!

Issue #2187, opened 2022-09-21
Idea: Support for TypedDicts


patrick91:
Python's type system provides a way to type dictionaries, using `TypedDict`, for example:

```python

GabrielTDS-dev:
I'm going to work on this feature along with @Ladol. I'll give updates on its implementation soon once a good solution is produced.

PR #2565, opened 2023-02-18
feat: support JSON type hints

### LINKED ISSUE #2513: support JSON type hints
Right now you can accept arbitrary metadata using the `JSON` scalar, which is great. But sometimes the json has a specific structure, for example: `JSON[dict[str, int]]` and it would be great if you could add type hints like that to validate the incoming data. 

This could hopefully also get around the linting errors like

```
Expression of type "dict[str, str]" cannot be assigned to declared type "JSON"
```

bellini666:
Fix #2513

### DIFF
--- a/strawberry/scalars.py
+++ b/strawberry/scalars.py
@@ -1,18 +1,35 @@
 from __future__ import annotations
 
 import base64
-from typing import TYPE_CHECKING, Any, Dict, NewType, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    Generic,
+    NewType,
+    Union,
+)
+from typing_extensions import TypeVar
 
 from .custom_scalar import scalar
 
 if TYPE_CHECKING:
     from .custom_scalar import ScalarDefinition, ScalarWrapper
 
-
+T = TypeVar("T", bound=Any, default=Any)
 ID = NewType("ID", str)
 
-JSON = scalar(
-    NewType("JSON", object),  # mypy doesn't like `NewType("name", Any)`
+
+# This is required for mypy/pyright identify JSON[T] as T and not JSON[T]
+# They will respect only the typings on `__getitem__` defined in the metaclass
+# and ignore `__class_getitem__` on both
+class JSONMeta(type):
+    def __getitem__(self, item: T) -> T:
+        return self.__class_getitem__(item)  # type: ignore
+
+
+@scalar(
+    name="JSON",
     description=(
         "The `JSON` scalar type represents JSON values as specified by "
         "[ECMA-404]"
@@ -24,6 +41,9 @@ JSON = scalar(
     serialize=lambda v: v,
     parse_value=lambda v: v,
 )
+class JSON(Generic[T], metaclass=JSONMeta):
+    ...
+
 
 Base16 = scalar(
     NewType("Base16", bytes),
--- a/tests/pyright/test_scalars.py
+++ b/tests/pyright/test_scalars.py
@@ -4,14 +4,25 @@ pytestmark = [skip_on_windows, requires_pyright]
 
 
 CODE = """
+from typing import Any, Dict
+from typing_extensions import TypedDict
 import strawberry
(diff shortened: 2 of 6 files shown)


botberry:



---

This release adds support for typing `JSON` values. For example, the following
can now be used to get better typing support on it:

```python
from typing import TypedDict
import strawberry


class SomeJSON(TypedDict):
    foo: int
    bar: str


@strawberry.type
class SomeType:
    json: JSON[SomeJSON]
```

Note that the scalar will still be exposed as `JSON` in the schema and strawberry
will not validate its value at runtime.


---




Here's the tweet text: 
```

```

tests/pyright/test_scalars.py
```diff
@@ -38,35 +57,57 @@ def test_pyright():
 
     # NOTE: This is also guaranteeing that those scalars could be used to annotate
     # the attributes. Pyright 1.1.224+ doesn't allow non-types to be used there
+    __import__("pprint").pprint(results)
```
BryceBeagle, line 60: Why are you using `__import__` here?

tests/schema/test_scalars.py
```diff
@@ -226,6 +227,146 @@ def echo_json_nullable(data: Optional[JSON]) -> Optional[JSON]:
     }
 
 
+def test_json_with_type_hints():
```
BryceBeagle, line 230: Could you possibly put these files in another file (say, `tests/schema/test_json_types.py`)?

strawberry/scalars.py
```diff
@@ -1,18 +1,35 @@
...
 
-JSON = scalar(
-    NewType("JSON", object),  # mypy doesn't like `NewType("name", Any)`
+
+# This is required for mypy/pyright identify JSON[T] as T and not JSON[T]
+# They will respect only the typings on `__getitem__` defined in the metaclass
+# and ignore `__class_getitem__` on both
+class JSONMeta(type):
+    def __getitem__(self, item: T) -> T:
+        return self.__class_getitem__(item)  # type: ignore
```
BryceBeagle, line 28: If I understand your problem properly, can you use a Generic Protocol instead?

tests/pyright/test_scalars.py
bellini666: loool, forgot to remove. This is just a macro I have to print things for debugging using `pprint`. In this case, I wanted to get the real output to replace here

tests/schema/test_scalars.py
bellini666: Sure

strawberry/scalars.py
bellini666: Will try here, makes sense

bellini666: I tried, but using a `Protocol` here, both as a multiple inheritance for `JSONMeta` and also to type `self: GenericProtocol` produced other errors (specially for pyright) and made it stop working.

I tried some other approaches, but couldn't get it to work. Do you have any suggestions that I can try here?

bellini666:
Thought of this PR today and had an idea.

What if we allow to use `TypedDicts` for annotations and expose them as `JSON` to the schema?

The pros is that it is a straightforward way of doing that. We can even add some kind of validation to the `TypedDict` for (de)serialization if we want.

The only con I can think right know is that it is not obvious that the typeddict will become a JSON in the schema, although maybe not a major issue?

I still want to try to fix this PR soon (both pyright and mypy evolved a lot since I created it), but in case it is still not possible that can be another approach.

What do you think @patrick91 ?

patrick91:
> Thought of this PR today and had an idea.
> 
> What if we allow to use `TypedDicts` for annotations and expose them as `JSON` to the schema?

I think I want to be able to support typed dicts as return types, like so: https://github.com/strawberry-graphql/strawberry/issues/2187

r614:
@patrick91 both this + #2187 haven't been updated in a while, was wondering if either of these is going to get merged? 

For reference, I have a use case where I need to return a JSON type of `{ aribitraryKey: CustomStruct() }` or the JS/TS equivalent of `Record<str, CustomStruct>`.

Issue #2168, opened 2022-09-13
Custom field API discussion


patrick91:
We've been discussing the API for custom fields quite a bit on discord, I'll summarize some of the thoughts here.

# Goal

We want to allow users of Strawberry to leverage code to change how fields work and **what** they return. 

Some use cases:

- Relay pagination
- Authenticated only fields (they would return `None` if users are not logged in)

## Worries

My personal worry is this might make it harder to understand how the final schema will look like, but I might be worrying too much 😊

# How it should work

In #920 we have a proposal that uses a field class to update information about the field. I was also thinking we could try to leverage python's dynamic features to do the same thing. Let's see both approaches:

# API 1

```python
class PaginatedField(StrawberryField):
    def get_arguments(self) -> dict[str, type]:
        arguments = super().get_arguments()
        if "first" not in arguments:
            arguments["first"] = int
        return arguments

    def get_result(
        self, source: Any, info: Any, arguments: dict[str, Any]
    ) -> Awaitable[Any] |  Any:
        first = arguments.pop("first")
        result = super().get_result(arguments, source, info)

        return result[:first]

@strawberry.type
class Query:
    @PaginatedField()
    def books() -> list[str]:
        return [
            "Pride and Prejudice",
            "Sense and Sensibility",
            "Persuasion",
            "Mansfield Park",
        ]
```

# API 2

```python
def paginated(function: Callable[[...], T]):
    @strawberry.wraps(function)
    def paginated_function(root: Any, info: Any, first: int, **kwargs) -> T:
        # strawberry.wraps will change the original function
        # to always accept root and info
        result = function(root, info=info, **kwargs)
        return result[:first]

    return paginated_function

@strawberry.type
class Query:
    @strawberry.field
    @paginated
    def books() -> list[str]:
        return [
            "Pride and Prejudice",
            "Sense and Sensibility",
            "Persuasion",
            "Mansfield Park",
        ]
```

I'm not 100% sure if this API is doable, but it might be worth trying

---

I haven't added any pro and cons for both APIs mostly because I want to hear other people's opinion first 😊

nrbnlulu:
### API 1 - Extending `StrawberryField`:

| Pros                                               | Cons                                       |
|----------------------------------------------|----------------------------------------|
| more understand able API, no need for rocket since to know how strawberry calls it | decorating with a class is awkward |
| allows to modify everything you can in one place |
| allows inheriting other custom fields and modifying upon them |

[...]

PR #2567, opened 2023-02-18
feat: add field extension support

erikwrede:
- fix(test): correct pytest.raises exception matching
- feat: add field extension support

## Description

Adds support for field extensions using the approach specified in PR #2168. Field extensions can support `sync` ,`async` or both. 

However, `sync` and `async` extensions cannot be mixed on the same field: The result of an async resolver would have to be awaited before calling the sync extension, making it impossible for the sync extension to modify any arguments before calling the resolvers.

`async-only` extensions can be applied to `sync` resolvers, because we automatically wrap sync resolvers in a `SyncToAsyncExtension`.

Additional checks to verify no async extensions are used in a sync server (django, flask) may be added in a later PR.
## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

* #2168

## Checklist


strawberry/extensions/field_extension.py
```diff
@@ -0,0 +1,105 @@
...
+    Inserts a SyncToAsyncExtension to be able to
+    use Async extensions on sync resolvers
+    Throws a TypeError otherwise.
+
+    Returns True if resolving should be async, False on sync resolving
+    based on the resolver and extensions
+    """
+    if not field.extensions or not len(field.extensions):
+        return field.is_async  # pragma: no cover
+
```
bellini666, line 62: I think we are not adding the `SyncToAsync` compatibility between all extensions that require one. What do you think about doing something like this to support that:

```python
supports_sync = [extension.supports_sync for extension in field.extensions] + [not field.is_async]
supports_async = [extension.supports_async for extension in field.extensions] + [field.is_async]

previous: Optional[Tuple[bool, bool]] = None
for i, supports_sync, supports_async in enumerate(zip(supports_sync, supports_async)):
    if previous is not None:
        previous_supports_sync, previous_supports_async = previous

        # When the field is async, all extensions must support async as well
        if field.is_async and not supports_async:
            non_async_extensions = ', '.join(
                e.__name__ for e in field.extensions if not e.supports_async
            )
            raise TypeError(
                f"Cannot add sync-only extension(s) {non_async_extension_names} "
                f"to the async resolver of Field {field.name}. "
                f"Please add a resolve_async method to the extension(s)."
            )

        # When previous is sync only and the current extension is async only,
        # there's no way 
        previous_sync_only = previous_supports_sync and not previous_supports_async
        async_only = supports_async and not supports_sync
        if previous_sync_only and async_only:
            # If this was the field itself, we would have raised TypeError
            # in the condition above.
            if i == len(field.extensions):  # pragma: no cover
                raise AssertionError("This should never happen.")

            extension = field.extensions[i]
            previous_extension = field.extensions[i - 1]
            raise TypeError(
                f"sync-only extension {previous_extension.__name__} followed by "
                f"async-only extension {extension.__name__}."
            )

        if previous_supports_async and not supports_async:
            field.extensions.insert(i - 1, SyncToAsyncExtension())


    previous = (supports_sync, supports_async)
```

This should cover all cases, except one: You have an extension that supports both sync and async, followed by a sync only in a sync field. For `schema.execute` it would work correctly, but `schema.execute_sync` would give an error because `SyncToAsync` doesn't have a `resolve` method. This can be fixed by adding one there:

```python
class SyncToAsyncExtension(FieldExtension):
    """Helper class for mixing async extensions with sync resolvers.
    Applied automatically"""

    def resolve(
        self, next: Callable[..., Any], source: Any, info: Info, **kwargs
    ) -> Any:
        return next(source, info, **kwargs)

    async def resolve_async(
        self, next: Callable[..., Any], source: Any, info: Info, **kwargs
    ) -> Any:
        return next(source, info, **kwargs)
```

strawberry/schema/schema_converter.py
```diff
@@ -524,17 +526,44 @@ def _get_result(_source: Any, info: Info, **kwargs):
...
+        def wrap_field_extensions() -> Callable[..., Any]:
+            """Wrap the provided field resolver with the middleware."""
+
+            if field.extensions is None:
+                return _get_result
+
+            for extension in field.extensions:
+                extension.apply(field)
+
+            use_async = ensure_field_extension_compatibility(field)
```
bellini666, line 538: I might be mistaken here, but in case we have a field that supports both sync and async this would force the execution to call `resolve_async`, which would be wrong when using `schema.execute_sync`

Probably we need a flag here and wrap it 2 times:

- One for sync, which should also ensure that all extensions support `execute` (even the field itself)
- One for async that doesn't matter if it calls `resolve` or `resolve_async` because the compatibility itself will handle all incompatibilities.

bellini666 (changes requested):

strawberry/schema/schema_converter.py
erikwrede: `use_async` only returns true if there is some extension or an async resolver that forces the use of async resolving. In case everything can be done sync (resolver is sync and extensions all support sync), `use_async` will always be `False`.

strawberry/extensions/field_extension.py
erikwrede: Having the ability to run sync-only extensions on a sync field, then followed by async only extensions is useful! I was worried it was going to add too much complexity, but this seems great, open to adapting it.

I don't exactly understand the case to add `resolve` to the `SyncToAsyncExtension`. This is only used to convert the sync result of a resolver or a sync extension to an `Awaitable`, so that the following `resolve_async` gets the result it expects. All of this is computed before execution of a request, during schema conversion, so we have a static contract of this only being used while converting a sync result to an async.

erikwrede: First point resolved in latest commit ☺️

docs/guides/field-extensions.md
```diff
@@ -0,0 +1,170 @@
...
+import strawberry
+from strawberry.extensions import FieldExtension
+
+
+class UpperCaseExtension(FieldExtension):
+    async def resolve(
+        self, next: Callable[..., Awaitable[Any]], source: Any, info: Info, **kwargs
+    ):
+        result = await next(source, info, **kwargs)
+        return str(result).upper()
```
jkimbo, line 156:
```suggestion
    def resolve(
        self, next: Callable[...,Any], source: Any, info: Info, **kwargs
    ):
        result = next(source, info, **kwargs)
        return str(result).upper()
```

docs/guides/field-extensions.md
```diff
@@ -0,0 +1,170 @@
...
+query {
+    string
+}
+---
+{
+  "string": "THIS IS A TEST!!"
+}
+```
+
+## Modifying the field
```
jkimbo, line 50: I think it's worth adding a warning here about `StrawberryField` since it's the first time that we're exposing it publicly and it's got some weird edges. We also (in the future) need ways to easily change the type of the field and modify its arguments.

jkimbo (changes requested):
This is really great! 👏  Just added a couple of suggestions for the docs but otherwise I say we ship it.

docs/guides/field-extensions.md
erikwrede: Do you have any pointers on what could be worthwhile to warn?

jkimbo: Maybe just something like:

Warning: the StrawberryField API is not stable and might change in the future without warning.

erikwrede: Done ☺️

jkimbo (approved):

bellini666 (approved):
LGTM 👍🏼 

There are some comments that I made at the other PR, but they are mostly for the permissions integration.

RELEASE.md
```diff
@@ -0,0 +1,36 @@
...
+
+
+@strawberry.type
+class Query:
+    @strawberry.field(extensions=[UpperCaseExtension()])
+    async def string(self) -> str:
+        return "This is a test!!"
+```
+
+```gql
```
patrick91, line 26:
```suggestion
```graphql
```

docs/guides/field-extensions.md
```diff
@@ -0,0 +1,172 @@
...
+        return "This is a test!!"
+```
+
+In this example, the `UpperCaseExtension` wraps the resolver of the `string`
+field (`next`) and modifies the resulting string to be uppercase.
+The extension will be called instead of the resolver and receives
+the resolver function as the `next` argument. Therefore, it is important
+to not modify any arguments that are passed to `next` in an incompatible way.
+
+```gql+json
```
patrick91, line 40:
```suggestion
```graphql+response
```

docs/guides/field-extensions.md
```diff
@@ -0,0 +1,172 @@
...
+        return do_expensive_computation()
+---
+type Client {
+    analyzedHours: Int! @Cached(time=200)
+}
+```
+
+## Combining multiple field extensions
+
+When chaining multiple field extensions, the outermost extension is called first,
```
patrick91, line 108: do you think we could replace `outermost` with last?

or could we make it so that the order is the same as the one you pass? I think the current behaviour might be confusing 😊

docs/guides/field-extensions.md
```diff
@@ -0,0 +1,172 @@
...
+import strawberry
+from strawberry.extensions import FieldExtension
+
+
+class UpperCaseExtension(FieldExtension):
+    def resolve(self, next: Callable[..., Any], source: Any, info: Info, **kwargs):
+        result = next(source, info, **kwargs)
+        return str(result).upper()
+
+    async def resolve_async(
```
patrick91, line 160: we might want to make a decision on the names here, in our schema we have schema.execute and schema.execute_sync, could do we do the same here to keep consistency?
/cc @BryceBeagle @jkimbo

erikwrede: To clarify why I chose to "flip" the naming: having `resolve` with async as the default case may cause more use of async-only extension on synchronous fields, because the naming implies that `resolve_sync` is the special case that is only needed when doing sync execution (even though the performance disadvantages of converting every resolver to async are stated in the docs - but we know that not everyone reads them). When the sync case is the default case, it might encourage more to implement the sync extension, too. 
That being said, I'm all for consistency, so I'm fine with adjusting to what you decide for ☺️

docs/guides/field-extensions.md
patrick91: let's use mermaid, it should work on github too: https://mermaid.js.org

erikwrede: Something like this?: https://mermaid.live/edit#pako:eNqNkVFLwzAQx79KOF8UWthzB0Kr86mipN2DLDJCe3XFNClpqi1j391rbJkOBPOU3O_3vwvJEQpTIkTwZmV7YClfC81odW5UyLI85jmraqWiqwe_VquAmDXveK5cRJ6e_5uY3B17XQ40aycAByx6VxvNCqMdDk7AbMREt22L9k52uBkc6o40wnO7hHhqPv_knDjHzqgPtOeqnxuGAuw32lc1qvL6RkAY3sazFE_C4wvjm2yb5h753E-s6a77OZcsV5pAMzLq3Sv3q2VymVmenvtZI-PnTLKGABq0jaxL-qvjJApwB2xQQETbEivpZaFPpMremWzUBUTO9hhA35bS4X0t6ZcbiCqpOjx9AUu0mVA

patrick91: yup, maybe vertical, as it might be easier to read on mobile 😊

erikwrede: Does the strawberry website Markdown support html with media queries? I think it's much easier to read horizontally on desktop, but much better vertically on mobile so best to have both, what do you think? 👀

erikwrede: Ok I just realized that doesn't work because the content inside those tags (the diagram in this case) won't be parsed by markdown then.

patrick91: @erikwrede we could make a custom component for that if it helps

or add a directive, like

````
```mermaid (mobile:vertical)
````

if it is easy to change the setting 😊

erikwrede: Ok it actually looks just fine on the github preview. The strawberry preview doesn't seem to update though.

patrick91: it's probably because we don't support mermaid yet on the site :D I'll see if I can add support to the current design in the next couple of days 😊

erikwrede: We could also add the graph later when the new website is ready

patrick91 (approved):
This is awesome 😊

jkimbo:
🎉

Issue #2406, opened 2022-12-19
A way to specify return value if permission fails.


MeRuslan:
Would be quite useful, IMO.
I would prefer to have an option to return some value in a restricted fields in case of a permission fail.

Possible solution:
A new permission base class
```
class QuietBasePermission(BasePermission):
    default_value: Any = None

```

And some additional logic to handle it in GraphQLCoreConverter: 
```
        async def _check_permissions_async(
            source: Any, info: Info, kwargs: Dict[str, Any]
        ) -> Optional[QuietBasePermission]:
            failed_perm = None
            for permission_class in field.permission_classes:
                permission = permission_class()
                has_permission: bool

                has_permission = await await_maybe(
                    permission.has_permission(source, info, **kwargs)
                )

                if not has_permission and not isinstance(permission, QuietBasePermission):
                    message = getattr(permission, "message", None)
                    raise PermissionError(message)
                elif not has_permission:
                    failed_perm = permission
            return failed_perm
....
        async def _async_resolver(_source: Any, info: GraphQLResolveInfo, **kwargs):
            strawberry_info = _strawberry_info_from_graphql(info)
            perm_failed = await _check_permissions_async(_source, strawberry_info, kwargs)
            if perm_failed:
                return perm_failed.default_value

            return await await_maybe(_get_result(_source, strawberry_info, **kwargs))
```

It returns the last default value for a permission check fail if there are multiple, and raises a PermissionError if required by some other permission.

MeRuslan:
At the first glimpse I thought permission fail on a field ruined the whole query, which is not the case, nonetheless, a default value for permission fail seems useful to me.

patrick91:
Potentially done in https://github.com/strawberry-graphql/strawberry/releases/tag/0.321.0

Issue #2399, opened 2022-12-13
Allow errors to set HTTP status codes


patrick91:
### Discussed in https://github.com/strawberry-graphql/strawberry/discussions/2304

<div type='discussions-op-text'>

<sup>Originally posted by **ShourovRoy** November  2, 2022</sup>
Apollo Graphql has a function to modify status code on error that can be overridden which plays an important role for implementing refresh token. Is there any way to do that here?</div>

Apollo server allows errors to change the status code, maybe we can do the same:

```
import { GraphQLError } from 'graphql';

const resolvers = {
  Query: {
    someField() {
      throw new GraphQLError('the error message', {
        extensions: {
          code: 'SOMETHING_BAD_HAPPENED',
          http: {
            status: 404,
            headers: new Map([
              ['some-header', 'it was bad'],
              ['another-header', 'seriously'],
            ]),
          },
        },
      });
    },
  },
};
```

https://www.apollographql.com/docs/apollo-server/data/errors/#setting-http-status-code-and-headers

erikwrede:
I'm somewhat critical of setting the http status right in the base GraphQL error, as the Schema is transport-agnostic. Maybe we can create a new Http or StatusCode GraphQLError with a middleware for the HTTP/Websocket/(in the future SSE) views that injects the code based on the new Error class.

PR #2570, opened 2023-02-18
refactor: Permissions using Field Extensions

### LINKED ISSUE #244: Improve permissions support
Right now permissions are working well, but I think we can improve them a bit. 

- [ ] (1) Return an exception or warning if field with permission classes is not optional
- [ ] (2) Right now we return errors for each field that you don't have permissions to (see screenshot)

For (1) we currently don't do anything if the field that has permission classes is not optional, this breaks when we don't return anything since GraphQL core checks if non-nullable fields have null values. We should either update the return type to make it optional or throw a warning/exception as soon as possible so that this can be fixed.

For (2) I'm a bit conflicted, as the current implementation makes sense in a way, but I'm wondering if we should provide a way to not throw the errors. Any idea @marcoacierno @jaydenwindle?

image

### LINKED ISSUE #2454: Update how we handle permissions by creating an extensions
After @nrbnlulu's work on extensions in #2428 I think we can update how we handle permissions to use an extension rather than adding the check in our custom resolver.

On top of that we could also skip adding this extension if permission classes aren't used :)

erikwrede:
- feat: add field extension support
- fix: revert imports broken by pycharm
- [pre-commit.ci] auto fixes from pre-commit.com hooks
- fix: match error message on test
- chore: refactor & add RELEASE.md
- chore: make mypy happy
- feat(draft): add permission extension

## Description
blocked by #2168 

Closes #244
Closes #2454

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist


botberry:



---

Adds support for a custom field using the approach specified in issue [#2168](abc).
Field Extensions may be used to change the way how fields work and what they return.
Use cases might include pagination, permissions or other behavior modifications.

```python
from strawberry.extensions import FieldExtension


class UpperCaseExtension(FieldExtension):
    async def resolve_async(
        self, next: Callable[..., Any], source: Any, info: Info, **kwargs
    ):
        result = await next(source, info, **kwargs)
        return str(result).upper()


@strawberry.type
class Query:
    @strawberry.field(extensions=[UpperCaseExtension()])
    async def string(self) -> str:
        return "This is a test!!"
```

```gql
query {
    string
}
```

```json
{
  "string": "THIS IS A TEST!!"
}
```



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Erik Wrede for the PR 👏

```

strawberry/permission.py
```diff
@@ -1,21 +1,115 @@
...
+            if not permission.has_permission(source, info, **kwargs):
+                permission.raise_error()
+
+    async def resolve_async(
+        self, next: Callable[..., Any], source: Any, info: Info, **kwargs
+    ) -> Any:
+        for permission in self.permissions:
+            has_permission: bool
+
+            has_permission = await await_maybe(
```
bellini666, line 109: Maybe instead of using `await_maybe` we can wrap non async permissions into async functions during `__init__` to cache those? The main reason is that the `inspect.isawaitable` call is very expensive.

strawberry/permission.py
```diff
@@ -1,21 +1,115 @@
...
+                permission.has_permission(source, info, **kwargs)
+            )
+
+            if not has_permission:
+                message = getattr(permission, "message", None)
+                raise PermissionError(message)
```
bellini666, line 115: Shouldn't this be `permission.raise_error()`?

strawberry/permission.py
```diff
@@ -1,21 +1,115 @@
...
+
+    @abc.abstractmethod
     def has_permission(
         self, source: Any, info: Info, **kwargs
     ) -> Union[bool, Awaitable[bool]]:
         raise NotImplementedError(
             "Permission classes should override has_permission method"
         )
+
+    def raise_error(self) -> None:
```
bellini666, line 49: An idea here: Instead of this being a function that should raise an error only, it could also be able to return something else?

Basically to solve [this issue](https://github.com/strawberry-graphql/strawberry/issues/2406), but also for some use cases I have

strawberry/permission.py
```diff
@@ -1,21 +1,115 @@
...
+    def resolve(
+        self, next: Callable[..., Any], source: Any, info: Info, **kwargs
+    ) -> Any:
+        """
+        Checks if the permission should be accepted and
+        raises an exception if not
+        """
+        for permission in self.permissions:
+            if not permission.has_permission(source, info, **kwargs):
+                permission.raise_error()
```
bellini666, line 101: Based on the feedback above, the this could do `return permission.raise_error()` (or some better name)

strawberry/permission.py
bellini666: But I just noticed that we use that on schema_converter for the resolver function. Probably it will be better to try to optimize all of those cases in a separated PR

bellini666: Actually, nevermind this comment. I thought about [this issue](https://github.com/graphql-python/graphql-core/issues/54), but it seems to not be a major issue for newer python versions anymore.

I made a quick test of returning 1 million objects with an async resolver that returned an int. It took ~52sec with the current `await_maybe` and ~51sec when I removed it. It also took ~51sec when I replaced `inspect.isawaitable` by [graphql-core's own implementation](https://github.com/graphql-python/graphql-core/blob/main/src/graphql/pyutils/is_awaitable.py). I'm using python 3.11

Even using an `asyncio.Future`, which [is the worst case to check](https://github.com/python/cpython/blob/main/Lib/inspect.py#L454) had the same differences.

So apparently there isn't too much overhead for newer python versions anymore.

jkimbo:
This is awesome @erikwrede ! Much better than I could have done.

strawberry/permission.py
erikwrede: What exactly would you like to return here? We would have to implement some mechanism to expose the type of that value from the permission to modify the `field`'s return type in `PermissionExtension::apply`. For typed errors, this is easy but supporting _any_ return is making things more complicated (e.g. parsing the type annotation of `Permission::has_permission`).

strawberry/permission.py
erikwrede: I left the `PermissionError` in here for backwards compatibility, but thinking about it again it should not break things to raise a GQL error instead.

strawberry/permission.py
erikwrede: Do you think it makes sense to `asyncio.gather` all async permissions here? It shouldn't break anything since permissions should not have side-effects, but make these checks run faster.

strawberry/permission.py
erikwrede: Thanks for the clarification! While seeing the use-case for this, I'm also concerned that this will encourage the use of returning `None` for permissions. Being explicit about errors is a better approach in many cases and follows the `failure-first` approach. I believe it's better to focus on the core permissions feature here, because a user can easily extend the PermissionsExtension to accomplish this. To make that easier, I can move the error raising to the extension.

Typed errors should be handled in a separate Extension, as this extension is focused on classic `GQLError`-style permission errors. However, that would mean having to create another base class for permissions. An alternative is to introduce another layer: `BasePermission -> GQLErrorPermission alias Permission | TypedErrorPermission`. That way we could still handle both in the default extension. 

Please let me know what you think ☺️

strawberry/permission.py
```diff
@@ -1,21 +1,125 @@
...
+
+    @abc.abstractmethod
     def has_permission(
         self, source: Any, info: Info, **kwargs
     ) -> Union[bool, Awaitable[bool]]:
         raise NotImplementedError(
             "Permission classes should override has_permission method"
         )
+
+    def get_error(self) -> GraphQLError:
```
bellini666, line 51: Wouldn't just a `handle_error` (or `handle_no_permission` be enough? And this function raise the error itself?

For my use case I can just use FieldExtension and override `resolve` directly, so it is not a major issue if you prefer to leave it like this :)

erikwrede: This way we're leaving things open for typed errors using the same interface in the future. I'm fine with both ways of handling it 🙂

strawberry/field.py
```diff
@@ -134,8 +135,20 @@ def __init__(
...
+        if len(self.permission_classes):
+            from .permission import PermissionExtension
+
+            if not self.extensions:
+                self.extensions = []
+            permission_instances = [
+                permission_class() for permission_class in permission_classes
+            ]
+            # Append to make it run first (last is outermost)
+            self.extensions.append(PermissionExtension(permission_instances))
```
erikwrede, line 151: We should decide wether we want to deprecate `permission_classes` as a way of adding Permissions to a field in favor of just using the `PermissionsExtension`. I think this is a duplicate way to do things and may lead to confusion for new users. Just using the extension is more straight forward, as it has no side-effects on extensions and enables + encourages customization as opposed to this piece of 'magic'.

strawberry/permission.py
```diff
@@ -1,21 +1,162 @@
...
+            if isinstance(field.type, StrawberryOptional):
+                if isinstance(field.type.of_type, StrawberryList):
+                    self.return_empty_list = True
+            elif isinstance(field.type, StrawberryList):
+                self.return_empty_list = True
+            else:
+                raise Exception(
+                    "Cannot use fail_silently=True with a non-optional "
+                    "or non-list field"
+                )
```
erikwrede, line 119: Should I move this to a custom exception in the exceptions module, or is it okay like this (e.g. with a value error)?
How can I get the fancy diagnostics for this?

strawberry/permission.py
erikwrede: We talked about performance drawbacks of using gather in the #performance channel today. I think it's best to leave it as is for now.

strawberry/permission.py
erikwrede: After thinking it through again, it seems to be cleaner to separate typed error permissions from normal GQLError permissions. If that changes, we can revisit that later.

erikwrede:
Hmm, coverage report seems to be broken. Locally, these lines seem to be covered - only when not using -n auto though.

skilkis:
@erikwrede I tried it once more with `-n auto` and it gave the correct coverage. Perhaps there is a race condition with `coverage.py` will investigate more later

strawberry/extensions/field_extension.py
```diff
@@ -0,0 +1,151 @@
...
+from typing import TYPE_CHECKING, Any, Awaitable, Callable, Union
+
+from strawberry.utils.cached_property import cached_property
+
+if TYPE_CHECKING:
+    from strawberry.field import StrawberryField
+    from strawberry.types import Info
+
+SyncExtensionResolver = Callable[..., Any]
+AsyncExtensionResolver = Callable[..., Awaitable[Any]]
```
bellini666, line 13:
```suggestion
SyncExtensionResolver: TypeAlias = Callable[..., Any]
AsyncExtensionResolver: TypeAlias = Callable[..., Awaitable[Any]]
```

strawberry/field.py
```diff
@@ -134,8 +135,22 @@ def __init__(
         self.is_subscription = is_subscription
 
         self.permission_classes: List[Type[BasePermission]] = list(permission_classes)
-        self.directives = directives
-
+        self.directives = list(directives)
+        self.extensions: Optional[List[FieldExtension]] = extensions
+
+        # Automatically add the permissions extension
+        if len(self.permission_classes):
```
bellini666, line 142: Are we going to deprecate the `permission_classes` usage and have only `extensions`? If so, it would be nice to add a deprecation warning here

strawberry/field.py
```diff
@@ -134,8 +135,22 @@ def __init__(
         self.is_subscription = is_subscription
 
         self.permission_classes: List[Type[BasePermission]] = list(permission_classes)
-        self.directives = directives
-
+        self.directives = list(directives)
+        self.extensions: Optional[List[FieldExtension]] = extensions
```
bellini666, line 139: Nitpick. By doing this you don't need to do `if not self.extensions` bellow

```suggestion
        self.extensions: List[FieldExtension] = extensions or []
```

strawberry/field.py
```diff
@@ -187,7 +202,11 @@ def is_basic_field(self) -> bool:
         an `Info` object and running any permission checks in the resolver
         which improves performance.
         """
-        return not self.base_resolver and not self.permission_classes
+        return (
+            not self.base_resolver
+            and not self.permission_classes
```
bellini666, line 207: Since we are converting permissions to extensions now, we don't need to check for this anymore. Just base_resolve and extensions

```suggestion
```

strawberry/field.py
```diff
@@ -208,7 +227,9 @@ def _python_name(self) -> Optional[str]:
     def _set_python_name(self, name: str) -> None:
         self.name = name
 
-    python_name: str = property(_python_name, _set_python_name)  # type: ignore[assignment]  # noqa: E501
+    python_name: str = property(  # type: ignore[assignment]
+        _python_name, _set_python_name
+    )
```
bellini666, line 232: Since you touched this code, is there a reason to why it is like this and not using `@property`/`@python_name.setter`?

strawberry/field.py
```diff
@@ -343,7 +368,7 @@ def _has_async_base_resolver(self) -> bool:
 
     @cached_property
     def is_async(self) -> bool:
-        return self._has_async_permission_classes or self._has_async_base_resolver
+        return self._has_async_base_resolver
```
bellini666, line 371: Don't we need to check if we have async extensions as well?

strawberry/permission.py
```diff
@@ -1,21 +1,176 @@
...
+        """
+        if self.use_directives:
+            for permission in self.permissions:
+                if permission.schema_directive:
+                    field.directives.append(permission.schema_directive)
+        # We can only fail silently if the field is optional or a list
+        if self.fail_silently:
+            if isinstance(field.type, StrawberryOptional):
+                if isinstance(field.type.of_type, StrawberryList):
+                    self.return_empty_list = True
```
bellini666, line 126: Don't know if this check is enabled on ruff yet, but it is something it would raise an issue

```suggestion
            if isinstance(field.type, StrawberryOptional) and isinstance(field.type.of_type, StrawberryList):
                self.return_empty_list = True
```

strawberry/permission.py
bellini666: I think you just need to create a subclass of `StrawberryException` and raise it. You can take a look at some examples like `InnvalidArgumentTypeError`

strawberry/permission.py
```diff
@@ -1,21 +1,176 @@
...
+                self.return_empty_list = True
+            else:
+                raise Exception(
+                    "Cannot use fail_silently=True with a non-optional "
+                    "or non-list field"
+                )
+
+    def _handle_no_permission(self, permission: BasePermission) -> Any:
+        if self.fail_silently:
+            if self.return_empty_list:
```
bellini666, line 137: Should we also return an empty list for `List[...] | None`? Or should `None` take precedence in this case?

bellini666 (changes requested):
This looks great! Some comments, mostly nitpicks

strawberry/field.py
erikwrede: No, since that is handled by the `SchemaConverter`. The field itself does not access the extensions, only the schema converters and extension creators access the field. An async extension might make a wrapped field resolver async, but since it's handled by the extension, this doesn't change the `is_async` property of the field.

strawberry/field.py
erikwrede: This was just `pre-commit` formatting, so I'm not sure

strawberry/field.py
patrick91: Between:

`@strawberry.field(extensions=[PermissionExtension(permissions=[IsAuthorized()])])`

and

`@strawberry.field(permission_classes=[IsAuthorized()])`

I still prefer the latter, it's less code to write. What do you think of only using `PermissionExtension` internally and keeping the same api we have now? :)

bellini666: There's a third option:

```python
@strawberry.field(extensions=[IsAuthorized()])
```

And have the extensions wrap the permission, just like it does when passing them as `permission_classes`.

Or even change the permissions to be a subclass of `FieldExtension`, this way permissions would basically be the same as field extensions without any special handling

Julianwustl:
How is the status on that. ??  Seems like an extremely useful feature.

strawberry/permission.py
```diff
@@ -1,21 +1,185 @@
...
+
+    def apply(self, field: StrawberryField) -> None:
+        """
+        Applies all of the permission directives to the schema
+        and sets up silent permissions
+        """
+        if self.use_directives:
+            for permission in self.permissions:
+                if permission.schema_directive:
+                    field.directives.append(permission.schema_directive)
```
bellini666, line 125: nit: to reduce nesting, but feel free to ignore :)

```suggestion
            field.directives.extend(
                p.schema_directive
                for p in self.permissions
                if p.schema_directive
            )
```

strawberry/permission.py
```diff
@@ -1,21 +1,185 @@
...
+                self.return_empty_list = True
+            else:
+                raise PermissionFailSilentlyRequiresOptionalError(field)
+
+    def _handle_no_permission(self, permission: BasePermission) -> Any:
+        if self.fail_silently:
+            if self.return_empty_list:
+                return []
+            else:
+                return None
```
bellini666, line 141: nit: single return

(I know that there's a ruff check that would automatically do that change, but I'm not sure if it is enabled)

```suggestion
            return [] if self.return_empty_list else None
```

strawberry/exceptions/permission_fail_silently_requires_optional.py
```diff
@@ -0,0 +1,37 @@
...
+        )
+
+        super().__init__(self.message)
+
+    @cached_property
+    def exception_source(self) -> Optional[ExceptionSource]:
+        # we can't return an exception source as currently, permission
+        # extensions are only linked
+        # to strawberryfields. these aren't linked to the source code.
+        return None
```
bellini666, line 37: I'm not really sure, but I think this could return the source from the field itself, which you have here.

Maybe this would work?

```python
        origin = self.field.origin
        if origin is None:
            return None

        source_finder = SourceFinder()
        return source_finder.find_class_attribute_from_object(
            origin,
            self.field.python_name,
        )
```

bellini666 (changes requested):
This looks awesome! :)

Can't wait to be able to use it on strawberry-django-plus and also when porting the permissions functionality to strawberry-graphql-django.

bellini666 (approved):
Awesome! :)

botberry:
# Pre-release
:wave:

Pre-release **0.190.0.dev.1687447182** [ae2453e7d7f6c88a83ba26256bc1925a138299d3] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.190.0.dev.1687447182
```

cpsnowden:
Is there a time frame on this feature? Super excited by having authorization self-documenting on the schema

erikwrede:
@cpsnowden going to /cc @patrick91 on this, IIRC the work on this is done and this is ready for merge

RELEASE.md
```diff
@@ -0,0 +1,54 @@
...
+
+For now, the old way of adding permissions using `permission_classes` is still
+supported via the automatic addition of a `PermissionExtension` on the field, but will
+be removed in a future release.
+
+Using the new `PermissionExtension` API, permissions support even more features:
+
+#### Silent errors
+
+To return `None` or `[]` instead of raising an error, the `silent_errors` keyword
```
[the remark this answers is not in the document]
erikwrede: That slipped through! Thanks Ester :)

RELEASE.md
```diff
@@ -0,0 +1,54 @@
...
+```python
+import strawberry
+from strawberry.permission import PermissionExtension, BasePermission
+
+
+class IsAuthorized(BasePermission):
+    message = "User is not authorized"
+    error_extensions = {"code": "UNAUTHORIZED"}
+
+    def has_permission(self, source, info, **kwargs) -> bool:
```
patrick91, line 15: ah, the other day I was thinking about refactoring this to remove the need of passing source, info and kwargs.

What do you think if we move them to be properties on the class? I think that would make the code much nicer 😊

strawberry/permission.py
```diff
@@ -1,21 +1,184 @@
...
+
+    @abc.abstractmethod
     def has_permission(
         self, source: Any, info: Info, **kwargs: Any
     ) -> Union[bool, Awaitable[bool]]:
         raise NotImplementedError(
             "Permission classes should override has_permission method"
         )
+
+    def handle_no_permission(self) -> None:
```
patrick91, line 58: shall we call this `handle_unauthorized`? or maybe `on_unauthorized` to be consistent with extensions?

RELEASE.md
patrick91: should we also allow users to set a status code somehow?

related: https://github.com/strawberry-graphql/strawberry/issues/2399

patrick91 (commented):
left some initial comments :)

RELEASE.md
erikwrede: @patrick91 since source and info change on every request, but the Permission class persists over requests I am not sure if this is a good idea. Take an async permission which yields to another coroutine. Then suddenly the source and info class properties change. We can make them optional for the user if you prefer, but `info.context` is often very crucial for the permission checks.

strawberry/permission.py
erikwrede: sounds good 👍

RELEASE.md
patrick91: we could have the same approach we have for resolvers for root/info, but not sure how we could do it for kwargs 🤔 

Also, I really don't like the fact we need to add kwargs there :D 

Oh I was also thinking about being able to check permission after resolving, would that be a valid use case? /cc @bellini666 

maybe after I'm done travelling we can pair on this?

erikwrede:
@patrick91 Py3.12 tests fail with this:

```
nox > Error: python is not installed into the virtualenv, it is located at /opt/hostedtoolcache/Python/3.11.5/x64/bin/python. Pass external=True into run() to explicitly allow this.
```

any idea? 👀

patrick91:
> @patrick91 Py3.12 tests fail with this:
> 
> ```
> nox > Error: python is not installed into the virtualenv, it is located at /opt/hostedtoolcache/Python/3.11.5/x64/bin/python. Pass external=True into run() to explicitly allow this.
> ```
> 
> any idea? 👀

Ah yes, this happens once in a while, not sure why, it is fine to retry them 😊

RELEASE.md
erikwrede: I agree on both points, but not sure it's in scope for this PR. Everything here is backwards-compatible. If we change the structure of `has_permission` it might break things for other users.
Already discussed permission after resolving with @bellini666, that's also something separate we can tackle building up on this.

Let's get this started after your travels 😊

patrick91: > I agree on both points, but not sure it's in scope for this PR. Everything here is backwards-compatible. If we change the structure of has_permission it might break things for other users.

makes sense, I didn't check all the code just yet I thought we were going to replace how users would use permissions 😊

XChikuX:
@patrick91 @erikwrede Are we ready to merge this?

I'm planning to start implementing permissions and I'd prefer to start with the new way.

erikwrede:
@XChikuX sorry for the delay, I didn't find any time to talk this through with patrick the last weeks. Trying to get it done this month. Does that work 4 u?

XChikuX:
I can't, in good conscience ask you to work faster 😁

I appreciate all the work you've put into the PR. Get it out at your time, I was mostly curious why it hasn't been merged.

I'll refactor my code later as necessary.

patrick91 (approved):

wlaub:
I'm suddenly getting snake case argument names instead of camel case in the kwargs passed to has_permissions. Is this an intentional change? I don't see it listed on the breaking changes page, but it's breaking some things. I can fix it locally, but it would be useful to know whether the change was intentional or whether it's likely to be reverted in the future.

patrick91:
@wlaub sorry for the late reply, I've opened a PR to update the release notes for this: https://github.com/strawberry-graphql/strawberry/pull/3371

PR #2580, opened 2023-02-21
feat: new input_mutation field

bellini666:
This is the input mutation functionality that I removed from the [relay implementation PR](https://github.com/strawberry-graphql/strawberry/pulls)


botberry:



---

This release adds `strawberry.input_mutation` as a way to make it easier to create
mutations that receive a single input type called `input`, while still being able
to define the arguments of that input in the resolver itself.

The following example:

```python
@strawberry.type
class Fruit:
    id: strawberry.ID
    name: str
    weight: float


@strawberry.type
class Mutation:
    @relay.input_mutation
    def update_fruit_weight(
        self,
        info: Info,
        id: strawberry.ID,
        weight: float,
    ) -> Fruit:
        fruit = ...  # retrieve the fruit with the given ID
        fruit.weight = weight
        ...  # maybe save the fruit in the database
        return fruit
```

Would generate a schema like this:

```graphql
input CreateFruitInput {
  id: ID!
  weight: Float!
}

type Fruit {
  id: ID!
  name: String!
  weight: Float!
}

type Mutation {
  updateFruitWeight(input: CreateFruitInput!): Fruit!
}
```

That pattern makes it easier to include/remove arguments without breaking the
whole API, which could happen when using positional arguments.



Here's the tweet text: 
```

```

RELEASE.md
```diff
@@ -0,0 +1,52 @@
...
+@strawberry.type
+class Fruit:
+    id: strawberry.ID
+    name: str
+    weight: float
+
+
+@strawberry.type
+class Mutation:
+    @relay.input_mutation
```
vamshiaruru-virgodesigns, line 19: Great feature!
Should this be `@strawberry.input_mutation`?

bellini666: Nice catch. I ported this from the relay PR and forgot to adjust after ctrl+c/ctrl+v

erikwrede:
I believe we can also achieve the same functionality with the new FieldExtensions, by utilizing the apply method to modify the input arguments and create the input type. And use resolve/resolve_async to transform the arguments for the resolver. This might require less change to the codebase.

bellini666:
@erikwrede @patrick91 just migrated the implementation to use field extensions instead of being a subclass of `StrawberryField`. What do you think?

docs/general/mutations.md
```diff
@@ -84,3 +84,47 @@ type Mutation {
...
 
 </Note>
+
+### The Input Mutation
+
+It is usually useful to use a pattern of defining a mutation that receives a single
+[input type](./input-types) argument called `input`.
+
+That pattern makes it easier to include/remove arguments without breaking the
+whole API, which could happen when using positional arguments.
```
erikwrede, line 94:
```suggestion
That pattern makes it easier to include/remove arguments without breaking the
whole API, which could happen when using positional arguments. Additionally, fields in an input type can be marked as deprecated, while field can't be deprecated.
```

strawberry/field.py
```diff
@@ -196,10 +198,11 @@ def is_basic_field(self) -> bool:
...
     @property
     def arguments(self) -> List[StrawberryArgument]:
-        if not self.base_resolver:
-            return []
+        arguments = self.default_arguments[:]
+        if self.base_resolver and not self.ignore_resolver_arguments:
+            arguments.extend(self.base_resolver.arguments)
 
-        return self.base_resolver.arguments
+        return arguments
```
erikwrede, line 205: I wonder if there's a more "elegant" solution to this than adding a `ignore_resolver_arguments` kwarg to `field`. Maybe move this from a property to something eagerly computed into a list on the field and then modify that list in the extension? We have the internal guarantee that field extensions are applied before the type is mapped to a gql-core type.

What do the others think?

strawberry/field.py
```diff
@@ -353,6 +356,7 @@ def is_async(self) -> bool:
 @overload
 def field(
     *,
+    field_class: Type[StrawberryField] = StrawberryField,
```
erikwrede, line 359: Do we still need the `field_class` change when using `FieldExtension`?

strawberry/mutation.py
```diff
@@ -1,8 +1,123 @@
...
+            annotation = annotations[arg.python_name]
+            if get_origin(annotation) is Annotated:
+                directives = tuple(
+                    d
+                    for d in get_args(annotation)[1:]
+                    if hasattr(d, "__strawberry_directive__")
+                )
+            else:
+                directives = ()
+
```
erikwrede, line 54: We should think about providing helper methods to make the api more accessible. That's not the scope of this PR though, so lets keep this as a reminder and todo for later. ☺️

strawberry/schema/schema_converter.py
```diff
@@ -468,9 +468,22 @@ def _get_arguments(
...
+            # This happens in mutation.InputMutationExtension
+            field_arguments = field.arguments
+            if field.base_resolver:
+                existing = [arg.python_name for arg in field_arguments]
+                field_arguments = field_arguments + [
+                    arg
+                    for arg in field.base_resolver.arguments
+                    if arg.python_name not in existing
+                ]
+
```
erikwrede, line 483: See my above comment on `field.arguments`. Maybe we can find a solution that fixes this in the same sweep 😊

erikwrede (commented):
I really like this! Added some comments on how we can optimize the internals, but other than that I really think this is the best approach. 😊

strawberry/field.py
bellini666: Yeah, I wrote this solution to avoid messing with the field API too much, and also to avoid changing the field extension code. I don't like it very much either.

Another solution could be adding an API to the field extension to resolve this. Something like this:

```python
@property
def arguments(self):
    arguments = self.base_resolver.arguments if self.base_resolver else []
    for extension in self.extensions:
        arguments = extension.resolve_arguments(self, arguments)
    return arguments
```

That `FieldExtension.resolve_arguments` would by default be implemented as:

```python
def resolve_arguments(self, field, arguments):
    return arguments
```

Meaning it would be a noop by default, but extensions could extend it or even remove arguments for advanced usage, like in this PR.

What do you think?

cc @patrick91 

obs. the `default_arguments` is also the same solution I used for the relay integration. Solving this in a more "elegant" way would also benefit it

strawberry/field.py
bellini666: Oh, we don't need it anymore. Forgot to remove it from the PR :)

strawberry/field.py
jkimbo: IMO we should leave `apply` as the only way to modify the attributes of a field. To make things a bit more ergonomic can we change the `arguments` property into a settable property? Something like this:

```python
    @property
    def arguments(self) -> List[StrawberryArgument]:
        if self._arguments:
            return self._arguments

        if not self.base_resolver:
            self._arguments = []
            return self._arguments

        self._arguments = self.base_resolver.arguments
        return self._arguments

    @arguments.setter
    def arguments(self, value: List[StrawberryArgument]):
        self._arguments = value
```

That way an extension can set the arguments in the apply method and it will be respected everywhere.

jkimbo (changes requested):
Some small documentation changes and I added my 2cents to https://github.com/strawberry-graphql/strawberry/pull/2580/files#r1181520479

Also IMO we should just expose the `InputMutationExtension` instead of adding a new top level `input_mutation` function. That way, to use it, you just need to do:

```python
@strawberry.mutation(extensions=[InputMutationExtension()])
def my_mutation(self, my_str: str) -> str:
    ...
```

strawberry/field.py
erikwrede: I agree with @jkimbo that `apply` should remain the only place for field extensions to modify the field. This prevents a lot of unknowns as to execution order of `apply` & `resolve_arguments` removes the need for extensions to know about each other (e.g. one extension needs to know about the next extension's resolve_arguments result).

The approach with the setter seems to solve the problem well for that matter ☺️

bellini666: The setter should work indeed! I'll refactor the code to use that approach :)

bellini666:
> > Also IMO we should just expose the InputMutationExtension instead of adding a new top level input_mutation function.
> 
> I agree with @jkimbo!

The function is really handy when someone only uses input mutation (me for example in my personal projects). But I don't have any issues in just having the `InputMutationExtension` if it is the preference for everyone since I can easily replicate the toplevel function in my projects and use it instead.

Also, would it be better in that situation to move it to another module? And maybe change its name as well? Maybe add a `strawberry/extensions/mutation.py` and add extensions related to mutations inside it? Then call it just `InputExtension`? Totally open for suggestions :)

erikwrede:
> The function is really handy when someone only uses input mutation (me for example in my personal projects). But I don't have any issues in just having the InputMutationExtension if it is the preference for everyone since I can easily replicate the toplevel function in my projects and use it instead.

This is an important point for consideration. Convenience is always great. In this case we should mention that `input_mutation` is just an alias of the extension usage in the docs, in case there are ever extensions that you want to apply _before_ the actual input type is created. Let's keep the `input_mutation`


> Also, would it be better in that situation to move it to another module? And maybe change its name as well? Maybe add a strawberry/extensions/mutation.py and add extensions related to mutations inside it? Then call it just InputExtension? Totally open for suggestions :)

As for that, I'm not too sure what the preference on modules in strawberry has been ^

patrick91:
> Some small documentation changes and I added my 2cents to [#2580 (files)](https://github.com/strawberry-graphql/strawberry/pull/2580/files#r1181520479)
> 
> Also IMO we should just expose the `InputMutationExtension` instead of adding a new top level `input_mutation` function. That way, to use it, you just need to do:
> 
> ```python
> @strawberry.mutation(extensions=[InputMutationExtension()])
> def my_mutation(self, my_str: str) -> str:
>     ...
> ```

yes, let's do this for now! 

We can think about expanding the public API in future 😊

strawberry/mutation.py
```diff
@@ -1,8 +1,123 @@
...
+from .field import StrawberryField, field
+
+if TYPE_CHECKING:
+    from strawberry.types.info import Info
+
+_T = TypeVar("_T")
+
+
+class InputMutationExtension(FieldExtension):
```
patrick91, line 30: if we keep only this extension and not the decorator then I think we can move this do another module, maybe `strawberry.utils.mutation` or `strawberry.field_extensions.input_mutation`?

strawberry/field.py
bellini666: It's done!

strawberry/mutation.py
bellini666: I actually didn't need it anymore. This code was ported from the relay PR, which was ported from my strawberry-django-plus lib. At that time strawberry probably didn't have `StrawberryArgument.directives`, but now it has :)

strawberry/schema/schema_converter.py
bellini666: I still had to keep this code even after changing `StrawberryField.arguments`. Don't know if there's a more elegant solution for this atm though

strawberry/mutation.py
bellini666: Moved it to `strawberry.field_extensions.input_mutation` :)

bellini666:
All suggestions applied :)

RELEASE.md
```diff
@@ -0,0 +1,64 @@
...
+}
+
+type Mutation {
+  updateFruitWeight(input: UpdateFruitInput!): Fruit!
+}
+```
+
+This pattern makes it easier to include/remove arguments without breaking the
+whole API, which could happen when using positional arguments. Additionally,
+fields in an input type can be marked as deprecated, while field can't be deprecated.
```
patrick91, line 64:
```suggestion

```

strawberry/field_extensions/input_mutation.py
```diff
@@ -0,0 +1,96 @@
...
+_T = TypeVar("_T")
+
+
+class InputMutationExtension(FieldExtension):
+    def apply(self, field: StrawberryField) -> None:
+        resolver = field.base_resolver
+        assert resolver
+
+        name = field.graphql_name or to_camel_case(resolver.name)
+        name_captalized = name[0].upper() + name[1:]
```
patrick91, line 33: let's use `capitalize_first` here 😊

```suggestion
        name_capitalized = name[0].upper() + name[1:]
```

strawberry/field_extensions/input_mutation.py
```diff
@@ -0,0 +1,96 @@
...
+                default=arg.default,
+                type_annotation=arg.type_annotation,
+                directives=tuple(arg.directives),
+            )
+            type_dict[arg_field.python_name] = arg_field
+            type_dict["__annotations__"][arg_field.python_name] = annotations[
+                arg.python_name
+            ]
+
+        caps_name = name[0].upper() + name[1:]
```
patrick91, line 54: same here 😊

patrick91 (approved):
Looks great, feel free to merge this when the code is updated 😊

erikwrede (approved):

PR #2592, opened 2023-02-27
feat: full support for forward refs

bellini666:
* Support | notation for older python versions when using `from __future__ import annotations`
* Fix `Annotated` not being correctly parsed when using `from __future__ import annotations`


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

bellini666:
@patrick91 ping :)

.github/workflows/test.yml
patrick91, line 1: we might want to revert this file 😊

strawberry/annotation.py
```diff
@@ -228,7 +184,7 @@ def create_union(self, evaled_type) -> StrawberryUnion:
     @classmethod
     def _is_async_type(cls, annotation: type) -> bool:
         origin = getattr(annotation, "__origin__", None)
-        return origin in ASYNC_TYPES
+        return isinstance(origin, type) and issubclass(origin, ASYNC_TYPES)
```
patrick91, line 187: how come this was changed?

strawberry/lazy_type.py
```diff
@@ -78,8 +78,16 @@ def __init__(self, module: str) -> None:
...
+    def resolve_forward_ref(self, forward_ref: Union[ForwardRef, str]) -> LazyType:
+        return LazyType(
+            (
+                forward_ref.__forward_arg__
+                if isinstance(forward_ref, ForwardRef)
+                else forward_ref
+            ),
+            self.module,
+            self.package,
+        )
```
patrick91, line 90: same as above, when would you receive a str here? and why it is ok to pass a string? 😊

strawberry/utils/typing.py
```diff
@@ -1,16 +1,59 @@
...
+    _SpecialForm,
+    cast,
+    overload,
 )
+from typing_extensions import Annotated, get_args, get_origin
+
+ast_unparse = getattr(ast, "unparse", None)
+# ast.unparse is only available on python 3.9+. For older versions we will
+# use `astunparse.unparse`.
+# We area also using "not TYPE_CHECKING" here because mypy gives an erorr
```
patrick91, line 30:
```suggestion
# We are also using "not TYPE_CHECKING" here because mypy gives an erorr
```

strawberry/utils/typing.py
```diff
@@ -1,16 +1,59 @@
...
+# on tests because "astunparse" is missing stubs, but the mypy action says
+# that the comment is unused.
+if not TYPE_CHECKING and ast_unparse is None:
+    import astunparse
+
+    ast_unparse = astunparse.unparse
+
+
+@lru_cache()
+def get_generic_alias(type_: Type) -> Type:
```
patrick91, line 40: can we add a docstring for this?

patrick91 (commented):
This looks great! I've left a couple of minor comments. 

I'll /cc @BryceBeagle just in case he's interested in taking a look at this too, but we can merge once we update the code :)

.github/workflows/test.yml
bellini666: The changes here was because of two things I noticed during my updates to the PR:

1) The `concurrency.group` change is because I noticed that when I pushed new commits, sometimes the unit tests did not get to run. I had to force push again to trigger them. Adding eh workflow prefix seems to have fixed that, but it might have been a coincidence. Np in reverting this change though :)

2) The skip of `poetry install` gave errors on some packages being missing (fastapi to be more precise). I already had this issue a lot in some private projects workflows and discovered that `poetry install` should be done even if a cache was found. It isn't too bad because, when the cache is found, this step will usually take ~5 seconds to complete.

strawberry/annotation.py
bellini666: I remember having an issue with this (which this change fixed) while writing this PR, but reverting this change doesn't seem to produce the error anymore.

Checking for `issubclass` have an advantage here that it would work for custom implementations of those types/abcs as well, but knowing now that the change is not actually required, I'm fine in reverting it if you prefer :)

strawberry/lazy_type.py
bellini666: Oh, totally here! Also something that I changed when starting the refactoring of the annotation class, but it is not required based on how the callsites work anymore. Reverting this!

strawberry/utils/typing.py
bellini666: Done! :)

strawberry/annotation.py
patrick91: let's revert 😊 we can reupdate it later if needed

.github/workflows/test.yml
patrick91: I think I fixed this in another PR! :D

bellini666: Oh, nice to know! :)

Reverted

strawberry/annotation.py
bellini666: Reverted! :)

patrick91 (approved):
🔥

PR #2600, opened 2023-03-01
graphql-transport-ws: Do not send "complete" after "error".

kristjanvalur:
## Description

The _graphql-transport-ws_ protocol specifies that an [error ](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#error )message should be the final
message for an _id_ coming from the server.  There were, however cases in strawberry where an
extra "complete" would be sent in violation of the protocol.

This PR cretates an Operation object which represents the state of an individual operation and ensures that
a "complete" or "error" message are the final messages for an active operation.

## Types of Changes

- [x] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
+++ b/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
@@ -205,6 +205,8 @@ class BaseGraphQLTransportWSHandler(ABC):
 
             result_source = get_result_source()
 
+        operation = Operation(self, message.id)
+
         # Handle initial validation errors
         if isinstance(result_source, GraphQLExecutionResult):
             assert result_source.errors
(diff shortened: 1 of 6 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -310,3 +314,24 @@ async def reap_completed_tasks(self) -> None:
         for task in tasks:
             with suppress(BaseException):
                 await task
+
+
+class Operation:
```
patrick91, line 319: can we add slots here? I imagine we might have a few instances of this class 😊

patrick91 (commented):
Looks good to me, other than a minor nitpick :)

Not sure if @DoctorJohn has time to review it, but I think we can merge once we update the operation class :)

kristjanvalur:
I intentionally didn't tack on some more refactoring which I have in mind for this code, to keep this PR as limited as possible.  But I have in preparation a few PRs which address subtle problems in the protocol, which will benefit from further refactoring.

kristjanvalur:
Also, please see my discussion #2610
A _lot_ of the unittests for the websocket protocol have nothing to do with the integration, and are just looking at
protocol behaviour, i.e. the _handler_ after a websocket has been established.  For those cases, I would prefer if we just added them to the aiohttp integration folder, rather than duplicate them to all three.

patrick91 (approved):

Issue #2605, opened 2023-03-02
Object Type Extensions


erikwrede:
Currently, Strawberry lacks a mechanism for extending types with reusable logic, such as pagination, database model mappers for libraries like sqlalchemy, or compatibility extensions like strawberry.experimental.pydantic. Much of the logic necessary for these to work has to be written separately for each use case. Opinionating the way we extend strawberry types could standardize behavior and improve compatibility and maintainability. This issue proposes an `ObjectTypeExtension` API that provides a foundation to address the mentioned points.

## API proposal

`ObjectTypeExtension`s provide custom functionality, can modify the underlying `TypeDefinition` or may offer the possibility to implement standardized hooks (more on that later). 
```python
class ObjectTypeExtension(ABC):

    def apply(strawberry_type: TypeDefinition):
		"""
		In this method, modifications can be made to TypeDefinition. 
		Custom fields or directives can be added here. It will be called 		 
         after the TypeDefinition is initialized
		"""
		pass

```
A `DjangoModelExtension` could use `apply` to setup all the automatic model fields, similar to a `PydanticModelExtension` or `SQLAlchemyModelExtension`

Similar to Field Extensions (#2168 , #2567), `ObjectTypeExtension` instances are passed to the `@strawberry.type` annotation:

```python
@strawberry.type(extensions=[DjangoModelExtension(model=CarModel)])
class Car:
	...
```

will resolve to:
```graphql
type Car{
	id: ID!
	manufacturer: String!
	model: String!
	doors: Int!
}
```
#### Extensions and Polymorphism
Using the annotation to define extensions is favorable over polymorphism of the actual `MyType` class, as that is a `dataclass` with _resolver logic_. The Extensions will provide behavioral logic and extended functionality and are a better fit for `StrawberryType`. Extensions themselves support polymorphism. The `DjangoExtension` could natively support `OffsetPaginationExtension` or `RelayPaginationExtension`.

#### Initialization
Extensions are initialized after `TypeDefinition` initialization. Additionally, we can provide helper methods to make dealing with polymorphic extensions easier:

```python
@dataclass
class TypeDefinition(StrawberrryType):
    extensions: List[ObjectTypeExtension]

    def __post_init__():
    	for extension in self.extensions:
        	extension.apply(self)
   			 ...
    		#rest of current post init

    def get_extension(extension_type: Type[ObjectTypeExtension]):
[...]

PR #2609, opened 2023-03-05
Add support for `supported_schemas`

vdurmont:
### Running multiple version of the API

While GraphQL APIs are usually not versioned and should strive to never introduce
breaking changes, the fact is that you may want to version your GraphQL API and
run multiple versions in parallel within the same server.

Instead of duplicating code or creating a complex class hierarchy, `supported_schemas`
enable you to annotate a field and specify in which version this field should be
available.

### Running multiple schemas using "similar" objects

Sometimes, you may want to serve multiple GraphQL APIs that use the same objects
but with slightly different fields. A good example is that your third party GraphQL 
API might expose the "core" fields of an object but your first party API might 
add more information and more fields for your first party apps.

Instead of duplicating code or creating a complex class hierarchy, `supported_schemas`
enable you to annotate a field and specify in which schemas those fields should
be available.

## How it works

Adds an optional `supported_schemas` feature to the fields.

One can decide to limit the support for a particular field to:
- A schema designated by name,
- A schema, until a specific version,
- A schema, from a specific version.

When the GraphQL schema is being built, we may pass an identifier for the current schema name and version and the fields will be filtered appropriately.

Examples of field definitions:

```python
@strawberry.type
class User:
    # [...]

    @strawberry.field(name="debugMessage", supported_schemas=[
        SupportedSchema(name="internal"),
    ])
    def get_debug_message(self) -> str:
        # This field will only appear in the schemas called `internal`
        # [...]

    @strawberry.field(name="riskScore", supported_schemas=[
        SupportedSchema(name="internal", until_version="1.2"),
    ])
    def get_old_risk_score(self) -> float:
        # This field will only appear in the schemas called `internal` that have
        # a version lower than or equal to `1.2`
        # [...]

    @strawberry.field(name="riskScore", supported_schemas=[
        SupportedSchema(name="internal", from_version="1.3"),
    ])
    def get_new_risk_score(self) -> float:
        # This field will only appear in the schemas called `internal` that have
        # a version higher than or equal to `1.3`
        # [...]
```

Examples of schema definition:

```python
internal_schema = Schema(
    query=Query,
    mutation=Mutation,
    schema_identifier=SchemaIdentifier(name="internal", version="1.4"),
)
```

## Checklist

- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.

I couldn't get the pre-commit hook to run but ran `black` manually:
```bash
$ poetry run pre-commit install
Command not found: pre-commit
```


botberry:



---

This release adds an optional `supported_schemas` feature to the fields.

One can decide to limit the support for a particular field to:
- A schema designated by name,
- A schema, until a specific version,
- A schema, from a specific version.

When the GraphQL schema is being built, we may pass an identifier for the current
schema name and version and the fields will be filtered appropriately.

Examples of field definitions:

```python
import strawberry
from strawberry.field import SupportedSchema

@strawberry.type
class User:
    # [...]

    @strawberry.field(name="debugMessage", supported_schemas=[
        SupportedSchema(name="internal"),
    ])
    def get_debug_message(self) -> str:
        # This field will only appear in the schemas called `internal`
        # [...]

    @strawberry.field(name="riskScore", supported_schemas=[
        SupportedSchema(name="internal", until_version="1.2"),
    ])
    def get_old_risk_score(self) -> float:
        # This field will only appear in the schemas called `internal` that have
        # a version lower than or equal to `1.2`
        # [...]

    @strawberry.field(name="riskScore", supported_schemas=[
        SupportedSchema(name="internal", from_version="1.3"),
    ])
    def get_new_risk_score(self) -> float:
        # This field will only appear in the schemas called `internal` that have
        # a version higher than or equal to `1.3`
        # [...]
```

Examples of schema definition:

```python
import strawberry
from strawberry.schema.identifier import SchemaIdentifier

# [...] Define `Query` and `Mutation`

internal_schema = strawberry.Schema(
    query=Query,
    mutation=Mutation,
    schema_identifier=SchemaIdentifier(name="internal", version="1.4"),
)
```


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to @vdurmont for the PR 👏

```

vdurmont:
Looks like I have some failures. I'm happy to work on cleaning things up and fix them if the feature itself and the direction of the implementation make sense to the strawberry maintainers.

jkimbo:
@vdurmont this is an interesting proposal but we'd rather avoid adding new parameters to the `strawberry.field` method if possible. Do you think this could be solved by using the proposed Object Type extensions: https://github.com/strawberry-graphql/strawberry/issues/2605 ?

(cc @erikwrede )

erikwrede:
@jkimbo interesting suggestion. While an `ObjectTypeExtension` could help with the object type, it still needs some info about the fields that are affected. This is hard to achieve conveniently without additional info on each field. `FieldExtensions` in their current design are mainly intended to affect the field's behavior, so they aren't a good place to store field-specific info either. Using `SchemaDirectives` on the fields and types could be an alternative, as they are the perfect place to store metadata on a field:

```python
@strawberry.type(directives=[SupportsSchemas(schemas=("internal","external")), SchemaVersion(lte="1.4")]))
class User:

    @strawberry.field(name="debugMessage", 
			directives=[SupportsSchemas(schemas=("internal",)), SchemaVersion(lte="1.4")])
	def debugField():
		pass
```
However, that doesn't solve the problem of building the schema. That schema-building would still need to involve custom logic to pass the current schema version to the type extension (unless we figure out an API for `SchemaExtensions` (currently `Extension`) to hook into the converter).

On the topic of multiple schemas using the same types, I do have some concerns about the maintenance and clarity of the codebase. While I understand the use case for schema versioning, I worry that allowing multiple named schemas (e.g. "internal", "external"...) with the same types and fields could lead to more complex and harder to maintain code.  Instead, it might be better to provide customers adjusted SDLs for each API and ensure that permissions are handled properly in the background. That being said, I do recognize that there are situations where this feature might be necessary or appropriate.

jkimbo:
> While an ObjectTypeExtension could help with the object type, it still needs some info about the fields that are affected. 

@erikwrede you could use the `metadata` argument to fields to store this information.

> That schema-building would still need to involve custom logic to pass the current schema version to the type extension (unless we figure out an API for SchemaExtensions (currently Extension) to hook into the converter).

Good point and IMO we should extend schema extensions to include hooks into the converter.

> On the topic of multiple schemas using the same types, I do have some concerns about the maintenance and clarity of the codebase. While I understand the use case for schema versioning, I worry that allowing multiple named schemas (e.g. "internal", "external"...) with the same types and fields could encourage design of more complex and harder to maintain code.

I completely agree with this and it's something to be mindful of @vdurmont . I do think we should provide enough functionality through extensions so that people can experiment with functionality like this to see how it might work in practice.

erikwrede:
@jkimbo Field Metadata is even better, I agree. 
Maybe we should look at the two concerns this PR solves (schema versioning, schema variants) separately.

Revisiting the post in `graphql-js` I linked yesterday, it might be worthwhile investigating visibility modifiers instead of the  schema variants implementation of this PR. Inspired by graphql-ruby, a visibility modifier decides based on context wether a field is visible to a certain user in introspection and queries: https://graphql-ruby.org/authorization/visibility.html

Using these modifiers, internal and external schemas can be defined without building two separate schemas from the same types. This would reduce complexity by disabling the use of two fields with the same name that have different resolvers or return types. At the same time, we enable users to easily implement static and dynamic feature toggles, which is an important use case. 

The problem with these modifiers might be a deep integration into GQL-Core (to ensure introspection & validation are behaving correctly).

In any way, if we come up with a design for `SchemaExtensions` that enables modifying the exposed fields, I'd favor that approach over integrating the feature right into the core schema converter.

patrick91:
@vdurmont sorry for keeping this PR in flight for so long, I totally forgot about it 😊

I've implement something similar using metadata here #3274, it's a much simpler approach with less opinions from us, so I'll close this issue, hope that's ok 😊

PR #2612, opened 2023-03-06
prototype: pydantic object type extension

erikwrede:
Prototype for #2605

### DIFF
--- a/strawberry/object_type.py
+++ b/strawberry/object_type.py
@@ -1,8 +1,11 @@
+from __future__ import annotations
+
 import dataclasses
 import inspect
 import sys
 import types
 from typing import (
+    TYPE_CHECKING,
     Callable,
     Dict,
     List,
@@ -27,6 +30,8 @@ from .utils.dataclasses import add_custom_init_fn
 from .utils.str_converters import to_camel_case
 from .utils.typing import __dataclass_transform__
 
+if TYPE_CHECKING:
+    from .extensions.type_extensions import TypeExtension
 T = TypeVar("T", bound=Type)
 
 
@@ -136,6 +141,7 @@ def _process_type(
     is_interface: bool = False,
     description: Optional[str] = None,
     directives: Optional[Sequence[object]] = (),
+    extensions: Optional[List[TypeExtension]] = None,
     extend: bool = False,
 ):
     name = name or to_camel_case(cls.__name__)
@@ -152,6 +158,7 @@ def _process_type(
         description=description,
         directives=directives,
         origin=cls,
+        extensions=extensions or [],
         extend=extend,
         _fields=fields,
         is_type_of=is_type_of,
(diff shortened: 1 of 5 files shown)


strawberry/object_type.py
```diff
@@ -193,6 +200,7 @@ def type(
     is_interface: bool = False,
     description: Optional[str] = None,
     directives: Optional[Sequence[object]] = (),
+    extensions: Optional[List[TypeExtension]] = None,
```
thejaminator, line 203: lame nitpick -> should at least follow directies and have a default of empty sequence. e.g.

```
directives: Optional[Sequence[object]] = (),
extensions: Optional[Sequence[TypeExtension]] = (),
```
though idk why we even needed the None case, it could just be 
```
extensions: Sequence[TypeExtension]] = (),
```

strawberry/types/types.py
```diff
@@ -46,6 +48,8 @@ class TypeDefinition(StrawberryType):
     )
 
     def __post_init__(self):
+        for extension in self.extensions:
+            extension.apply(self)
```
thejaminator, line 52: cool stuff

thejaminator:
looks great to meas a prototype for exploration. 

I guess something to think about is how we'll retrieve some static information of the transformation. E.g. how do we get our static typecheckers to know that `from_pydantic` and `to_pydantic` exist now?  
Related to discussions in https://github.com/strawberry-graphql/strawberry/issues/2168. I did try to play around to see how we could do that, but no strong ideas so far. so LGTM :)

tests/experimental/pydantic/test_pydantic_extension.py
```diff
@@ -0,0 +1,817 @@
...
+    definition: TypeDefinition = UserType._type_definition
+    assert definition.name == "UserType"
+
+    [field1, field2] = definition.fields
+
+    assert field1.python_name == "private"
+    assert field1.metadata["admin_only"]
+
+    assert field2.python_name == "public"
+    assert not field2.metadata
```
thejaminator, line 817: oh, i think you'll need some tests calling `from pydantic` or `to_pydantic` seems like `PydanticTypeExtension` doesn't define it too?

erikwrede: Yes, these were just the basic tests to see if the general idea works. Will add more soon 🙂 
PydanticTypeExtension should also set the from and to methods, but type checking is indeed an issue

erikwrede:
> I guess something to think about is how we'll retrieve some static information of the transformation. E.g. how do we get our static typecheckers to know that from_pydantic and to_pydantic exist now?
Related to discussions in https://github.com/strawberry-graphql/strawberry/issues/2168. I did try to play around to see how we could do that, but no strong ideas so far. so LGTM :)

Good point. Maybe we could make all of these model dataclasses extend `StrawberryTypeFromPydantic`. That comes with the drawback of reducing explicitness, so I'm not yet sure if I like it or not. Though, it does only make a stylistic difference, not a functional difference.

```python
  @strawberry.type(extensions=[PydanticTypeExtension(model=User)])
  class UserType(StrawberryTypeFromPydantic):
      age: strawberry.auto
      password: strawberry.auto
```

patrick91:
Going to close this, since it's been a while and we might do a different Pydantic integration "soon" 😊

PR #2617, opened 2023-03-07
Add handler method to process websocket connections

kristjanvalur:
Allow an application to process the connection request for `graphql_transport_ws` and `graphql_ws`
protocols, perform authentication, reject a connection and return custom response payload

## Description

A previous PR, #2380 added support for `connection_params`.  However, they were simply taken as provided and
injected into the _context_ object.  This PR adds an overridable hook in the `AsyncBaseHTTPView`, `on_ws_connect()`
which allows an application to examine the connection parameters when a connection is made.
The method receives an `WSConnectionParams` object, `params`.
The application can then

- choose to reject the connection (by calling `params.reject()`)
- add additional information into the `params.connection_params` or modify it.
- set `params.rsponse_params` to return a _payload_ to the client.

This is based on similar behaviour in the reference implementation for `graphql_transport_ws`, namely
this code here:
https://github.com/enisdenjo/graphql-ws/blob/972ed07ea263624bb556970cfd85bb05ea2c6eac/src/server.ts#L614
except that we use a parameter object to operate on.

This approach allows us to split **authentication** and **authorization** into two steps.   Authentication can be done when establishing connection and authorization can be performed for each resolver.

Documentation is modified and also a piece added about how to perform authentication for _single result operations_ (queries and mutations) since they can be invoked both via http and websockets.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation

*

## Checklist


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

kristjanvalur:
3.10 test runner seems to have FastAPI version problems.

patrick91:
@kristjanvalur you can add reject to the .alexignore file 😊

kristjanvalur:
**ruff** and **mypy** can't agree on whether to use `return` (ruff) or `return None` (mypy).  How to resolve this?

patrick91:
> **ruff** and **mypy** can't agree on whether to use `return` (ruff) or `return None` (mypy). How to resolve this?

@kristjanvalur feel free to disable the rule that's causing issues in Ruff :)

kristjanvalur:
Sounds fine, I wasn't aware of views until now :).  Thank you for pointing that out.
I'll see if I can grok it and make the relevant change.

patrick91:
@kristjanvalur no worries, the idea is to unify most of the logic in a base class/view, for now I'm only tackling http.

I'm also fine with having duplication for on_ws_connect for the time being, as long as it is inside the views

patrick91:
@kristjanvalur thanks for keeping this PR up to date, do you think we can move the ws_connect method on the BaseHTTPView (or create a BaseWSView?)?

kristjanvalur:
So, I'm not quite sure what you have in mind with views, and how I could use them.

Just looking at `graphql_transport_ws`, notice how we have to catch the event in `handle_connection_init()` in `subpscription/protocols/graphql_transport_ws/handlers.py`.
The only customizable context I can get at in this case is the schema, which is available as `self.schema` in the protocol handler.  I am unsure how to invoka a "view" method.

Also, how would an application customize provide a customized view class for strawberry to use?

patrick91:
I'll try to make an example tomorrow 😊

kristjanvalur:
I'll start by providing the view to the handler class creation.

kristjanvalur:
So, I've started to make changes to the initialization of the handler classes, by providing a "view" constructor argument.  

But I'll have to keep the old behaviour until we implement views for "Channels".  The view is still an "Optional".   I guess we have to omit that integration from the relevant unit-tests for the time being.

DoctorJohn:
Interesting changes @kristjanvalur! I agree that the new `on_ws_connect` method should live in the base integration view classes next to `get_context` and `get_root_value` which have been put there to be easily modifyable by "users" of integrations.

I saw a `view` parameter has been added to the ws base classes contructors. I was thinking about that for a while the past few months. The motivation is to reduce the amount of constructor arguments and to standardize these params for all integration views. After @patrick91 just did a great job unifying all views I think it's a good time to make that change.

```python
def __init__(
    self,
    view: AsyncBaseHTTPView,
    schema: BaseSchema, # -> view.schema
    debug: bool, # -> view.debug
    keep_alive: bool, # -> view.keep_alive
    keep_alive_interval: float, # -> view.keep_alive_interval
    get_context, # -> view.get_context
    get_root_value, # -> view.get_root_value
    ws: WebSocket,
):
```

Of course we need to make sure that all these params are available on the base view integration class. If this would bloat this PR to much I'd suggest going with a `on_ws_connect` param instead of a `view` param for now to keep things consistent. In a separate PR we would then standardize all these params on the base class and then replace them in the constructor with a single `view` param 🤔 What do both of you think?

kristjanvalur:
> What do both of you think?

Fine by me, I was thinking along the same lines.  I'll see if I can roll that change into this one.  I personally don't have a problem with PRs sometimes taking on additional changes that make sense in the context of what is being done.

kristjanvalur:
Oh, and I noticed that the "Starlite" integration isn't included in my unified websocket test framework parametrization.  I'll try adding that in a separate pr.

DoctorJohn:
> The problem is that one of the integrations, Channels, doesn't have a "view" yet. Which is why the parameter is currently still "Optional". Once @patrick91 does his magic with the Channels integration, we can consolidate those params. Until then, I don't see a problem with passing an Optional[View] as oppsed to an Optional[Callable]

Ah I see, thanks. Should be fine then since we plan to do it eventually :+1:

kristjanvalur:
If you agree with this pattern, it just remains for me to adjust my documentation changes to reflect this new and better approach, i.e. subclass the view rather than the schema.  I'll make that change tomorrow.

DoctorJohn:
> If you agree with this pattern

I think the introduced method is useful, just wasn't sure about whether returning False or a Dict was the best API we could come up with. I know Apollo does it but I was wondering whether we could do better. That being said, so far I didn't come up with a better idea so I'm fine with the proposed pattern.

I talked to Patrick for a minute about this PR two days ago. He stumbled accross the optional view parameter, so I kinda want him to have another look.

kristjanvalur:
Yes


> I think the introduced method is useful, just wasn't sure about whether returning False or a Dict was the best API we could come up with. I know Apollo does it but I was wondering whether we could do better. That being said, so far I didn't come up with a better idea so I'm fine with the proposed pattern.
 
I agree with that,  It also doesn't sit quite right with me, it seems a bit un-pythonic.  An alternative is returning a tuple, or a dict, or a special pydantic type, but that starts to feel a bit heavy-weight.   

> I talked to Patrick for a minute about this PR two days ago. He stumbled accross the optional view parameter, so I kinda want him to have another look.

Sure, I am in no particular hurry with this right now, although it will be useful to get something like this in eventually.  Are there technical reasons why "channels" didn't get a "view", or simply lack of time?

FYI, we're using Strawberry in this project:  https://playpaxdei.com/en-us

kristjanvalur:
By the way, ASGI supports the extension "websocket.http.response", and I have several PRs active for supporting that, including this here with Starlette:  https://github.com/encode/starlette/pull/2041

The idea is that an application may reject a websocket connection with an actual HTTP response including informative payload.  The Starlette pr aims to abstract that into a `send_response()` method which will send such a response, if the active web server supports the extension, otherwise, it will close the connection.

What I'm getting at here is that perhaps the connection handler should actively reject the connection, perhaps by calling a method on the view, rather than returning a literal False. 

Perhaps a `View.ws_reject(reason: str)` which could later be extended to send a proper HTTPResponse if supported.

kristjanvalur:
Please see my latest commit.  The `on_ws_connect()` now receives a class instance as an argument, which has both
`connection_params` and `response_params` members.  It also has a `reject()` method. 

I believe this is cleaner and allows for future expansion, e.g. if we ever want to include the original Upgrade request headers too.

nrbnlulu:
IMO the public API for this should be in extensions, also http integrations.

kristjanvalur:
> IMO the public API for this should be in extensions, supporting http integrations as well.

Not sure what you mean, by "http integrations".  What exactly do you have in mind?  This is specifically for _websockets_ so that a connection can be accepted or rejected based on the payload provided in the sub-protocol's greeting message.

This PR has been edited so that the change is on the `view` not the schema, but I clearly forgot to change the RELEASE.md.

nrbnlulu:
> Not sure what you mean, by "http integrations". What exactly do you have in mind?

currently there is no hook for when a graphql connection has started on extensions
this is not something that is specific to websockets...

kristjanvalur:
Note, as pointed out, a different approach _might_ be to do this via an _extension_.  There is already an extension PR for websocket operations (e.g. #2784 and #2825).  We could extend the callbacks to add an on_ws_connect() hook there instead of doing it in the View.

DoctorJohn:
> what do you think about this approach to the api? Getting an object passed with a reject() method?

IMHO it's an interesting approach with clearer semantics compared to the approach of returning None, dict or the the False literal. That being said it's an extra object we create on every connect with only one method, a pattern we don't really use anywhere else. I should talk to @patrick91 about this API. At this point I feel like there is three options: 1. special return values 2. params object with a reject method 3. maybe a exception the user could raise to reject the connection (just an idea, not suggesting it's superior at all) 4. maybe use an extension instead

I would be interested in seeing this as an extension. Sounds like a good usecase for an extension which could allow us to keep the implementation lean. Sorry for yet another delay, I'll take a look at the proposed subscription extension APIs.

kristjanvalur:
I hope your objection isn't performance related. An "extra object on every connection" is not problematic for websockets connections where the connection is long lived and serves to support a number of requests and subscriptions. Performance always needs to be considered in perspective.

DoctorJohn:
> I hope your objection isn't performance related

Not trying to prematurely optimize here, it's purely about introducing a new pattern into the code. It's also less of an objection and more something I'd like to pitch to Patrick first since he has a good sense for API design that'll make devs happy ;)

kristjanvalur:
Right.   I'm happy to change the api however you think is best.
There is also a separate suggestion that all of this could be done via the "Extensions" mechanism, as I suggested here:  https://github.com/strawberry-graphql/strawberry/pull/2617#issuecomment-1598478744

XChikuX:
Does this allow for custom headers?

kristjanvalur:
> Does this allow for custom headers? I need to add

This is just about looking at the "params" from the graphql_transport_ws connection request.  Nothing to do with HTTP headers.  What exactly is preventing you from getting websockets to work on IOS?  Can you explain a bit better the problem you are having?

XChikuX:
> > Does this allow for custom headers? I need to add
> 
> This is just about looking at the "params" from the graphql_transport_ws connection request. Nothing to do with HTTP headers. What exactly is preventing you from getting websockets to work on IOS? Can you explain a bit better the problem you are having?

I've been trying to triage a secure websocket issue I've been facing with iOS. Initially I thought there were extra security measures taken such as adding user-agent to the headers.

Turns out its a misconfiguration on apple's part.

[See Here](https://stackoverflow.com/questions/62352790/websockets-not-working-on-ios-and-safari-ossstatus-error-9837)

I switched to TLSv1.2 and everything works now. Thanks for getting back to me though :)

PR #2618, opened 2023-03-07
Update mypy.ini and pipeline

thejaminator:
## Description
Updates mypy.ini so that developers don't need to run the specific incantation of the pipeline commands to replicate the mypy pipeline.

All you'll need to do is
```
poetry run mypy
```

### DIFF
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -39,4 +39,4 @@ jobs:
         if: steps.cache-poetry-dependencies.outputs.cache-hit != 'true'
 
       - run: mkdir -p .mypy_cache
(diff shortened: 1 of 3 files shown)


.github/workflows/mypy.yml
```diff
@@ -39,4 +39,4 @@ jobs:
         if: steps.cache-poetry-dependencies.outputs.cache-hit != 'true'
 
       - run: mkdir -p .mypy_cache
-      - run: poetry run mypy --ignore-missing-imports --config-file mypy.ini --install-types --non-interactive --show-traceback strawberry
+      - run: poetry run mypy --config-file mypy.ini
```
thejaminator, line 42: pipeline simply uses the mypy.ini file

patrick91 (approved):

PR #2627, opened 2023-03-09
Enable Ruff ANN201 (missing-return-type-public-function) rule

JadHADDAD92:
## Description
Added return types to public functions and activated the Ruff rule ANN201

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* #2477

## Checklist


botberry:



---

Add return types annotations and lint return types with Ruff.


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Jad Haddad for the PR 👏

```

strawberry/custom_scalar.py
```diff
@@ -32,7 +32,7 @@
     _T = TypeVar("_T", bound=type)
 
 
-def identity(x):
+def identity(x) -> Any:
```
patrick91, line 35: can you type this as:

```

def identity(x: T) -> T:
    return x
```

https://mypy-play.net/?mypy=latest&python=3.11&gist=fcf488a41be8a6f87e3aef4a0d188818

strawberry/django/test/client.py
```diff
@@ -9,7 +9,7 @@ def request(
         body: Dict[str, object],
         headers: Optional[Dict[str, object]] = None,
         files: Optional[Dict[str, object]] = None,
-    ):
+    ) -> Any:
```
patrick91, line 12: we should probably type these better in future, since we roughly know what we'll be getting

strawberry/experimental/pydantic/conversion.py
```diff
@@ -87,7 +89,7 @@ def convert_pydantic_model_to_strawberry_class(cls, *, model_instance=None, extr
     return cls(**kwargs)
 
 
-def convert_strawberry_class_to_pydantic_model(obj):
+def convert_strawberry_class_to_pydantic_model(obj) -> Any:
```
patrick91, line 92: @thejaminator does this type make sense?

strawberry/experimental/pydantic/error_type.py
```diff
@@ -23,13 +34,13 @@
...
 
-def get_type_for_field(field: ModelField):
+def get_type_for_field(field: ModelField) -> Union[Any, Type[None], Type[List]]:
     type_ = field.outer_type_
     type_ = normalize_type(type_)
     return field_type_to_type(type_)
 
 
-def field_type_to_type(type_):
+def field_type_to_type(type_) -> Union[Any, Type[None], Type[List]]:
```
patrick91, line 43: uhm, can we remove any here?

what's type[none]?

strawberry/experimental/pydantic/object_type.py
```diff
@@ -40,7 +41,7 @@
     from pydantic.fields import ModelField
 
 
-def get_type_for_field(field: ModelField, is_input: bool):
+def get_type_for_field(field: ModelField, is_input: bool) -> Union[Any, Type[None]]:
```
patrick91, line 44: same as above

strawberry/federation/scalar.py
```diff
@@ -10,7 +10,7 @@
     _T = TypeVar("_T", bound=type)
 
 
-def identity(x):  # pragma: no cover
+def identity(x) -> Any:  # pragma: no cover
```
patrick91, line 13: same as above, maybe we can reuse the one from above instead of declaring it again?

strawberry/federation/schema.py
```diff
@@ -171,7 +171,7 @@ def _add_entities_to_query(self):
         self._schema.query_type = query_type
         self._schema.type_map[query_type.name] = query_type
 
-    def entities_resolver(self, root, info, representations):
+    def entities_resolver(self, root, info, representations) -> List[Any]:
```
patrick91, line 174: maybe List[object] is better here

strawberry/sanic/views.py
```diff
@@ -79,7 +79,7 @@ def __init__(
 
             self.json_encoder = json.JSONEncoder
 
-    def get_root_value(self):
+    def get_root_value(self) -> None:
```
patrick91, line 82: I think we want Any or object here since we want subclasses to override this method

strawberry/utils/typing.py
```diff
@@ -103,7 +103,7 @@ def is_type_var(annotation: Type) -> bool:
     return isinstance(annotation, TypeVar)
 
 
-def get_parameters(annotation: Type):
+def get_parameters(annotation: Type) -> Union[Any, Tuple[()]]:
```
patrick91, line 106: I think this might be tuple of objects

tests/aiohttp/test_graphql_transport_ws.py
```diff
@@ -879,7 +879,7 @@ async def test_single_result_operation_error(aiohttp_client):
         assert ws.closed
 
 
-async def test_single_result_operation_exception(aiohttp_client):
+async def test_single_result_operation_exception(aiohttp_client) -> None:
```
patrick91, line 882: can we skip this rule for tests? I don't think it adds any value :)

patrick91 (commented):
Nice work so far! I've left a few comments :)

I think we should disable this rule for the test 😊

strawberry/experimental/pydantic/conversion.py
thejaminator: yes, the function name seems abit of a misnomer, but its more like recursively converting the fields of a strawberry type into the fields of a pydantic model.

which can be anything

strawberry/sanic/views.py
JadHADDAD92: then I guess `Any` would be more adequate.

I don't really understand the use of `object` in typing, since it is the base class for everything else, so might as well use `Any` IMO

tests/aiohttp/test_graphql_transport_ws.py
JadHADDAD92: right on, will fix that

JadHADDAD92:
I'm having troubles passing the mypy tests. I've temporarily put a `# noqa` and didn't put return typings on the related methods (`get_type_for_field()`, and `await_maybe()`)

strawberry/utils/typing.py
JadHADDAD92: It should be `Union[Tuple[object], Tuple[()]]` then, otherwise mypy will break because of the `return ()` in this function

patrick91:
@JadHADDAD92 I'll try to take a look at this today or tomorrow!

patrick91 (approved):
Thank you so much for this! We can definitely update some types in future, but for now this looks great!

botberry:

PR #2680, opened 2023-03-31
Move fastapi handlers to instance methods.

mattalbr:
## Description

This came up while I was trying to implement required persisted queries in my application, where I wanted to block regular queries in certain environments. But, because the handlers were closures, I had to do some janky stuff to overwrite them only in certain environments, which made testing a challenge. Moving them to instance methods should make it easy to subclass and override the handler.

NOTE: One downside to this change is that any changes to the handler function signatures (e.g. adding a new Fastapi.Depends) would not be backwards compatible.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ x ] Enhancement/optimization
- [ ] Documentation

## Issues Fixed or Closed by This PR

*

## Checklist

I'm a new contributor and am still working on getting the environment setup. Wanted to throw this by you preliminarily and if you're directionally aligned, I'll finish getting it setup and run the tests. I don't think we should need any new tests for this, unless you want to explicitly test subclassing the handlers.

- [ x ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ x ] I have read the CONTRIBUTING document.
- [ ] I have added tests to cover my changes.
- [ ] I have tested the changes and verified that they work and don't break anything (as well as I can manage).

PR #2681, opened 2023-03-31
Refactor all* HTTP views

patrick91:
This PR refactors all* the HTTP views, it's still a work in progress, but it's a good start.

I created a base http view (async and non) and used that in the various views/router/controllers 
we have.

This should make it easier to add support to new features to all integrations at once, ~~I think
by doing this we also gained upload support on chalice~~ :)

There's some work to do around temporal views (do we want to have a generic class that we can reuse),
and in the interface in general.

* this doesn't yet include the HTTP part of the django channels integarion


botberry:

<table>
<thead>
</thead>
<tbody>
</tbody>
</table>

Learn more:

strawberry/http/base.py
```diff
@@ -0,0 +1,62 @@
...
+from .typevars import Request
+
+
+class BaseRequestProtocol(Protocol):
+    @property
+    def query_params(self) -> Mapping[str, Optional[Union[str, List[str]]]]:
+        ...
+
+    @property
+    def method(self) -> str:
```
nrbnlulu, line 17: Maybe this can be an enum

patrick91: or a literal too :D

that might be quite nice as I won't need to use .lower then!

mattalbr (approved):
This looks like what it will do what I was looking for out of #2680, thanks!

patrick91:
Breaking changes:

1. Flask `get_root_value` now receives the request
2. Flask `get_context` now receives the request
 (Same for the async version)

3. Sanic `get_root_value` now receives the request and it is async

botberry:



---

This release updates all\* the HTTP integration to use the same base class,
which makes it easier to maintain and extend them in future releases.

While this doesn't provide any new features, it does make it easier to extend
the HTTP integrations in the future. So, expect some new features in the next
releases!

**Breaking changes:**

Unfortunately, this release does contain some breaking changes, but they are
minimal and should be quick to fix.

1. Flask `get_root_value` and `get_context` now receive the request
2. Sanic `get_root_value` now receives the request and it is async

\* The only exception is the channels http integration, which will be updated in
a future release.


---




Here's the tweet text: 
```

```

kkroening:
A couple of other subtle breaking changes (AFAICT) with `strawberry.fastapi.GraphQLRouter`:

#### 1. Removal of `get_graphiql_response`

Use of `GraphQLRouter.get_graphiql_response()` must be changed to `GraphQLRouter.render_graphiql(response)`.

#### 2. Explicit type annotations

Strawberry 0.169.0 seems to make `mypy` quite a bit more finicky with type annotations around use of `strawberry.fastapi.GraphQLRouter`.

For example, the [official FastAPI Strawberry examples](https://strawberry.rocks/docs/integrations/fastapi#fastapi) previously passed with `mypy` out of the box with Strawberry 0.168.2 but now fails as follows:
```
$ mypy strawberry_fastapi_example.py
strawberry_fastapi_example.py:29: error: Need type annotation for "graphql_app"  [var-annotated]
Found 1 error in 1 file (checked 1 source file)
```

When running `mypy` in non-strict mode, adding a simple `graphql_app: GraphQLRouter = ...` annotation makes the problem go away (even though this was unnecessary in <=0.168.2).

However, `mypy --strict` (as many of us use in production apps) produces a more ominous error, even with the explicit annotation from above:
```
strawberry_fastapi_example.py:29: error: Missing type parameters for generic type "GraphQLRouter"  [type-arg]
```

It's not clear what the generic parameters should be for `strawberry.fastapi.GraphQLRouter` - it doesn't seem to be mentioned in the release notes or documentation, and is different from <=0.168.2.

Not sure if this is intentional or a bug, but mentioning it in case others run into similar issues when upgrading to Strawberry 0.169.0+.

patrick91:
@kkroening oh, I totally missed your comment, sorry! Did you ever get around the FastAPI typing part?

I want to make that better in the future, but not yet sure how 😊 The generic types are for context and root type 😊

@sstur thanks for the suggestion! just pushed it to main 😊

PR #2677, opened 2023-03-31
Fix typing in experimental generation models from Pydantic models.

Rubikoid:
## Description

Fix type hint for `StrawberryTypeFromPydantic._pydantic_type` to match real state and don't confuse type checkers.

Also add `StrawberryTypeFromPydantic._original_model` to make extending models easier.

MRE/Use case for `._original_model`:
```python
from pydantic import BaseModel

import strawberry
from strawberry.experimental.pydantic.conversion_types import StrawberryTypeFromPydantic

class User(BaseModel):
    name: str

@strawberry.experimental.pydantic.type(model=User, all_fields=True)
class UserType:
    @strawberry.field
    def upper_name(self: StrawberryTypeFromPydantic[User]) -> str:
        # we have to ignore E1101 error, bc pylint 2.17.1 don't agree with StrawberryTypeFromPydantic protocol
        self_model = self._original_model  # noqa: E1101
        return self_model.name.upper()

def get_users() -> UserType:
    return UserType.from_pydantic(User(name="test"))  # noqa: E1101

@strawberry.type
class Query:
    user: UserType = strawberry.field(resolver=get_users)

schema = strawberry.Schema(query=Query)
```

Anyway I personally don't like approach with Protocol with class decorator because it makes type checkers and linters go crazy.
So I think that this mechanism can be made more convenient.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [x] New feature
- [ ] Enhancement/optimization
- [ ] Documentation

## Issues Fixed or Closed by This PR

Didn't found right now.

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the CONTRIBUTING document.
- [ ] I have added tests to cover my changes.
- [ ] I have tested the changes and verified that they work and don't break anything (as well as I can manage).

PR #2683, opened 2023-04-03
Add release file for mr 2677, revert protocol exposure of _original_model

thejaminator:
## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/strawberry/experimental/pydantic/conversion_types.py
+++ b/strawberry/experimental/pydantic/conversion_types.py
@@ -35,7 +35,3 @@ class StrawberryTypeFromPydantic(Protocol[PydanticModel]):
     @property
     def _pydantic_type(self) -> Type[PydanticModel]:
         ...
-
-    @property
-    def _original_model(self) -> PydanticModel:
-        ...
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,30 +1,37 @@
 release type: patch
 
-Fixes type hint for StrawberryTypeFromPydantic._pydantic_type to be a Type instead of a class.
+Fixes type hint for StrawberryTypeFromPydantic._pydantic_type to be a Type instead of an instance of the Pydantic model.
+As it is a private API, we still highly discourage using it, but it's now typed correctly.
 
 ```python
 from pydantic import BaseModel
-from typing import Type
+from typing import Type, List
 
 import strawberry
+from strawberry.experimental.pydantic.conversion_types import StrawberryTypeFromPydantic
 
 
 class User(BaseModel):
     name: str
-    
+
     @staticmethod
-    def available_names() -> list[str]:
+    def foo() -> List[str]:
         return ["Patrick", "Pietro", "Pablo"]
+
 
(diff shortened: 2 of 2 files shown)


patrick91 (approved):

botberry:



---

Fixes type hint for StrawberryTypeFromPydantic._pydantic_type to be a Type instead of an instance of the Pydantic model.
As it is a private API, we still highly discourage using it, but it's now typed correctly.

```python
from typing import Type, List

import strawberry
from strawberry.experimental.pydantic.conversion_types import StrawberryTypeFromPydantic


class User(BaseModel):
    name: str

    @staticmethod
    def foo() -> List[str]:
        return ["Patrick", "Pietro", "Pablo"]


@strawberry.experimental.pydantic.type(model=User, all_fields=True)
class UserType:
    @strawberry.field
    def foo(self: StrawberryTypeFromPydantic[User]) -> List[str]:
        # This is now inferred correctly as Type[User] instead of User
        # We still highly discourage using this private API, but it's
        # now typed correctly
        pydantic_type: Type[User] = self._pydantic_type
        return pydantic_type.foo()


def get_users() -> UserType:
    user: User = User(name="Patrick")
    return UserType.from_pydantic(user)


@strawberry.type
class Query:
    user: UserType = strawberry.field(resolver=get_users)


```


---




Here's the tweet text: 
```

```

strawberry/experimental/pydantic/conversion_types.py
```diff
@@ -35,7 +35,3 @@ def _type_definition(self) -> TypeDefinition:
     @property
     def _pydantic_type(self) -> Type[PydanticModel]:
         ...
-
-    @property
-    def _original_model(self) -> PydanticModel:
-        ...
```
thejaminator, line 41: @Rubikoid sorry but i'll need to revert this change made in https://github.com/strawberry-graphql/strawberry/pull/2677

`_original_model` is not always available. It only gets set when you convert directly from pydantic. As we also support directly instantiating from the strawberry type, e.g.

```python
class User(BaseModel):
     name: str

     @staticmethod
     def foo() -> List[str]:
         return ["Patrick", "Pietro", "Pablo"]


 @strawberry.experimental.pydantic.type(model=User, all_fields=True)
 class UserType:
    ...

UserType(name="Patrick"). _original_model 
```
will cause a runtime exception.


I'll suggest for your example in the previous MR, that you do
```
@strawberry.experimental.pydantic.type(model=User, all_fields=True)
class UserType:
    @strawberry.field
    def upper_name(self: StrawberryTypeFromPydantic[User]) -> str:
        self_model = self.to_pydantic()
        return self_model.name.upper()
```
instead. It'll need to convert the model back into pydantic, which is kind of silly, but thats what you'll have to do if you want that guarantee. `to_pydantic` is a public method so that api would less likely be broken, unlike private methods.

patrick91:
@thejaminator feel free to merge this when ready :)

PR #2699, opened 2023-04-08
Move tasks into Operation object for GraphqlTransportWS handler

kristjanvalur:
Simplify the handling of tasks/generators for the GraphQLTransportWS handler.  This helps as a preparation for some restructuring I have planned to fix some actual problems soon.

## Description

- The handler now contains a single `operations` member which holds `Operation` instances.
  The `task` is an attribute of the object
- We no longer keep track of the `AsyncGenerator` objects especially.  They are closed inside the task when handling base exceptions and do not require special guard constructs.  The operation tasks themselves only need to be cancelled.
- We make necessary changes to test harnesses.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/tests/starlite/schema.py
+++ b/tests/starlite/schema.py
@@ -134,7 +134,7 @@ class Subscription:
     @strawberry.subscription
     async def debug(self, info) -> typing.AsyncGenerator[DebugInfo, None]:
         active_result_handlers = [
-            task for task in info.context["tasks"].values() if not task.done()
+            task for task in info.context["get_tasks"]() if not task.done()
         ]
 
         connection_init_timeout_task = info.context["connectionInitTimeoutTask"]
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+`graphql-transport-ws` handler now uses a single dict to manage active operations.
--- a/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
+++ b/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
@@ -51,8 +51,7 @@ class BaseGraphQLTransportWSHandler(ABC):
         self.connection_init_received = False
         self.connection_acknowledged = False
         self.connection_timed_out = False
-        self.subscriptions: Dict[str, AsyncGenerator] = {}
-        self.tasks: Dict[str, asyncio.Task] = {}
+        self.operations: Dict[str, Operation] = {}
         self.completed_tasks: List[asyncio.Task] = []
         self.connection_params: Optional[Dict[str, Any]] = None
 
@@ -85,7 +84,7 @@ class BaseGraphQLTransportWSHandler(ABC):
             with suppress(asyncio.CancelledError):
                 await self.connection_init_timeout_task
 
-        for operation_id in list(self.subscriptions.keys()):
+        for operation_id in list(self.operations.keys()):
             await self.cleanup_operation(operation_id)
         await self.reap_completed_tasks()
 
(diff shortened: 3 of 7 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

tests/starlite/schema.py
```diff
@@ -134,7 +134,7 @@ async def flavors(self) -> typing.AsyncGenerator[Flavor, None]:
     @strawberry.subscription
     async def debug(self, info) -> typing.AsyncGenerator[DebugInfo, None]:
         active_result_handlers = [
-            task for task in info.context["tasks"].values() if not task.done()
+            task for task in info.context["get_tasks"]() if not task.done()
```
DoctorJohn, line 137: Interesting, could you elaborate on putting a get_tasks method instead of the tasks list in the context? My first thought was that it would make sure that the tasks are always up to date when the debug subscription is used. Is that the intention? In that case I'm wondering (I'm not sure asking to make sure) whether the list would always be updated even without that change since it's never copied but passed by reference 🤔

DoctorJohn:
Thanks for yet another PR @kristjanvalur . I appreciate the changes and the Operation abstraction. Clearly there is more refactoring to be done and I'm looking forward to easier maintainable code. I left one comment just to make sure where on the same page.

tests/starlite/schema.py
kristjanvalur: So, yes:  Previously, both handler classes had a `tasks` member which was just a `dict`.  A reference to this dict was in the context.  And a fresh view of the active tasks could be got at any point by executing its `values()` member.

With this change, the tasks are not immediately available like that.  It also, imho, is too coupled to the implementation, even for testing code.  So instead, we have a "getter" method in the context.

The change to the tests isn't really that big, instead of having to call  `values()` on the property, it just needs to _call_ the property.

kristjanvalur:
> Thanks for yet another PR @kristjanvalur . I appreciate the changes and the Operation abstraction. Clearly there is more refactoring to be done and I'm looking forward to easier maintainable code. I left one comment just to make sure where on the same page.

Thanks.  To explain better, because I'm afraid I'm maybe just wading in without too much warning, here is what I have in mind, and what these PRs are building up to:

1. Moving all potentially long running steps of subscriptions into the task.
2. Fixing error handling for subscriptions, such that they behave like for Appollo
3. making handling of errors inside tasks more robust

Along the way, I am finding myself constantly pulled into different and facinating rabbit holes, such as the one opened by #2702 which makes this all so interesting.  I'll be making a PR for that one, which likely will include some small refactorings as well.  This in turn might prompt me to make a PR for Starlite.  Or Litestar.  But maybe that project just isn't ripe enough just yet.

tests/starlite/schema.py
DoctorJohn: Thanks for clearifying. So to summarize: The idea behind the `get_tasks` getting is to abstract the internals of task handling away from the test code shared by the different integrations. (correct? :eyes: )

kristjanvalur:
This PR is currently a prerequisite for two other prs which I have in progress (more socket blocking fixes and error handling).  Having a single "operation" object simplifies logistics.

tests/starlite/schema.py
kristjanvalur: Hi, sorry for late answer.  It's that, yes, and also so the the test code for the individual integrations need not know exactly how the handlers are storing their asyncio tasks, and if they do so differently.  It is enough to call the special get_tasks() method on the context created by the debuggable schema.

DoctorJohn:
> This PR is currently a prerequisite for two other prs which I have in progress

Sorry for the delay. Not sure why I didn't merge it after approving. I'll take another look to see whether I had any second thoughts or other reasons not to merge it yet

tests/starlite/schema.py
DoctorJohn: Thanks for clearifying!

DoctorJohn:
Okay this is good to go. Thanks again for the PR @kristjanvalur!

PR #2700, opened 2023-04-08
Add support for passing through custom dataclasses

### LINKED ISSUE #2688: Allow passing dataclass kwargs to the underlying dataclass
## Feature Request Type

- [ ] Core functionality
- [ ] New behavior

## Description

Currently `_wrap_dataclass` doesn't allow setting kwargs to the dataclass it creates. This leads to the creation (among other things) of a `__repr__` method, which may not be desired.

Case in point, I have classes that are SQLAlchemy models and wrapped with `type` and who inherits a generic `__repr__`.
However, the underlying dataclass creates a new `__repr__` method which makes me load all my attributes. (Which is obviously undesired).

jkimbo:
## Description

This PR adds support for `strawberry.type` to handle dataclasses that have custom parameters (e.g. `repr=False`) rather than recreating the dataclass from scratch.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* Fixes #2688

## Checklist

### DIFF
--- a/strawberry/experimental/pydantic/object_type.py
+++ b/strawberry/experimental/pydantic/object_type.py
@@ -223,7 +223,11 @@ def type(
         if hasattr(cls, "resolve_reference"):
             namespace["resolve_reference"] = cls.resolve_reference
 
-        kwargs: Dict[str, object] = {}
+        # Copy the dataclass params from the underlying pydantic dataclass
+        dclass_params = wrapped.__dataclass_params__  # type: ignore
+        dclass_kwargs: Dict[str, object] = {
+            slot: getattr(dclass_params, slot) for slot in dclass_params.__slots__
+        }
 
         # Python 3.10.1 introduces the kw_only param to `make_dataclass`.
         # If we're on an older version then generate our own custom init function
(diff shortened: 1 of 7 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

patrick91:
ah! nice one!

Corentin-Br:
With your PR, I have a problem in my code. I believe it's related to generic types (and multiple decorations). I'll check it out to understand the exact issue. Maybe it can be solved from my code.

jkimbo:
@Corentin-Br do you have an example that’s failing?

Corentin-Br:
```python
T = TypeVar("T")

@dataclass
class MyGeneric(Generic[T]):
    attr: T

T = TypeVar("T")

@strawberry.type
class MyGenericChild(MyGeneric[T], Generic[T]):
    attr: T

def resolve() -> MyGenericChild[int]:
    ...

@strawberry.type
class Query:
        concrete_resolver = strawberry.field(resolver=resolve)
``` 

=> `KeyError: ~T`

Wheras without the dataclass it works.

In my case, it happens because I import `MyGeneric` and redefine a TypeVar (effectively leading to an equivalence with the minimal example I gave).
In my specific case again, I can easily work around it, but it is a breaking change.

patrick91:
Someone mentioned this on Discord not long ago, I wonder if we should still do it? 

I think I'm worried a bit about having to maintain all the kwargs, but maybe it's not an issue? /cc @bellini666 @DoctorJohn

bellini666:
> Someone mentioned this on Discord not long ago, I wonder if we should still do it?
> 
> I think I'm worried a bit about having to maintain all the kwargs, but maybe it's not an issue? /cc @bellini666 @DoctorJohn

I think it should be ok. At the end of the day, strawberry types are "fancy" dataclasses =P, so why not allow them to behave as such and allow those kind of customizations, like `repr=True`? 😊

Checking the current `kwargs` a dataclass decorator can receive, I don't think any of those would conflict with how strawberry works. The only concerning one IMO would be `slots`, but this PR is already handling that

Issue #2702, opened 2023-04-09
Issue with starlite websockets test


kristjanvalur:
Just a heads up, there is an issue with the Starlite websockets test client which causes unittests to sometimes hang.
This is some issue with starlite itself, because the websocket_connect() method deadlocks if it raises a WebsocketDisconnect.

I'm working on a PR to disable the test in question, and to fix some race conditions regartig websocket connect timeouts.

## System Information

 - Operating system:
 - Strawberry version (if applicable):

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/strawberry-graphql) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/strawberry-graphql/strawberry/issues/2702">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/2702/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/2702/pledge.svg">
</picture>
</a>

kristjanvalur:
I guess this is closed with the merging of 2703

PR #2703, opened 2023-04-10
Fix problems with timeouts in graphql_transport_ws

kristjanvalur:
This fixes an issue mentioned in #2702

## Description

After merging uniform websocket tests for Starlite extension, sporadic deadlocks were observed.
It turned out that  the timeout trigger, part of the `graphql_transport_ws` protocol, could trigger too early,
when the initial `websockets` handshake was still being done.  This caused the whole websocket connection
attempt to be rejected and this triggered a deadlock in Starlite, which still appears a bit un-robust.

This PR does a few things

1. provides an API to start the timeout, after the integration specific handler has completed the Websockets handshake.
2. Add initial synchronization (`connection_timed_out:bool`) to ensure that there is never a race between a timeout and accepting a connection.
3. Robustly terminate the timeout task when it is done, or no longer needed.
4. Add error handing around the timeout task.  
5. Add a Task error handling hook, to report unhandled errors in Tasks.  This is not yet in use, need to consult on which log handler to use, perhaps "strawberry.task"
6. Provide a shutdown() api to the integrations, rather than the former, rather unwieldly, boilerplate code.

## Types of Changes

- [x] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #2702 

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+`grapql-transport-ws`: Don't start the sub-protocol timeout until the `websockets` connection handshake is complete.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,3 @@
 Release type: patch
 
-`grapql-transport-ws`: Don't start the sub-protocol timeout until the `websockets` connection handshake is complete.
+This release improves the `graphql-transport-ws` implementation by starting the sub-protocol timeout only when the connection handshake is completed.
--- a/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
+++ b/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
@@ -102,7 +102,7 @@ class BaseGraphQLTransportWSHandler(ABC):
             await asyncio.sleep(delay=delay)
 
             if self.connection_init_received:
-                return
+                return  # pragma: no cover
 
             self.connection_timed_out = True
             reason = "Connection initialisation timeout"
@@ -110,7 +110,7 @@ class BaseGraphQLTransportWSHandler(ABC):
         except asyncio.CancelledError:
             raise
         except Exception as error:
-            await self.handle_task_exception(error)
+            await self.handle_task_exception(error)  # pragma: no cover
         finally:
             # do not clear self.connection_init_timeout_task
             # so that unittests can inspect it.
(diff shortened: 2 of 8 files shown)


kristjanvalur:
Note that I am not exactly sure why the deadlock occurs, it could be a result of the Strawberry integration.  I was uncsuccessful in reproducing it in the plain Starlite repo, and indeed, it contains unittests for websocket rejection using 'close' before an 'accept' message is sent.
But Strawberry always accepts a websocket connection.  The protocols for GraphQL instead provide for a rejection to be done as a result of the sub-protocol handshake.

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -73,19 +74,51 @@ async def handle_request(self) -> Any:
...
+            self.connection_init_timeout_task.cancel()
+            with suppress(asyncio.CancelledError):
+                await self.connection_init_timeout_task
+
+        for operation_id in list(self.subscriptions.keys()):
+            await self.cleanup_operation(operation_id)
+        await self.reap_completed_tasks()
+
+    def on_request_accepted(self) -> None:
+        #  handle_request should call this once it has sent the
```
patrick91, line 90:
```suggestion
        # handle_request should call this once it has sent the
```

RELEASE.md
```diff
@@ -0,0 +1,3 @@
+Release type: patch
+
+`grapql-transport-ws`: Don't start the sub-protocol timeout until the `websockets` connection handshake is complete.
```
```suggestion
This release improves the `graphql-transport-ws` implementation by starting the sub-protocol timeout only when the connection handshake is completed.
```

kristjanvalur:
One question:  
I added a skeleton error handler for background tasks.  The only sensible thing to do is to log them.  but I didn't implement that for this PR.  I could do that now, if we agree on a log channel to use.  I suggested `strawberry.task`, maybe that is ok? 
These are the only tasks created by strawberry, all other top level task logging is done by the web framework in question.

patrick91:
> One question: I added a skeleton error handler for background tasks. The only sensible thing to do is to log them. but I didn't implement that for this PR. I could do that now, if we agree on a log channel to use. I suggested `strawberry.task`, maybe that is ok? These are the only tasks created by strawberry, all other top level task logging is done by the web framework in question.

when are these errors happening?

kristjanvalur:
> when are these errors happening?

Never, hopefully.  But whenever you create a background task, it is prudent to install a top-level error handler to catch and log it.  If you don't, Python will create a warning about an task with an exception not being "awaited" but that warning may end up anywhere.  In case of the timeout thread, there really aren't many things which can go wrong. But for the subscription thread, all kinds of errors can occur and it is best to handle them in-task by a top level error handler.  

The alternative is to have the main task "await" all background tasks and catch and log any errors which occur there.

PR #2717, opened 2023-04-19
graphql_transport_ws: Fix task shutdown/cancellation issues evident from testsuite warnings

kristjanvalur:
Examining the warnings after running the unittests for `test_graphql_transport_ws.py`
showed some problems in how we were handling task cleanup and cancellation.

## Description

- AsyncGenerator doesn't need to be explicitly closed, but for expedient shutdown it should be done from witin
  within the task, using Finally semantics.
- The current `Task` instance should be retrieved at the start, not during shutdown when the event loop may be gone.
- CancelledError should raise out of the task.
- The cleanup of current operation id can race with an external `cleanup_operation` so must not fail if the id isn't on record.
- Added a unittest to verify that the subscription AsyncGenerators are properly finalized when an operation is cancelled.

As this is not a feature enhancement, a Release file is probably not warranted.  Also, this PR belongs in the group of
"subscription enhancements" that I am working on.

## Types of Changes

- [x] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
+++ b/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
@@ -276,11 +276,6 @@ class BaseGraphQLTransportWSHandler(ABC):
             await self.handle_async_results(result_source, operation)
         except BaseException:  # pragma: no cover
             # cleanup in case of something really unexpected
-            # wait for generator to be closed to ensure that any existing
-            # 'finally' statement is called
-            result_source = self.subscriptions[operation.id]
-            with suppress(RuntimeError):
-                await result_source.aclose()
             if operation.id in self.subscriptions:
                 del self.subscriptions[operation.id]
                 del self.tasks[operation.id]
@@ -320,6 +315,12 @@ class BaseGraphQLTransportWSHandler(ABC):
             await operation.send_message(error_message)
             self.schema.process_errors([error])
             return
+        finally:
+            # Make sure the AsyncGenerator is closed promptly, not as part
+            # of the garbage collection process
+            with suppress(RuntimeError):
+                # Guard us against an asynchronous shutdown
+                await result_source.aclose()
 
     def forget_id(self, id: str) -> None:
         # de-register the operation id making it immediately available
@@ -340,7 +341,7 @@ class BaseGraphQLTransportWSHandler(ABC):
     async def cleanup_operation(self, operation_id: str) -> None:
         if operation_id not in self.subscriptions:
             return
-        result_source = self.subscriptions.pop(operation_id)
+        del self.subscriptions[operation_id]
         task = self.tasks.pop(operation_id)
         task.cancel()
         # do not await the task here, lest we block the main
--- a/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
+++ b/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
@@ -304,7 +304,7 @@ class BaseGraphQLTransportWSHandler(ABC):
                     next_message = NextMessage(id=operation.id, payload=next_payload)
                     await operation.send_message(next_message)
         except asyncio.CancelledError:
-            # CancelledErrors are expected during task cleanup.
+            # Not a BaseException until 3.8.
             raise
         except Exception as error:
             # GraphQLErrors are handled by graphql-core and included in the
--- a/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
+++ b/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
@@ -304,7 +304,9 @@ class BaseGraphQLTransportWSHandler(ABC):
                     next_message = NextMessage(id=operation.id, payload=next_payload)
                     await operation.send_message(next_message)
         except asyncio.CancelledError:
-            # Not a BaseException until 3.8.
+            # Need to catch and raise this separately because in versions <= 3.7
+            # asyncio.CancelledError is an Exception, not a direct subclass of
+            # BaseException
             raise
         except Exception as error:
             # GraphQLErrors are handled by graphql-core and included in the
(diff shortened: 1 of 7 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -228,15 +228,12 @@ async def operation_task(
         Operation task top level method.  Cleans up and de-registers the operation
         once it is done.
         """
+        task = asyncio.current_task()
+        assert task is not None
```
patrick91, line 232: is this just for mypy?

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -267,7 +262,7 @@ async def handle_async_results(
                     next_message = NextMessage(id=operation.id, payload=next_payload)
                     await operation.send_message(next_message)
         except asyncio.CancelledError:
-            # CancelledErrors are expected during task cleanup.
+            # Not a BaseException until 3.8.
```
patrick91, line 265: not sure what this comment means 😊

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -298,7 +299,7 @@ async def send_message(self, message: GraphQLTransportMessage) -> None:
     async def cleanup_operation(self, operation_id: str) -> None:
         if operation_id not in self.subscriptions:
             return
-        result_source = self.subscriptions.pop(operation_id)
+        del self.subscriptions[operation_id]
```
patrick91, line 302: what's the difference between these two?

tests/websockets/test_graphql_transport_ws.py
```diff
@@ -33,6 +33,28 @@ async def ws_raw(http_client: HttpClient) -> AsyncGenerator[WebSocketClient, Non
     await ws.close()
     assert ws.closed
 
+    # loop cleanup code, similar to code from asyncio.run()
+    # disabled, can be enabled for debugging
+    return
+
+    loop = asyncio.get_running_loop()
```
patrick91, line 40: could we put this under a `if DEBUG_ENABLED` flag? just to make it clearer that it can be used for that?

tests/websockets/test_graphql_transport_ws.py
```diff
@@ -789,3 +811,76 @@ async def test_subsciption_cancel_finalization_delay(ws: WebSocketClient):
...
+            continue
+        assert response["payload"]["data"]["finalizerState"] is True
+        break
+
+    # cancel the subscription
+    await ws.send_json(CompleteMessage(id="sub1").as_dict())
+
+    # wait until context is dead.
+    # We don't know exactly how many packets will arrive or how long it will take.
+    # Need manual timeout because async timeout doesn't work on async integrations
```
patrick91, line 858: > Need manual timeout because async timeout doesn't work on async integrations

how come?

patrick91 (commented):
I've left some comments, I'm not super familiar with the WS work, though, so I'd wait for @DoctorJohn for a final review.

Oh and since we are changing running code, we should make a release 😊

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
kristjanvalur: Yes, exactly.  
Btw, in a future PR, I aim to get rid of this code, since it isn't necessary to "reap" background tasks.  This is something I realized recently, I was confused with how "coroutines" must always be awaited.  But this PR focuses on getting rid of current warnings from the unit tests.

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
kristjanvalur: Oh.  I'm commenting on the fact that we are catching an exception just to re-raise it.  In python 3.8 this code isn"t necessary because CancelledError is a `BaseException` and there is no BaseException clause.    But in 3.7, we need to catch and raise CancelledError so that it isn't caught by the `Exception`  clause.
Normally, CancelledErrors aren't caught at all in more recent python versions except in very rare cases.  Do you have a suggestion on how to better phrase the comment?

tests/websockets/test_graphql_transport_ws.py
kristjanvalur: Sure.  The thinking here is that if we are having strange errors/warnings in the unittests, enabling this code to forcefully cleanup tasks after each test fixture should help with isolating the problem.  I can also expand on the comment.

tests/websockets/test_graphql_transport_ws.py
kristjanvalur: Because the actual IO calls are actually blocking, and the event loop running the test will stop until the call returns, so no "async" timeouts can be delivered.  And not all of the integrations support a "timeout" parameter to their receive methods.

kristjanvalur:
Let's not review this any more until I have had time to clean it up a bit.

patrick91:
@kristjanvalur is this ready for a review?

kristjanvalur:
I think, since this is a test suite pr, that I'll roll in a fix to the timeout failure for the cancellation test which sometimes shows up. Allow until tomorrow please for me, I have a change to the unit test in mind which doesn't require timeout checks, they are doomed to occasionally fail in a virtualize environment.

kristjanvalur:
Ok, I found that the problems with the finalizer termination tests are hard to solve right now.  The problem is that the iterator/generator created by graphql-core, for iterating over subscriptions, has such weird semantics.  The finalizer is actually called on a Task, and may get _cancelled_ during execution.  I cannot make good tests for this until we adopt something like PR #2724

PR #2724, opened 2023-04-23
Use our own iterator to iterate and transform subscription results

kristjanvalur:
Provide a generator based iterator to iterate over and transform the subscription results from graphql core.

## Description

In https://github.com/graphql-python/graphql-core/pull/197 I descibe how the iterator provided by graphql core is
problematic.  Its `aclose()` semantics are unexpected and iteration happens on a subtask.  A much simpler approach is possible which is both more performant (no subtasks), more easily debuggable (call stack isn't broken on `__anext__()` and where `await acancel()` actually only returns when the iterator has been fully cancelled.

A concrete problem with the current iterator is that during the handling of a `GeneratorExit` exception, e.g. in a `finally:` clause entered because the generator is closed, it is likely that one receives a `CancelledError` if one tries to do any `async` operations.  This makes it hard to write test which test the finalization of subscriptions, and will make it impossible to do proper cleanup of async resources in case of subscription cancellation.  This does not happen using regular async-generators.

We _inspect_ the iterator we get from graphql-core.  If it appears to be the old type with its internal `_close_event`, we replace it with our own implementation.

This is intended as an interim solution until graphql-core provides a better iterator which more closely conforms to the semantics one is to expect from Python's AsyncGenerators.

The change helps avoid the strange "RuntimeErrors()" we were seeing when calling `aclose()` despite doing so in a legal
way from the same task which was iterating, and in a sense is a bugfix.

It should also be noted that an iterator implemented this way, with an Async Generator, does not _require_ an explicit call to "aclose()" since it will be scheduled when it goes out of scope.  But expliticly doing so may be prudent if one wants the cleanup to happen promptly.

Upstream I have submitted both In https://github.com/graphql-python/graphql-core/pull/197 and https://github.com/graphql-python/graphql-core/pull/199.  It is unclear if and when these will ever be accepted.  Regardless of that, this PR _also_ provides a framework for adding support for _extensions_ to subscriptions, since they would need a custom iterator anyway.

## Types of Changes

- [x] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist


kristjanvalur:
Can't find where the coverage tests are failing...  Are these tests actually correct?

patrick91:
@kristjanvalur I don't this so, not sure if it is a misconfiguration or just coverage being bad

kristjanvalur:
@jthorniley something like this?

The current iterator truly is a can of worms.  Cancelling it can result in unpredictable behaviour.  For example, during processing of a `GeneratorExit` in the finalizer, a `CancelledError` may also arrive.  This makes any reasonable logic operations in the finalizer hard to implement.

the error currently visible in the the unit tests is actually due to this.  We are trying to measure the responsiveness of code which isn't well predictable.  I already tried making this unit test work without timing, but was foiled by the above, CancelledErrors coming out of nowhere.

jthorniley:
Yep, in terms of the processing being done inside subscribe, this is in line with what I was thinking - thanks!

kristjanvalur:
After fixing the subscription iterator, it is finally possible to get deterministic behaviour for subscription finalizers.  Previously they could get bot GeneratorExit _and_ a CancelledError.  it made it impossible to do any async logic in there, like wait for condition variables etc.  I have fixed the test verifying that subscription cancellation does not block the connection, by using synchronization primitives rather than timeout measruements.

kristjanvalur:
> Yep, in terms of the processing being done inside subscribe, this is in line with what I was thinking - thanks!

This can also now be expanded upon to add extension processing to subscriptions.

kristjanvalur:
graphql-core has this functionality in https://github.com/graphql-python/graphql-core/releases/tag/v3.3.0a3
Since upcoming release also changes the signature of "subscribe()" to be hybrid-async, I think I think I should probably 
close this PR in favour of #2784

DoctorJohn:
Closing this one because the upstream patch landed and extension support for subscriptions has been added in another PR.

PR #2744, opened 2023-05-01
feat: support lazy references with forward references

### LINKED ISSUE #2504: new strawberry.lazy not compatible with from __future__ import annotations
# Description

The new way of lazy referencing via strawberry.lazy is not compatible with 
`from __future__ import annotations`

It would be nice to have compatibility as typing.Self is not available for python <=3.10 (the main reason to have `from __future__ import annotations`)

bellini666:
This should make resolving `Annotated[SomeType, strawberry.lazy("some.module")]` work when using `from __future__ import annotations`.

This also means that it is important for supporting python 3.12, which will work with future annotations by default.

Fix #2504

### DIFF
--- a/strawberry/utils/typing.py
+++ b/strawberry/utils/typing.py
@@ -223,7 +223,7 @@ def _ast_replace_union_operation(
 
 
 _annotated_re = re.compile(
-    r"(typing\.Annotated|Annotated\[)(?P<type>\w*),(?P<args>.*)(\])",
+    r"(Annotated\[)(?P<type>\w*),(?P<args>.*)(\])",
 )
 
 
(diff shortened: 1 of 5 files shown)


botberry:



---

This release adds support for properly resolving lazy references
when using forward refs.

For example, this code should now work without any issues:

```python

if TYPE_CHECKING:
    from some.module import OtherType


@strawberry.type
class MyType:
    @strawberry.field
    async def other_type(self) -> Annotated[OtherType, strawberry.lazy("some.module")]:
        ...
```


---




Here's the tweet text: 
```

```

strawberry/utils/typing.py
```diff
@@ -249,6 +255,36 @@ def eval_type(
...
+        # and trying to _eval_type on it will fail. Take a different approach
+        # here to resolve lazy types by execing the annotated args and resolving
+        # the type directly.
+        annotated_match = _annotated_re.match(type_.__forward_arg__)
+        if annotated_match:
+            gdict = annotated_match.groupdict()
+            ldict = (localns or {}).copy()
+            # Exec the remaining annotated args to get their real values and
+            # put the result in ldict["args"]
+            exec(f'args = ({gdict["args"]}, )', globalns, ldict)
```
patrick91, line 269: could we use eval here?

btw: this makes me a bit antsy :D

strawberry/utils/typing.py
```diff
@@ -249,6 +255,36 @@ def eval_type(
...
+            args = ldict["args"]
+
+            for arg in args:
+                if isinstance(arg, StrawberryLazyReference):
```
ThirVondukr, line 273: I would prefer to use `not isinstance(...)` to avoid nesting, maybe we could do the same with `if annotated_match` check?

strawberry/utils/typing.py
```diff
@@ -221,6 +222,11 @@ def _ast_replace_union_operation(
     return expr
 
 
+_annotated_re = re.compile(
+    r"(typing\.Annotated|Annotated\[)(?P<type>\w*),(?P<args>.*)(\])",
```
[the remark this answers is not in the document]
patrick91: > Maybe it would make more sense to just match Annotated and use re.search? People could alias typing import as t or as something else

that's a valid point!

strawberry/utils/typing.py
bellini666: Sure!Totally forgot about `eval` lol

> btw: this makes me a bit antsy :D

hahaha me too. The major issue here is that the whole thing is a string, but we need its value to resolve the lazy part. Don't know if there's a better option for this, it was the best I could think of =P (but open to suggestions)

patrick91: we should check how pydantic deals with these things, but we can do that another time 😊

bellini666: I did a quick check, and they have a similar solution to us: https://github.com/pydantic/pydantic/blob/c600c010418c46cec1f3a0fe8fb6ca4be4a7949f/pydantic/_internal/_typing_extra.py#L319

I don't think they workaround this specific issue (maybe they don't have it?). They call [ForwardRef._evaluate](https://github.com/python/cpython/blob/main/Lib/typing.py#L862) directly, but we can't do that because we don't have the module in globals/locals, and the only way to retrieve it is to execute the `strawberry.lazy(...)` argument (or parse it and call `__import__` there).

We might want to revisit how lazy works another time to allow us to better support this :)

strawberry/utils/typing.py
bellini666: I changed it to use `next(...)` to avoid nesting. Should solve this and the issue above

bellini666:
@patrick91 @ThirVondukr anything missing for this?

Also, any idea why the coverage is not hitting 100% in this PR? I don't see any comments in the diff page of any line not being covered

ThirVondukr:
@bellini666 Code looks good to me, seems like `utils/typing.py -> is_union` is missing coverage now?

bellini666:
> @bellini666 Code looks good to me, seems like `utils/typing.py -> is_union` is missing coverage now?
> 
> https://app.codecov.io/gh/strawberry-graphql/strawberry/pull/2744/blob/strawberry/utils/typing.py#L92

So strange, I did not change anything that would affect that code.

Also, that code is not covered in the `main` branch itself (just tested it locally at the main branch). Don't know why codecov thinks that this PR is the culprit here =P

patrick91 (approved):
Thanks!

Ronjea:
thanks!

PR #2749, opened 2023-05-04
Docs: Ensure syntax highlighting is present (and consistent)

watfordkcf:
## Description

I noticed some code fences weren't being syntax highlighted, so I did a quick inventory and made sure they were present and consistent.

Question: is it worth standardizing the naked code fences with shell code to `bash` or `shell`? I noticed this inconsistency, but did not include those wider changes here.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [X] Documentation

## Checklist

- [X] I have updated the documentation accordingly.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -108,7 +108,7 @@ Poetry.
 You will first need to clone the repository using `git` and place yourself in
 its directory:
 
-```bash
+```shell
 $ git clone git@github.com:strawberry-graphql/strawberry.git
 $ cd strawberry
 ```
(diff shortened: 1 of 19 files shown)


patrick91:
> Question: is it worth standardizing the naked code fences with shell code to bash or shell? I noticed this inconsistency, but did not include those wider changes here.

let's do that, let's go with shell since those commands are for people to run in the shell 😊

watfordkcf:
I left out the changelog as I'm not quite sure if that was auto-generated or by hand.

patrick91 (approved):
Awesome! Thank you!

patrick91:
> I left out the changelog as I'm not quite sure if that was auto-generated or by hand.

this is fine, they are autogenerated, and not used yet on the site 😊

PR #2773, opened 2023-05-17
Allow additional tags in DatadogTracingExtension

ronaldnwilliams:
This PR adds an optional parameter, `additional_tags_func`, to the `DatadogTracingExtension` class.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

N/A

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -5,13 +5,14 @@ This release adds a method on the DatadogTracingExtension class called `create_s
 ```python
 from ddtrace import Span
 
-from strawberry.extensions.tracing import DatadogTracingExtension, LifeCycleStep
+from strawberry.extensions import LifecycleStep
+from strawberry.extensions.tracing import DatadogTracingExtension
 
 
 class DataDogExtension(DatadogTracingExtension):
     def create_span(
         self,
-        lifecycle_step: LifeCycleStep,
+        lifecycle_step: LifecycleStep,
         name: str,
         **kwargs,
     ) -> Span:
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,22 @@
+Release type: patch
+
+This release adds a method on the DatadogTracingExtension class called `create_span` that can be overridden to create a custom span or add additional tags to the span.
+
+```python
+from ddtrace import Span
+
+from strawberry.extensions.tracing import DatadogTracingExtension, LifeCycleStep
+
+
+class DataDogExtension(DatadogTracingExtension):
+    def create_span(
+        self,
+        lifecycle_step: LifeCycleStep,
+        name: str,
+        **kwargs,
+    ) -> Span:
+        span = super().create_span(lifecycle_step, name, **kwargs)
+        if lifecycle_step == LifeCycleStep.OPERATION:
+            span.set_tag("graphql.query", self.execution_context.query)
+        return span
+```
--- a/docs/extensions/datadog.md
+++ b/docs/extensions/datadog.md
@@ -66,13 +66,14 @@ Example:
 ```python
 from ddtrace import Span
 
-from strawberry.extensions.tracing import DatadogTracingExtension, LifeCycleStep
+from strawberry.extensions import LifecycleStep
+from strawberry.extensions.tracing import DatadogTracingExtension
 
 
 class DataDogExtension(DatadogTracingExtension):
     def create_span(
         self,
-        lifecycle_step: LifeCycleStep,
+        lifecycle_step: LifecycleStep,
         name: str,
         **kwargs,
     ) -> Span:
--- a/docs/extensions/datadog.md
+++ b/docs/extensions/datadog.md
@@ -50,3 +50,34 @@ schema = strawberry.Schema(
 ```
 
 </Note>
+
+## API reference:
+
+_No arguments_
+
+## Extending the extension
+
+### Overriding the `create_span` method
+
+You can customize any of the spans or add tags to them by overriding the `create_span` method.
+
+Example:
+
+```python
+from ddtrace import Span
+
+from strawberry.extensions.tracing import DatadogTracingExtension, LifeCycleStep
+
+
+class DataDogExtension(DatadogTracingExtension):
+    def create_span(
+        self,
+        lifecycle_step: LifeCycleStep,
+        name: str,
+        **kwargs,
+    ) -> Span:
+        span = super().create_span(lifecycle_step, name, **kwargs)
+        if lifecycle_step == LifeCycleStep.OPERATION:
+            span.set_tag("graphql.query", self.execution_context.query)
+        return span
+```
--- a/strawberry/extensions/base_extension.py
+++ b/strawberry/extensions/base_extension.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from enum import Enum
 from typing import TYPE_CHECKING, Any, Callable, Dict, Set
 
 from strawberry.utils.await_maybe import AsyncIteratorOrIterator, AwaitableOrValue
@@ -8,6 +9,13 @@ if TYPE_CHECKING:
     from graphql import GraphQLResolveInfo
 
     from strawberry.types import ExecutionContext
+
+
+class LifecycleStep(Enum):
+    OPERATION = "operation"
+    VALIDATION = "validation"
+    PARSE = "parse"
+    RESOLVE = "resolve"
 
 
 class SchemaExtension:
--- a/strawberry/extensions/tracing/datadog.py
+++ b/strawberry/extensions/tracing/datadog.py
@@ -40,7 +40,7 @@ class DatadogTracingExtension(SchemaExtension):
         self,
         lifecycle_step: LifecycleStep,
         name: str,
-        **kwargs,
+        **kwargs: Any,
     ) -> Span:
         """
         Create a span with the given name and kwargs.
--- a/strawberry/extensions/tracing/datadog.py
+++ b/strawberry/extensions/tracing/datadog.py
@@ -4,9 +4,9 @@ import hashlib
 from inspect import isawaitable
 from typing import TYPE_CHECKING, Any, Callable, Generator, Iterator, Optional
 
-from ddtrace import tracer
+from ddtrace import Span, tracer
 
-from strawberry.extensions import SchemaExtension
+from strawberry.extensions import LifecycleStep, SchemaExtension
 from strawberry.extensions.tracing.utils import should_skip_tracing
 from strawberry.utils.cached_property import cached_property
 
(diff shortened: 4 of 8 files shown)


RELEASE.md
```diff
@@ -0,0 +1,25 @@
...
+
+
+schema = strawberry.Schema(
+    Query,
+    extensions=[
+        DatadogTracingExtension(
+            additional_tags_func=additional_tags_func,
+        ),
+    ],
+)
```
patrick91, line 25:
````suggestion
)
```
````

**Review by patrick91 — COMMENTED** [review]
thanks for doing this PR! 

I think we should make it easy to add tags to span, but I'm unsure if this is the right approach 😊

my main issue is that this function is only used for the request span, while we do have more spans (for example the parsing span)

For your use case, do you think you could subclass the extension? If so we can provide more hooks for getting the request span or getting the tags for them?

**Comment by patrick91** [comment]
what the extension had a function/property for the request tag (and for other tags), for example:

```python
    def _get_span(name: str)
        # here we could use name or additional kwargs to add more tags?
        return tracer.trace(
            name,
            resource=self._resource_name,
            span_type="graphql",
            service="strawberry",
        )


    def on_operation(self) -> Iterator[None]:
        self._operation_name = self.execution_context.operation_name
        span_name = (
            f"{self._operation_name}" if self._operation_name else "Anonymous Query"
        )

        self.request_span = self._get_span(span_name)
        self.request_span.set_tag("graphql.operation_name", self._operation_name)

```

ronaldnwilliams:
Hello, @patrick91! Thank you for the comments. I am open to a different solution. Do you have any ideas on how I could add a tag that adds the query of every request? 

`request_span.set_tag('graphql.query', execution_context.query)`

patrick91:
sure, I'd go with a design like this:

```
    def _get_span(self, lifecycle_step: Literal["request", "validation", ...], span_name: str, **kwargs):
        # here we could use name or additional kwargs to add more tags?
        span = tracer.trace(
            span_name,
            **kwargs
        )

        if lifecycle_step == "request":        
            span.set_tag("graphql.operation_name", self._operation_name)

        return span


    def on_operation(self) -> Iterator[None]:
        ...
        self.request_span = self._get_span(span_name, 
            resource=self._resource_name,
            span_type="graphql",
            service="strawberry",
        )
```

does it make sense? :)

RELEASE.md
```diff
@@ -0,0 +1,21 @@
+Release type: patch
+
+This release adds a method on the DatadogTracingExtension class called `_create_span` that can be overridden to create a custom span or add additional tags to the span.
+```python
+from ddtrace import Span
+
+from strawberry.extensions.tracing import DatadogTracingExtension, LifeCycleStep
```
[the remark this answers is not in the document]
patrick91: can we also call it `LifecycleStep`? :D 

same for the variable below `lifecycle_step` :)

docs/extensions/datadog.md
```diff
@@ -50,3 +50,34 @@ schema = strawberry.Schema(
...
 
 </Note>
+
+## API reference:
+
+_No arguments_
+
+## Extending the extension
+
+### Overriding the `_create_span` method
```
patrick91, line 60: 💯

patrick91 (commented):
Looks great, I've only added a couple of minor notes 😊

strawberry/extensions/tracing/datadog.py
```diff
@@ -45,34 +78,38 @@ def on_operation(self) -> Iterator[None]:
             f"{self._operation_name}" if self._operation_name else "Anonymous Query"
         )
 
-        self.request_span = tracer.trace(
+        self.request_span = self._create_span(
+            LifeCycleStep.REQUEST,
```
nrbnlulu, line 82: Note that this is not really a request it is operation, one request can have many operations. especially if you use websockets and your connection is alive for as long as the client might query for something.

RELEASE.md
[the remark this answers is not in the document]
nrbnlulu: @patrick91 

> can we move LifeCycleStep to strawberry.extensions

I think it's quite trivial to implement this yourself no?, is that a pattern you see frequently?

nrbnlulu (changes requested):

strawberry/extensions/tracing/datadog.py
patrick91: ah, yes, good point 😊

RELEASE.md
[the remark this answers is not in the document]
patrick91: we'll probably use this for all the tracing extensions, so I think it is worth keeping in our code base and inside extensions 😊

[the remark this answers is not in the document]
nrbnlulu: I can't think of an easy way to do this, this would be either to add another argument on the hook signature or to use some decoration mechanism.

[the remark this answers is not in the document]
patrick91: oh, I don't think we need to integrate it deeply, we can just keep the variable 😊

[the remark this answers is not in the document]
nrbnlulu: Is that async safe?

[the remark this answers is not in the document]
patrick91: yup, I'm talking about this:

```
class LifeCycleStep(Enum):
    REQUEST = "request" # use operation here
    VALIDATION = "validation"
    PARSE = "parse"
    RESOLVE = "resolve"
``` 

we keep this for extension author to use, but we don't much with it :)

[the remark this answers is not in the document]
nrbnlulu: oh cool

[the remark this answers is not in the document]
ronaldnwilliams: Renamed and moved the enum to strawberry/extensions/base_extension.py. Let me know if you would prefer it in a different file.

strawberry/extensions/tracing/datadog.py
ronaldnwilliams: Renamed this to OPERATION.

ronaldnwilliams:
@patrick91 @nrbnlulu I made the requested changes. Great suggestions. Looks like there is a similar enum defined in the opentelemetry.py tracer here: https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/extensions/tracing/opentelemetry.py#L39

Do you want me to look into having the opentelemetry us this LifecycleStep enum? Looks like it is only used as keys to a dictionary so changing it to the new enum should not have any side effects.

patrick91:
> @patrick91 @nrbnlulu I made the requested changes. Great suggestions. Looks like there is a similar enum defined in the opentelemetry.py tracer here: [`main`/strawberry/extensions/tracing/opentelemetry.py#L39](https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/extensions/tracing/opentelemetry.py?rgh-link-date=2023-05-18T12%3A41%3A40Z#L39)
> 
> Do you want me to look into having the opentelemetry us this LifecycleStep enum? Looks like it is only used as keys to a dictionary so changing it to the new enum should not have any side effects.

oh, that would be amazing if you have time, thanks!

ronaldnwilliams:
@patrick91 @nrbnlulu ✅ Hopefully all jobs pass and this should be ready to 🚀

strawberry/extensions/base_extension.py
```diff
@@ -10,6 +11,13 @@
     from strawberry.types import ExecutionContext
 
 
+class LifecycleStep(Enum):
```
nrbnlulu, line 14: I think it would fit better in https://github.com/strawberry-graphql/strawberry/blob/51eb183a317722480bc08cfc1032bfdad048e253/strawberry/extensions/context.py#L201

And the name might be `ExtensionLifecycle` / `OperationLifeCycle`
Also I would add a comment there for it's purpose.

@patrick91 What do you think?

strawberry/extensions/tracing/datadog.py
```diff
@@ -36,6 +37,31 @@ def _resource_name(self):
 
         return query_hash
 
+    def _create_span(
```
nrbnlulu, line 40: Why is that `_`private if it is intended to be overridden? Am I missing something?

nrbnlulu (changes requested):

strawberry/extensions/tracing/datadog.py
ronaldnwilliams: Good call, I think this should be public.

docs/extensions/datadog.md
```diff
@@ -50,3 +50,34 @@ schema = strawberry.Schema(
...
+### Overriding the `create_span` method
+
+You can customize any of the spans or add tags to them by overriding the `create_span` method.
+
+Example:
+
+```python
+from ddtrace import Span
+
+from strawberry.extensions.tracing import DatadogTracingExtension, LifeCycleStep
```
patrick91, line 69:
```suggestion
from strawberry.extensions import LifecycleStep
from strawberry.extensions.tracing import DatadogTracingExtension
```

docs/extensions/datadog.md
```diff
@@ -50,3 +50,34 @@ schema = strawberry.Schema(
...
+
+
+class DataDogExtension(DatadogTracingExtension):
+    def create_span(
+        self,
+        lifecycle_step: LifeCycleStep,
```
patrick91, line 75:
```suggestion
        lifecycle_step: LifecycleStep,
```

strawberry/extensions/base_extension.py
patrick91: I think `LifecycleStep` should be ok, since it is namespaced inside the extensions module already 😊

RELEASE.md
```diff
@@ -0,0 +1,22 @@
+Release type: patch
+
+This release adds a method on the DatadogTracingExtension class called `create_span` that can be overridden to create a custom span or add additional tags to the span.
+
+```python
+from ddtrace import Span
+
+from strawberry.extensions.tracing import DatadogTracingExtension, LifeCycleStep
```
patrick91, line 8:
```suggestion
from strawberry.extensions import LifecycleStep
from strawberry.extensions.tracing import DatadogTracingExtension
```

RELEASE.md
```diff
@@ -0,0 +1,22 @@
...
+
+
+class DataDogExtension(DatadogTracingExtension):
+    def create_span(
+        self,
+        lifecycle_step: LifeCycleStep,
```
patrick91, line 14:
```suggestion
        lifecycle_step: LifecycleStep,
```

patrick91 (approved):
Looks great! Thank you!

patrick91 (approved):
Looks great! Thank you!

strawberry/extensions/tracing/datadog.py
```diff
@@ -36,6 +36,31 @@ def _resource_name(self):
 
         return query_hash
 
+    def create_span(
+        self,
+        lifecycle_step: LifecycleStep,
+        name: str,
+        **kwargs,
```
patrick91, line 43:
```suggestion
        **kwargs: Any,
```

tests/schema/extensions/test_datadog.py
```diff
@@ -252,3 +252,40 @@ def test_uses_operation_type_sync(datadog_extension_sync):
...
+    from strawberry.extensions.tracing.datadog import LifecycleStep
+
+    extension, mock = datadog_extension
+
+    class CustomExtension(extension):
+        def create_span(
+            self,
+            lifecycle_step: LifecycleStep,
+            name: str,
+            **kwargs,
```
patrick91, line 268:
```suggestion
            **kwargs,  # type: ignore
```

tests/schema/extensions/test_datadog.py
```diff
@@ -252,3 +252,40 @@ def test_uses_operation_type_sync(datadog_extension_sync):
...
+    from strawberry.extensions.tracing.datadog import LifecycleStep
+
+    extension, mock = datadog_extension
+
+    class CustomExtension(extension):
+        def create_span(
+            self,
+            lifecycle_step: LifecycleStep,
+            name: str,
+            **kwargs,  # type: ignore
```
patrick91, line 268:
```suggestion
            **kwargs,  # noqa: ANN003
```

botberry:

Issue #2097, opened 2022-08-14
Extenstions support for subscriptions.


nrbnlulu:
Provide extenstions support for subscriptions.

## Feature Request Type

- [ ] Core functionality
- [ ] Alteration (enhancement/optimization) of existing feature(s)
- [x] New behavior

## Description
currently with subscriptions won't hook to any of
 - `on_request_start`/`end`
 - `resolve`
 - `get_result`
 - `on_validation_start`/ `end`
 - `on_parsing_start` / `end`
 - `on_executing_start` / `end`
 
It would be nice if we can hook it there too
maybe `on_request_start`/`end` would hook to `connect` / `disconnect` of the application.
for the other I am not quite sure..

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/strawberry-graphql) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/strawberry-graphql/strawberry/issues/2097">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/2097/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/2097/pledge.svg">
</picture>
</a>

ArtemConstantinov:
In my project, I also need a symmetrical interface of extensions between ``schema.execute(...)`` and ``schema.subscribe(...)``. I come up with a not ideal solution that works for my project. There is definitely a need for more discussion about extensions 
interface for giving answers to questions as: 
- "Does it relevant to use on_request_... if we wanna use unified extensions for subscriptions and queries?"
- "Do we need to add a hook as `before_subscribe_result` and `after_subscribe_result`?"
- ...

**Due to an open issue [#1864](https://github.com/strawberry-graphql/strawberry/issues/1864) I will not make a pull request with my solution.** 

## Module structure

```
.../strawberry_patch/
    |- ./__init__.py
    |- ./subscribe.py
    |- ./schema.py
```

``` python
# subscribe.py

from typing import (
    AsyncGenerator, 
    Sequence, 
    Type, 
    Union,
)
from strawberry.extensions import Extension
from strawberry.schema.exceptions import InvalidOperationTypeError
from strawberry.schema.execute import _run_validation
[...]

PR #2784, opened 2023-05-27
Support Schema Extensions in subscriptions

kristjanvalur:
Subscriptions are now executed like queries and mutations, and extension hooks are called.
Result extension hooks are called for each result yielded by the subscription

__NOTE:__ #2825 is an updated version of this, with a less intrusive change to the `schema` API.

## Description

the `Schema.subscribe()` is turned into an `AsyncGenerator` yielding `ExecutionResult` objects.
In case the validation of the requests results in a single result instead of a subscription,  e.g. when the inner
`graphql-core.subscribe()` does not return an iterator, a special exception is thrown, `SubscribeSingleResult` which' `value`
attribute contains the single `ExecutionResult`.

This pattern, always being a `AsyncGenerator` which aborts under certain circumstances, makes for a cleaner interface
and allows for much simpler use of context managers.  Following the pattern of `graphql-core` which either returns
an iterator __or__ an object, makes it unnecessarily tricky to implement with the Extensions context managers.

The `payload` part of the returned messages in both protocols is augmented with an `extensions` member if added by the corresponding hooks.

Tests are added to tests for result extensions.

A side effect of this change is that initial validation now happens on a worker task.  This feature was a part of a separate PR in progress, so corresponding tests are also added to verify that functionality.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

* #2097

## Checklist

### DIFF
--- a/strawberry/schema/execute.py
+++ b/strawberry/schema/execute.py
@@ -29,6 +29,7 @@ from strawberry.exceptions import MissingQueryError
 from strawberry.extensions.runner import SchemaExtensionsRunner
 from strawberry.types import ExecutionResult
 
+from .base import SubscribeSingleResult
 from .exceptions import InvalidOperationTypeError
 
 if TYPE_CHECKING:
@@ -283,7 +284,7 @@ async def subscribe(
     extensions: Sequence[Union[Type[SchemaExtension], SchemaExtension]],
     execution_context: ExecutionContext,
     process_errors: Callable[[List[GraphQLError], Optional[ExecutionContext]], None],
-) -> AsyncGenerator[Tuple[bool, ExecutionResult], None]:
+) -> AsyncGenerator[ExecutionResult, None]:
     """
     The graphql-core subscribe function returns either an ExecutionResult or an
     AsyncGenerator[ExecutionResult, None].  The former is returned in case of an error
@@ -319,32 +320,34 @@ async def subscribe(
             except GraphQLError as error:
                 execution_context.errors = [error]
                 process_errors([error], execution_context)
-                yield False, ExecutionResult(
-                    data=None,
-                    errors=[error],
-                    extensions=await extensions_runner.get_extensions_results(),
+                raise SubscribeSingleResult(
+                    ExecutionResult(
+                        data=None,
+                        errors=[error],
+                        extensions=await extensions_runner.get_extensions_results(),
+                    )
                 )
-                # the generator is usually closed here, so the following is not
-                # reached
-                return  # pragma: no cover
 
             except Exception as error:  # pragma: no cover
                 error = GraphQLError(str(error), original_error=error)
                 execution_context.errors = [error]
                 process_errors([error], execution_context)
-                yield False, ExecutionResult(
-                    data=None,
-                    errors=[error],
-                    extensions=await extensions_runner.get_extensions_results(),
+
+                raise SubscribeSingleResult(
+                    ExecutionResult(
+                        data=None,
+                        errors=[error],
+                        extensions=await extensions_runner.get_extensions_results(),
+                    )
                 )
-                return  # pragma: no cover
 
         async with extensions_runner.validation():
             _run_validation(execution_context)
             if execution_context.errors:
                 process_errors(execution_context.errors, execution_context)
-                yield False, ExecutionResult(data=None, errors=execution_context.errors)
-                return  # pragma: no cover
+                raise SubscribeSingleResult(
+                    ExecutionResult(data=None, errors=execution_context.errors)
+                )
 
         async def process_result(result: GraphQLExecutionResult):
             execution_context.result = result
@@ -397,13 +400,13 @@ async def subscribe(
                 )
 
             if isinstance(result, GraphQLExecutionResult):
-                yield False, await process_result(result)
-                return
+                raise SubscribeSingleResult(await process_result(result))
 
             aiterator = result.__aiter__()
             try:
                 async for result in aiterator:
-                    yield True, await process_result(result)
+                    yield await process_result(result)
             finally:
+                # grapql-core's iterator may or may not have an aclose() method
                 if hasattr(aiterator, "aclose"):
                     await aiterator.aclose()
--- a/strawberry/schema/execute.py
+++ b/strawberry/schema/execute.py
@@ -4,6 +4,8 @@ from asyncio import ensure_future
 from inspect import isawaitable
 from typing import (
     TYPE_CHECKING,
+    AsyncGenerator,
+    AsyncIterable,
     Awaitable,
     Callable,
     Iterable,
@@ -17,8 +19,10 @@ from typing import (
     cast,
 )
 
+from graphql import ExecutionResult as GraphQLExecutionResult
 from graphql import GraphQLError, parse
 from graphql import execute as original_execute
+from graphql import subscribe as original_subscribe
 from graphql.validation import validate
 
 from strawberry.exceptions import MissingQueryError
@@ -31,7 +35,6 @@ if TYPE_CHECKING:
     from typing_extensions import NotRequired, Unpack
 
     from graphql import ExecutionContext as GraphQLExecutionContext
-    from graphql import ExecutionResult as GraphQLExecutionResult
     from graphql import GraphQLSchema
     from graphql.language import DocumentNode
     from graphql.validation import ASTValidationRule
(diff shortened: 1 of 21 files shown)


botberry:



---

Subscriptions now support Schema Extensions.


---




Here's the tweet text: 
```

```

kristjanvalur:
Current unit test flakiness is addressed by #2785

kristjanvalur:
Now, this is an __internally__ breaking change in that the results from `schema.subscribe()` have changed.  but is this a documented interface?

We __could__ change the interface but it would impact the possibilities of what kind of context it is possible to maintain during the evaluation of a subscription.

patrick91:
> Now, this is an **internally** breaking change in that the results from `schema.subscribe()` have changed. but is this a documented interface?
> 
> We **could** change the interface but it would impact the possibilities of what kind of context it is possible to maintain during the evaluation of a subscription.

I think breaking it is fine, as long as we write it down :)

kristjanvalur:
Right, so please advise, if you like the approach I'm suggesting in this PR, then should I add the breakage to the RELEASE.md, similar to what was done with the recent `channels` PR?

patrick91:
> Right, so please advise, if you like the approach I'm suggesting in this PR, then should I add the breakage to the RELEASE.md, similar to what was done with the recent `channels` PR?

we have a folder for breaking changes now: https://github.com/strawberry-graphql/strawberry/tree/main/docs/breaking-changes

I'll update the version number before merging 😊

strawberry/schema/execute.py
```diff
@@ -256,3 +259,135 @@ def execute_sync(
...
+        execution_context=execution_context,
+        extensions=list(extensions),
+    )
+
+    # unlike execute(), the entire operation, including the results hooks,
+    # is run within the operation() hook.
+    async with extensions_runner.operation():
+        # Note: In graphql-core the schema would be validated here but in
+        # Strawberry we are validating it at initialisation time instead
+        assert execution_context.query is not None
```
nrbnlulu, line 294: Should raise here MissingQueryError

strawberry/schema/execute.py
```diff
@@ -256,3 +259,135 @@ def execute_sync(
...
+    execution_context: ExecutionContext,
+    process_errors: Callable[[List[GraphQLError], Optional[ExecutionContext]], None],
+) -> AsyncGenerator[Tuple[bool, ExecutionResult], None]:
+    """
+    The graphql-core subscribe function returns either an ExecutionResult or an
+    AsyncGenerator[ExecutionResult, None].  The former is returned in case of an error
+    during parsing or validation.
+    Because we need to maintain execution context, we cannot return an
+    async generator, we must _be_ an async generator.  So we yield a
+    (bool, ExecutionResult) tuple, where the bool indicates whether the result is an
```
nrbnlulu, line 277: This tuple hack is kinda hard to understand... You might wanna check what I did at https://github.com/strawberry-graphql/strawberry/pull/2810/files#diff-88aa6fd17e4c6feac6e7152ebd3f2b8f972544c444a071b550e4d23061b97a3fR215 where if there is an error I return `ExecutionResultError`

strawberry/schema/execute.py
```diff
@@ -256,3 +259,135 @@ def execute_sync(
...
+    )
+
+    # unlike execute(), the entire operation, including the results hooks,
+    # is run within the operation() hook.
+    async with extensions_runner.operation():
+        # Note: In graphql-core the schema would be validated here but in
+        # Strawberry we are validating it at initialisation time instead
+        assert execution_context.query is not None
+
+        async with extensions_runner.parsing():
```
nrbnlulu, line 296: This is duplicated from async execution, probably can be reused... see https://github.com/nrbnlulu/strawberry/blob/support_extensions_on_subscriptions/strawberry/schema/execute.py#L192

strawberry/schema/execute.py
```diff
@@ -256,3 +259,135 @@ def execute_sync(
...
+                return  # pragma: no cover
+
+        async with extensions_runner.validation():
+            _run_validation(execution_context)
+            if execution_context.errors:
+                process_errors(execution_context.errors, execution_context)
+                yield False, ExecutionResult(data=None, errors=execution_context.errors)
+                return  # pragma: no cover
+
+        async def process_result(result: GraphQLExecutionResult):
```
nrbnlulu, line 333: Why is that defined here, on every execution?

strawberry/schema/execute.py
```diff
@@ -256,3 +259,135 @@ def execute_sync(
...
+                    ],
+                    result_or_awaitable,
+                )
+
+            if isinstance(result, GraphQLExecutionResult):
+                yield False, await process_result(result)
+                return
+
+            aiterator = result.__aiter__()
+            try:
```
nrbnlulu, line 388: why not `contextlib.suppress`?

nrbnlulu (changes requested):
Hey, I have worked on this PR long time ago though it got outdated... added some notes, if you want I can help.

strawberry/schema/execute.py
kristjanvalur: Should it?  I removed that code because it didn't get hit by coverage testing.  It is my understanding that that can only happen if a query parameter is missing from a "query string", and we don't have these for subscriptions.  Under what conditions could that possibly happen?

strawberry/schema/execute.py
kristjanvalur: That is a good idea.  Yes, tuples are problematic and thought this might be a sticking point.  I'll create a special exception class instead, much nicer.

strawberry/schema/execute.py
kristjanvalur: Its a closure.  Avoids writing a compicated external function and passing arguments, as well as scoping functionality.
Closures aren't actually _defined_ when run, merely instantiated with bindings.  I guess it is down to style, I very much favor closures for their conciseness and locality.  Happy to change it if you like.

strawberry/schema/execute.py
kristjanvalur: I'll see if I can do that after I change the tuple semantics.

strawberry/schema/execute.py
nrbnlulu: I think it's a weak point for performance, I would be happy if we would merge #2810 here or vice-versa. There I added a base class for async execution and it is done for subscription and async execution via   https://github.com/nrbnlulu/strawberry/blob/support_extensions_on_subscriptions/strawberry/schema/execute.py#LL225C15-L225C39

strawberry/schema/execute.py
nrbnlulu: hmm I'm not sure I think @jthorniley added this.

strawberry/schema/execute.py
kristjanvalur: Well, either an exception, or a special result class..  I think the exception might be cleaner, since one _expects_ a subscription and the failure to get one is an exception of sorts.  I'll see which one is nicer.

strawberry/schema/execute.py
kristjanvalur: I wasn't aware of #2810, I'll look at that.

I just want to add, as a general comment, that I think you will find that performance is not affected.
First of all, instantiate a closure a drop in the ocean of all the stuff that goes on to actually start executing the subscription.
Secondly, all of the execution time of a subscription actually happens parsing and processing the multiple results.  Setting up of an subscription is not performance critical.

Finally, you suggested using `contextlib.suppress`, and that involves not only instantiating a class on every query, but also invoking two methods on it.  There is quite a bit of machinery involved in a a `contextlib.contextmanager`.

I understand when people are concerned about performance, but one always needs to look at that in context.

strawberry/schema/execute.py
nrbnlulu: I don't understand.
> Because that exception might come from anywhere

What's wrong with
  
```py
with contextlib.supress(BaseException):
    async for result in aiterator:
        yield True, await process_result(result)
finally:
if hasattr(aiterator, "aclose"):
    await aiterator.aclose()
```

strawberry/schema/execute.py
nrbnlulu: > I understand when people are concerned about performance, but one always needs to look at that in context.

Thanks. ig it is just something that bugs me... unless it is in a decorator I won't do it...
Sorry if i'm being too strict.

strawberry/schema/execute.py
kristjanvalur: No, you don't want to suppress BaseException.  You never want to do that.  CancelledError is one reason.
No, try-finally exists precisely do do this kind of thing.  In fact, that is how `contextlib.aclosing()` is implemented and it would be appropriate there, except that a) it requires python 3.8 and b), it _assumes_ that `aclose()` is present.

strawberry/schema/execute.py
kristjanvalur: Sure, its moved out of the function in the latest commit.

strawberry/schema/execute.py
kristjanvalur: I can't see how this code can be reused.  Subscription in my case is an _async generator_ and necessarily _yields_ its result.
Unless, you are willing to make everything use async generator context and create a special single-instance yielder when performing `exeucte()`

kristjanvalur: I did refactor the common validation tests

kristjanvalur:
Probably want to add ruff as dev-dependency and pin it to a version

tests/views/schema.py
```diff
@@ -20,10 +21,121 @@ def has_permission(self, source: Any, info: Info[Any, Any], **kwargs: Any) -> bo
...
+
+    def on_execute(self):
+        self.lifecycle_called("execute", "before")
+        self.active_counter += 1
+        yield
+        self.lifecycle_called("execute", "after")
+        self.active_counter -= 1
+
+
+class MyAsyncExtension(SchemaExtension):
```
nrbnlulu, line 87: Can this use https://github.com/strawberry-graphql/strawberry/blob/b7982878f1b4ec73d74af4a86ab85fdeb921763b/tests/schema/extensions/test_extensions.py#LL194C7-L194C23 ?

tests/views/schema.py
```diff
@@ -20,10 +21,121 @@ def has_permission(self, source: Any, info: Info[Any, Any], **kwargs: Any) -> bo
...
+
+    async def has_permission(self, source, info, **kwargs: Any) -> bool:
+        self.sleep = kwargs.get("sleep", None)
+        self.fail = kwargs.get("fail", True)
+        if self.sleep is not None:
+            await asyncio.sleep(kwargs["sleep"])
+        return not self.fail
+
+
 class MyExtension(SchemaExtension):
```
nrbnlulu, line 37: Can this use https://github.com/strawberry-graphql/strawberry/blob/b7982878f1b4ec73d74af4a86ab85fdeb921763b/tests/schema/extensions/test_extensions.py#LL194C7-L194C23 ?

nrbnlulu: > It would be preferable to do it the other way round
sure

kristjanvalur:
By the way, I offer a different kind of resolving the "either return an ExecutionResult or an Iterator" thing, to more
closely match what `graphql-core` does.  Basically, it involves an outer helper around an inner generator.
The code looks like this:

```python
import asyncio
import contextlib
import inspect
from typing import AsyncGenerator, Optional, Union, cast

incontext = False


@contextlib.asynccontextmanager
async def mycontext() -> AsyncGenerator[None, None]:
    global incontext
    incontext = True
    try:
        yield
    finally:
        incontext = False


async def inner_gen(dofail: bool) -> AsyncGenerator[Optional[str], None]:
    async with mycontext():

        # yield an initial value, which will be the single result value,
        # or None if we're going to yield more values
        if dofail:
            yield "failure"
            return  # typically not reached
        else:
            yield None

        # yield more values
        yield "success"
        yield "success2"


async def outer(dofail: bool) -> Union[str, AsyncGenerator[str, None]]:
    # check the first value, and if it's not None, return it,
    # aborting the generator
    gen = inner_gen(dofail)
    first = await gen.__anext__()
    if first is not None:
        await gen.aclose()
        return first
    else:
        return cast(AsyncGenerator[str, None], gen)


async def main() -> None:
    assert not incontext
    v = await outer(True)
    assert not incontext
    assert v == "failure"

    v = await outer(False)
    assert inspect.isasyncgen(v)
    assert incontext
    async for val in v:
        assert "success" in val
        assert incontext
    assert not incontext


if __name__ == "__main__":
    asyncio.run(main())

```

Using this pattern, we could avoid the `SubscribeSingleResult` exception, but we would then, again, have to check the
type of `schema.subscribe()` result to determine if it is a `ExecutionResult` or an iterator.

Both work.  Which one is peferred by the maintainers?  I'm happy to convert.

nrbnlulu (approved):

kristjanvalur:
Important, before mergint this, please consider #2825, which is a less intrusive change

patrick91:
@kristjanvalur can you rebase this? I'll make a pre-release so @Speedy1991 can test it :D

kristjanvalur:
> @kristjanvalur can you rebase this? I'll make a pre-release so @Speedy1991 can test it :D

Sure, although you may want to look at #2825 instead, it maintains essentially the same API as before (returning either a `GraphQLResult` or an `iterator`.   I think it has a smaller "footprint.".  I'll rebase both.

There is also still some flakiness in one of the unittests (testing the order of the hooks bein called), it appears that despite my efforts, some straggling tests may be interfering with it...  I'll figure out a way to fix it.

botberry:
# Pre-release
:wave:

Releasing commit [abfd60e6968a9cdc5632a78536eefc3acf67a070] to PyPi as pre-release! :package:

botberry:
# Pre-release
:wave:

Pre-release **0.196.0.dev.1689676990** [abfd60e6968a9cdc5632a78536eefc3acf67a070] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.196.0.dev.1689676990
```

patrick91:
from @Speedy1991 on discord:

> btw. from a programmer perspective the https://github.com/strawberry-graphql/strawberry/pull/2784 is much cleaner to implement in a ws view 🙂

kristjanvalur:
> from @Speedy1991 on discord:
> 
> > btw. from a programmer perspective the #2784 is much cleaner to implement in a ws view 🙂

Interesting.
So, _this_ deviates a bit from the api which graphql-core has, which is trying hard to stay consistent with the original js implementation, returning either a "graphql-result" or an iterator.  That approach is a bit un-pythonic.
Strawberry-graphql generally sticks to an identical api in `schema.py`

This pr changes that approach to raise an `Exception` in the exceptional case.  I originally did this because it was easier for me to  support the context managers for the extensions.  But it resulted in a semantic change in `schema.subscribe`, more than just the returned types changing from graphql-core ExecutionResult to strawberry ExecutionResult.

I later figured out how to keep the old api, while only changing the type.  That was #2825.  Code which calls `schema.subscribe` now does not need to be re-structured, only expect different types.

Everything else is the same difference between these two prs.  I guess it comes down to which api style is considered cleaner.

patrick91:
@kristjanvalur sorry for my lazyness, could you show me the two different APIs here? (or even another issue)

kristjanvalur:
here we go.  2784 style:

```python
try:
    async for result in self.schema.subscribe(...):
        await self.send_result(operation, result, False)
except SubscribeSingleResult as single_result:
    await self.send_result(operation, single_result.value, True)
finally:
        await result_source.aclose()
```

2825  style (and the current style in code, inherited from graphql-core)

```python
result_source = self.schema.subscribe(...)
if isinstance(result_source, ExecutionResult):
    await self.send_result(operation, result_source, True)
    return
try:
    async for result in result_source:
        await self.send_result(operation, result, False)
finally:
        await result_source.aclose()
```

Basically, switching on the result value type, or treating a single result (usually validation error) as an "exceptional" case.

XChikuX:
@patrick91 Can we merge this?

kristjanvalur:
let me see if I can bring this up to date

kristjanvalur:
Still some minor fixes needed.
Please consider also if you think pr #2825 is a better api, see discussion above.

kristjanvalur:
There have recently been made some changes to extension error handling.
It is unclear to me how to translate that to this PR.  
Note that errors during evaluation of the request need to be raised as SingleValueException with an ExecutionResult.
After that, ExecutionResults are not fatal.  
But unexpected exceptions, such as should indeed be fatal during iteration.

patrick91:
@kristjanvalur thank you so much for keeping this PR up to date, @nrbnlulu made a new PR recently and we merged that, hope that's ok!

patrick91:
@kristjanvalur probably, there's been a lot of refactoring recently, I'll check with @DoctorJohn soon :)

PR #2802, opened 2023-06-01
Allow the usage of fragment directives in codegen.

### LINKED ISSUE #2801: Codegen does not support non-inline `fragment`
## Feature Request Type

- [ ] Core functionality
- [x] Alteration (**enhancement**/optimization) of existing feature(s)
- [ ] New behavior

## Description

Fragments in queries would be pretty difficult to support in general, however, I think that a usecase like:

```
fragment Point on Bar {
   id
   x
   y
}

query GetPoints {
  circlePoints {
    ...Point
  }
  squarePoints {
    ...Point
  }
}
```

Should be reasonable to support.  In this case, I'd expect to get some sort of type classes like:

```py
class Point:
    id: str
    x: float
    y: float

class GetPointsResult:
    circle_points: List[Point]
    square_points: List[Point]
```

Because it's obvious from the spread operator that `circlePoints` and `squarePoints` are the same type.

I don't think we should have to support any additional fields.  e.g. I think it would be Ok if the following would raise some kind of (hopefully helpful) error:

```
fragment Point on Bar {
   id
   x
   y
}

query GetPoints {
  circlePoints {
    ...Point
    z
  }
  squarePoints {
    ...Point
    z
  }
}
```

It would also be really nice if custom plugins would be able to tell whether a `GraphQLObjectType` comes from a fragment or not (I think it will have to be able to make that distinction in order for the print plugin to function properly anyway)

In my envisioned use-case, I plan on generating the queries via templates with a consistent set of fragments that I can inline -- then my client-generator can stick the fragments in separate files (based on the names) and my objects will be de-duplicated when my full client is finished being generated.

mgilson:
## Description

This PR enables the use of non-inline `fragment` in client generation.

The limitations/behaviors so far are that:

* fragments can only be included by themselves (they cannot be used with other fields)
* fragments are emitted as new types in the generated types files and they replace the types that would have been generated in that position otherwise.

See the associated issue for more details.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* fixes #2801

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,39 @@
+Release type: patch
+
+In this release codegen no longer chokes on queries that use a fragment.
+
+There is one significant limitation at the present.  When a fragment is included via the spread operator in an object, it must be the only field present.  Attempts to include more fields will result in a ``ValueError``.
+
+However, there are some real benefits.  When a fragment is included in multiple places in the query, only a single class will be made to represent that fragment:
+
+```
+fragment Point on Bar {
+   id
+   x
+   y
+}
+
+query GetPoints {
+  circlePoints {
+    ...Point
+  }
+  squarePoints {
+    ...Point
+  }
+}
+```
+
+Might generate the following types
+
+```py
+class Point:
+    id: str
+    x: float
+    y: float
+
+class GetPointsResult:
+    circle_points: List[Point]
+    square_points: List[Point]
+```
+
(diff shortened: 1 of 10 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

patrick91 (approved):

PR #2805, opened 2023-06-02
Fix fragment codegen.

mgilson:
## Description

Need to find the fields in the correct type (which is the type that the fragment is referencing.

Sadly the tests didn't pick this up because the fragment in the test is on a type called "Query".  Oops.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/strawberry/codegen/query_codegen.py
+++ b/strawberry/codegen/query_codegen.py
@@ -199,7 +199,11 @@ class QueryCodegen:
             if isinstance(definition, FragmentDefinitionNode)
         )
         for fd in fragment_definitions:
-            query_type = self.schema.get_type_by_name("Query")
+            query_type = self.schema.get_type_by_name(fd.type_condition.name.value)
+            if query_type is None:
+                raise ValueError(
+                    f"{fd.type_condition.name.value!r} is not a type in the graphql schema!"
+                )
             assert isinstance(query_type, TypeDefinition)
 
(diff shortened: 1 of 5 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

strawberry/codegen/query_codegen.py
```diff
@@ -198,14 +198,13 @@ def _populate_fragment_types(self, ast: DocumentNode) -> None:
             if isinstance(definition, FragmentDefinitionNode)
         )
         for fd in fragment_definitions:
-            query_type = self.schema.get_type_by_name("Query")
-            assert isinstance(query_type, TypeDefinition)
-
-            def graph_ql_object_type_factory(name: str, on: str):
-                return GraphQLFragmentType(name, on=on)
+            query_type = self.schema.get_type_by_name(fd.type_condition.name.value)
```
patrick91, line 201: I should have noticed this, sorry!

patrick91 (approved):
Do you think we should add another test for this case?

mgilson:
> Do you think we should add another test for this case?

It's easy enough to exercise this case with the test resources we have if we just change around what the query is returning.  I've updated it.

patrick91 (approved):
Thanks!

PR #2819, opened 2023-06-07
Fix codegen missing fields

### LINKED ISSUE #2818: Codegen drops all but the last field when working with inline fragments.
## Describe the Bug

I noticed that not all fields are included in the generated types when the query includes an inline fragment.  e.g.

```
query OperationName {
  getPersonOrAnimal {
    ... on Person {
      name
      age
    }
  }
}
```

assuming we add the following to the appropriate conftest for query-generation

```py
    @strawberry.field
    def get_person_or_animal(self) -> Union[Person, Animal]:
        """Randomly get a person or an animal."""
        p_or_a = random.choice([Person, Animal])()
        p_or_a.name = "Howard"
        p_or_a.age = 7
        return p_or_a
```

would only generate an object with an `age` property -- no `name`.  It took me a little while to come up with a MCVE that tripped this up, but the above should deterministically generate objects with missing fields.

## System Information

 - Operating system:
 - Strawberry version (if applicable):

mgilson:
## Description

When we want to add a field to an inline fragment `GraphQLObjectType` (e.g. `current_type`), the first thing we do is set the `current_type.fields` to a list that contains the common-types between _this_ inline fragment and any sibling inline fragments.  This has the effect of stomping on any fields which have already been added and only the last field makes it into the result `GraphQLObjectType`.

Instead, we should just add the common fields in the beginning before adding any additional fields and then we don't ever need to do it again.

It's still important that we make a copy of the list in the `GraphQLObjectType` construction to make sure that we aren't sharing that list between multiple `GraphQLObjectType`.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* Fix #2818

## Checklist

### DIFF
--- a/strawberry/codegen/query_codegen.py
+++ b/strawberry/codegen/query_codegen.py
@@ -504,7 +504,9 @@ class QueryCodegen:
         selected_field = self.schema.get_field_for_type(
             selection.name.value, parent_type.name
         )
-        assert selected_field
+        assert (
+            selected_field
+        ), f"Couldn't find {parent_type.name}.{selection.name.value}"
 
         selected_field_type, wrapper = self._unwrap_type(selected_field.type)
         name = capitalize_first(to_camel_case(selection.name.value))
--- a/strawberry/codegen/query_codegen.py
+++ b/strawberry/codegen/query_codegen.py
@@ -640,13 +640,12 @@ class QueryCodegen:
 
         for fragment in fragments:
             fragment_class_name = class_name + fragment.type_condition.name.value
-            current_type = GraphQLObjectType(fragment_class_name, [])
+
+            current_type = GraphQLObjectType(fragment_class_name, list(common_fields))
 
             for sub_selection in fragment.selection_set.selections:
                 # TODO: recurse, use existing method ?
                 assert isinstance(sub_selection, FieldNode)
-
-                current_type.fields = list(common_fields)
 
                 parent_type = cast(
                     TypeDefinition,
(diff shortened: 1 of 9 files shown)


strawberry/codegen/query_codegen.py
```diff
@@ -504,7 +504,9 @@ def _field_from_selection_set(
         selected_field = self.schema.get_field_for_type(
             selection.name.value, parent_type.name
         )
-        assert selected_field
+        assert (
+            selected_field
+        ), f"Couldn't find {parent_type.name}.{selection.name.value}"
```
mgilson, line 509: I hit this assertion a bunch when testing and I make this change every single time and revert before pushing it up.  This error actually gets exposed to the user when they misspell a field name in the GraphQL query, so it makes sense to me to at least give it a more helpful error message so that the user knows something about what the codegen was looking for.

It would probably be better to _not_ use `assert` for this at some point and raise a nicer exception overall, but that felt more invasive than I wanted for this particular PR.

patrick91: let's keep the assert, and improve the message in another PR 😊

patrick91 (approved):
Looks good! It's only missing the release notes 😊

mgilson:
Oops, forgot to git add/push it  🤕 .  Sorry about that.

patrick91:
not sure what's up with codecov, I'll try again in 30 minutes 😊

PR #2824, opened 2023-06-08
Fix duplicated type when using generics and lazy types

patrick91:
This PR changes how `get_specialized_type_var_map` to only work on strawberry types, this prevents us thinking that types extending specialized generics that aren't types should be copied :)

It also removes support for `include_type_vars` as this was used in only one place

### DIFF
--- a/strawberry/types/types.py
+++ b/strawberry/types/types.py
@@ -16,7 +16,6 @@ from typing import (
 from typing_extensions import Self
 
 from strawberry.type import StrawberryType, StrawberryTypeVar
-from strawberry.utils.inspect import get_specialized_type_var_map
 from strawberry.utils.typing import is_generic as is_type_generic
 
 if TYPE_CHECKING:
@@ -116,12 +115,7 @@ class TypeDefinition(StrawberryType):
 
     @property
     def is_specialized_generic(self) -> bool:
-        if not self.is_generic:
-            return False
-
-        type_var_map = get_specialized_type_var_map(self.origin)
-
-        return not type_var_map and not getattr(self.origin, "__parameters__", None)
+        return self.is_generic and not getattr(self.origin, "__parameters__", None)
 
     @property
     def type_params(self) -> List[TypeVar]:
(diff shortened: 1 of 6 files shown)


strawberry/utils/inspect.py
```diff
@@ -104,6 +114,14 @@ def get_specialized_type_var_map(cls: type, *, include_type_vars: bool = False):
 
     type_var_map = {}
 
+    # only get type vars for base generics (ie. Generic[T]) and for strawberry types
+
+    orig_bases = [
+        b
+        for b in orig_bases
+        if hasattr(b, "_type_definition") or type(b) is _GenericAlias
+    ]
```
patrick91, line 123: @bellini666 does this make sense?

this ignores all the generics on non strawberry types

bellini666: If I understood this correctly, this is to fix that issue reported on discord, right?

I think this makes sense as non strawberry generics in the inheritance shouldn't create different types.

bellini666 (approved):

strawberry/types/types.py
```diff
@@ -119,10 +119,9 @@ def is_specialized_generic(self) -> bool:
         if not self.is_generic:
             return False
 
-        type_var_map = get_specialized_type_var_map(self.origin, include_type_vars=True)
-        return type_var_map is None or not any(
-            isinstance(arg, TypeVar) for arg in type_var_map.values()
-        )
+        type_var_map = get_specialized_type_var_map(self.origin)
+
+        return not type_var_map and not getattr(self.origin, "__parameters__", None)
```
patrick91, line 124: @bellini666 can you double check this? 😊

bellini666: Yeah, this seems correct.

Also, I tested this branch locally with this change:

```python
@property
def is_specialized_generic(self) -> bool:
    return self.is_generic and not getattr(self.origin, "__parameters__", None)
```

And all tests worked fine! So probably we don't actually need to check for `type_var_map` in here

bellini666 (approved):
There's one suggestion to make `is_specialized_generic` even simpler. Other than that everything else is fine :)

strawberry/types/types.py
patrick91: I'll update thanks!

do you also want to test a pre-release on your projects or shall we just ship the release? :D

PR #2825, opened 2023-06-08
Support schema extensions for subsctriptions - Alternative

kristjanvalur:
Subscriptions are now executed like queries and mutations, and extension hooks are called.
Result extension hooks are called for each result yielded by the subscription

Note, This is a separate PR from #2784, because it contains an alternative, less intrusive, way to do this.
It avoids changing the signature of `schema.subscribe` which now, like `graphql-core`, continues to return __either__
an `ExcecutionResult` or an `AsyncGenerator`.

## Description

the `Schema.subscribe()`, when _awaited_ returns a `Union[ExeuctionResult, AsyncGenerator[ExecutionResult,None]]`.
This is similar to before, but previously it would return a raw `GraphQLExecutionResult` object, or an `AsyncIterator` over said type, directly from `graphql-core`.

Internally, an `AsyncGenerator` is created which yields an extra _initial result_, which is either a `ExecutionResult` in case
the inner subscribe method failed and no iterator is forthcoming, or a `None` if subscription is about to proceed.  An
outer method examines this initially yielded value and either returns that to the caller, or the already started `AsyncGenerator`, which the caller then continues to iterate from, maintaining all context which that iterator had already established.

Tests are added to tests for result extensions.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

* #2097

## Checklist

- [x] I have tested the changes and verified that they work and don't break anything (as we

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: minor
+
+Subscriptions now support Schema Extensions.
--- a/docs/guides/custom-extensions.md
+++ b/docs/guides/custom-extensions.md
@@ -36,8 +36,6 @@ resolvers.
 If you need to wrap only certain field resolvers with additional logic, please
 check out [field extensions](field-extensions.md).
 
-Note that `resolve` can also be implemented asynchronously.
-
 ```python
 from strawberry.types import Info
 from strawberry.extensions import SchemaExtension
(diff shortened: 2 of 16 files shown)


botberry:



---

Subscriptions now support Schema Extensions.


---




Here's the tweet text: 
```

```

botberry:
# Pre-release
:wave:

Releasing commit [cb3949a7b39ac1d4d5362bef01349863c9ae3d2e] to PyPi as pre-release! :package:

botberry:
# Pre-release
:wave:

Pre-release **0.196.0.dev.1689676980** [cb3949a7b39ac1d4d5362bef01349863c9ae3d2e] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.196.0.dev.1689676980
```

kristjanvalur:
Performance regression is expected when all of the schema extension boilerplate code is added on top of a trivial method.
I suspect that if we add a performance test which tests iterating over 1000 entries of a subscription, that will perform much more favourably

Issue #1754, opened 2022-03-21
get_context() is only resolved once per WS connection for FastAPI


kristjanvalur:
Unlike for http operations, the get_context() handler is set up as a FastAPI _dependency_ for the websocket endpoint on `GraphQLRouter` and resolved when the websocket connection is initialized. 
So, even though the protocol handler's `get_context()` member is evaluated for each subscription, or any other operation, it will ultimately return this single cached value for the duration of the websocket connection.

This may be by design, but it is a bit unexpected and is not mentioned in the documentation.  It also means that the dependencies cannot be used via the `yield` pattern to acquire and release resources for the operation.

tsmith023:
This is an interesting side-effect of the current implementation!

In my understanding, the problem arises due to https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/fastapi/router.py#L216 where, as you say, any `yield` functionality in the `self.context_getter` is resolved once by `Depends()` and thus never again subsequently.

Would the solution then be to simply remove the `Depends()` here? Would doing this have any unintended consequences? @Kludex, do you have thoughts regarding this issue?

Kludex:
Can you show me a code snippet on which this is a problem for `strawberry`?

kristjanvalur:
Well, I noticied this when many months ago when I was looking at authentication.  Obviously, authentication needs happen only once per connection.  But authorization might change.
Anyway, here is a trivial example:

```python
def custom_context_dependency() -> float:
    return time.time()


async def get_context(
    custom_value=Depends(custom_context_dependency),
):
    return {
        "custom_value": custom_value,
    }


@strawberry.type
class Query:
    @strawberry.field
    def example(self, info: Info) -> str:
        return f"Hello at {info.context['custom_value']}"
```

A more concrete example might be this.  Note that I didn't test the `yield` functionality, so I am making a guess, but my guess is that the transaction will be kept open for the entire connection, which might have unintended consequences:

```python
def custom_context_dependency():
    with db.get_session() as sesssion:
        yield check_authorization(session)
```

Depending on the type of connection made, these queries will work differently.

Kludex:
I see what you mean, but the way I see it, this is the intended behavior.

A note on the documentation with a meaningful example clarifying this may be helpful.

kristjanvalur:
In other words, dependencies **cannot be expected** to be evaluated for every request.
[...]

PR #2829, opened 2023-06-08
Do not block `graphql_transport_ws` operations while creating context or validating a single request operation

kristjanvalur:
Now both `Context` creation and validation of an operation (subscripton or single-result operation) are performed
on the worker task, freeing up the Websocket for other requests.

## Description

Previously both creating the context, and validating the request, were done on the same task which handles
incoming messages for the Websocket.  Both of these operations are async and can potentially take a long time.
Now they are performed on the worker background `Task` which is created to handle the operation.

We add tests to make sure that long operations in context and validation don't block the websocket connection.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
+++ b/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
@@ -14,9 +14,9 @@ from typing import (
     Dict,
     List,
     Optional,
+    Union,
 )
 
-from graphql import ExecutionResult as GraphQLExecutionResult
 from graphql import GraphQLError, GraphQLSyntaxError, parse
 
 from strawberry.subscriptions.protocols.graphql_transport_ws.types import (
@@ -246,7 +246,13 @@ class BaseGraphQLTransportWSHandler(ABC):
         root_value = await self.get_root_value()
 
         # Get an AsyncGenerator yielding the results
-        async def start_operation() -> AsyncGenerator[ExecutionResult, None]:
+        async def start_operation() -> Union[AsyncGenerator[Any, None], Any]:
+            # there is some type mismatch here which we need to gloss over with
+            # the use of Any.
+            # subscribe() returns
+            # Union[AsyncIterator[graphql.ExecutionResult], graphql.ExecutionResult]:
+            # whereas execute() returns strawberry.types.ExecutionResult.
+            # These execution result types are similar, but not the same.
             if operation_type == OperationType.SUBSCRIPTION:
                 return await self.schema.subscribe(
                     query=message.payload.query,
@@ -257,7 +263,7 @@ class BaseGraphQLTransportWSHandler(ABC):
                 )
             else:
                 # single results behave similarly to subscriptions,
-                # return either a GraphQLExecutionResult or an AsyncGenerator
+                # return either a ExecutionResult or an AsyncGenerator
                 result = await self.schema.execute(
                     query=message.payload.query,
                     variable_values=message.payload.variables,
@@ -265,7 +271,8 @@ class BaseGraphQLTransportWSHandler(ABC):
                     root_value=root_value,
                     operation_name=message.payload.operationName,
                 )
-                if isinstance(result, GraphQLExecutionResult):
+                # Both validation and execution errors are handled the same way.
+                if result.errors:
                     return result
 
                 # create AsyncGenerator returning a single result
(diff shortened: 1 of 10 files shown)


botberry:



---

Operations over `graphql-transport-ws` now create the Context and perform validation on
the worker `Task`, thus not blocking the websocket from accepting messages.


---




Here's the tweet text: 
```

```

kristjanvalur:
No rush, just keeping things tidy :)

kristjanvalur:
PR is updated but I cannot see from the codecov that there is any missing coverage for my changes.

kristjanvalur:
This PR needs to be completely refactored because of the upstream changes.
The "context" and "root value" are now created in a different place, long before the task is created.  These really need to be evaluated in the context of the Task to not block the socket.  I will attempt to re-factor this again.

DoctorJohn:
> This PR needs to be completely refactored because of the upstream changes. The "context" and "root value" are now created in a different place, long before the task is created. These really need to be evaluated in the context of the Task to not block the socket. I will attempt to re-factor this again.

Hi Kristján, thanks for bringing this up again!

During a brief discussion of this PR at PyconIT earlier this year, I realized that the strawberry core devs I talked to generally perceive `get_context` and `get_root_value` as lightweight methods. After recent refactors, both methods are only being called once per WebSocket connection.

Based on the assumption that these operations are lightweight, we were less concerned about `get_context`/`get_root_value` than the request validation mentioned in this PR's description. Feel free to split the request validation optimization in a separate PR so that we do not delay the merging of either part of this PR by pondering the other.

kristjanvalur:
Sounds good.   I'll leave the context/root as they are and rework this to just do the validation.

kristjanvalur:
It appears that validation now happens in the task ,because the `result_source` is awaited inside the Task now.  So presumably, validation now happens on the thread.  Since context is now created once for each connection, this leaves nothing left for this PR to do.  All it could contribute is perhaps tests that validation does not block other messages for the connection.   I'm inclined to simply close this.

kristjanvalur:
Actually, my tests as written for the old PR do indicate a discrepancy in validation handling for queries vs subscriptions.
I'll create a new PR with just the tests, showing the problem, without proposing a solution.

kristjanvalur:
By the way, speaking of "context" being created once for each WS connection, I want to draw your attention to this bug: #1754

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -107,7 +108,7 @@ def on_request_accepted(self) -> None:
 
     async def handle_connection_init_timeout(self) -> None:
         task = asyncio.current_task()
-        assert task
+        assert task is not None  # for typecheckers
```
bellini666, line 111: nit: since we enforce typing a lot here, `assert <something>` is a well known practice for type checkers, so I don't think the comment is required

```suggestion
        assert task is not None
```

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -333,6 +342,11 @@ async def handle_async_results(
                         ]
                     next_message = NextMessage(id=operation.id, payload=next_payload)
                     await operation.send_message(next_message)
+                await operation.send_message(CompleteMessage(id=operation.id))
+            finally:
+                # Close the AsyncGenerator in case of errors or cancellation
+                await result_source.aclose()
```
bellini666, line 348: Should this be wrapped in a `with suppress(asyncio.CancelledError)`?

bellini666 (commented):
My $.02 cents in here, leaving for @DoctorJohn to actually approve this once it is ready

DoctorJohn:
> It appears that validation now happens in the task ,because the `result_source` is awaited inside the Task now. So presumably, validation now happens as part of that, in a separate Task. Since context is now created once for each connection, this leaves nothing left for this PR to do. All it could contribute is perhaps tests that validation does not block other messages for the connection. **I'm inclined to simply close this.**

Thanks for verifying! It looks like this was changed when extension support for subscriptions was added. Let's close this PR then.

> Actually, my tests as written for the old PR do indicate a discrepancy in validation handling for queries vs subscriptions.
I'll create a new PR with just the tests, showing the problem, without proposing a solution.
see https://github.com/strawberry-graphql/strawberry/pull/3671. Please have a look at this and give me your opinion.

Thank you so much for highlighting that. I'll check the specs and the PR introducing this discrepancy again. It looks like something slipped through the review and had no proper tests.

> By the way, speaking of "context" being created once for each WS connection, I want to draw your attention to this bug: https://github.com/strawberry-graphql/strawberry/issues/1754

:+1: Currently reading through it.

PR #2836, opened 2023-06-11
rename `_type_definition` to `__strawberry_definition__`

nrbnlulu:
## Description
Rename `_type_definition` to `__strawberry_definition__`.

I understand that @BryceBeagle intended that `StrawberryObject` would be merged with `TypeDefinition` somehow in the future.
Currently I exposed `WithStrawberryDefinition` as a protocol that only includes `__strawberry_definition__` in it. If in the future it would be a real type the annotations are already in place.

<s>I am not very sure about the name `__strawberry_object__` since it can be `InputObject` / `Interface` / `ObjectType`
and enums use `_enum_definition`, open to suggestions...</s> As [discussed on discord ](https://discord.com/channels/689806334337482765/750088532479180840/1117689267649515530) every strawberry decorated type would have this dunder. Follow #2841 for more details.
 
TODO:
- [x] Add alias for `_type_definition` backward compat with warning. 

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

strawberry/annotation.py
```diff
@@ -57,7 +57,10 @@ def __init__(
 
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, StrawberryAnnotation):
-            return NotImplemented
+            raise NotImplementedError(
+                f"Comparing {StrawberryAnnotation.__name__} "
+                f"with {type(other)} is ambiguous."
```
patrick91, line 62:
```suggestion
                f"with {type(other)} is not supported."
```

strawberry/annotation.py
```diff
@@ -124,11 +127,8 @@ def set_namespace_from_field(self, field: StrawberryField) -> None:
         self.namespace = module.__dict__
 
     def create_concrete_type(self, evaled_type: type) -> type:
-        if _is_object_type(evaled_type):
-            type_definition: TypeDefinition
-            type_definition = evaled_type._type_definition  # type: ignore
-            return type_definition.resolve_generic(evaled_type)
-
+        if has_strawberry_object(evaled_type):
```
patrick91, line 130: I'm not 100% I like this name, `has_strawberry_object` somewhat implies that anything could have a StrawberryObject 😊

I think `is_strawberry_object` might be better 😊

/cc @BryceBeagle

strawberry/arguments.py
```diff
@@ -171,8 +171,10 @@ def convert_argument(
         enum_definition: EnumDefinition = type_._enum_definition
         return convert_argument(value, enum_definition, scalar_registry, config)
 
-    if hasattr(type_, "_type_definition"):  # TODO: Replace with StrawberryInputObject
-        type_definition: TypeDefinition = type_._type_definition
+    if hasattr(
+        type_, "__strawberry_object__"
+    ):  # TODO: Replace with StrawberryInputObject
+        type_definition: StrawberryObject = type_.__strawberry_object__
```
patrick91, line 177: should we reuse the function from above?

strawberry/object_type.py
```diff
@@ -151,6 +148,7 @@ def _process_type(
         _fields=fields,
         is_type_of=is_type_of,
     )
+    cls._type_definition = cls.__strawberry_object__
```
patrick91, line 151: could we make this a deprecated property? 😊

strawberry/schema/compat.py
```diff
@@ -40,7 +41,7 @@ def is_scalar(
 
 
 def is_object_type(type_: Union[StrawberryType, type]) -> TypeGuard[type]:
-    return hasattr(type_, "_type_definition")
+    return has_strawberry_object(type_)
```
patrick91, line 44: should we deprecate this?

strawberry/tools/merge_types.py
```diff
@@ -23,9 +23,7 @@ class SuperQuery(B, A):
     if not types:
         raise ValueError("Can't merge types if none are supplied")
 
-    fields = chain(
-        *(t._type_definition.fields for t in types)  # type: ignore[attr-defined]
-    )
+    fields = chain(*(t.__strawberry_object__.fields for t in types))  # type: ignore
```
patrick91, line 26: can we still keep the error type on the type ignore?

strawberry/type.py
```diff
@@ -84,8 +105,8 @@ def copy_with(
         of_type_copy: Union[StrawberryType, type] = self.of_type
 
         # TODO: Obsolete with StrawberryObject
```
patrick91, line 107: I think we can remove this todo now 😊

strawberry/types/types.py
```diff
@@ -167,3 +174,33 @@ def is_implemented_by(self, root: Union[type, dict]) -> bool:
...
+
+class WithTypeDefinition(Protocol):
+    __strawberry_object__: StrawberryObject
+
+
+def has_strawberry_object(klass: Any) -> TypeGuard[Type[WithTypeDefinition]]:
+    if hasattr(klass, "__strawberry_object__"):
+        return True
+    # Generics remove dunder members here
+    # https://github.com/python/cpython/blob/3a314f7c3df0dd7c37da7d12b827f169ee60e1ea/Lib/typing.py#L1104
```
patrick91, line 187: I think this comment is pointing to the wrong line 😊

patrick91 (commented):
This is amazing! Thank you so much 😊

I left a few comments 😊

strawberry/annotation.py
nrbnlulu: It follows attrs `has_attrs` but I'm cool with both.

strawberry/arguments.py
nrbnlulu: I want to replace it with `__strawberry_input__` in the future. What do you think?

strawberry/object_type.py
nrbnlulu: Since it is used on a class the property would need to use `self`, It is only possible with a metaclass and I wanted to avoid that.

strawberry/schema/compat.py
nrbnlulu: probably :smile:

strawberry/type.py
nrbnlulu: Finally! :smiling_face_with_tear:

skilkis:
@nrbnlulu nice work! I was about to comment last night that I would personally prefer the name `strawberry_type_definition` as opposed to `strawberry_object` but I see you already changed it to `strawberry_definition` which is even better. @patrick91 if you would like to me have a look as well let me know 😊

strawberry/types/types.py
```diff
@@ -167,3 +184,30 @@ def is_implemented_by(self, root: Union[type, dict]) -> bool:
...
+
+
+def get_strawberry_definition(klass: Any) -> Optional[StrawberryObjectDefinition]:
+    if has_strawberry_definition(klass):
+        return klass.__strawberry_definition__
+    return None
+
+
+# TODO: remove when deprecating _type_definition
+TypeDefinition = StrawberryObjectDefinition
```
skilkis, line 213: Good to see the alias here, was worried about it for a second 😁

strawberry/types/types.py
```diff
@@ -167,3 +184,30 @@ def is_implemented_by(self, root: Union[type, dict]) -> bool:
 
         # All field mappings succeeded. This is a match
         return True
+
+
+class WithStrawberryDefinition(Protocol):
+    __strawberry_definition__: StrawberryObjectDefinition
+
+
+def has_strawberry_definition(klass: Any) -> TypeGuard[Type[WithStrawberryDefinition]]:
```
skilkis, line 193: Would prefer `obj` as opposed to `klass` as these functions can be called on any object, not just classes:
https://docs.python.org/3/reference/datamodel.html#data-model

skilkis: Also I believe `WithStrawberryDefinition` and these functions should be moved to `strawberry/type.py` with the latter becoming methods defined on `StrawberryType`. It would also avoid having to perform the deferred import of this module there. Do you agree?

skilkis:
Once again great work @nrbnlulu!! Just left two minor comments, the rest looks good to me 🚀

strawberry/types/types.py
nrbnlulu: Since this method would in the future be true only for object type (input / interface / object) I putted it there. But I think you are right fits better there!

nrbnlulu:
Something is wrong with the workflows https://github.com/strawberry-graphql/strawberry/actions/runs/5251640452/jobs/9486786990?pr=2836#step:7:410 I can't see that line not localy nor on github workspaces. Any ideas?

strawberry/types/types.py
patrick91: yeah, let's move it to strawberry/type.py

I want to get rid of strawberry/types/types.py at some point :D

strawberry/types/types.py
bellini666: I don't know if this would have other implications, but it would be nice to mark this as [deprecated](https://peps.python.org/pep-0702/) like this:

```python
if TYPE_CHECKING:
    @deprecated("Use StrawberryObjectDefinition instead")
    class TypeDefinition(StrawberryObjectDefinition):
        ...
else:
    TypeDefinition = StrawberryObjectDefinition
```

That decorator can be used with the latest version of `typing_extensions`: https://typing-extensions.readthedocs.io/en/latest/#deprecated

bellini666 (changes requested):
This is an awesome change in general! :)

Made some minor suggestions

strawberry/object_type.py
bellini666: Yeah, I looked for it and unfortunately the current `dataclass_transform` specs doesn't allow to define extra attributes that the final type would have. Would be awesome it it allowed though, we would then also be able to mark `__strawberry_object__` as an extra attribute in the type.

Having said that, maybe we could inject a descriptor in `_type_definition` that returns `cls.__strawberry_object__` and raises a deprecation warning when used?

strawberry/annotation.py
nrbnlulu: I called it `StrawberryObjectDefinition` because of GraphQL calls "types" objects, [Input Object](http://spec.graphql.org/October2021/#sec-Input-Object-Values) *type* [Object type](http://spec.graphql.org/October2021/#sec-Objects) [Input object](http://spec.graphql.org/October2021/#sec-Input-Objects) and [interfaces](http://spec.graphql.org/October2021/#sec-Interfaces) all of them are closely related because they have fields that can have other types. Also as far as GraphQL concern enum / scalar are also "types"

nrbnlulu: > I personally prefer is_strawberry_object as well.

So how would you call `get_object_definition`? `get_strawberry_object_definition`?

strawberry/object_type.py
nrbnlulu: That's a good idea

nrbnlulu: I have never implemented a descriptor so I hope I have done it right...

strawberry/schema/schema_converter.py
```diff
@@ -340,7 +344,7 @@ def get_graphql_input_fields(
         )
 
     def from_input_object(self, object_type: type) -> GraphQLInputObjectType:
```
bellini666, line 346: Just notice that `from_input_object` is the only converter that receives a type instead of `StrawberryObjectDefinition`

@patrick91 should we try to normalize this? Don't know if in this PR, but since the breaking change is happening it would be nice. Either this function should receive the object definition, like the others (e.g. from_object, from_interface, etc), or those others should receive the `type` directly (even typing it as `type: WithObjectDefinition`)

bellini666 (changes requested):

strawberry/type.py
```diff
@@ -151,3 +179,35 @@ def __eq__(self, other: object) -> bool:
 
     def __hash__(self):
         return hash(self.type_var)
+
+
+class WithStrawberryObjectDefinition(Protocol):
+    __strawberry_definition__: StrawberryObjectDefinition
```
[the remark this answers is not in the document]
nrbnlulu: True

strawberry/schema/schema_converter.py
patrick91: let's do in another PR 😊

bellini666 (approved):

RELEASE.md
```diff
@@ -0,0 +1,7 @@
+Release type: minor
+
+**BREAKING CHANGES:**
+
+This release renames `_type_definition` to `__strawberry_definition__`
+if you don't know what `_type_definition` is, there is no need to worry 🙃.
+However, if you have used it, it would be aliased for now until deprecation.
```
patrick91, line 7:
```suggestion

This release renames `_type_definition` to `__strawberry_definition__`. This doesn't change the public API of Strawberry, but if you were using `_type_definition` you can still access it, but it will be removed in future.
```

patrick91 (approved):
Thank you so much for doing this! <3

PR #2838, opened 2023-06-11
refactor: resolve `relay.NodeID` annotations when creating the schema itself

### LINKED ISSUE #2833: NameError with lazy types
## Describe the Bug

When I use lazy types as described in https://strawberry.rocks/docs/types/lazy I get a NameError like the following:
```
NameError: name 'Attribute' is not defined
```

Relevant code:
```
from strawberry_django_plus import gql
...

if TYPE_CHECKING:
    from api.attribute.types.attribute import Attribute
    
...
@gql.django.type(models.ProductType)
class ProductType(gql.relay.Node):
    @gql.relay.connection
    def attributes(self) -> list[Annotated["Attribute", gql.lazy("api.attribute.types.attribute")]]:
        pass
```

The error goes away if I use a regular forward reference without `Annotation` and `lazy`, like so:
```
    @gql.relay.connection
    def attributes(self) -> list["Attribute"]:
        pass
```
(Although I'm guessing this would prevent the schema from being generated correctly. By the way... Is it actually necessary to use `Annotated` and `lazy`? It feels weird to write out the import path twice... Why not treat all forward references as lazy types?)

## System Information

Python 3.10.7

bellini666:
Fix #2833

### DIFF
--- a/strawberry/relay/types.py
+++ b/strawberry/relay/types.py
@@ -399,40 +399,41 @@ class Node:
 
     @classmethod
     def resolve_id_attr(cls) -> str:
-        if cls._id_attr is None:
-            candidates: list[str] = []
-            for base in cls.__mro__:
-                base_namespace = sys.modules[base.__module__].__dict__
+        if cls._id_attr is not None:
+            return cls._id_attr
 
-                for attr_name, attr in getattr(base, "__annotations__", {}).items():
-                    evaled = eval_type(attr, globalns=base_namespace)
+        candidates: list[str] = []
+        for base in cls.__mro__:
+            base_namespace = sys.modules[base.__module__].__dict__
 
-                    if get_origin(evaled) is Annotated and any(
-                        isinstance(a, NodeIDPrivate) for a in get_args(evaled)
-                    ):
-                        candidates.append(attr_name)
+            for attr_name, attr in getattr(base, "__annotations__", {}).items():
+                evaled = eval_type(attr, globalns=base_namespace)
 
-                # If we found candidates in this base, stop looking for more
-                # This is to support subclasses to define something else than
-                # its superclass as a NodeID
-                if candidates:
-                    break
+                if get_origin(evaled) is Annotated and any(
+                    isinstance(a, NodeIDPrivate) for a in get_args(evaled)
+                ):
+                    candidates.append(attr_name)
 
-            if len(candidates) == 0:
-                raise NodeIDAnnotationError(
-                    f'No field annotated with `NodeID` found in "{cls.__name__}"', cls
-                )
-            if len(candidates) > 1:
-                raise NodeIDAnnotationError(
-                    (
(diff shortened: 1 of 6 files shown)


botberry:



---

This release changes the way we check for `relay.NodeID` annotations to be done
later instead of in class initialization. This solves problems with trying to
evaluate type annotations before they are totally defined, and also allows
integrations to inject code for it in the type.


---




Here's the tweet text: 
```

```

strawberry/relay/types.py
```diff
@@ -426,6 +397,44 @@ def _id(cls, root: Node, info: Info) -> GlobalID:
...
         return GlobalID(type_name=type_name, node_id=str(node_id))
 
+    @classmethod
+    def resolve_id_attr(cls) -> str:
+        if cls._id_attr is None:
+            candidates: list[str] = []
+            for base in cls.__mro__:
+                base_namespace = sys.modules[base.__module__].__dict__
+
+                for attr_name, attr in getattr(base, "__annotations__", {}).items():
```
devkral, line 407: wouldn't it be better to do it like in:
types/type_resolver/_get_fields
(use dataclasses.fields)

bellini666: In theory (and it was actually my first idea as well =P), but the `NodeID` is a `Private` field, meaning it will be skipped and not added to the type's fields

strawberry/relay/types.py
```diff
@@ -426,6 +397,44 @@ def _id(cls, root: Node, info: Info) -> GlobalID:
         # If node_id is not str, GlobalID will raise an error for us
         return GlobalID(type_name=type_name, node_id=str(node_id))
 
+    @classmethod
+    def resolve_id_attr(cls) -> str:
+        if cls._id_attr is None:
+            candidates: list[str] = []
```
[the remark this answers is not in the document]
bellini666: Sure! I tried to have a single return clause, but the code in the scope is actually large =P

strawberry/schema/schema.py
```diff
@@ -163,6 +164,7 @@ def __init__(
         self._schema._strawberry_schema = self  # type: ignore
 
         self._warn_for_federation_directives()
+        self._validate_types_node_id_attr()
```
patrick91, line 167: shall we call this `_resolve_node_ids`?

patrick91: also, do you think we can merge this with the code at line 128?

CleanShot 2023-06-12 at 20 50 17@2x

bellini666: > shall we call this `_resolve_node_ids`?

Sure! I didn't like the name I gave it, that is better :)

> also, do you think we can merge this with the code at line 128?

I initially tried to do that, but the iterated `types` in there are just extra types that you can pass to the schema and not all existing types :(

patrick91 (approved):
Thanks!

PR #2839, opened 2023-06-11
refactor: copy field arguments instead of extending them

bellini666:


botberry:



---

Refactor `ConnectionExtension` to copy arguments instead of extending them.
This should fix some issues with integrations which override `arguments`,
like the django one, where the inserted arguments were vanishing.


---




Here's the tweet text: 
```

```

patrick91 (approved):
Looks good to me, do you think it is worth adding a test for this?

bellini666:
> Looks good to me, do you think it is worth adding a test for this?

Sure! :)

Added a test that would fail without this change. Without this change, the `fruitCustomField` would have only `foo` argument, and the relay ones would be lost. This mimics what strawberry-graphql-django does on arguments

PR #2840, opened 2023-06-11
Implement unions as annotated types

Ambro17:
## Description

Replace `strawberry.union` direct usage with `typing.Annotated` for better type hints

- [ ] Add test for warning over multiple args in `tests/types/resolving/test_unions.py`
- [ ] Add tests for unions new api equivalent to those of the old api in `tests/types/resolving/test_unions.py`

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Nahuel Ambrosini for the PR 👏

```

strawberry/annotation.py
```diff
@@ -183,6 +186,20 @@ def create_union(self, evaled_type: Type) -> StrawberryUnion:
         union = StrawberryUnion(
             type_annotations=tuple(StrawberryAnnotation(type_) for type_ in types),
         )
+
+        union_args = [arg for arg in args if isinstance(arg, StrawberryUnion)]
+        if len(union_args) > 1:
+            logging.warning(
```
patrick91, line 192: shall we add a test for this? 😊

patrick91 (commented):
Thank you so much! <3 

I've left only one comment 😊

strawberry/annotation.py
Ambro17: Good idea, will add

patrick91:
I've worked on this a bit more, I think it is almost ready to go, I want to improve the errors, but shouldn't take too long 😊

botberry:
# Pre-release
:wave:

Releasing commit [f6f04d2bdd952a7abd9f6a6d5d642dcf6878635d] to PyPi as pre-release! :package:

botberry:
# Pre-release
:wave:

Pre-release **0.189.1.dev.1687473609** [f6f04d2bdd952a7abd9f6a6d5d642dcf6878635d] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.189.1.dev.1687473609
```

patrick91:
This should be ready now 😊

strawberry/annotation.py
```diff
@@ -182,11 +187,61 @@ def create_union(self, evaled_type: Type) -> StrawberryUnion:
...
+            union._source_file = arg._source_file
+            union._source_line = arg._source_line
+
+        self.validate_union_members(types, union)
+
         return union
 
+    def validate_union_members(
+        self, types: Tuple[type], union: StrawberryUnion
+    ) -> None:
```
bellini666, line 217: Since this doesn't use `self`, I wonder if it would be better to define it as a separated function.

Even better, define it as a method in `StrawberryUnion` as `def validate_members(types: Tuple[type])`, or even without `types`, since the union already have those inside of it.

strawberry/union.py
```diff
@@ -254,9 +254,36 @@ def union(
...
+        )
+
+    warnings.warn(
+        (
+            "Passing types to `strawberry.union` is deprecated. Please use "
+            f'{name} = Annotated[Union[A, B], strawberry.union("{name}")] instead'
+        ),
+        DeprecationWarning,
+        stacklevel=2,
+    )
```
bellini666, line 285: How about reducing one nested lvl here by putting thr `warnings` in a `if types is not None` and with an early return, and the main code with no indentation?

strawberry/union.py
```diff
@@ -53,11 +56,16 @@ def __init__(
         type_annotations: Tuple[StrawberryAnnotation, ...] = tuple(),
         description: Optional[str] = None,
         directives: Iterable[object] = (),
+        # used for better error messages
+        _source_file: Optional[str] = None,
+        _source_line: Optional[int] = None,
```
bellini666, line 61: I personally would prefer to not have those in the `__init__`, since it is mostly for internal usage and it will pollute code completions for people not working in the internals.

If we must have this it would at least be nice to add those after a `*` to make sure they are forced to be passed as keywords, otherwise adding an extra arg in here would either break some APIs or we would be forced to add those after `_source_file`/`_source_line` and make it mandatory to pass those.

But maybe we could just set those in the __init__ like:

```python
self.source_file = Optional[str] = None
self.source_line = Optional[int] = None
```

And set them by hand after instantializing the object.

Also, don't know if this makes sense, but since this type inherits from `StrawberryType`, and a lot of other subclasses might want to use something similar (e.g. `StrawberryList`, `StrawberryOptional`, etc), we could add those lines there so that all `StrawberryType` subclasses can have a way to provide info for rich exceptions?

cc @patrick91

bellini666 (changes requested):
My 2 cents :)

strawberry/union.py
patrick91: the `_source_file` is only for things that don't have an origin so not sure if it makes sense to add it everywhere :)

I like your suggestion, I'll try to implement it today

strawberry/annotation.py
```diff
@@ -182,11 +187,61 @@ def create_union(self, evaled_type: Type) -> StrawberryUnion:
...
+                union_members = get_args(inner_type)
+
+                union_definition = (
+                    metadata[0]
+                    if metadata and isinstance(metadata[0], StrawberryUnion)
+                    else None
+                )
+
+                for member in union_members:
+                    if isinstance(member, scalars):
```
bellini666, line 237: Also, it is not just those scalars (int, str and float) that cannot be used. We also have custom scalars which would need to be checked.

Maybe the best way in here is to check that the members are strawberry types (i.e. probably by using `has_object_definition`?) and not inputs because inputs cannot be used in unions as well.

strawberry/union.py
patrick91: it's pretty much the same 😊 since the rest of the code will go in the if :)

strawberry/annotation.py
patrick91: I removed this code, we already had another place that was doing a similar check 😊

patrick91:
this is ready!

bellini666 (approved):
No more suggestions, LGTM 👍🏼

patrick91:
@bellini666 let's merge yours first and then this one later today or tomorrow 😊

PR #2834, opened 2023-06-10
Fix: don't require NodeId if resolve_id is overwritten

devkral:
If resolve_id is overwritten it is valid that no NodeId annotation is found (custom logic). Don't fail in this case

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement/optimization
- [ ] Documentation

## Issues Fixed or Closed by This PR

* #2816

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the CONTRIBUTING document.
- [x] I have added tests to cover my changes.
- [x] I have tested the changes and verified that they work and don't break anything (as well as I can manage).

Issue #2816, opened 2023-06-06
relay: subclasses need relay.NodeId even resolve_id is specified


devkral:
In case of a overwritten resolve_id the subclass should not be checked for the relay.NodeId  annotation

patrick91:
Fixed in https://github.com/strawberry-graphql/strawberry/pull/2844

PR #2844, opened 2023-06-12
fix: relay.Node with overwritten resolve_id and without NodeId 

devkral:
Using a Node without a NodeId annotation should be possible in case of the resolve_id method is overwritten

Fix a remaining bug and add tests for this case

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #2816
* #2834

## Checklist

### DIFF
--- a/strawberry/schema/schema.py
+++ b/strawberry/schema/schema.py
@@ -26,6 +26,7 @@ from graphql import (
 from graphql.execution import subscribe
 from graphql.type.directives import specified_directives
 
+from strawberry import relay
 from strawberry.annotation import StrawberryAnnotation
 from strawberry.extensions.directives import (
     DirectivesExtension,
@@ -328,6 +329,21 @@ class Schema(BaseSchema):
             # them needs to do that.
             if type_def.is_interface:
                 continue
+
+            # Call resolve_id_attr in here to make sure we raise provide
+            # early feedback for missing NodeID annotations
+            origin = type_def.origin
+            if issubclass(origin, relay.Node):
+                has_custom_resolve_id = False
+                for base in origin.__mro__:
+                    if base is relay.Node:
+                        break
+                    if "resolve_id" in base.__dict__:
+                        has_custom_resolve_id = True
+                        break
+
+                if not has_custom_resolve_id:
+                    origin.resolve_id_attr()
 
     def _warn_for_federation_directives(self):
         """Raises a warning if the schema has any federation directives."""
--- a/strawberry/schema/schema.py
+++ b/strawberry/schema/schema.py
@@ -334,7 +334,16 @@ class Schema(BaseSchema):
             # early feedback for missing NodeID annotations
             origin = type_def.origin
             if issubclass(origin, relay.Node):
-                origin.resolve_id_attr()
+                has_custom_resolve_id = False
+                for base in origin.__mro__:
+                    if base is relay.Node:
+                        break
+                    if "resolve_id" in base.__dict__:
+                        has_custom_resolve_id = True
+                        break
+
+                if not has_custom_resolve_id:
+                    origin.resolve_id_attr()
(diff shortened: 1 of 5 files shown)


botberry:



---

relay: don't require NodeId annotation if resolve_id is overwritten
relay: add tests for resolve_id overwrites


---




Here's the tweet text: 
```

```

patrick91 (commented):
@bellini666 what do you think of this?

bellini666 (changes requested):
@patrick91 oh, this indeed makes sense.

@devkral sorry for not taking a look at this earlier. Left a couple of comments in here

devkral:
ok?

tests/relay/test_schema.py
```diff
@@ -356,3 +356,47 @@ def fruits(self) -> List[Fruit]:
...
+"""An object with a Globally Unique ID"""
+interface Node {
+  """The Globally Unique ID of this object"""
+  id: GlobalID!
+}
+
+type Query {
+  fruit: Fruit!
+}
+    '''
```
bellini666, line 398: nitpick to keep this using the same style as the other tests in this file

```suggestion
    expected_type = textwrap.dedent(
        '''
        type Fruit implements Node {
        """The Globally Unique ID of this object"""
        id: GlobalID!
        color: String!
      }

      """__GLOBAL_ID_DESC__"""
      scalar GlobalID @specifiedBy(url: "https://relay.dev/graphql/objectidentification.htm")

      """An object with a Globally Unique ID"""
      interface Node {
        """The Globally Unique ID of this object"""
        id: GlobalID!
      }

      type Query {
        fruit: Fruit!
      }
      '''
```

bellini666 (changes requested):
@devkral thank you! All good now!

Left a small nitpick and a suggestion for the changelog. Other than that it is a 👍🏼 for me

bellini666 (approved):
LGTM 👍🏼 

Unless @patrick91 has any comment in here, this should be fine to merge

strawberry/schema/schema.py
```diff
@@ -334,7 +334,16 @@ def _resolve_node_ids(self):
...
+                has_custom_resolve_id = False
+                for base in origin.__mro__:
+                    if base is relay.Node:
+                        break
+                    if "resolve_id" in base.__dict__:
+                        has_custom_resolve_id = True
+                        break
+
+                if not has_custom_resolve_id:
+                    origin.resolve_id_attr()
```
patrick91, line 346: could this be baked into Node itself?

devkral: maybe with __init_subclass__

devkral: can you please test the change? Currently because of pydantic errors I cannot test

devkral:
seems there are some problems with the tests

strawberry/schema/schema.py
devkral: this would prevent non-final Node subclasses.

bellini666: I may be wrong, but I imagine what @patrick91 suggested is to move the logic to move this logic to inside `resolve_id_attr` itself

devkral: this breaks the guarantee that `resolve_id_attr` always returns an attribute or error.

It would be better to have an hook on Node and possible other types: `add_to_schema(cls, schema)`. This would allow a type agnostic customization

devkral: but I think this should be scope of a different PR.

devkral:
ping

devkral:
ping?

patrick91 (approved):

bellini666:
Ty! :)

devkral:
thanks

PR #2847, opened 2023-06-14
Fix directive typehint

### LINKED ISSUE #2846: `strawberry.directive` type hints appear to be incorrect.
## Describe the Bug

A function like:

```py
@strawberry.directive(
    locations=[DirectiveLocation.FRAGMENT_DEFINITION],
    description="Do nothing, but add a base class for query generation in the python client.",
)
def identity(value: str, a: str, b: str | None = None) -> str:
    """Do nothing, but add a directive so that the python client can use this to add metadata."""
    return value

reveal_type(identity)
```

`mypy` will say that the revealed type here is `builtins.str`.  Looking at the code for `directive`, it appears that the type hints imply that the decorated object is going to return the same type that the function returns (in this case, the return type of `identity`)

## System Information

 - Operating system:
 - Strawberry version (if applicable):

mgilson:
## Description

Correct the type hints on `strawberry.directive`.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* fixes #2846 

## Checklist

### DIFF
--- a/tests/mypy/test_directives.yml
+++ b/tests/mypy/test_directives.yml
@@ -1,0 +1,20 @@
+- case: test_directive_decorator
+  main: |
+    import strawberry
+    from strawberry.directive import DirectiveLocation
+
+    @strawberry.directive(
+        locations=[DirectiveLocation.FRAGMENT_DEFINITION],
+        description="description.",
+    )
+    def make_int(value: str) -> int:
+        """description."""
+        try:
+            return int(value)
+        except ValueError:
+            return 0
+
+    reveal_type(make_int)
+
+  out: |
+    main:15: note: Revealed type is "strawberry.directive.StrawberryDirective[builtins.int]"
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,5 @@
+Release type: patch
+
+Correct a type-hinting bug with `strawberry.directive`.
+This may cause some consumers to have to remove a `# type: ignore` comment
+or unnecessary `typing.cast` in order to get `mypy` to pass.
--- a/strawberry/directive.py
+++ b/strawberry/directive.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import dataclasses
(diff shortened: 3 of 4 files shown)


botberry:



---

Correct a type-hinting bug with `strawberry.directive`.
This may cause some consumers to have to remove a `# type: ignore` comment
or unnecessary `typing.cast` in order to get `mypy` to pass.


---




Here's the tweet text: 
```

```

patrick91 (commented):
Nice! Would it be possible to add a test for mypy/pyright? so we make sure we don't break this in future?

patrick91:
> That's a pretty neat testing setup that you have.
> 
> I've added the tests. The pyright tests 🦘 by default until I did a `poetry add pyright` -- that updated the project packaging though -- I'm not sure if there is a particular group that I should be using or a different command (`poetry run python -m pip install pyright`?). Anyway, you might want to check that those tests are actually running all the time when you expect them to be.

ah, yes they only run if pyright is installed, maybe we can add it to our dependencies! I didn't realise we could install it via pip 😊

we do install it on GitHub: https://github.com/strawberry-graphql/strawberry/blob/main/.github/workflows/test-type-checkers.yml#L64

tests/mypy/test_directives.yml
```diff
@@ -0,0 +1,20 @@
...
+        """description."""
+        try:
+            return int(value)
+        except ValueError:
+            return 0
+
+    reveal_type(make_int)
+
+  out: |
+    main:15: note: Revealed type is "strawberry.directive.StrawberryDirective[builtins.int]"
```
patrick91, line 20: this might be for another PR, but I wonder if we should return the function as the type, it is true that we wrap it but that should just be an internal thing (similar to what we do for scalars, types, etc)

I don't remember why we wrap the function instead of monkeypatching it.

I'll make a note to check this in future, I don't want to block this PR 😊

patrick91 (approved):

Issue #2799, opened 2023-06-01
Add the option in ChannelsConsumer to confirm topic subscription


moritz89:
In order to inform the subscriber (GraphQL client) that the GQL subscription has been activated, it would be useful to return a null message. In order to avoid a race condition, it would be useful to configure `ChannelsConsumer.channel_listen` to return None as soon as the subscription on the channels layer has been activated.

## Feature Request Type

- [ ] Core functionality
- [x] Alteration (enhancement/optimization) of existing feature(s)
- [ ] New behavior

## Description

In `ChannelsConsumer.channel_listen`, I would like to add the option of yielding None after the `channel_layer.group_add` loop:

```py
# strawberry/channels/handlers/base.py:150
for group in groups:
    await self.channel_layer.group_add(group, self.channel_name)
    added_groups.append(group)
```

This would allow the subscriber to be written like this:

```py
@strawberry.type
class FooSubscription:
    @strawberry.subscription
    @staticmethod
    async def foo_subscription(info: Info) -> AsyncGenerator[FooType | None, None]:
        # yield None
        async for message in info.context.ws.channel_listen("foo_type", groups=["a"]):
            if message is None:
                yield None
                continue
            yield FooType(message["payload"])
```

The commented out `yield None` is a workaround but confirms the subscription before the topic has been subscribed to. This is noticeable in tests when testing the Channels communication via REDIS.

patrick91:
@DoctorJohn what do you think? 😊

DoctorJohn:
Interesting. Yielding `None` from `channel_listen` would certainly work, but I feel like it's not obvious to newcomers why the first yielded value would always be `None`. After a quick local experiement I think it's possible to make the `channel_listen` method return an AsyncGenerator after all groups have been added to the channel layer. The method could then be used like this:

```python
@strawberry.type
class FooSubscription:
    @strawberry.subscription
    @staticmethod
    async def foo_subscription(info: Info) -> AsyncGenerator[FooType | None, None]:
        gen = await info.context.ws.channel_listen("foo_type", groups=["a"])
        yield None
        async for message in gen:
            yield FooType(message["payload"])
```

However, this would still be a breaking change since `channel_listen` would now return `Awaitable[AsyncGenerator]` instead of `AsyncGenerator`.

moritz89:
Thanks for investigating possible solutions. Would an idea be to have a separate method that returns the awaitable AsyncGenerator but is used by the original channel_listen method?

[...]

PR #2856, opened 2023-06-16
Reimplement channel_listen as context manager

moritz89:
Why:

- Allow user code to be run after Channels subscription
- Avoids race condition when confirming GQL subscriptions

This change addresses the need by:

- Returning the async generator instead of runnning it

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* https://github.com/strawberry-graphql/strawberry/issues/2799

## Checklist


moritz89:
Addresses issue #2799

botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to @moritz89 for the PR 👏

```

moritz89:
This is the simplest implementation possibility. However, it has the problem that a leak can occur if the async generator is never started. Will create a second approach that uses a context manager to ensure that the cleanup will be done.

moritz89:
Added a second option of starting channel_listen with a context manager to ensure that group_discard is called after it is started.

moritz89:
@DoctorJohn @patrick91 Wanted to ask which approach you'd prefer? ContextManager or the simple approach

DoctorJohn:
> @DoctorJohn @patrick91 Wanted to ask which approach you'd prefer? ContextManager or the simple approach

Hi @moritz89 thanks for the PR. I prefer the ContextManager approach. It's the one I did my quick proof of concept with. Main argument for it is the cleanup ability you already mentionened in your comment. Feel free to remove the "simple approach".

moritz89:
Isn't this fun, adding breaking changes :sweat_smile: Anything else that should be updated?

patrick91:
@moritz89 it is a lot of work to make it backwards compatibile with a deprecation? 😊

moritz89:
> @moritz89 it is a lot of work to make it backwards compatibile with a deprecation? blush

Not at all. The question is only how this should be handled. It would result in two functions in the API and the question is how they should be named. One option is to append `_ctx` to the new function.

moritz89:
@patrick91 I've had a look at unifying the calls in the following style, but also that did not work and going down that line of thought makes the library code itself more and more tricky. If you have any suggestions, would be glad to take them up.

```
class ChannelListen:
    def __init__(
        self, type: str, *, timeout: Optional[float] = None, groups: Sequence[str] = ()
    ) -> None:
        ...

    async def __call__(
        self, type: str, *, timeout: Optional[float] = None, groups: Sequence[str] = ()
    ) -> Any:
        ...

    async def __aenter__(self) -> Awaitable[AsyncGenerator[Any, None]]:
        ...

    async def __aexit__(self) -> None:
        ...

    async def _channel_listen_generator(
        self, queue: asyncio.Queue, timeout: Optional[float]
    ) -> AsyncGenerator[Any, None]:
        ...

class ChannelsConsumer(AsyncConsumer):
    ...
    channel_listen = ChannelListen
```

moritz89:
Fixed mypy issues and updated existing tests and added a new one, but don't know why one is failing:

`ERROR 😡  tests/cli/test_server.py::test_debug_server_routes - KeyError: 'STRAWBERRY_DEBUG_SERVER_SCHEMA'`

Would appreciate help in fixing it

moritz89:
How should the deprecation be communicated to the library users?

DoctorJohn:
> I'll create a commit assuming you meant the former (deprecate the existing approach).

Yup, that's what I meant. I think the new approach is superior and covers all usecases of the old one.

> How should the deprecation be communicated to the library users?

Using the `warnings` module, just like here for example:

https://github.com/strawberry-graphql/strawberry/blob/19be5751c888bea76c03401be105a4a1b80c723c/strawberry/types/info.py#L52-L56

moritz89:
@patrick91 The MR won't break compatibility but adds a deprecation warning

docs/integrations/channels.md
```diff
@@ -136,13 +136,15 @@ class Subscription:
...
-                )
+        async with ws.channel_listen("chat.message", groups=room_ids) as cm:
+            # Optionally: Confirm subscription with `yield None`
+            async for message in cm:
+                if message["room_id"] in room_ids:
+                    yield ChatRoomMessage(
+                        room_name=message["room_id"],
+                        message=message["message"],
+                        current_user=user,
+                    )
```
DoctorJohn, line 147: These docs need an update now :)

```suggestion
        async with ws.listen_to_channel("chat.message", groups=room_ids) as cm:
            async for message in cm:
                if message["room_id"] in room_ids:
                    yield ChatRoomMessage(
                        room_name=message["room_id"],
                        message=message["message"],
                        current_user=user,
                    )
```

docs/integrations/channels.md
```diff
@@ -153,7 +155,8 @@ message to all the channel_layer groups (specified in the subscription argument
 <Note>
 
 The `ChannelsConsumer` instance is shared between all subscriptions created in
-a single websocket connection. The `ws.channel_listen` function will yield all
+a single websocket connection. The `ws.channel_listen` context manager will return
```
DoctorJohn, line 158:
```suggestion
a single websocket connection. The `ws.listen_to_channel` context manager will return
```

Here as well :)

DoctorJohn (changes requested):

patrick91:
> @patrick91 The MR won't break compatibility but adds a deprecation warning

Thank you so much! <3

moritz89:
@DoctorJohn @patrick91 Does this MR require anything else or can it be merged?

strawberry/channels/handlers/base.py
```diff
@@ -144,6 +146,79 @@ async def channel_listen(
...
+                at the end of the execution.
+        """
+
+        # Code to acquire resource (Channels subscriptions)
+        if self.channel_layer is None:
+            raise RuntimeError(
+                "Layers integration is required listening for channels.\n"
+                "Check https://channels.readthedocs.io/en/stable/topics/channel_layers.html "  # noqa:E501
+                "for more information"
+            )
```
patrick91, line 181: 👍

RELEASE.md
```diff
@@ -0,0 +1,43 @@
...
+class MyDataType:
+   name: str
+
+@strawberry.type
+class Subscription:
+   @strawberry.subscription
+   async def my_data_subscription(
+      self, info: Info, groups: list[str]
+   ) -> AsyncGenerator[MyDataType | None, None]:
+      yield None
```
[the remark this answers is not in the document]
moritz89: It's to show how GQL subscription confirmations (with null data) should be migrated, i.e., that confirmation messages should be sent before the `async for` loop. However, for every else not confirming their GQL subscriptions the `yield None` can be omitted in both new and old code variants.

How should the doc be updated to reflect that?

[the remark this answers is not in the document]
moritz89: Good point. Added a section in the channels.md doc

botberry:

PR #2858, opened 2023-06-16
Fix deprecation warning

### LINKED ISSUE #2857: Tests failing from DeprecationWarnings in strawberry code
## Describe the Bug

Our tests have started failing with a later release of Strawberry (0.183.6) because of deprecation warnings. They fail the moment they try to import a strawberry module. We run our tests to error on warnings and we do not ignore deprecation warnings. 

```
../xxx/strawberry/types/fields/resolver.py:199: in reserved_parameters
    return {spec: spec.find(parameters, self) for spec in self.RESERVED_PARAMSPEC}
../xxx/strawberry/types/fields/resolver.py:199: in <dictcomp>
    return {spec: spec.find(parameters, self) for spec in self.RESERVED_PARAMSPEC}
../xxx/strawberry/types/fields/resolver.py:128: in find
    warnings.warn(warning, stacklevel=3)
E   DeprecationWarning: Argument name-based matching of 'info' is deprecated and will be removed in v1.0. Ensure that reserved arguments are annotated their respective types (i.e. use value: 'DirectiveValue[str]' instead of 'value: str' and 'info: Info' instead of a plain 'info').

```

I've tracked this down and it seems that some strawberry code is actually triggering the deprecation warning. This is happening in `strawberry/relay/types.py:365`. 
https://github.com/strawberry-graphql/strawberry/blob/d4ae4a5f4504c34fd5c7234f0bb7266e8cfbe674/strawberry/relay/types.py#L365
The `info:Info` parameter looks like it is being declared in the recommended way (given the warning above), however it still triggers the warning. If I change the `info` variable to be named `info2` for example, the warning goes away.

As a test, I've also run the strawberry test locally but have removed the `ignore:DeprecationWarning:strawberry.*.resolver` flag (i.e. https://github.com/strawberry-graphql/strawberry/blob/d4ae4a5f4504c34fd5c7234f0bb7266e8cfbe674/pyproject.toml#LL164C1-L164C56) and also set to error on warning. In this case I get the same failures due to deprecation warnings in the strawberry tests.

## System Information

 - Operating system: Centos 7 
 - Strawberry version (if applicable): 0.183.6 (have not tested exactly which release introduced this issue). It is still present in 0.186.0

patrick91:
Closes #2857

This is probably not the best way to fix this deprecation warning, as I think the code that checks for `Info` doesn't support forward refs when `Info` is not defined. I think we could fix it by checking the name of the type, but this is probably an ok workaround for the time being

~EDIT:~

~I've updates the PR to check the name of the class 😊~

/cc @bellini666 @skilkis

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,4 @@
+Release type: minor
+
+This release fixes a deprecation warning generated inside our codebase and
+adds support for using `Info` when declared in a `if TYPE_CHECKING` block
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,4 +1,0 @@
-Release type: minor
-
-This release fixes a deprecation warning generated inside our codebase and
-adds support for using `Info` when declared in a `if TYPE_CHECKING` block
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,4 @@
+Release type: patch
+
+This release fixes a depreaction warning being triggered
+by the relay integration.
--- a/strawberry/relay/types.py
+++ b/strawberry/relay/types.py
@@ -38,6 +38,7 @@ from strawberry.object_type import interface, type
 from strawberry.private import StrawberryPrivate
 from strawberry.relay.exceptions import NodeIDAnnotationError
 from strawberry.type import StrawberryContainer
+from strawberry.types.info import Info  # noqa: TCH001
 from strawberry.types.types import TypeDefinition
 from strawberry.utils.aio import aenumerate, aislice, resolve_awaitable
 from strawberry.utils.inspect import in_async_context
@@ -47,7 +48,6 @@ from .utils import from_base64, to_base64
 
 if TYPE_CHECKING:
     from strawberry.scalars import ID
-    from strawberry.types.info import Info
     from strawberry.utils.await_maybe import AwaitableOrValue
 
 _T = TypeVar("_T")
--- a/strawberry/relay/types.py
+++ b/strawberry/relay/types.py
@@ -38,7 +38,6 @@ from strawberry.object_type import interface, type
 from strawberry.private import StrawberryPrivate
 from strawberry.relay.exceptions import NodeIDAnnotationError
 from strawberry.type import StrawberryContainer
-from strawberry.types.info import Info  # noqa: TCH001
 from strawberry.types.types import TypeDefinition
 from strawberry.utils.aio import aenumerate, aislice, resolve_awaitable
 from strawberry.utils.inspect import in_async_context
@@ -48,6 +47,7 @@ from .utils import from_base64, to_base64
 
 if TYPE_CHECKING:
     from strawberry.scalars import ID
+    from strawberry.types.info import Info
     from strawberry.utils.await_maybe import AwaitableOrValue
 
 _T = TypeVar("_T")
--- a/strawberry/types/fields/resolver.py
+++ b/strawberry/types/fields/resolver.py
@@ -135,14 +135,6 @@ class ReservedType(NamedTuple):
         if origin is Annotated:
             # Handle annotated arguments such as Private[str] and DirectiveValue[str]
             return any(isinstance(argument, self.type) for argument in get_args(other))
-        elif type(origin) is str:
-            # Handle forward references, there might be conflicts with the name,
-            # for example is someone makes an input type called "Info", but this
-            # is a rare case and we can ignore it for now
-            # another option would be to check if strawberry.types.info.Info is
-            # imported in a `if TYPE_CHECKING:` block, but that feels like too much
-            # work for a rare case
-            return origin == self.type.__name__
         else:
             # Handle both concrete and generic types (i.e Info, and Info[Any, Any])
             return (
--- a/strawberry/types/fields/resolver.py
+++ b/strawberry/types/fields/resolver.py
@@ -135,6 +135,14 @@ class ReservedType(NamedTuple):
         if origin is Annotated:
             # Handle annotated arguments such as Private[str] and DirectiveValue[str]
             return any(isinstance(argument, self.type) for argument in get_args(other))
+        elif type(origin) is str:
+            # Handle forward references, there might be conflicts with the name,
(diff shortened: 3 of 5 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

bellini666 (approved):

bellini666:
Really interesting!

> This is probably not the best way to fix this deprecation warning, as I think the code that checks for Info doesn't support forward refs when Info is not defined. I think we could fix it by checking the name of the type, but this is probably an ok workaround for the time being

I wonder if we should inject `Info` in the resolvers namespace: https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/types/fields/resolver.py#L318

I mean, specially with this ruff's rule being used everywhere now, `info: Info` has a large probability of being rewritten to the inside of a `if TYPE_CHECKING` check by ruff itself.

Something like this maybe would also solve the issue for the relay module, and also for user defined code doing something similar?

```python
    @cached_property
    def _namespace(self) -> Dict[str, Any]:
        namespace = sys.modules[self._unbound_wrapped_func.__module__].__dict__
        namespace.setdefault("info", Info)
        return namespace
```

patrick91:
@bellini666 wouldn't it be easier to check the name? 😊

bellini666:
> @bellini666 wouldn't it be easier to check the name? blush

Hrmm, maybe, yes =P

But not sure what would happen for some "weird" cases, like "info.Info" (not actually weird) or "from strawberry.types.info import Info as StrawberryInfo"

patrick91:
I think falling back to checking the string should be fine, there's a case in where someone might to an input type called `Info`, but that would be rare 😊

patrick91:
@bellini666 updated to check the name of the class 😊

bellini666 (approved):
Nice! :)

patrick91:
Somehow this changed:

CleanShot 2023-06-16 at 20 27 08@2x

bellini666:
> 

Wow, no idea what happened there. The strangest part is that it didn't remove the `id`/`ids` argument, it just removed the description for some reason 😯

patrick91:
you'll investigate after dinner :)

skilkis:
@patrick91 @bellini666 there's indeed nothing robust we can do in the scenario where `Info` has been imported in a `TYPE_CHECKING` block. Only viable solution I see would be to run a type-checker on the source code but that would come with an import time penalty or we make our own tiny AST parser localized to `TYPE_CHECKING` blocks. However, the source code might not even be available if someone is compiling with Cython or Nuitka.

I propose that we keep the current behavior but come up with a new warning for this case so that user's are aware of how to fix it, so instead of the deprecation warning something like:

**warning: ambiguous parsing result, parameter 'foo' annotated with forward referenced `Info` that is unresolvable from module `myFoo.py`. Ensure that the `Info` type is accessible from the module scope and not imported within an `is TYPE_CHECKING` block`**

This ensures that we don't have a false positive which could be quite annoying for the user to debug, wondering why their custom `Info` input type is not showing up in their API. We enforce resolvability for user-types otherwise we wouldn't be able to generate the schema types, so then why would we not enforce it for our own types as well?

skilkis:
> Somehow this changed:
> CleanShot 2023-06-16 at 20 27 08@2x

@patrick91 could you give me the repro for this? I wrote the refactored resolver argument parsing so I could perhaps help with debugging this ☺

patrick91:
> > Somehow this changed:
> 
> > CleanShot 2023-06-16 at 20 27 08@2x
> 
> 
> 
> @patrick91 could you give me the repro for this? I wrote the refactored resolver argument parsing so I could perhaps help with debugging this ☺

it’s the failing test on the actions ☺️

patrick91:
I think I figured it out 😊

CleanShot 2023-06-16 at 23 12 35@2x

this code is converting the string annotation to standard annotation, so when the code later on tries to add the strawberry argument to the resolver (by instanciating it) it still has the annotation a string:

CleanShot 2023-06-16 at 23 14 11@2x

patrick91:
Adding the `Info` to the namespace won't work, since `Node` is also not on the namespace:

CleanShot 2023-06-16 at 23 16 59@2x

patrick91:
I wonder if we could move the parsing of the annotated later on in the code, but I'll go to sleep for now :D

skilkis:
> I wonder if we could move the parsing of the annotated later on in the code, but I'll go to sleep for now :D

I always agree with that sentiment, I don't think we should parse anything until `strawberry.Schema` is called

patrick91:
> @patrick91 @bellini666 there's indeed nothing robust we can do in the scenario where `Info` has been imported in a `TYPE_CHECKING` block. Only viable solution I see would be to run a type-checker on the source code but that would come with an import time penalty or we make our own tiny AST parser localized to `TYPE_CHECKING` blocks. However, the source code might not even be available if someone is compiling with Cython or Nuitka.
> 
> I propose that we keep the current behavior but come up with a new warning for this case so that the user is aware of how to fix it. Thus, instead of the deprecation warning something like:
> 
> **warning: ambiguous parsing result, parameter 'foo' annotated with forward referenced `Info` that is unresolvable from module `myFoo.py`. Ensure that the `Info` type is accessible from the module scope and not imported within an `is TYPE_CHECKING` block`**
> 
> This ensures that we don't have a false positive which could be quite annoying for the user to debug, wondering why their custom `Info` input type is not showing up in their API. We enforce resolvability for user-types otherwise we wouldn't be able to generate the schema types, so then why would we not enforce it for our own types as well?

Good suggestion, I'd say we do this: if the type of the param is `"Info"` we assume this is our info, and put a warning to ask users to move the import for info outside the `if TYPE_CHECKING` block

bellini666:
> Good suggestion, I'd say we do this: if the type of the param is "Info" we assume this is our info, and put a warning to ask users to move the import for info outside the if TYPE_CHECKING block

I agree with this as well!

patrick91:
I've pushed the initial fix, removing the check for string names for now at least, I want to ship this fix asap 😊

PR #2868, opened 2023-06-20
refactor: refactor the way we resolve annotations to better support future refs

devkral:
Currently only against schemes without `from __future__ import annotations` is tested.
This PR adds a simple test against a schema with `from __future__ import annotations`

## Why does it crash?

Because `from __future__ import annotations` causes bugs which are not tested against yet.

For example in strawberry/relay/fields.py it is assumed in line 228  that field.base_resolver.signature.return_annotation is an ForwardRef or the type.

This is not the case. It can be a string

I hope this PR improves the `from __future__ import annotations` compatibility


tests/relay/test_schema.py
```diff
@@ -25,6 +25,21 @@ def test_schema():
     assert schema_output == expected
 
 
+def test_schema_future_annotations():
+    from .schema_future_annotations import schema
+
+    schema_output = str(schema).strip("\n").strip(" ")
+    output = pathlib.Path(__file__).parent / "schema.gql"
```
bellini666, line 32: I think it is probably better to export it to a different path than the base one, or else any change done there will also need to be done in the future annotations file.

```suggestion
    output = pathlib.Path(__file__).parent / "future_annotations_schema.gql"
```

bellini666 (changes requested):
Just a minor change request in there.

Also, you test might have caught an issue an based on the fact that the tests broke

tests/relay/test_schema.py
devkral: the idea is to plainly copy schema.py and add `from __future__ import annotations` to it.
So it should match schema.gql

devkral: it may should be done more sophisticated

botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

bellini666:
@devkral thanks for pointing this out. I used this PR to refactor the way we resolve field types to properly support specialized generics (and other corner cases) when using forward refs.

@patrick91 can you take a look?

devkral:
@bellini666 thank you!

strawberry/field.py
```diff
@@ -307,45 +310,97 @@ def type_params(self) -> List[TypeVar]:
...
+        # If this is a generic field, try to resolve it using its origin's
+        # specialized type_var_map
+        if _is_generic(resolved):  # type: ignore
+            specialized_type_var_map = get_specialized_type_var_map(
+                cast(type, self.origin)
+            )
+            if specialized_type_var_map and isinstance(resolved, StrawberryType):
+                resolved = resolved.copy_with(specialized_type_var_map)
+            elif specialized_type_var_map and has_object_definition(resolved):
+                resolved = resolved.__strawberry_definition__.copy_with(
```
patrick91, line 353: can we add a test for this and the code below? 😊

patrick91 (commented):
did a quick review, I think it is important for this to have full coverage 😊

strawberry/field.py
bellini666: Oh, just noticed that this check is actually impossible to happen =P

The code below should actually be covered (except for the same check that it does here). Codecov says that no paths there are covered, but if I remove the code there a lot of tests break.

I'll try to refactor it to see if it gets fixed

tests/relay/test_schema.py
```diff
@@ -23,6 +23,19 @@ def test_schema():
     assert schema_output == expected
 
 
+def test_schema_future_annotations():
+    from .schema_future_annotations import schema
+
+    schema_output = str(schema).strip("\n").strip(" ")
+    output = pathlib.Path(__file__).parent / "schema_future_annotations.gql"
+    if not output.exists():
+        output.write_text(schema_output + "\n")
```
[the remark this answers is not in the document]
bellini666: Oh, didn't know about that one. Nice! :)

RELEASE.md
```diff
@@ -0,0 +1,7 @@
+Release type: patch
```
patrick91, line 1: shall we mark this as minor? 😊

pyproject.toml
```diff
@@ -142,7 +142,7 @@ tests/codegen/snapshots/
 '''
 
 [tool.pytest.ini_options]
-addopts = "-s --emoji --mypy-ini-file=mypy.ini --benchmark-disable"
+addopts = "--emoji --mypy-ini-file=mypy.ini --benchmark-disable"
```
patrick91, line 145: any reason why this was removed? 😊

strawberry/field.py
```diff
@@ -307,45 +310,94 @@ def type_params(self) -> List[TypeVar]:
...
+        type_definition: Optional[StrawberryObjectDefinition] = None,
+    ) -> Union[  # type: ignore [valid-type]
+        StrawberryType,
+        Type[WithStrawberryObjectDefinition],
+        Literal[UNRESOLVED],
+    ]:
+        # We return UNRESOLVED by default, which means this case will raise a
+        # MissingReturnAnnotationError exception in _check_field_annotations
+        resolved = UNRESOLVED
+        allow_generic = True
```
patrick91, line 325: do you think we can remove this allow generic?

We can maybe use a property? but my main issue with this is that I'm not sure how it is used

strawberry/schema/schema_converter.py
```diff
@@ -83,7 +87,56 @@
...
-    from strawberry.type import StrawberryType
+
+
+FieldType = TypeVar(
+    "FieldType", bound=Union[GraphQLField, GraphQLInputField], covariant=True
+)
+
+
+class FieldConverterProtocol(Generic[FieldType], Protocol):
+    def __call__(  # pragma: nocover
```
patrick91, line 98:
```suggestion
    def __call__(  # pragma: no cover
```

strawberry/types/type_resolver.py
```diff
@@ -138,23 +122,25 @@ class if one is not set by either using an explicit strawberry.field(name=...) o
...
+                and field.type_annotation.namespace is None
+            ):
+                field.type_annotation.set_namespace_from_field(field)
+
+            # If the field origin is not cls, copy it and set cls on it.
+            # This is important for us to be able to resolve specialized
+            # generic when generating the schema.
+            if field.origin != cls:
+                field = copy.copy(field)  # noqa: PLW2901
+                field.origin = cls
```
patrick91, line 136: do we need to do this all the time? shouldn't we check if it is generic?

strawberry/types/type_resolver.py
```diff
@@ -138,23 +122,25 @@ class if one is not set by either using an explicit strawberry.field(name=...) o
...
 
         # Create a StrawberryField for fields that didn't use strawberry.field
         else:
             # Only ignore Private fields that weren't defined using StrawberryFields
             if is_private(field.type):
                 continue
 
-            field_type = _resolve_specialized_type_var(cls, field.type)
-
             origin = origins.get(field.name, cls)
```
patrick91, line 144: is this not used anymore?

tests/types/cross_module_resolvers/test_cross_module_resolvers.py
```diff
@@ -73,12 +73,12 @@ def test_c_inheritance():
...
-    assert a_is_of_full_age.origin == a_mod.AObject
-    assert b_name.origin == b_mod.BBase
-    assert b_age.origin == b_mod.BObject
-    assert b_is_of_full_age.origin == b_mod.BObject
+    assert a_name.origin == c_mod.CInheritance
+    assert a_age.origin == c_mod.CInheritance
+    assert a_is_of_full_age.origin == c_mod.CInheritance
+    assert b_name.origin == c_mod.CInheritance
+    assert b_age.origin == c_mod.CInheritance
+    assert b_is_of_full_age.origin == c_mod.CInheritance
```
patrick91, line 81: is this a good change? :D

patrick91 (commented):
left some more comments 😊

pyproject.toml
bellini666: I'm an old school "print debugger" guy =P, and `-s` avoids tests being able to capture stdout/stderr to show those, and instead they get displayed while running the tests.

I'm not sure if there was an specific reason for having `-s` in here, but I thought it would be ok to remove. I'm fine with keeping it if you prefer :)

patrick91: no that's fine, I actually thought it was doing the opposite 🤣

strawberry/field.py
bellini666: Explaining what is happening here:

If you look at the previous code, `type_annotation.resolve()` got returned regardless, while `base_resolver.type` would force `UNRESOLVED` if it was generic.

I tried handling handling `base_resolver` both ways, as allowing generic to be returned and not, but those broke some tests and I did not get too deep to try to understand why, so I decided to keep the old behaviour in this case, while still trying to solve both when they are specialized.

I can probably add a TODO comment in here

RELEASE.md
bellini666: Hrmmm, although this is a fix the refactor is large enough to be considered a minor. Will change it :)

strawberry/relay/types.py
```diff
@@ -407,7 +408,16 @@ def resolve_id_attr(cls) -> str:
...
+                try:
+                    evaled = eval_type(
+                        ForwardRef(attr) if isinstance(attr, str) else attr,
+                        globalns=base_namespace,
+                    )
+                except TypeError:
+                    # Some ClassVar might raise TypeError when being resolved
+                    # on some python versions. This is fine to skip since
+                    # we are not interested in ClassVars here
+                    continue
```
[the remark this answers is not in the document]
bellini666: I actually tried to check for `ClassVar` using the same code that dataclasses use, but thought it was clenaer to `except TypeError`.

I'll add the check for `ClassVar` so you can tell me which one you prefer :)

strawberry/types/type_resolver.py
bellini666: I wanted to do that, but it all falls back to the same "forward ref" issues. There's no way to be sure that this is generic when using forward refs at this point, because we can't properly resolve the type until schema creation.

strawberry/field.py
patrick91: let's do that 😊

strawberry/types/type_resolver.py
patrick91: ah sorry, I meant the `origin` variable 😊 below you're using `cls` instead of that (also why? :D)

tests/types/cross_module_resolvers/test_cross_module_resolvers.py
bellini666: Your comments made me realize that I can use `type_definition` directly and I don't need the `origin` for this to work :)

This change will get reverted

strawberry/field.py
bellini666: Hrmm, interestingly enough once I refactored the code to not replace the `origin` anymore all tests are now fine D:

Removing the `allow_generic` as it is not required anymore.

strawberry/relay/types.py
```diff
@@ -408,17 +408,14 @@ def resolve_id_attr(cls) -> str:
...
-                try:
-                    evaled = eval_type(
-                        ForwardRef(attr) if isinstance(attr, str) else attr,
-                        globalns=base_namespace,
-                    )
-                except TypeError:
-                    # Some ClassVar might raise TypeError when being resolved
-                    # on some python versions. This is fine to skip since
-                    # we are not interested in ClassVars here
+                if is_classvar(base, attr):
```
patrick91, line 411: I'd retain the comment about this though :D

patrick91 (approved):
Looks great! I've left one not about a missing comment 😊

strawberry/relay/types.py
bellini666: Done! :)

devkral:
This looks really nice. Could strawberry-graphql-django pickup this and use it instead of its custom logic?

patrick91:
@devkral I think that's the goal!

patrick91:
This seems to have triggered this issue https://github.com/strawberry-graphql/strawberry-graphql-django/issues/276#issuecomment-1621724809 /cc @bellini666

Issue #2874, opened 2023-06-21
Codegen should be able to know the `__typename` for an object


mgilson:
## Feature Request Type

- [ ] Core functionality
- [x] Alteration (**backward compatible enhancement**/optimization) of existing feature(s)
- [ ] New behavior

## Description

The code generator doesn't (currently) know very much about the GraphQL types.  e.g. a query like:

```
query OperationName {
  __typename
  union {
    ... on Animal {
      age
    }
    ... on Person {
      name
    }
  }
}
```

Will generate classes like:

```py
class OperationNameResultUnionAnimal:
    # typename: Animal
    age: int

class OperationNameResultUnionPerson:
    # typename: Person
    name: str

OperationNameResultUnion = Union[OperationNameResultUnionAnimal, OperationNameResultUnionPerson]

class OperationNameResult:
    union: OperationNameResultUnion
```

In principle, the `__typename` field of the response should tell us whether we are going to get an `Animal` or a `Person`, but it would be very handy for client generators to be able to easily construct the following mapping:

```py
{"Animal": OperationNameResultUnionAnimal, "Person": OperationNameResultUnionPerson}
```

Currently, this is very hard to do with the information exposed to the client generator.  Let's plumb that information through so that clients can start to take advantage of it.

PR #2875, opened 2023-06-21
Plumb the GraphQL typename through to clients if known.

mgilson:
## Description

Issue #2874 does a pretty good job explaining what I'm trying to accomplish here, but the tl;dr is that GraphQL often makes response objects that look like:

```py
OperationNameResultUnion = Union[OperationNameResultUnionAnimal, OperationNameResultUnionPerson]

class OperationNameResult:
    union: OperationNameResultUnion
```

The query result JSON might have a `__typename` field that indicates whether the response is for an `Animal` or a `Person`, however, it isn't easy for a client to know whether `Animal` maps to `OperationNameResultUnionAnimal` or `OperationNameResultUnionPerson`.  (In principle, you _could_ look at the suffix in the class name and try to figure it out that way, but that feels flaky).

This change just adds the typename to the `GraphQLObjectType` that is passed to the client generator plugins.  The plugin can then choose to use this information or to disregard it.

The python plugin has been updated to include the typename in a comment in the class body -- this information might be useful for humans looking at the generated classes, but it is also a convenient way to test that the new functionality is working.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] **Enhancement**/optimization
- [ ] Documentation

* #2874 

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,5 @@
+Release type: minor
+
+This release gives codegen clients the ability to inquire about the `__typename`
+of a `GraphQLObjectType`.  This information can be used to automatically select
+the proper type to hydrate when working with a union type in the response.
--- a/strawberry/codegen/query_codegen.py
+++ b/strawberry/codegen/query_codegen.py
@@ -293,8 +293,11 @@ class QueryCodegen:
                 query_type, StrawberryObjectDefinition
             ), f"{fd.type_condition.name.value!r} is not a type in the graphql schema!"
 
+            typename = fd.type_condition.name.value
             graph_ql_object_type_factory = partial(
-                GraphQLFragmentType, on=fd.type_condition.name.value
+                GraphQLFragmentType,
+                on=typename,
+                graphql_typename=typename,
             )
 
             self._collect_types(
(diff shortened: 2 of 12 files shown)


patrick91 (approved):

RELEASE.md
```diff
@@ -0,0 +1,5 @@
+Release type: patch
```
```suggestion
Release type: minor
```

this is technically a new feature 😊

strawberry/codegen/query_codegen.py
```diff
@@ -751,7 +755,13 @@ def _collect_types_using_fragments(
         for fragment in fragments:
             fragment_class_name = class_name + fragment.type_condition.name.value
 
-            current_type = GraphQLObjectType(fragment_class_name, list(common_fields))
+            typename = None
+            with contextlib.suppress(AttributeError):
+                typename = fragment.type_condition.name.value
```
patrick91, line 760: can we change this to use get attribute? I think it is a bit more explicit of what's happening :)

patrick91 (approved):
I forgot to push the comments before approving :)

I think this looks good! We can merge after resolving my two nit picks 😊

strawberry/codegen/query_codegen.py
mgilson: It turns out that we don't need to worry about the `AttributeError` here -- we use the exact same `fragment.type_condition.name.value` a couple lines up (and then again later in the loop).  This makes sense ... I think that fragments _have_ to be `on` _something_.

I guess there's an [edge case](https://spec.graphql.org/October2021/#sec-Inline-Fragments) for inline fragments with `@include(if: ...)` directives but the current code doesn't handle those (and I'm not really sure how I would expect the codegen to handle an `@include(if: ...)` anyway...

In any case, I'll leave that handling for a future PR if that's OK with you 😅

patrick91: sure! thanks 😊

PR #2876, opened 2023-06-21
Support subscription errors without terminating

kristjanvalur:
`strawberry-transport-graphql` protocol now creates a `Next` message and doesn't terminate the operation
if a non-fatal error is generated during execution of a `subscribe` request.

## Description

The graphql `ExecutionResult` struct contained in an `Next` message contains both  `data` and `errors` members.
When a non-fatal error occurs when producing a result, such an `ExecutionResult` is produced by Appollo server,
a `Next` message is created, and the subscription continues.    This PR replicates that behaviour.

Single result operations (`query`, `mutate`) which cause errors during execution, continue to produce `Error` messages
as before, again, similar to Appollo.

## Note:

this PR builds on top of #2699, so we may consider merging that first, to reduce the footprint of the change.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

- [x]  I have tested the changes and verified that they work and don't break anything (as well as I can manage).

### DIFF
--- a/strawberry/channels/testing.py
+++ b/strawberry/channels/testing.py
@@ -140,9 +140,6 @@ class GraphQLWebsocketCommunicator(WebsocketCommunicator):
 
 def process_errors(errors: List[GraphQLFormattedError]) -> List[GraphQLError]:
     return [
-        GraphQLError(
-            message=error["message"],
-            extensions=error.get("extensions", None),
-        )
+        GraphQLError(str(error), original_error=error)
         for error in errors
     ]
--- a/strawberry/channels/testing.py
+++ b/strawberry/channels/testing.py
@@ -127,19 +127,17 @@ class GraphQLWebsocketCommunicator(WebsocketCommunicator):
                 payload = NextMessage(**response).payload
                 ret = ExecutionResult(payload["data"], None)
                 if "errors" in payload:
-                    ret.errors = process_errors(payload["errors"])
+                    ret.errors = self.process_errors(payload["errors"])
                 ret.extensions = payload.get("extensions", None)
                 yield ret
             elif message_type == ErrorMessage.type:
                 error_payload = ErrorMessage(**response).payload
-                yield ExecutionResult(data=None, errors=process_errors(error_payload))
+                yield ExecutionResult(
+                    data=None, errors=self.process_errors(error_payload)
+                )
                 return  # an error message is the last message for a subscription
             else:
                 return
 
-
-def process_errors(errors: List[GraphQLFormattedError]) -> List[GraphQLError]:
-    return [
-        GraphQLError(str(error), original_error=error)
-        for error in errors
-    ]
+    def process_errors(self, errors: List[GraphQLFormattedError]) -> List[GraphQLError]:
+        return [GraphQLError(str(error), original_error=error) for error in errors]
--- a/strawberry/channels/testing.py
+++ b/strawberry/channels/testing.py
@@ -14,7 +14,7 @@ from typing import (
     Union,
 )
 
-from graphql import GraphQLError
+from graphql import GraphQLError, GraphQLFormattedError
 
 from channels.testing.websocket import WebsocketCommunicator
 from strawberry.subscriptions import GRAPHQL_TRANSPORT_WS_PROTOCOL, GRAPHQL_WS_PROTOCOL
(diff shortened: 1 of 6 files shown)


botberry:



---

`graphql-transport-ws` protocol allows for subscriptions to yield errors, without terminating
the subscription.  Non-fatal errors produced by subscriptions now produce `Next` messages containing
an error payload and don't terminate the subscription.  This is in accordance to the behaviour of Apollo server.


---




Here's the tweet text: 
```

```

DoctorJohn:
> this PR builds on top of https://github.com/strawberry-graphql/strawberry/pull/2699

Thanks for the heads up! I just merged #2699, taking a look at this PR next. Looks like it could use a rebase, lemme know whether you have time for it or whether I should take care of it :+1:

kristjanvalur:
Thank's, I'll rebase presently.

strawberry/channels/testing.py
```diff
@@ -125,22 +125,24 @@ async def subscribe(
...
-                        for message in error_payload
-                    ],
-                )
+                yield ExecutionResult(data=None, errors=process_errors(error_payload))
                 return  # an error message is the last message for a subscription
             else:
                 return
+
+
+def process_errors(errors: List[GraphQLFormattedError]) -> List[GraphQLError]:
```
DoctorJohn, line 141: nit: Should we make this a member of the `GraphQLWebsocketCommunicator` class? Would make it easier to understand where it belongs to and makes it easier to override if needed.

strawberry/channels/testing.py
```diff
@@ -125,22 +125,24 @@ async def subscribe(
...
+    return [
+        GraphQLError(
+            message=error["message"],
+            extensions=error.get("extensions", None),
+        )
```
DoctorJohn, line 146: Not specific to this change since it has been like this before, but I notice we transform errors differently in other places in our code. For example:

https://github.com/strawberry-graphql/strawberry/blob/8f18beba507e56973a555967ca584e3b3db2ae36/strawberry/schema/execute.py#L109

https://github.com/strawberry-graphql/strawberry/blob/8f18beba507e56973a555967ca584e3b3db2ae36/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py#L314

Looking at the graphql-core source we can see this is also preserving extensions:

https://github.com/graphql-python/graphql-core/blob/0c93b8452eed38d4f800c7e71cf6f3f3758cd1c6/src/graphql/error/graphql_error.py#L169-L173

We might want to use this approach here as well :)


```suggestion
        GraphQLError(str(error), original_error=error)
```

DoctorJohn (changes requested):
Thanks for yet another nice PR @kristjanvalur 
Overall looks good to me, I left some minor comments.

strawberry/channels/testing.py
kristjanvalur: Sure, I wasn't paying too much attention to it since this is essentially testing code.  I'll make the change.

strawberry/channels/testing.py
kristjanvalur: Hm, this change breaks typing, I'll investigate.

kristjanvalur: Ah, this is a different case.  The code in the channels testing client is transforming the error for the _client_, whereas that other code you saw is creating GraphQLErrors on the server side.

I am frankly unsure why the channels testing client is doing this, since the error is already a GraphQLFormattedError.   Except as some sort of convenience for the testing framework.   I think this transformation should be dropped, in fact.

kristjanvalur: Oh, the reason is that it is yielding an "ExecutionResult".  
The thing is, ExecutionResult is specified as containing "GraphQLError".  
But this is weird, because, on the wire, it really should be GraphQLFormattedError.

I think this stems from a confusion between python types and graphql types.  You cannot in fact create an ExecutionResult from the NextMessage.  An ExecutionResult is a server internal format, whereas the NextMessage contains formatted data, including formatted errors...

It should be a `FormattedExecutionResult` type...

kristjanvalur:
I reverted the `process_errors` function to how it was before, how it constructed a `GraphQLError`.  The testing client is returning `ExecutionResult` objects but they are not available on the client after going over the protocol.  It could just return `FormattedExecutionResult` but those are dicts, not objects with attributes.  So I am leaving this as it was, which is a bit hacky, since the objects in question aren't the "real" error objects as raised server-side.

strawberry/channels/testing.py
DoctorJohn: Awesome, thanks

strawberry/channels/testing.py
DoctorJohn: Oh, interesting, I see. Thanks for the investigation and write up! We can adress this in another PR. This one is good to go now.

PR #2879, opened 2023-06-21
fix: fix a regression on StrawberryAnnotation.__eq__ method

bellini666:
Fix a regression when comparing a `StrawberryAnnotation` instance with
anything that is not also a `StrawberryAnnotation` instance, which
caused it to raise a `NotImplementedError`.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,10 @@
+Release type: patch
+
+This release fixes a regression when comparing a `StrawberryAnnotation`
+instance with anything that is not also a `StrawberryAnnotation` instance,
+which caused it to raise a `NotImplementedError`.
+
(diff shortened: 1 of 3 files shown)


botberry:



---

This release fixes a regression when comparing a `StrawberryAnnotation`
instance with anything that is not also a `StrawberryAnnotation` instance,
which caused it to raise a `NotImplementedError`.

This reverts its behavior back to how it worked before, where it returns
`NotImplemented` instead, meaning that the comparison can be delegated to
the type being compared against or return `False` in case it doesn't define
an `__eq__` method.


---




Here's the tweet text: 
```

```

patrick91 (approved):
💯

PR #2880, opened 2023-06-22
Add missing arguments to create_type

patrick91:
This PR adds missing arguments to `create_type` so people don't need to create
their own utility: https://github.com/dagger/dagger/blob/68cfddb53730cd3ec132bd115d6717013dd60062/sdk/python/src/dagger/server/_commands.py#L52-L58

### DIFF
--- a/strawberry/tools/create_type.py
+++ b/strawberry/tools/create_type.py
@@ -1,11 +1,19 @@
 import types
-from typing import List, Type
+from typing import List, Optional, Sequence, Type
 
 import strawberry
 from strawberry.field import StrawberryField
 
 
-def create_type(name: str, fields: List[StrawberryField]) -> Type:
+def create_type(
+    name: str,
+    fields: List[StrawberryField],
+    is_input: bool = False,
+    is_interface: bool = False,
+    description: Optional[str] = None,
+    directives: Optional[Sequence[object]] = (),
+    extend: bool = False,
+) -> Type:
     """Create a Strawberry type from a list of StrawberryFields
 
     >>> @strawberry.field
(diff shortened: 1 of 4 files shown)


botberry:



---

This release updates `create_type` to add support for all arguments
that `strawberry.type` supports. This includes: `description`, `extend`,
`directives`, `is_input` and `is_interface`.


---




Here's the tweet text: 
```

```

strawberry/tools/create_type.py
```diff
@@ -1,11 +1,19 @@
...
-def create_type(name: str, fields: List[StrawberryField]) -> Type:
+def create_type(
+    name: str,
+    fields: List[StrawberryField],
+    is_input: bool = False,
+    is_interface: bool = False,
+    description: Optional[str] = None,
+    directives: Optional[Sequence[object]] = (),
+    extend: bool = False,
+) -> Type:
```
bellini666, line 16: Maybe return `Type[WithStrawberryObjectDefinition]` in here since we have that now?

tests/tools/test_create_type.py
```diff
@@ -3,16 +3,93 @@
...
-def test_create_decorator_type():
+def test_create_type():
     @strawberry.field
     def name() -> str:
         return "foo"
 
-    MyType = create_type("MyType", [name])
-    definition = MyType.__strawberry_definition__
+    MyType = create_type("MyType", [name], description="This is a description")
+    definition = get_object_definition(MyType, strict=True)
```
bellini666, line 18: Then you can do this as well:

```python
from typing_extensions import assert_type

assert_type(MyType, WithStrawberryObjectDefinition)
```

tests/tools/test_create_type.py
```diff
@@ -28,7 +105,7 @@ def get_name() -> str:
     name = strawberry.field(name="name", resolver=get_name)
 
     MyType = create_type("MyType", [name])
-    definition = MyType.__strawberry_definition__
+    definition = get_object_definition(MyType, strict=True)
```
bellini666, line 108: With the suggestion this could be done directly without requiring the `get_object_definition` and type-safe

bellini666 (approved):
Looks good! :)

A minor suggestion, but feel free to ignore if you prefer to keep it the way it is

tests/tools/test_create_type.py
patrick91: I think I prefer using get_object_definition 🍡

Issue #2181, opened 2022-09-19
First class support for Pydantic


patrick91:
Hi folks :) I've been thinking that maybe at some point we should add **first class** support for Pydantic.

## What does **first class** support mean?

Currently we support Pydantic via an, albeit experimental, integration, but the API is a cumbersome, as we need to create two classes, one for Pydantic and one for the Strawberry type.

Example:

```python
class User(BaseModel):
    id: int
    name: str
    signup_ts: Optional[datetime] = None
    friends: List[int] = []

@strawberry.experimental.pydantic.type(model=User)
class UserType:
    id: strawberry.auto
    name: strawberry.auto
    friends: strawberry.auto
```

This great if you already have Pydantic models and want to add an API on top of them, and I still think this API makes a lot of sense for converting database models to a GraphQL API.

With **first class** support I mean having a way to define both the Strawberry type and the Pydantic model at once, like this:

```python
@strawberry.type
class User(BaseModel):
    id: int
    name: str
    signup_ts: Optional[datetime] = None
    friends: List[int] = []
```

An API like this should make it much easier to define types that need to be validated (input types might need this often)

I don't know when we'll have time to work on this, I'd definitely work on improving the code for the integrations (making sure we reduce the code duplication we have now) first and we should also finalise the internal API too.

There's also question on how we can make sure we type this correctly, as we currently use a `dataclass` transform decorator to tell pyright what happens to the class. Maybe we introduce a new decorator just for pydantic?

gazorby:
Hi @patrick91 

This is a great idea!

I implemented something similar in [gazorby/strawberry-sqlalchemy-mapper](https://github.com/gazorby/strawberry-sqlalchemy-mapper) , but for strawberry inputs.

The idea is to automatically handle data validation of query/mutation inputs and return any error raised by pydantic in dedicated strawberry error types.

Basically I needed:
- A custom pydantic `BaseModel` that allow postponed validation so we can trigger when a resolver is called rather than during model instantiation
- A decorator that hook `strawberry.field` to validate pydantic based inputs before passing them to the resolver


If ported to strawberry, usage would look like this:

```python
  import strawberry
  from pydantic import BaseModel, validator
  
  # The input decorator actually return a new UserCreate type with the custom BaseModel (with postponed validation) as base
  @strawberry.input
  class UserCreate(BaseModel):
      username: str
      age: int
  
      @validator("age", check_fields=False)
      def check_age(cls, v):
          if v < 18:
[...]

PR #2888, opened 2023-06-25
WIP: Pydantic v2 compat

thejaminator:
This PR provides basic functionality for pydantic v2. I do not recommend people to actually use it, but would like to merge this in first so that others can get work started on it.

This is done by creating a new directory inside strawberry, that is a sister subdirectory to pydantic v1.
This does duplicate a lot of code, but the upside is that
- this avoids a lot of `if pydantic version == 1` logic
- something in pydantic v1 does always have an equivalent in pydantic v2. E.g. the ModelInfo class in pydantic v1 does not have the field name in pydantic v2. In pydantic v2 its a dict instead. So if the code is shared, the v1 code needs to be changed to add this additional information, even though its not needed. so that becomes quite confusing.
- pydantic v2 has some different resolution logic for types than v1
- `if pydantic version == 1` at run time messes up your IDE and type checkers.  (or at least it did for me)

FastAPI did take the other approach [of having a compat module](https://github.com/tiangolo/fastapi/blob/master/fastapi/_compat.py), and a bunch of  `if pydantic version == 1` logic scattered around tests and files. I'm not sure which of these ways are better. Having a shared module like them could have been better if we really are commiting to continue to support pydantic v1 when we introduce new fixes. But I do think that it would slow us down in the sense of developing something new on pydantic v2 (because we are scared to break pydantic v1 stuff)

For tests, i've only added some of them so far. I'll add the rest of them in from pydantic v1 over the next few MRs. Not all the tests have the same behavior between pydantic v1 and v2.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist


strawberry/experimental/pydantic2/object_type.py
```diff
@@ -0,0 +1,343 @@
...
+    name: Optional[str] = None,
+    is_input: bool = False,
+    is_interface: bool = False,
+    description: Optional[str] = None,
+    directives: Optional[Sequence[object]] = (),
+    all_fields: bool = False,
+    use_pydantic_alias: bool = True,
+) -> Callable[..., Type[StrawberryTypeFromPydantic[PydanticModel]]]:
+    def wrap(cls: Any) -> Type[StrawberryTypeFromPydantic[PydanticModel]]:
+        model_fields: Dict[str, FieldInfo] = model.model_fields
```
thejaminator, line 136: in pydantic v1 this was ModelFields, but its now FieldInfo in pydantic v2. The `FieldInfo` also does not contain the name of the field itself, so now its a dit.

botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

tests/experimental/pydantic2/schema/test_basic.py
```diff
@@ -0,0 +1,429 @@
...
+
+import strawberry
+
+
+def test_all_fields():
+    class UserModel(pydantic.BaseModel):
+        age: int
+        password: Optional[str]
+
+    @strawberry.experimental.pydantic2.type(UserModel, all_fields=True)
```
thejaminator, line 15: for now its just `pydantic2`, for the lack of a better name

strawberry/experimental/pydantic2/utils.py
```diff
@@ -0,0 +1,144 @@
...
+        return default_factory
+
+    # if we have a default, we should return it
+
+    if has_default:
+        return lambda: smart_deepcopy(default)
+
+    # Note that unlike pydantic v1, pydantic v2 does not add a default of None when
+    # the field is Optional[something]
+    # so there is no need to handle that case here
```
thejaminator, line 128: in v1 there was some logic of adding a default of `None` when its Optional. pydantic v2 doesn't do that anymore

noxfile.py
```diff
@@ -23,10 +23,10 @@ def tests(session: Session) -> None:
         "not django",
         "-m",
         "not starlite",
-        "-m",
-        "not pydantic",
         "--ignore=tests/mypy",
         "--ignore=tests/pyright",
+        "--ignore=tests/experimental/pydantic",
+        "--ignore=tests/experimental/pydantic2",
     )
```
thejaminator, line 30: otherwise pytest on the pydantic2 folder would go boom. since it tries to import stuff directly from pydantic v2.

strawberry/experimental/__init__.py
```diff
@@ -1,6 +1,14 @@
...
-else:
-    __all__ = ["pydantic"]
+try:
+    from . import pydantic2
+
+    # Support for pydantic2 is highly experimental and the interface will change
+    # We don't recommend using it yet
+    __all__ = ["pydantic2"]
+except ImportError as e:
+    pass
```
thejaminator, line 14: explicitly a separate module

ppease:
@thejaminator I was just curious if we had any idea on timeline for when we would feel comfortable with someone using pydantic v2 in a production type environment?  I know your comment above mentioned that this MR is just adding initial support, and we wouldn't recommend people using it.  

Not sure this is the right place to ask, but I was also wondering if there has been any talk of removing the experimental tag from pydantic in general?  Our team has found a ton of value out of using the pydantic integration with strawberry.  For better or worse we've been using the existing experimental pydantic support in our production environments for over a year without any issues.

patrick91:
> @thejaminator I was just curious if we had any idea on timeline for when we would feel comfortable with someone using pydantic v2 in a production type environment? I know your comment above mentioned that this MR is just adding initial support, and we wouldn't recommend people using it.

I’m going to chat with James next week about this :) 

> 
> Not sure this is the right place to ask, but I was also wondering if there has been any talk of removing the experimental tag from pydantic in general? Our team has found a ton of value out of using the pydantic integration with strawberry. For better or worse we've been using the existing experimental pydantic support in our production environments for over a year without any issues.

That’s good to know, I think we could do that, but we’d need to make sure we have enough bandwidth to fix issues once this is marked as non experimental

there’s also an idea of doing first class support for pydantic in future, so maybe this integration could not be that useful anymore? 

Also @ppease do you think your company could be interested in sponsoring us to work a bit on this? that’d be very helpful ☺️

ppease:
Thanks for the quick response.  Is this the right issue to follow to get updates on work being done with pydantic v2? Also I'd be willing to try an early version of stuff and report any bugs if that would help you all out.

> there’s also an idea of doing first class support for pydantic in future, so maybe this integration could not be that useful anymore?

I'm assuming [this issue ](https://github.com/strawberry-graphql/strawberry/issues/2181) is what you mean by first class support?  If so I do think first class support would help us because that is one of the pain points right now with using the experimental support is having to create duplicate types.  There will probably still be cases where we end up creating duplicate classes because we want to expose different items at the graphql layer vs. our data model layer so maybe the experimental support would still be useful for those use cases?  I could also see different ways of doing that with a more first class built in way as well.

> Also @ppease do you think your company could be interested in sponsoring us to work a bit on this? that’d be very helpful

We are not too big of a company so I can't make any promises, but I'll bring it up.  At a minimum we might be able to commit some time from myself or another dev to try and help out as well.

thejaminator:
Making the MR here isntead (did a compat file instead of a different folder)
https://github.com/strawberry-graphql/strawberry/pull/2972

patrick91:
@ppease can you try this release? `strawberry-graphql==0.196.0.dev.1690222024`

ppease:
@patrick91 I just saw this.  I probably won't have time this week, and I'm OOO next week.  However, when I get back, I should be able to prioritize trying this out.

cvarner-shippo:
Hi there, I'm currently running `strawberry-graphql==0.196.0.dev.1690222024`. Is the following expected behavior?

```python3
@strawberry.experimental.pydantic.type(model=PydanticOne, all_fields=True)
class One:
    @strawberry.field
    def two(self: 'One') -> 'Two':
        return Two()

@strawberry.experimental.pydantic.type(model=PydanticTwo, all_fields=True)
class Two:
    @strawberry.field
    def one(self: 'Two') -> 'One':
        return One()
```
Type hints work as expected, but at runtime whichever type is defined second fails to resolve the forward ref on the strawberry.field. 
```
strawberry.exceptions.unresolved_field_type.UnresolvedFieldTypeError: Could not resolve the type of 'one'. Check that the class is accessible from the global module scope.
```
Side note, this error message was difficult to debug as nowhere in the stack trace does it specify the filename or class the unresolved field belongs to.

I can use `LazyType` to make this work but it mucks up type hints and feels unnecessary (considering the type hints were working before).

```python3
@strawberry.experimental.pydantic.type(model=PydanticOne, all_fields=True)
class Two:
    @strawberry.field
    def one(self: 'Two') -> LazyType['One', __name__]:
        return One()
```

ppease:
@patrick91 Sorry for taking a while to respond to this.  I tried the latest version of strawberry, 0.208.1, with my graphql project, and I was able to generate a schema that matched my existing schema aside from some stuff moving around which was related to alphabetical order.  I did have to comment out two fields.  One that was using conint, and another one that was using EmailStr.  It looks like inside of strawberry/experimental/pydantic/fields.py we already have a FIELDS_MAP to map from pydantic specific types to the base type.  If both of those fields are fully supported in pydantic v2, could I just add those to the pydantic v2 field map?

Another data point is I ran all of our unit tests for our graphql code and those passed as well so this seems like a really promising start!  I plan on doing more testing in our dev environment this week, and I'll report back any issues I find.

patrick91:
@ppease feel free to open issues or send PRs 😊

PR #2897, opened 2023-06-27
Add support for setting cookies in TemporalResponse

### LINKED ISSUE #2832: Sanic get_context returning a TemporalResponse
## Describe the Bug

Ref.: https://strawberry.rocks/docs/integrations/sanic#get_context

I am trying to set cookies in the response as discussed here:
https://strawberry.rocks/docs/integrations/asgi#setting-response-headers

```python
class MyGraphQLView(GraphQLView):
    async def get_context(self, request: Request, response: Response) -> Any:
        return {"request": request, "response": response}
```

and I am getting an error because I am getting a TemporalResponse instead of a Sanic response object.

I have tried to hack it this way but it doesn't work:

```python
class SanicGraphQLView(GraphQLView):
    async def get_context(self, request: Request, response: TemporalResponse) -> Any:
        return {"request": request, "response": self.create_response(GraphQLHTTPResponse(), response)}
```

## System Information

 - Operating system: linux/amd64 python3.11
 - Strawberry version (if applicable): 0.183.5

cdroege:
WIP

This PR adds support for setting cookies in integrations that use a `TemporalResponse` in their context, e.g.:

* Sanic
* Chalice
* Channels

## Description

For Chalice and Sanic, not many changes were needed.

For Channels, I had to change the type of the headers field in `ChannelsResponse` to a list of tuples, so that it is possible to set multiple cookies in one response. Channels itself supports a list of tuples as headers in the `send_response` method automatically.

This also adds a breaking change to the `TemporalResponse` class. Header values can be a list now.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* Fixes #2832

## Checklist

### DIFF
--- a/strawberry/http/temporal_response.py
+++ b/strawberry/http/temporal_response.py
@@ -1,8 +1,51 @@
+import datetime
 from dataclasses import dataclass, field
-from typing import Dict
+from typing import Dict, List, Optional, Union
+
+from .cookie import SameSite, generate_cookie_header_value
 
 
 @dataclass
 class TemporalResponse:
     status_code: int = 200
-    headers: Dict[str, str] = field(default_factory=dict)
+    headers: Dict[str, Union[str, List[str]]] = field(default_factory=dict)
+
+    def set_cookie(
+        self,
+        key: str,
+        value: str,
+        max_age: Optional[int] = None,
+        expires: Optional[datetime.datetime] = None,
+        path: str = "/",
+        domain: Optional[str] = None,
+        secure: bool = True,
+        httponly: bool = False,
+        samesite: Optional[SameSite] = None,
+    ) -> None:
+        header_value = generate_cookie_header_value(
+            key,
+            value,
+            max_age=max_age,
+            expires=expires,
+            path=path,
+            domain=domain,
+            secure=secure,
+            httponly=httponly,
+            samesite=samesite,
+        )
+
+        # It is possible to have multiple Set-Cookie headers
+        # so we need to support multiple values per key in the
+        # self.headers dictionary for this header.
+        if "Set-Cookie" in self.headers:
+            if isinstance(self.headers["Set-Cookie"], list):
+                self.headers["Set-Cookie"].append(header_value)
+            else:
+                self.headers["Set-Cookie"] = [self.headers["Set-Cookie"], header_value]
+        else:
+            self.headers["Set-Cookie"] = header_value
+
+    def delete_cookie(
+        self, key: str, path: str = "/", domain: Union[None, str] = None
+    ) -> None:
+        self.set_cookie(key, "", path=path, max_age=0, domain=domain)
--- a/strawberry/chalice/views.py
+++ b/strawberry/chalice/views.py
@@ -111,7 +111,9 @@ class GraphQLView(
         return Response(
             body=self.encode_json(response_data),
             status_code=status_code,
-            headers=sub_response.headers,
+            # The type definition of the headers field of the Response class is wrong,
+            # see https://github.com/aws/chalice/pull/2050
+            headers=sub_response.headers,  # type: ignore
         )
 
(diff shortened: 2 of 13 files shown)


botberry:



---

This release adds a `set_cookie` method to the `TemporalResponse` class so that the integrations Sanic, Django Channels and Chalice have a method for setting cookies.

To support multiple cookies per response, the `headers` dict in the `TemporalResponse` class was also changed. The values of the dict can (optionally) be a list of strings.


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Christian Dröge for the PR 👏

```

cdroege:
@patrick91 If you have some time, can you have a look at this? But it is not really urgent

RELEASE.md
```diff
@@ -0,0 +1,5 @@
+Release type: patch
+
+This release adds a `set_cookie` method to the `TemporalResponse` class so that the integrations Sanic, Django Channels and Chalice have a method for setting cookies.
```
patrick91, line 3: this is only for integration that don't support this already (via their response), right?

I wonder if we should remove our temporal response and use the standard response from all the integration, what do you think?

RELEASE.md
```diff
@@ -0,0 +1,5 @@
+Release type: patch
```
```suggestion
Release type: minor
```

RELEASE.md
patrick91: I'm mostly thinking about consistency 😊

strawberry/http/temporal_response.py
```diff
@@ -1,8 +1,88 @@
...
+        # be present)
+        if "Set-Cookie" in self.headers:
+            if isinstance(self.headers["Set-Cookie"], list):
+                self.headers["Set-Cookie"].append(header_value)
+            else:
+                self.headers["Set-Cookie"] = [self.headers["Set-Cookie"], header_value]
+        else:
+            self.headers["Set-Cookie"] = header_value
+
+    def delete_cookie(
```
patrick91, line 85: I'd prefer calling this remove 😊

RELEASE.md
cdroege: I would really like removing the `TemporalResponse` class and I looked at the integrations:

1. Sanic has a [`Response`](https://github.com/sanic-org/sanic/blob/main/sanic/response/types.py#L244) class that supports adding / removing cookies, so using that is no problem.
2. Chalice has a [`Response`](https://github.com/aws/chalice/blob/79838b02dc330cfe549823899d2662ded0538015/chalice/app.py#L468) class but that doesn't support cookies. [Tests of that framework just set cookies via headers directly](https://github.com/aws/chalice/blob/79838b02dc330cfe549823899d2662ded0538015/tests/unit/test_local.py#L176).
3. Django Channels has no concept of a `Response` class. A consumer has to implement the [`handle`](https://github.com/django/channels/blob/2db92ccb57ff75b8f7535c3bc818f47d43a1b8bf/channels/generic/http.py#L55) method. The Strawberry implementation of that method uses its own [`ChannelsResponse`](https://github.com/strawberry-graphql/strawberry/blob/5503591baa28c5cea5ee2303f3fa00ca4a30f693/strawberry/channels/handlers/http_handler.py#L173) class.

So maybe this is a good idea moving forward:

1. Sanic: Use its own `Response` class.
2. Channels: Use Strawberry's `ChannelsResponse` and extend it with a `set_cookie` method. For that, I would extract the `set_cookie` method in this PR into a reuseable function.
3. Chalice: Use its `Response` class and create a small helper function (something like `set_cookie(response, value, ...)`) that adds the correct header to the `response` with the help of the function from 2.

What do you think? I can implement that. If you are OK with this idea, I would create a new PR for point 1 and reuse this PR for point 2. At a later time (after the PR for point 2 is merged), I would create a new PR for point 3.

PR #2901, opened 2023-06-28
Fix unresolved arguments using typing.Annotated

### LINKED ISSUE #2855: Annotated not working properly when using `Info` and future annotations?
This is a bit of a weird one, the following code snippet

```python
import strawberry
import uuid
from typing_extensions import Annotated
from strawberry.types import Info


@strawberry.type
class Query:
    @strawberry.field
    def get_testing(
        self,
        info: Info[None, None],
        id_: Annotated[uuid.UUID, strawberry.argument(name="id")],
    ) -> str | None:
        return None


print(schema)
```

prints this:

```graphql
type Query {
  getTesting(id_: UUID!): String # ⬅️ see `id_`
}

scalar UUID
```

but if you remove info, it prints correctly:

```graphql
type Query {
  getTesting(id: UUID!): String 
}

scalar UUID
```

skilkis:
Reserved argument parsing, which interally also resolves annotations, exits early after detecting a reserved argument like an ``info`` argumnent. As a result, any subsequent annotations are not reserved arguments are unresolved. This results in `StrawberryArgument` not being able to determine that an argument annotation makes use of `typing.Annotated`.

This behavior is fixed by by ensuring that `StrawberryArgument` makes use of the new `StrawberryAnnotation.evaluate` method instead of consuming the raw annotation from it.

An added benefit of this fix is that by moving annotation resolving code from `StrawberryResolver` to `StrawberryAnnotation`, it makes it possible for `StrawberryArgument` and `StrawberryResolver` to share the same type evaluation cache.

Summary of Changes:
- Move type evaluation and caching logic to `StrawberryAnnotation`
- Separate type evalution from resolution with the new `StrawberryAnnotation.evaluate` method. This allows `StrawberryArgument` to access the evaluated annotation before it gets converted to `StraberryType` specializations
- Remove type evaluation cache from `StrawberryResolver` and move it to `StrawberryAnnotation`.
- Fix typing issues in `resolver.py`

Closes #2855

## Types of Changes

- [x] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #2855

## Checklist

### DIFF
--- a/strawberry/annotation.py
+++ b/strawberry/annotation.py
@@ -10,6 +10,7 @@ from typing import (
     Any,
     Dict,
     ForwardRef,
+    List,
     Optional,
     Type,
     TypeVar,
@@ -31,11 +32,7 @@ from strawberry.type import (
 )
 from strawberry.types.types import StrawberryObjectDefinition
 from strawberry.unset import UNSET
-from strawberry.utils.typing import (
-    eval_type,
-    is_generic,
-    is_type_var,
-)
+from strawberry.utils.typing import eval_type, is_generic, is_type_var
 
 if TYPE_CHECKING:
     from strawberry.field import StrawberryField
@@ -55,11 +52,18 @@ ASYNC_TYPES = (
 
 
 class StrawberryAnnotation:
+    __slots__ = "annotation", "namespace", "__eval_cache__"
+
     def __init__(
-        self, annotation: Union[object, str], *, namespace: Optional[Dict] = None
+        self,
+        annotation: Union[object, str],
+        *,
+        namespace: Optional[Dict[str, Any]] = None,
     ):
         self.annotation = annotation
         self.namespace = namespace
+
+        self.__eval_cache__: Optional[Type[Any]] = None
 
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, StrawberryAnnotation):
@@ -72,7 +76,7 @@ class StrawberryAnnotation:
 
     @staticmethod
     def from_annotation(
-        annotation: object, namespace: Optional[Dict] = None
+        annotation: object, namespace: Optional[Dict[str, Any]] = None
     ) -> Optional[StrawberryAnnotation]:
         if annotation is None:
             return None
@@ -81,18 +85,29 @@ class StrawberryAnnotation:
             return StrawberryAnnotation(annotation, namespace=namespace)
         return annotation
 
-    def resolve(self) -> Union[StrawberryType, type]:
+    def evaluate(self) -> type:
+        """Return evaluated annotation using `strawberry.util.typing.eval_type`."""
+        evaled_type = self.__eval_cache__
+        if evaled_type:
+            return evaled_type
+
         annotation = self.annotation
         if isinstance(annotation, str):
             annotation = ForwardRef(annotation)
 
-        args = []
-
         evaled_type = eval_type(annotation, self.namespace, None)
+
+        self.__eval_cache__ = evaled_type
+        return evaled_type
+
+    def resolve(self) -> Union[StrawberryType, type]:
+        """Return resolved (transformed) annotation."""
+        evaled_type = cast(Any, self.evaluate())
 
         if is_private(evaled_type):
             return evaled_type
 
+        args: List[Any] = []
         if get_origin(evaled_type) is Annotated:
             evaled_type, *args = get_args(evaled_type)
 
@@ -102,7 +117,7 @@ class StrawberryAnnotation:
             return evaled_type
 
         if self._is_generic(evaled_type):
-            if any(is_type_var(type_) for type_ in evaled_type.__args__):
+            if any(is_type_var(type_) for type_ in get_args(evaled_type)):
                 return evaled_type
             return self.create_concrete_type(evaled_type)
 
@@ -131,6 +146,8 @@ class StrawberryAnnotation:
         module = sys.modules[field.origin.__module__]
         self.namespace = module.__dict__
 
+        self.__eval_cache__ = None  # Invalidate cache to allow re-evaluation
+
     def create_concrete_type(self, evaled_type: type) -> type:
         if has_object_definition(evaled_type):
             return evaled_type.__strawberry_definition__.resolve_generic(evaled_type)
@@ -143,15 +160,16 @@ class StrawberryAnnotation:
             raise NotAStrawberryEnumError(evaled_type)
 
     def create_list(self, evaled_type: Any) -> StrawberryList:
+        item_type, *_ = get_args(evaled_type)
         of_type = StrawberryAnnotation(
-            annotation=evaled_type.__args__[0],
+            annotation=item_type,
             namespace=self.namespace,
         ).resolve()
 
         return StrawberryList(of_type)
 
     def create_optional(self, evaled_type: Any) -> StrawberryOptional:
-        types = evaled_type.__args__
+        types = get_args(evaled_type)
         non_optional_types = tuple(
             filter(
                 lambda x: x is not type(None) and x is not type(UNSET),
@@ -184,7 +202,7 @@ class StrawberryAnnotation:
         if isinstance(evaled_type, StrawberryUnion):
             return evaled_type
 
-        types = evaled_type.__args__
+        types = get_args(evaled_type)
 
         union = StrawberryUnion(
             type_annotations=tuple(StrawberryAnnotation(type_) for type_ in types),
@@ -240,7 +258,7 @@ class StrawberryAnnotation:
         if not cls._is_union(annotation):
             return False
 
-        types = annotation.__args__
+        types = get_args(annotation)
 
         # A Union to be optional needs to have at least one None type
         return any(x is type(None) for x in types)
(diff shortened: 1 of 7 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

strawberry/types/fields/resolver.py
```diff
@@ -190,6 +187,18 @@ def __call__(self, *args: str, **kwargs: Any) -> T:
     def signature(self) -> inspect.Signature:
         return Signature.from_callable(self._unbound_wrapped_func, follow_wrapped=True)
 
+    # TODO: find better name
```
skilkis, line 190: Not sure about this name, seems a bit silly to have `annotations` and `strawberry_annotations` even though `strawberry_annotations` includes reserved parameters as well unlike `annotations`. Could we consider removing the `annotations` property below @patrick91 ? Any ideas welcome!

strawberry/annotation.py
```diff
@@ -249,7 +267,7 @@ def _is_optional(cls, annotation: Any) -> bool:
     def _is_list(cls, annotation: Any) -> bool:
         """Returns True if annotation is a List"""
 
-        annotation_origin = getattr(annotation, "__origin__", None)
+        annotation_origin = get_origin(annotation)
```
skilkis, line 270: Also @patrick91 was there a reason such as performance to get these attributes from the annotations directly instead of using `get_origin`, `get_args` ? If so I can revert these changes quickly

strawberry/types/fields/resolver.py
```diff
@@ -94,27 +100,23 @@ class ReservedType(NamedTuple):
     """
 
     name: str
-    type: Type
+    type: Type[Any]
```
skilkis, line 103: For consistency I'll also type these as `type` instead of `Type[Any]`

strawberry/annotation.py
```diff
@@ -55,12 +52,19 @@
 
 
 class StrawberryAnnotation:
+    __slots__ = "annotation", "namespace", "__eval_cache__"
```
patrick91, line 55: 🎰

strawberry/annotation.py
```diff
@@ -81,18 +85,29 @@ def from_annotation(
...
         annotation = self.annotation
         if isinstance(annotation, str):
             annotation = ForwardRef(annotation)
 
-        args = []
-
         evaled_type = eval_type(annotation, self.namespace, None)
 
+        self.__eval_cache__ = evaled_type
+        return evaled_type
```
patrick91, line 101: couldn't we use cached property here?

strawberry/annotation.py
```diff
@@ -131,6 +146,8 @@ def set_namespace_from_field(self, field: StrawberryField) -> None:
         module = sys.modules[field.origin.__module__]
         self.namespace = module.__dict__
 
+        self.__eval_cache__ = None  # Invalidate cache to allow re-evaluation
```
patrick91, line 149: why is this needed?

strawberry/annotation.py
patrick91: no, I don't think there was a particular reason 😊

strawberry/arguments.py
```diff
@@ -94,48 +89,46 @@ def __init__(
             _deprecated_UNSET if default is inspect.Parameter.empty else default
         )
 
-        if self._annotation_is_annotated(type_annotation):
-            self._parse_annotated()
+        try:
+            evaled_type = type_annotation.evaluate()
+        except NameError:
+            pass
```
patrick91, line 95: can we add a comment on when this might happen? (I'm assuming when using if TYPE_CHECKING)

strawberry/types/fields/resolver.py
patrick91: I wish it was the default :D

strawberry/types/fields/resolver.py
```diff
@@ -68,17 +68,23 @@ class ReservedName(NamedTuple):
     name: str
 
     def find(
-        self, parameters: Tuple[inspect.Parameter, ...], _: StrawberryResolver
+        self,
+        parameters: Tuple[inspect.Parameter, ...],
+        resolver: StrawberryResolver[Any],
     ) -> Optional[inspect.Parameter]:
+        del resolver
```
patrick91, line 75: why? :D

strawberry/types/fields/resolver.py
patrick91: Could we just use annotations?

patrick91 (commented):
left some comments!

patrick91 (commented):
left some comments!

strawberry/annotation.py
bellini666: I think it is because `get_origin` and `get_args` were introduced on python 3.8, and probably weren't available on `typing_extensions` when this code was written.

strawberry/arguments.py
bellini666: Also, try/except with pass can be rewritten as:

```python
with contextlib.suppress(NameError):
    evaled_type = type_annotation.evaluate()
```

strawberry/types/fields/resolver.py
```diff
@@ -190,6 +187,18 @@ def __call__(self, *args: str, **kwargs: Any) -> T:
...
 
+    # TODO: find better name
+    @cached_property
+    def strawberry_annotations(
+        self,
+    ) -> Dict[inspect.Parameter, Union[StrawberryAnnotation, None]]:
+        return {
+            p: StrawberryAnnotation(p.annotation, namespace=self._namespace)
+            if p.annotation is not inspect.Signature.empty
+            else None
```
bellini666, line 198: A small nit here, otherwise the alignment makes it harder to read

```suggestion
            p: (
                StrawberryAnnotation(p.annotation, namespace=self._namespace)
                if p.annotation is not inspect.Signature.empty
                else None
            )
```

tests/schema/test_arguments.py
```diff
@@ -199,3 +200,40 @@ def hello(
...
+    }
+
+    scalar UUID
+    """
+
+    schema_a = strawberry.Schema(type_a.Query)
+    schema_b = strawberry.Schema(type_b.Query)
+
+    assert str(schema_a) == str(schema_b)
+    assert f"\n{schema_a}\n" == textwrap.dedent(expected)
```
bellini666, line 239: nit

```suggestion
    assert str(schema_a) == textwrap.dedent(expected).strip()
```

bellini666 (changes requested):

strawberry/types/fields/resolver.py
skilkis: Do you mean just get rid of this and use the annotations property below?

strawberry/annotation.py
skilkis: To keep it consistent, probably there isn't a use-case for it. But if the
field somehow changes and it lives in a different module, then the previously
evaluated type could be incorrect

strawberry/annotation.py
skilkis: We'd have to remove the `__slots__` declaration above which is something I've
been wanting to introduce to Strawberry classes for a while. Especially `Info`
as it would reduce memory usage

skilkis: We can use cached property if you want to 😊

strawberry/types/fields/resolver.py
skilkis: To satisfy Pyright's hatred of unused arguments 😂 It's the suggested way by PyLint to
fix unused argument warnings / errors. Can't use `_` here as it violates Liskov
since the signature of find would be different 😥

tests/schema/test_arguments.py
skilkis: Thanks! Much better indeed 🚀

strawberry/arguments.py
skilkis: Good idea! I've added one now

strawberry/annotation.py
patrick91: oh, I didn't realise __slots__ don't make cachedproperty work! that's fine then

strawberry/types/fields/resolver.py
patrick91: yeah, I was suggesting that 😊 do you think it is worth doing it?

bellini666 (approved):
No more comments, LGTM! :)

skilkis:
> No more comments, LGTM! :)

Thanks @bellini666 !

strawberry/types/fields/resolver.py
skilkis: @patrick91 I don't think it's worth it to go back to how it was implemented
previously with the resolved annotations being injected by `ReservedType`, it's a nicer flow to create the `StrawberryAnnotation` instances
first. Maybe we can consider replacing the current `annotation` property below?
It would be a breaking change though.

patrick91: let's do that in future, if we remember ;)

patrick91 (approved):

PR #2907, opened 2023-06-30
chore: remove eol python 3.7

devkral:
Remove python 3.7 and add python 3.11 to remaining test matrices 

This PR also updates the testrunner docker image to python 3.8 and fully qualifies the source

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

- [x] I have tested the changes and verified that they work and don't break anything (as well as I can manage). (maximal damage: broken tests)

### DIFF
--- a/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
+++ b/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
@@ -108,8 +108,6 @@ class BaseGraphQLTransportWSHandler(ABC):
             self.connection_timed_out = True
             reason = "Connection initialisation timeout"
             await self.close(code=4408, reason=reason)
-        except asyncio.CancelledError:
-            raise
         except Exception as error:
             await self.handle_task_exception(error)  # pragma: no cover
         finally:
@@ -314,9 +312,6 @@ class BaseGraphQLTransportWSHandler(ABC):
                         ]
                     next_message = NextMessage(id=operation.id, payload=next_payload)
                     await operation.send_message(next_message)
-        except asyncio.CancelledError:
-            # CancelledErrors are expected during task cleanup.
-            raise
         except Exception as error:
             # GraphQLErrors are handled by graphql-core and included in the
             # ExecutionResult
--- a/.github/release-check-action/Dockerfile
+++ b/.github/release-check-action/Dockerfile
@@ -1,4 +1,4 @@
-FROM docker.io/python:3.7-alpine
+FROM docker.io/python:3.8-alpine
 
 RUN pip install httpx==0.7.*
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,7 +53,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: wntrblm/nox@main
         with:
-          python-versions: "3.7, 3.8, 3.9, 3.10, 3.11"
+          python-versions: "3.8, 3.9, 3.10, 3.11"
 
       - name: Pip and nox cache
         id: cache
@@ -63,8 +63,7 @@ jobs:
             ~/.cache
             ~/.nox
             .nox
-          key:
-            ${{ runner.os }}-nox-${{ matrix.session.session }}-${{
+          key: ${{ runner.os }}-nox-${{ matrix.session.session }}-${{
             hashFiles('**/poetry.lock') }}-${{ hashFiles('**/noxfile.py') }}
           restore-keys: |
             ${{ runner.os }}-nox-${{ matrix.session.session }}-
@@ -132,8 +131,7 @@ jobs:
             ~/.cache
             ~/.nox
             .nox
-          key:
-            ${{ runner.os }}-nox-lint-${{ matrix.session.session }}-${{
+          key: ${{ runner.os }}-nox-lint-${{ matrix.session.session }}-${{
             hashFiles('**/poetry.lock') }}-${{ hashFiles('**/noxfile.py') }}
           restore-keys: |
             ${{ runner.os }}-lint-nox-${{ matrix.session.session }}-
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,7 +1,7 @@
 Release type: minor
 
-chore: remove now eol python 3.7 from test matrices
-chore: remove python 3.7 quirks or mark them
-chore: update test runner docker image to python:3.8-alpine (and fully qualify it)
-chore: update dependencies
-fix: add python 3.11 to all test matrices
+This release removes support for Python 3.7 as its end of life
+was on 27 Jun 2023.
+
+This will allow us to reduce the number of CI jobs we have,
+and potentially use newer features of Python. ⚡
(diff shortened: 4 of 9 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

devkral:
this PR also updates the dependencies (required because of python update)

kristjanvalur:
If 3.7 is dropped, then the special handling of `asyncio.CancelledError` can be removed, since it becomes a direct child of  `BaseException` with 3.8

devkral:
as far as I can see, asyncio.CancelledError are not special handled (only for cleanup they are suppressed but for a good reason)

kristjanvalur:
> as far as I can see, asyncio.CancelledError are not special handled (only for cleanup they are suppressed but for a good reason)

Code like this:
```python
        try:
            ...
            raise
        except Exception as error:
            await self.handle_task_exception(error)  # pragma: no cover
        
```
is written for 3.7 compatibility, where `asyncio.CancelledError` is a subclass of `Exeption` yet should not be handled but re-raised.
In 3.8 it is a direct `BaseException` and the above clause can be omitted (because there is no `except BaseException` anywhere in sight)

devkral:
as said:

asyncio.CancelledError is propagated

the other errors are handled in:

handle_task_exception

It is quite a change of the logic if you handle now asyncio.CancelledError in handle_task_exception

devkral:
is something like 
``` patch
--- a/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
+++ b/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
@@ -108,8 +108,6 @@ class BaseGraphQLTransportWSHandler(ABC):
             self.connection_timed_out = True
             reason = "Connection initialisation timeout"
             await self.close(code=4408, reason=reason)
-        except asyncio.CancelledError:
-            raise
         except Exception as error:
             await self.handle_task_exception(error)  # pragma: no cover
         finally:
@@ -118,7 +116,10 @@ class BaseGraphQLTransportWSHandler(ABC):
             self.completed_tasks.append(task)
 
     async def handle_task_exception(self, error: Exception) -> None:
-        self.task_logger.exception("Exception in worker task", exc_info=error)
+        if isinstance(error, asyncio.CancelledError):
+            self.task_logger.exception("Worker task cancelled", exc_info=error)
+        else
+            self.task_logger.exception("Exception in worker task", exc_info=error)
 
     async def handle_message(self, message: dict) -> None:
         handler: Callable

```
right?
Handling CancelledErrors  in handle_task_exception is quite a change

kristjanvalur:
Just the first part. In 3.8 CancelledError won't be caught by "except Exception" because it isn't an "Exception"

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -116,14 +116,16 @@ async def handle_connection_init_timeout(self) -> None:
...
-        except Exception as error:
+        except BaseException as error:
+            # asyncio.CancelledError don't inherit from Exception
             await self.handle_task_exception(error)  # pragma: no cover
         finally:
             # do not clear self.connection_init_timeout_task
             # so that unittests can inspect it.
             self.completed_tasks.append(task)
 
-    async def handle_task_exception(self, error: Exception) -> None:
```
kristjanvalur, line 120: This change should be removed.

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -116,14 +116,16 @@ async def handle_connection_init_timeout(self) -> None:
...
+    async def handle_task_exception(self, error: BaseException) -> None:
+        # asyncio.CancelledError don't inherit from Exception
         if isinstance(error, asyncio.CancelledError):
```
kristjanvalur, line 122: This code can be reverted.

kristjanvalur:
I think I may have confused you with my talk about "CancelledError".  Basically, we never want to catch it.
Previously, for 3.7, to ensure that we didn't, we had to add in `except asyncio.CancelledError: raise` clauses, in order to avoid catching it.   for 3.8 and above, this is no longer necessary, as long as you are not catching `BaseException` (which you should almost never do).

So, the only thing required, is to simply __drop__ the `except asyncio.CancelledError: raise` clauses

devkral:
> I think I may have confused you with my talk about "CancelledError". Basically, we never want to catch it. Previously, for 3.7, to ensure that we didn't, we had to add in `except asyncio.CancelledError: raise` clauses, in order to avoid catching it. for 3.8 and above, this is no longer necessary, as long as you are not catching `BaseException` (which you should almost never do).
> 
> So, the only thing required, is to simply **drop** the `except asyncio.CancelledError: raise` clauses

n some new python versions asyncio.CancelledError inherits from Exception, so we need it

devkral:
cleaned up and rebased.

Also add the updated dependencies in the release.md

kristjanvalur:
> in some new python versions asyncio.CancelledError inherits from Exception, so we need it

This is incorrect.  as the documentation clearly states:  "Changed in version 3.8: [CancelledError](https://docs.python.org/3/library/asyncio-exceptions.html?highlight=cancellederror#asyncio.CancelledError) is now a subclass of [BaseException](https://docs.python.org/3/library/exceptions.html#BaseException)."

devkral:
technically this stanca would be also right, if they inherit from Exception, so it is implementation specific

kristjanvalur:
> technically this stanca would be also right, if they inherit from Exception, so it is implementation specific

No it isn't.  Yes, you could _deliberately_ mis-interpret it, but it indicates a _change_ from the 3.7 behaviour.  It is not implementation specific.  I don't quite understand why you are being so stubborn about this, it is a well known fact of Python asyncio programming.   If you like, I can have it fixed in the documentation.

devkral:
Thing is: vscode says something different. Maybe the stubs are wrong.

If you say it works, then I give it a try.

I just don't want to break anything as I have not so much time with two children

devkral:
this too?
````python
            # CancelledErrors are expected during task cleanup.
            pass
```

kristjanvalur:
It is fine, I can clean this all up later, I just wanted to mention this in the PR since this was about deprecating 3.7 and I have been wanting to get this out of the way for a long time.

Fwiw, I made a change to the docs, and here you have it from the horse's mouth:
https://github.com/python/cpython/pull/106453#pullrequestreview-1514751783

maximyurevich:
Also uvicorn can be updated with editing this PR

patrick91:
@devkral can you rebase this? 😊

devkral:
done

devkral:
I hope it is ok to add an editorconfig as my defaults always wreck the yaml formatting.

devkral:
there were some new refs to python 3.7 which had to be removed (tests/pyright).

Hope we can end the never ending PR soon ;)

patrick91 (approved):
Let's do this 😊

Issue #2828, opened 2023-06-08
Allow codegen to process multiple queries


mgilson:
## Feature Request Type

- [ ] Core functionality
- [x] Alteration (enhancement/optimization) of existing feature(s)
- [x] New behavior

## Description

codegen currently involves importing the schema from the user's app.  Depending on the structure of the app, that could be a very expensive operation (in my case, this pulls in our ORM and a number of other dependencies).  In my current application with only 5 or 6 queries, **this is currently taking about 19s to generate all the output files**.  If I only import the schema once and run the generator on all the queries in the same process, **this time drops to about 4s**.

I propose a couple minor changes to the `strawberry codegen` CLI (and underlying user-facing APIs).  First, I propose we add `nargs=-1` to the query CLI argument.  This will allow multiple queries to be passed on the command line.

The existing plugins use a hardcoded output file name.  e.g. `types.py` or `types.js`.  Changing this would likely be problematic because scripts might assume this is the output file and move it to a different name per query.  Because of this, the builtin plugins do not support processing multiple query files without some additional shenanigans.  There are two ways we could handle this:

* If there is 1 query, use the existing pattern of a hardcoded `types.py`/`types.js`.  If there are more than one query passed, use the `query.stem + extension` (where extension would vary by plugin).
* Add a flag to the plugin class that dictates whether the plugin will output unique files per input query file.

I have a branch downstream where I implemented the second, but implementing the first would be relatively simple as well.
Generally, to make this work, I have adjusted the `QueryCodegenPlugin` to accept a file path in it's `__init__`.  It's unlikely that people are overriding this currently, but if they are, we can `try`/`except` the `TypeError` and slap the `query` property on the class after initializing it but before doing anything else with it.  This should solve most backward compatibility issues (we'd only have problems if they call `__init__` directly with no arguments -- e.g. from `super().__init__()` and ever that _could_ be alleviated by having a default value that could be sniffed out (e.g. `Path.home()` -- it doesn't make sense to have a query's filename be a directory).

A potentially simpler alternative is we could just keep everything with a no-args `__init__` and always just add the `query` property in a loop after we've constructed all the user-space plugins.  From the perspective of the exposed API, the `query` filename will always be present and not `None`.  This probably requires a little white lie to the type-system for a very shot period, but in my experience, `mypy` is willing to believe a class has the properties that you claim in spite of what `__init__` does.

[...]

PR #2911, opened 2023-06-30
Allow codegen to process multiple query files in a single command

mgilson:
## Description

#2828 hopefully does a good job communicating the use-case for this feature.  Currently `codegen` can be very slow if the schema import is an expensive operation.  This allows a user to codegen multiple files with a single command (and a single schema import).  There are a few minor (backward compatible) extensions to the exposed `QueryCodegenPlugin` and `ConsolePlugin` interfaces.  The `RELEASE.md` should describe these changes and the class behaviors in their entirety.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* #2828

## Checklist

### DIFF
--- a/strawberry/cli/commands/codegen.py
+++ b/strawberry/cli/commands/codegen.py
@@ -19,7 +19,7 @@ from strawberry.codegen.plugins.typescript import TypeScriptPlugin
 def _is_codegen_plugin(obj: object) -> bool:
     return (
         inspect.isclass(obj)
-        and issubclass(obj, QueryCodegenPlugin)
+        and issubclass(obj, (QueryCodegenPlugin, ConsolePlugin))
         and obj is not QueryCodegenPlugin
     )
 
--- a/strawberry/cli/commands/codegen.py
+++ b/strawberry/cli/commands/codegen.py
@@ -12,8 +12,6 @@ import typer
 from strawberry.cli.app import app
 from strawberry.cli.utils import load_schema
 from strawberry.codegen import ConsolePlugin, QueryCodegen, QueryCodegenPlugin
-from strawberry.codegen.plugins.python import PythonPlugin
-from strawberry.codegen.plugins.typescript import TypeScriptPlugin
 
 
 def _is_codegen_plugin(obj: object) -> bool:
@@ -134,15 +132,6 @@ def codegen(
     for q in query:
         plugins = _load_plugins(selected_plugins, q)
         console_plugin.query = q  # update the query in the console plugin.
-
-        for p in plugins:
-            # Adjust the names of the output files for the buildin plugins if there
-            # are multiple files being generated.
-            if len(query) > 1:
-                if isinstance(p, PythonPlugin):
-                    p.outfile_name = q.stem + ".py"
-                elif isinstance(p, TypeScriptPlugin):
-                    p.outfile_name = q.stem + ".ts"
 
         code_generator = QueryCodegen(
             schema_symbol, plugins=plugins, console_plugin=console_plugin
--- a/strawberry/cli/commands/codegen.py
+++ b/strawberry/cli/commands/codegen.py
@@ -1,19 +1,19 @@
 from __future__ import annotations
 
+import functools
 import importlib
 import inspect
 from pathlib import Path  # noqa: TCH003
-from typing import TYPE_CHECKING, List, Optional, Type
+from typing import List, Optional, Type
 
 import rich
 import typer
 
 from strawberry.cli.app import app
 from strawberry.cli.utils import load_schema
-from strawberry.codegen import QueryCodegen, QueryCodegenPlugin
-
-if TYPE_CHECKING:
-    from strawberry.codegen import CodegenResult
+from strawberry.codegen import ConsolePlugin, QueryCodegen, QueryCodegenPlugin
+from strawberry.codegen.plugins.python import PythonPlugin
+from strawberry.codegen.plugins.typescript import TypeScriptPlugin
 
 
 def _is_codegen_plugin(obj: object) -> bool:
@@ -62,6 +62,7 @@ def _import_plugin(plugin: str) -> Optional[Type[QueryCodegenPlugin]]:
     return None
 
 
+@functools.lru_cache
 def _load_plugin(plugin_path: str) -> Type[QueryCodegenPlugin]:
     # try to import plugin_name from current folder
     # then try to import from strawberry.codegen.plugins
@@ -78,43 +79,21 @@ def _load_plugin(plugin_path: str) -> Type[QueryCodegenPlugin]:
     return plugin
 
 
-def _load_plugins(plugins: List[str]) -> List[QueryCodegenPlugin]:
-    return [_load_plugin(plugin)() for plugin in plugins]
+def _load_plugins(plugin_ids: List[str], query: Path) -> List[QueryCodegenPlugin]:
+    plugins = []
+    for ptype_id in plugin_ids:
+        ptype = _load_plugin(ptype_id)
+        plugin = ptype(query)
+        plugins.append(plugin)
 
-
-class ConsolePlugin(QueryCodegenPlugin):
-    def __init__(
-        self, query: Path, output_dir: Path, plugins: List[QueryCodegenPlugin]
-    ):
-        self.query = query
-        self.output_dir = output_dir
-        self.plugins = plugins
-
-    def on_start(self) -> None:
-        rich.print(
-            "[bold yellow]The codegen is experimental. Please submit any bug at "
-            "https://github.com/strawberry-graphql/strawberry\n",
-        )
-
-        plugin_names = [plugin.__class__.__name__ for plugin in self.plugins]
-
-        rich.print(
-            f"[green]Generating code for {self.query} using "
-            f"{', '.join(plugin_names)} plugin(s)",
-        )
-
-    def on_end(self, result: CodegenResult) -> None:
-        self.output_dir.mkdir(parents=True, exist_ok=True)
-        result.write(self.output_dir)
-
-        rich.print(
-            f"[green] Generated {len(result.files)} files in {self.output_dir}",
-        )
+    return plugins
 
 
 @app.command(help="Generate code from a query")
 def codegen(
-    query: Path = typer.Argument(..., exists=True, dir_okay=False),
+    query: Optional[List[Path]] = typer.Argument(
+        default=None, exists=True, dir_okay=False
+    ),
     schema: str = typer.Option(..., help="Python path to the schema file"),
     app_dir: str = typer.Option(
         ".",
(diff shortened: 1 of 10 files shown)


botberry:



---

`strawberry codegen` can now operate on multiple input query files.
The previous behavior of naming the file `types.js` and `types.py`
for the builtin `typescript` and `python` plugins respectively is
preserved, but only if a single query file is passed.  When more
than one query file is passed, the code generator will now use
the stem of the query file's name to construct the name of the
output files.  e.g. `my_query.graphql` -> `my_query.js` or
`my_query.py`.  Creators of custom plugins are responsible
for controlling the name of the output file themselves.  To
accomodate this, if the `__init__` method of a `QueryCodegenPlugin`
has a parameter named `query` or `query_file`, the `pathlib.Path`
to the query file will be passed to the plugin's `__init__`
method.

Finally, the `ConsolePlugin` has also recieved two new lifecycle
methods.  Unlike other `QueryCodegenPlugin`, the same instance of
the `ConsolePlugin` is used for each query file processed.  This
allows it to keep state around how many total files were processed.
The `ConsolePlugin` recieved two new lifecycle hooks: `before_any_start`
and `after_all_finished` that get called at the appropriate times.


---




Here's the tweet text: 
```

```

botberry:



---

`strawberry codegen` can now operate on multiple input query files.
The previous behavior of naming the file `types.js` and `types.py`
for the builtin `typescript` and `python` plugins respectively is
preserved, but only if a single query file is passed.  When more
than one query file is passed, the code generator will now use
the stem of the query file's name to construct the name of the
output files.  e.g. `my_query.graphql` -> `my_query.js` or
`my_query.py`.  Creators of custom plugins are responsible
for controlling the name of the output file themselves.  To
accomodate this, if the `__init__` method of a `QueryCodegenPlugin`
has a parameter named `query` or `query_file`, the `pathlib.Path`
to the query file will be passed to the plugin's `__init__`
method.

Finally, the `ConsolePlugin` has also recieved two new lifecycle
methods.  Unlike other `QueryCodegenPlugin`, the same instance of
the `ConsolePlugin` is used for each query file processed.  This
allows it to keep state around how many total files were processed.
The `ConsolePlugin` recieved two new lifecycle hooks: `before_any_start`
and `after_all_finished` that get called at the appropriate times.


---




Here's the tweet text: 
```

```

tests/cli/test_codegen.py
```diff
@@ -80,6 +124,87 @@ def test_codegen(cli_runner: CliRunner, query_file_path: Path, tmp_path: Path):
...
+    )
+
+    assert result.exit_code == 0
+
+    for path in expected_paths:
+        assert path.exists()
+        assert " GetUserResult" in path.read_text()
+
+
+def test_codegen_pass_no_query(cli_runner: CliRunner, tmp_path: Path):
```
mgilson, line 190: I don't have an opinion on the proper behavior here.  I went with the behavior recommended by `click` for [variadic arguments](https://click.palletsprojects.com/en/8.1.x/arguments/#variadic-arguments)

> ... we believe scripts should gracefully degrade into becoming noops if a variadic argument is empty. The reason for this is that very often, scripts are invoked with wildcard inputs from the command line and they should not error out if the wildcard is empty.

strawberry/cli/commands/codegen.py
```diff
@@ -143,12 +178,50 @@ def codegen(
...
-    plugins = _load_plugins(selected_plugins)
-    plugins.append(console_plugin(query, output_dir, plugins))
+    plugins = _load_plugins(selected_plugins, query[0])
+    console_plugin = console_plugin_type(query[0], output_dir, plugins)
+
+    if not isinstance(console_plugin, ConsolePlugin):
+        warnings.warn(
+            "The ConsolePlugin should inherit from ``{__name__}.ConsolePlugin``.",
+            DeprecationWarning,
+            stacklevel=1,
```
mgilson, line 195: `ruff` wants the explicit `stacklevel` 🤷‍♂️ .

patrick91:
@mgilson I'll try to review this during this week 😊 thanks for your patience!

strawberry/cli/commands/codegen.py
```diff
@@ -78,8 +84,26 @@ def _load_plugin(plugin_path: str) -> Type[QueryCodegenPlugin]:
...
+        sig = inspect.signature(ptype)
+        kwargs = {}
+        if "query" in sig.parameters:
+            kwargs["query"] = query
+        if "query_file" in sig.parameters:
+            kwargs["query_file"] = query
+
+        plugin = ptype(**kwargs)
+        if not hasattr(plugin, "query"):
+            plugin.query = query
```
patrick91, line 102: is this to keep backwards compatibility?

strawberry/cli/commands/codegen.py
patrick91: yeah, that's fine :D

strawberry/cli/commands/codegen.py
```diff
@@ -143,12 +178,50 @@ def codegen(
...
+            DeprecationWarning,
+            stacklevel=1,
+        )
 
-    code_generator = QueryCodegen(schema_symbol, plugins=plugins)
-    code_generator.run(query.read_text())
+    with contextlib.suppress(AttributeError):
+        # This method was not part of the original implementation of the ConsolePlugin
+        # and the ConsolePlugin is overridable on the CLI.  It isn't guaranteed that
+        # custom ConsolePlugin implementations inherit from ``ConsolePlugin``, so if
```
patrick91, line 201:
```suggestion
        # custom ConsolePlugin implementations inherit from `ConsolePlugin`, so if
```

strawberry/cli/commands/codegen.py
```diff
@@ -143,12 +178,50 @@ def codegen(
...
+
     schema_symbol = load_schema(schema, app_dir)
 
-    console_plugin = _load_plugin(cli_plugin) if cli_plugin else ConsolePlugin
+    console_plugin_type = _load_plugin(cli_plugin) if cli_plugin else ConsolePlugin
 
-    plugins = _load_plugins(selected_plugins)
-    plugins.append(console_plugin(query, output_dir, plugins))
+    plugins = _load_plugins(selected_plugins, query[0])
+    console_plugin = console_plugin_type(query[0], output_dir, plugins)
```
patrick91, line 189: I'm not sure I like having to pass the first query here 🤔

strawberry/cli/commands/codegen.py
```diff
@@ -143,12 +178,50 @@ def codegen(
...
+        plugins.append(console_plugin)
+
+        for p in plugins:
+            # Adjust the names of the output files for the buildin plugins if there
+            # are multiple files being generated.
+            if len(query) > 1:
+                if isinstance(p, PythonPlugin):
+                    p.outfile_name = q.stem + ".py"
+                elif isinstance(p, TypeScriptPlugin):
+                    p.outfile_name = q.stem + ".ts"
```
patrick91, line 217: I wonder if in future we should give the output name to the plugins

patrick91 (commented):
left some comments, I think this is almost ready to merge :)

strawberry/cli/commands/codegen.py
mgilson: yeah, I did a lot of shenanigans to try to preserve compatibility with any existing plugins that might be in the wild.

If we don't care about that, then we can probably simplify this a good bit.

strawberry/cli/commands/codegen.py
mgilson: Yeah, I dislike this as well.  I'd rather have the `console_plugin_type` accept a collection of queries, but anyone who has made a custom console plugin type would potentially be broken in that case :(.

strawberry/cli/commands/codegen.py
mgilson: I'd love that ... I guess I'll raise the question of compatibility and how much we care about making breaking changes in these APIs.

On my side, I'm quite happy to break the API to make it a little cleaner for this direction, but I don't have any idea how impactful that would be for downstream consumers (I don't know how much use this feature actually gets outside of my org 😅 ).

strawberry/cli/commands/codegen.py
patrick91: I think it might be fine to break compatibility for this, we can just add a note in our breaking changes files. We could do a codemod too but it is probably not worth doing 😊

strawberry/cli/commands/codegen.py
patrick91: those changes are quite simple, so if we provide a guide it should be ok!

mgilson: Excellent.  If you aren't worried about breaking changes then I'll get started on those in the next day or two.

Thanks!

patrick91: uhm, that's a good point!

I guess we can allow plugins to return a list of files, but they should still be able to name them, right?

maybe we can take inspiration from https://the-guild.dev/graphql/codegen ?

strawberry/cli/commands/codegen.py
```diff
@@ -1,19 +1,21 @@
...
 from strawberry.cli.app import app
 from strawberry.cli.utils import load_schema
-from strawberry.codegen import QueryCodegen, QueryCodegenPlugin
+from strawberry.codegen import ConsolePlugin, QueryCodegen, QueryCodegenPlugin
+from strawberry.codegen.plugins.python import PythonPlugin
+from strawberry.codegen.plugins.typescript import TypeScriptPlugin
 
 if TYPE_CHECKING:
-    from strawberry.codegen import CodegenResult
+    from pathlib import Path
```
mgilson, line 18: AHH!!!

What tool is actually moving this import?

In this case, the tool is _thinking_ that `Path` should be in an `if TYPE_CHECKING` because it's only used in annotations, however, `typer` is introspecting the types so the import statement needs to be present at runtime.

I'm not sure how to prevent our tool from moving this (other than putting in some sort of useless expression that actually _uses_ `Path` somewhere).

strawberry/cli/commands/codegen.py
mgilson: I'm not exactly sure how that thing works, but I _think_ it's similar to what I'm doing in my custom plugin.

We're using jinja templates to include fragments into the query -- when the codegen sees a named fragment, it will emit an object for that fragment and then that same object can be shared across multiple queries.

(currently, that object gets re-generated with each query that includes it -- it feels sloppy, but since the object generation is deterministic, it works out Ok).

In this paradigm, the plugin is choosing names for the files that include the types that describe the fragments as well as the file that includes the other query specific things.

strawberry/cli/commands/codegen.py
patrick91: that's ruff, I think that rule is becoming quite annoying, `# noqa: TCH001` should fix it

strawberry/cli/commands/codegen.py
patrick91: could we give the plugin the source file name and let it decide what to do? 😊

mgilson: > could we give the plugin the source file name and let it decide what to do? 😊

That's what we're doing ... except that the plugin doesn't know how many files they are going to process.  They're designed to process one file at a time.  It could _always_ get the output filename and then decide what to do, but then we don't have enough information to play the "If there is only one file, we'll name the output `types.py`, if there are more than one file, we'll name the output `<stem>.py`".

I'm happy to get rid of the default `types.py` output filename and try to update the corresponding documentation to let people know that the filename will be `<stem>.py` from here on -- is that the direction you'd rather take it?

patrick91: give me some time to think about this properly, I'm clearly tired today :D

patrick91:
@mgilson really sorry for the delay on this :)

I think we can merge as it is now, I've changed the plugins to use the filename of the query to generate the files 😊

I think maybe in future we can customise it more to make so that the generated files are generated near the source ones and not passing the output directory at all 😊 

let me know if the changes look good to you and I'll merge

mgilson (commented):
Looks great to me!  :shipit:

PR #2922, opened 2023-07-05
fix: only resolve a field's type after field extensions has been applied to it

### LINKED ISSUE #2921: `FieldExtension` changes to return type aren't reflected in schema
## Describe the Bug

Since upgrading to latest Strawberry, `FieldExtension`s don't appear to be able to modify the return type (`.type`) of a field.

I've attached a code sample to reproduce:
```python
import strawberry
from strawberry.annotation import StrawberryAnnotation
from strawberry.extensions import FieldExtension


class ChangeReturnTypeExtension(FieldExtension):
    def apply(self, field: StrawberryField) -> None:
        print(f'Converting field return type to int... (Currently: {field.type} / {field.type_annotation.resolve()})')
        field.type_annotation = StrawberryAnnotation.from_annotation(int)
        field.type = int
        print(f'Done. (New type: {field.type} / {field.type_annotation.resolve()})')

    # Not part of the example, needed so Strawberry can resolve
    def resolve(self, next_, source, info, **kwargs):
        return next_(source, info, **kwargs)


@strawberry.type
class Query:
    @strawberry.field(extensions=[ChangeReturnTypeExtension()])
    def test_changing_return_type(self) -> bool:
        # doesn't matter that type checkers complain - this is just to see the field type
        return 1


```

Run with `strawberry server main:schema`.

When ran on my laptop, this is the output:

```
$ strawberry server main:schema
Converting field return type to int... (Currently: <class 'bool'> / <class 'bool'>)
Done. (New type: <class 'int'> / <class 'int'>)
Running strawberry on http://0.0.0.0:8000/graphql 🍓
Converting field return type to int... (Currently: <class 'bool'> / <class 'bool'>)
Done. (New type: <class 'int'> / <class 'int'>)
Converting field return type to int... (Currently: <class 'bool'> / <class 'bool'>)
Done. (New type: <class 'int'> / <class 'int'>)
```

This _appears_ to successfully change the field's return type from a `bool` to an `int`. I've tried updating `field.type` and `field.type_annotation` together, and also each one independently. In previous versions of Strawberry, updating just `field.type_annotation` as shown above worked.

Checking field.type/type_annotation immediately after updating it does appear to show the new type - however the GraphQL schema won't actually reflect this. I've attached a screenshot of GraphiQL when running the above sample code.

image

I'm not sure if updating a field's type in a `FieldExtension` is officially supported behavior, but regardless it appears to be a regression compared to previous versions.

## System Information

 - Operating system: macOS / Python 3.10
 - Strawberry version (if applicable): `0.193.0`

bellini666:
0.190.0 changed `from_field` to receive an optional `override_type` as the already resolved
type for that field.

The issue is that field extensions will only be applied when calling `from_resolver`, which happens
inside `from_field` itself.

This still works as expected for specialized generics, but ensures that any changes to
the type annotation made by field extensions is not ignored.

Fix #2921

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,5 @@
+Release type: patch
+
+This fixes a regression from 0.190.0 where changes to the
+return type of a field done by Field Extensions would not
+be taken in consideration by the schema.
--- a/strawberry/schema/schema_converter.py
+++ b/strawberry/schema/schema_converter.py
@@ -56,7 +56,6 @@ from strawberry.type import (
     StrawberryList,
     StrawberryOptional,
     StrawberryType,
-    WithStrawberryObjectDefinition,
     has_object_definition,
 )
 from strawberry.types.info import Info
@@ -99,9 +98,7 @@ class FieldConverterProtocol(Generic[FieldType], Protocol):
         self,
         field: StrawberryField,
         *,
-        override_type: Optional[
-            Union[StrawberryType, Type[WithStrawberryObjectDefinition]]
-        ] = None,
+        type_definition: Optional[StrawberryObjectDefinition] = None,
     ) -> FieldType:
         ...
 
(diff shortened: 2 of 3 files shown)


patrick91 (approved):

botberry:



---

This fixes a regression from 0.190.0 where changes to the
return type of a field done by Field Extensions would not
be taken in consideration by the schema.


---




Here's the tweet text: 
```

```

PR #2925, opened 2023-07-06
Restore evaled type access in `StrawberryAnnotation`

skilkis:
Previously reserved argument parsing would evaluate stringified annotations when possible and cache these results. Therefore, when `StrawberryAnnotation` instances were created within `StrawberryResolver`, their `annotation` attribute could become an evaluated type. To restore this behavior, an `annotation` property is added to `StrawberryAnnotation` that calls
`StrawberryAnnotation.evaluate()`. On success the evaluated type is returned. If a `NameError` is raised due to an unresolvable annotation, the stringified annotation is returned instead.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/tests/experimental/pydantic/test_fields.py
+++ b/tests/experimental/pydantic/test_fields.py
@@ -169,7 +169,7 @@ def test_constrained_list_nested():
     if sys.version_info >= (3, 9):
         expected_annotation = list[list[int]]
     else:
-        expected_annotation = List[List[str]]
+        expected_annotation = List[List[int]]
 
     assert (
         UserType.__strawberry_definition__.fields[0].type_annotation.annotation
--- a/tests/experimental/pydantic/test_fields.py
+++ b/tests/experimental/pydantic/test_fields.py
@@ -1,4 +1,5 @@
 import re
+import sys
 from typing import List
 from typing_extensions import Literal
 
@@ -131,9 +132,15 @@ def test_constrained_list():
         ...
 
     assert UserType.__strawberry_definition__.fields[0].name == "friends"
+
+    if sys.version_info >= (3, 9):
+        expected_annotation = list[str]
+    else:
+        expected_annotation = List[str]
+
     assert (
         UserType.__strawberry_definition__.fields[0].type_annotation.annotation
-        == List[str]
+        == expected_annotation
     )
 
     data = UserType(friends=[])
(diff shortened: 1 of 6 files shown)


tests/experimental/pydantic/test_fields.py
```diff
@@ -131,9 +132,15 @@ class UserType:
         ...
 
     assert UserType.__strawberry_definition__.fields[0].name == "friends"
+
+    if sys.version_info >= (3, 9):
+        expected_annotation = list[str]
+    else:
+        expected_annotation = List[str]
+
     assert (
```
skilkis, line 141: Peculiar that this is necessary now

skilkis:
@patrick91 @bellini666 the error was caused by an implicit assumption that `StrawberryAnnotation.annotation` would be an evaluated annotation. The line in which the extension was failing can be traced to [object_type.py](https://github.com/strawberry-graphql/strawberry/blob/1de74778854671604be50e53dfad44ad2dc6728e/strawberry/object_type.py#L70) in which the annotation inside the `__annotations__` class variable is replaced with `StrawberryAnnotation.annotation` which before this fix has become the stringified annotation. To mimick the previous behavior I've created an `annotation` property that returns the evaluated annotation when possible.

patrick91:
@skilkis looks good to me :) I think it might be worth investigating all the places where we need actual annotations in future so we consolidate some of the logics

bellini666:
LGTM as well! :)

skilkis:
> @skilkis looks good to me :) I think it might be worth investigating all the places where we need actual annotations in future so we consolidate some of the logics

I completely agree, I also believe we should postpone evaluation of all annotations until the schema is converted as much as possible

tests/experimental/pydantic/test_fields.py
skilkis: @patrick I figured it out, previously the test-case was using `StrawberryAnnotation.annotation` which is now calling `eval_type` which has some code in there calling `get_origin` which on Python >= 3.9 becomes `list` instead of `typing.List`.

I'll make a separate PR to address this for `eval_type`. For now I've changed the test to use the `raw_annotation`

botberry:



---

Restore evaled type access in `StrawberryAnnotation`

Prior to Strawberry 192.2 the `annotation` attribute of `StrawberryAnnotation`
would return an evaluated type when possible due reserved argument parsing.
192.2 moved the responsibility of evaluating and caching results to the
`evaluate` method of `StrawberryAnnotation`. This introduced a regression when
using future annotations for any code implicitely relying on the `annotation`
attribute being an evaluated type.

To fix this regression and mimick pre-192.2 behavior, this release adds an
`annotation` property to `StrawberryAnnotation` that internally calls the
`evaluate` method. On success the evaluated type is returned. If a `NameError`
is raised due to an unresolvable annotation, the raw annotation is returned.


---




Here's the tweet text: 
```

```

patrick91 (approved):

PR #2926, opened 2023-07-06
improve performance of `strawberry.lazy()` with relative imports

karimsa:
## Description

I've been profiling a few tests in my company's codebase and realized that some tests that actually only perform about 1s of business logic take ~12s to run. A large part of that time (~8-9s) is spent resolving forward refs in `strawberry.lazy()`. The largest culprit there was the use of `inspect.stack()` which seems to be incredibly slow.

This PR updates the method of retrieving the target stack frame. It seems that the stack frames are stored as a linkedlist anyways, so retrieving the active frame and then working backwards is significantly faster (likely because `.stack()` collects more information for every frame that is unused by strawberry). This one line change took my test down from 12s to 2.5s.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

N/a

## Checklist

### DIFF
--- a/strawberry/lazy_type.py
+++ b/strawberry/lazy_type.py
@@ -68,11 +68,11 @@ class LazyType(Generic[TypeName, Module]):
 
 
 class StrawberryLazyReference:
-    def __init__(self, module: str) -> None:
+    def __init__(self, module: str, package: Optional[str] = None) -> None:
         self.module = module
-        self.package = None
+        self.package = package
 
-        if module.startswith("."):
+        if module.startswith(".") and not package:
             frame = inspect.stack()[2][0]
             # TODO: raise a nice error if frame is None
             assert frame is not None
@@ -82,5 +82,5 @@ class StrawberryLazyReference:
         return LazyType(forward_ref.__forward_arg__, self.module, self.package)
 
 
-def lazy(module_path: str) -> StrawberryLazyReference:
-    return StrawberryLazyReference(module_path)
+def lazy(module_path: str, package: Optional[str] = None) -> StrawberryLazyReference:
+    return StrawberryLazyReference(module_path, package)
--- a/strawberry/lazy_type.py
+++ b/strawberry/lazy_type.py
@@ -68,12 +68,12 @@ class LazyType(Generic[TypeName, Module]):
 
 
 class StrawberryLazyReference:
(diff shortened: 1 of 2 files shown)


strawberry/lazy_type.py
```diff
@@ -73,7 +73,7 @@ def __init__(self, module: str) -> None:
         self.package = None
 
         if module.startswith("."):
-            frame = inspect.stack()[2][0]
+            frame = inspect.currentframe().f_back.f_back
```
karimsa, line 76: I didn't bother to check for nulls at every location, since this should only be accessed via `.lazy()` and also the previous code would've hit a bounds check in the same condition anyways

botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Karim Alibhai for the PR 👏

```

strawberry/lazy_type.py
patrick91: nice! we also use `sys._getframe(x)` in another place, would you mind trying with that and see if performs the same? 😊

patrick91:
as a not for myself, I think we should add some benchmark for this :)

strawberry/lazy_type.py
karimsa: Yupp! It has the same effect - actually `_getframe` is in C, so looks like it is even faster. Whipped up a small benchmark to look at the difference:

Screenshot 2023-07-06 at 5 34 35 PM

<details>

	import time
	import inspect
	import sys

	def run_benchmark(name, fn):
		start_time = time.time()

		for i in range(1000000):
			fn()

		end_time = time.time()
		duration = end_time - start_time

		print(f"{name} completed in {duration}s")

	def main():
		run_benchmark("inspect.stack()[2]", lambda: inspect.stack()[2])
		run_benchmark("inspect.currentframe().f_back.f_back", lambda: inspect.currentframe().f_back.f_back)
		run_benchmark("sys._getframe(2)", lambda: sys._getframe(2))

	main()

</details>

patrick91: awesome! thanks for that!

patrick91 (approved):
Thank you so much! 🏃‍♂️💨

botberry:

PR #2955, opened 2023-07-15
Skip optimised check when returning generic types

### LINKED ISSUE #2954: Regression using interfaces with generic types.
## Describe the Bug
If a generic type implements an interface, and is registered as an extra type, it is not being recognized in the schema.

```python
from typing import Generic, TypeVar
import strawberry

T = TypeVar('T')

@strawberry.interface
class Abstract:
    x: str = ''

@strawberry.type
class Real(Generic[T], Abstract):
    y: T

@strawberry.type
class Query:
    @strawberry.field
    def real(self) -> Abstract:
        return Real[int](y=0)

schema = strawberry.Schema(Query, types=[Real[int]])
print(schema.execute_sync('{ real { __typename } }'))
```
raises `Abstract type 'Abstract' was resolved to a type 'Real' that does not exist inside the schema.`.

## System Information

 - Operating system: not applicable
 - Strawberry version (if applicable): 0.195

## Additional Context
This worked prior to v0.195. `IntReal` is present in the schema, but the error references type `Real`.

patrick91:
Closes #2954 

There's probably a better way to do this :) but I'm happy to do that in another PR.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,4 @@
+Release type: patch
+
+This release fixes a bug where returning a generic type from a field
+that was returning an interface would throw an error.
--- a/strawberry/schema/schema_converter.py
+++ b/strawberry/schema/schema_converter.py
@@ -60,6 +60,7 @@ from strawberry.type import (
     StrawberryList,
     StrawberryOptional,
     StrawberryType,
+    get_object_definition,
     has_object_definition,
 )
 from strawberry.types.info import Info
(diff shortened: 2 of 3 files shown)


patrick91:
One way to make this better could be to add some metadata when doing things like this: `Real[int](y=0)` (__class_getitem__ could use that)

but it won't fix the cases where some one does `Real(y=0)`, I'm sure we have some code that checks the right type somewhere though 😊

botberry:



---

This release fixes a bug where returning a generic type from a field
that was returning an interface would throw an error.


---




Here's the tweet text: 
```

```

PR #2990, opened 2023-07-28
fix: do not strip the type when using auto

bellini666:
`Annotated[auto, ...]` produce issues on python 3.10 and older
due to the fact that it tries to check that the type is `callable`.

This was being masked by the issue that 5bd41027366a04f644feb2679c752cd51b120e08
fixed, due to the fact that extra `Annotated` args were being thrown
away.

This change the evaluation to avoid throwing the original type (usually
`Any`) away, fixing that issue.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,11 @@
+Release type: patch
+
+This release fixes an issue when trying to use `Annotated[strawberry.auto, ...]`
+on python 3.10 or older, which got evident after the fix from 0.196.1.
+
+Previously we were throwing the type away, since it usually is `Any`, but python
+3.10 and older will validate that the first argument passed for `Annotated`
+is callable (3.11+ does not do that anymore), and `StrawberryAuto` is not.
+
+This changes it to keep that `Any`, which is also what someone would expect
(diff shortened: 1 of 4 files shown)


RELEASE.md
```diff
@@ -0,0 +1,4 @@
+Release type: patch
+
+This release fixes an issue when trying to use `Annotated[strawberry.auto, ...]`
+on python 3.10 or older, which got unmasked after the fix from 0.196.1.
```
patrick91, line 4: can we explain what unmasked is here?

botberry:



---

This release fixes an issue when trying to use `Annotated[strawberry.auto, ...]`
on python 3.10 or older, which got unmasked after the fix from 0.196.1.


---




Here's the tweet text: 
```

```

RELEASE.md
bellini666: Hrm, maybe the "unmasked" term is not the correct one in english?

Basically the issue was there, but was not happening because of the typo that 0.196.1 fixed.

bellini666: like, 0.196.1 made that issue evident

patrick91 (approved):

PR #2993, opened 2023-07-30
Fix new generics in Python 3.12

patrick91:
This PR changes how we handle TypeVars in generics to make it work in Python 3.12's new syntax 😊

I'm not 100% sure why it wasn't working before, but the issue was due to the fact that two TypeVars with the same name aren't equal:

```
Python 3.12.0b4 (main, Jul 12 2023, 21:58:01) [Clang 14.0.0 (clang-1400.0.29.202)] on darwin
>>> from typing import TypeVar
>>> TypeVar("T") == TypeVar("T")
False
```

We weren't encountering this issue previous because we are reusing the TypeVars, but it looks like this isn't the case in Python 3.12's new syntax.

To fix this I've changed the code use the name for the TypeVar (TypeVar("T").__name__), which seemed to have worked 😊

TODO: I need to skip the 3.12 tests because older versions of Python aren't unable to parse the code, so they'll fail


botberry:



---

This release changes how we handle generic type vars, bringing
support to the new generic syntax in Python 3.12 (which will be out in October).

This now works:

```python
@strawberry.type
class Edge[T]:
    cursor: strawberry.ID
    node_field: T


@strawberry.type
class Query:
    @strawberry.field
    def example(self) -> Edge[int]:
        return Edge(cursor=strawberry.ID("1"), node_field=1)


```


---




Here's the tweet text: 
```

```

PR #3006, opened 2023-08-02
Allow null payload in ws

patrick91:
We were rejecting payloads that had an explicit null value, while they are supported in the Protocol:

https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#connectionack

### DIFF
--- a/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
+++ b/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
@@ -21,6 +21,7 @@ from strawberry.subscriptions.protocols.graphql_transport_ws.types import (
     SubscribeMessagePayload,
 )
 from strawberry.types.graphql import OperationType
+from strawberry.unset import UNSET
 from strawberry.utils.debug import pretty_print_graphql_operation
 from strawberry.utils.operation import get_operation_type
 
@@ -162,7 +163,11 @@ class BaseGraphQLTransportWSHandler(ABC):
         if self.connection_init_timeout_task:
             self.connection_init_timeout_task.cancel()
 
-        payload = message.payload or {}
+        payload = (
+            message.payload
+            if message.payload is not None and message.payload is not UNSET
+            else {}
+        )
 
         if not isinstance(payload, dict):
             await self.close(code=4400, reason="Invalid connection init payload")
--- a/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
+++ b/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
@@ -21,7 +21,6 @@ from strawberry.subscriptions.protocols.graphql_transport_ws.types import (
     SubscribeMessagePayload,
 )
 from strawberry.types.graphql import OperationType
-from strawberry.unset import UNSET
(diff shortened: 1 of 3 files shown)


tests/websockets/test_graphql_transport_ws.py
```diff
@@ -183,7 +183,8 @@ async def test_close_twice(
     ) as ws:
         transport_close = mocker.patch.object(ws, "close")
 
-        await ws.send_json(ConnectionInitMessage(payload=None).as_dict())
+        # TODO: payload is set to 1 to force close, but is this right?
+        await ws.send_json(ConnectionInitMessage(payload=1).as_dict())
```
patrick91, line 187: @bellini666 can you take a look at this? I think you added this test 😊

patrick91:
this is draft only because it is missing the release file :)

tests/websockets/test_graphql_transport_ws.py
bellini666: Yeah, it was in [this PR](https://github.com/strawberry-graphql/strawberry/pull/2341), which had more changes other than the test itself but got superseded by another fix and we decided to keep just the tests.

This seems to be fine, what we want from this test is to send a wrong payload to `ConnectionInitMessage`, which would cause it to be closed immediatly, and then check that the init timeout handler doesn't try to close it again.

Passing `1` or any invalid payload is fine! :). Maybe just make it more specific that `1` is an arbitrary invalid value, because otherwise it could make someone think that `1` means close =P

tests/websockets/test_graphql_transport_ws.py
```diff
@@ -799,9 +800,9 @@ async def test_rejects_connection_params_not_dict(ws_raw: WebSocketClient):
     ws.assert_reason("Invalid connection init payload")
 
 
-async def test_rejects_connection_params_not_unset(ws_raw: WebSocketClient):
+async def test_rejects_connection_params_with_wrong_type(ws_raw: WebSocketClient):
```
bellini666, line 803: Maybe parametrize this with some other options? Like: a string, an integer, a list with any value in it (an empty list would work because the code is doing `payload or {}`), etc

bellini666 (approved):
Looks good! :)

Left 2 suggestions

botberry:



---

This release fixes an issue in the `graphql-ws` implementation
where sending a `null` payload would cause the connection
to be closed.


---




Here's the tweet text: 
```

```

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -163,11 +163,17 @@ async def handle_connection_init(self, message: ConnectionInitMessage) -> None:
...
             self.connection_init_timeout_task.cancel()
 
-        if message.payload is not UNSET and not isinstance(message.payload, dict):
+        payload = (
+            message.payload
+            if message.payload is not None and message.payload is not UNSET
+            else {}
+        )
+
+        if not isinstance(payload, dict):
```
kristjanvalur, line 172: Maybe additional sanitation.  We could have a helper function for this, since string dicts are common in the protocol.
`if not isinstance(payload, dict) or not all(isinstance(key, str) for key in payload.keys())`

PR #3011, opened 2023-08-03
Fix incorrect type narrowing on generic interfaces

skilkis:
## Description

Previously the narrowed type of a generic interface was not checked. Therefore, the first concrete type of the generic interface would always return `True` from the `is_type_of` method.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #2954

## Checklist

### DIFF
--- a/strawberry/schema/schema_converter.py
+++ b/strawberry/schema/schema_converter.py
@@ -18,7 +18,7 @@ from typing import (
     Union,
     cast,
 )
-from typing_extensions import Protocol
+from typing_extensions import Protocol, get_args
 
 from graphql import (
     GraphQLAbstractType,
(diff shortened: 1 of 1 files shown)


skilkis:
Relevant test cases:

- Multi-inheritance
- Unions
- Annotated types

skilkis:
@patrick91 the tests I just pushed highlight an issue with this approach with respect to `Annotated` types. If the user passes the "raw" type then the equality comparison fails. We could fix this in two ways, extending logic in `StrawberryType.__eq__` or by comparing "raw" types within `is_type_of` by stripping all `Annotated` and `StrawberryUnion` / `StrawberryAnnotation` down to their resolved types

PR #3018, opened 2023-08-08
[strawberry.ext.mypy_plugin] Fix import error for users who don't use pydantic

davidnemec:
This should fix import error in strawberry.ext.mypy_plugin for users who don't use pydantic introduced in latest version. Hopefully the new location of the import is fine.

`Error importing plugin "strawberry.ext.mypy_plugin": No module named 'pydantic'`

### DIFF
--- a/strawberry/ext/mypy_plugin.py
+++ b/strawberry/ext/mypy_plugin.py
@@ -52,7 +52,6 @@ from mypy.types import (
 )
 from mypy.typevars import fill_typevars
 from mypy.util import get_unique_redefinition_name
-from strawberry.experimental.pydantic._compat import IS_PYDANTIC_V1
 
 # Backwards compatible with the removal of `TypeVarDef` in mypy 0.920.
 try:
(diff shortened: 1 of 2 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to David Němec for the PR 👏

```

patrick91 (commented):
Ah good catch! Can you add a release file for this?

strawberry/ext/mypy_plugin.py
```diff
@@ -64,6 +63,8 @@
 try:
     from pydantic.mypy import METADATA_KEY as PYDANTIC_METADATA_KEY
     from pydantic.mypy import PydanticModelField
+
+    from strawberry.experimental.pydantic._compat import IS_PYDANTIC_V1
 except ImportError:
     PYDANTIC_METADATA_KEY = ""
```
patrick91, line 69: also I think it is safe to add a default to false 😊

so

```
IS_PYDANTIC_V1 = False
```

davidnemec: `IS_PYDANTIC_V1` is used in `if-else`. Should I update it to `if-elif` and also set (and check in elif) `IS_PYDANTIC_V2=False`? :thinking:

patrick91: I meant only in the expect `except ImportError:` if we can't import pydantic it means we don't have it installed, so the default for `IS_PYDANTIC_V1` doesn't matter that much :)

patrick91 (approved):
Thank you!

PR #3023, opened 2023-08-09
Fix incorrect TypingUnionType import check

tjeerddie:
- TypingUnionType import error check is set to `< 3.9` but 3.9 fails the import and it should be `< 3.10`

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/strawberry/experimental/pydantic/fields.py
+++ b/strawberry/experimental/pydantic/fields.py
@@ -26,13 +26,10 @@ try:
 except ImportError:
     import sys
 
-    # python < 3.9 does not have GenericAlias (list[int], tuple[str, ...] and so on)
+    # python < 3.10 does not have GenericAlias (list[int], tuple[str, ...] and so on)
     # we do this under a conditional to avoid a mypy :)
-    if sys.version_info < (3, 9):
-        TypingGenericAlias = ()
-    else:
-        raise
     if sys.version_info < (3, 10):
+        TypingGenericAlias = ()
         TypingUnionType = ()
     else:
         raise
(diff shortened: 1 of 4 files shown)


strawberry/experimental/pydantic/fields.py
```diff
@@ -26,13 +26,10 @@
...
 
-    # python < 3.9 does not have GenericAlias (list[int], tuple[str, ...] and so on)
+    # python < 3.10 does not have GenericAlias (list[int], tuple[str, ...] and so on)
     # we do this under a conditional to avoid a mypy :)
-    if sys.version_info < (3, 9):
-        TypingGenericAlias = ()
-    else:
-        raise
     if sys.version_info < (3, 10):
+        TypingGenericAlias = ()
```
patrick91, line 32: CleanShot 2023-08-09 at 10 50 22@2x

this seems to be available for me

botberry:



---

Fix incorrect TypingUnionType import check of python version `< 3.9` to `< 3.10` in `experimental.pydantic.fields.py`


---




Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to tjeerddie for the PR 👏

```

strawberry/experimental/pydantic/fields.py
tjeerddie: yea, separated the `TypeUnionAlias` and `TypeGenericAlias` version checks in their own try-catch

patrick91 (commented):
do you think we can add a test for this?

tjeerddie:
> do you think we can add a test for this?

not sure how to test this, will attempt to test it

patrick91 (commented):
I've changed the pydantic tests to run on all integrations, so the codepath is being triggered now without needing to write new tests 😊

PR #3025, opened 2023-08-10
refactor: support node_id values containing colons

bellini666:
The node_id can be anything depending on how the user is handling it, so colons can appear in it, specially when it is some kind of encrypted value.

This changes how we split the `GlobalID` string to only split the first colon in it, meaning that the remaining of the string will be considered as a whole for the node_id.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,9 @@
+Release type: minor
+
+Add support for extra colons in the `GlobalID` string.
+
+Before, the string `SomeType:some:value` would produce raise an error saying that
+it was expected the string to be splited in 2 parts when doing `.split(":")`.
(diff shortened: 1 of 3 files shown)


botberry:



---

Add support for extra colons in the `GlobalID` string.

Before, the string `SomeType:some:value` would produce raise an error saying that
it was expected the string to be splited in 2 parts when doing `.split(":")`.

Now we are using `.split(":")`, meaning that the example above will consider
`SomeType` to be the type name, and `some:value` to be the node_id.


---




Here's the tweet text: 
```

```

patrick91 (approved):

patrick91:
:yay: :D

PR #3017, opened 2023-08-07
Add new strawberry.Parent annotation to support static resolvers with non-self types.

mattalbr:
## Description

Allows creating a resolver that depends on a parent which has a different type than the resolver's class, as in `self` isn't particularly accurate. This is much more descriptive (and allows type checking) than calling the parameter "self" and knowing it will be a different type. See description in #515 

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ x ] New feature
- [ ] Enhancement/optimization
- [ ] Documentation

## Issues Fixed or Closed by This PR

* #515 

## Checklist

- [ x ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ x ] I have read the CONTRIBUTING document.
- [ x ] I have added tests to cover my changes.
- [ x ] I have tested the changes and verified that they work and don't break anything (as well as I can manage).

Unsure if this needs documentation!

PR #3028, opened 2023-08-10
Exclude unit tests themselves from code coverage results.

mattalbr:
I noticed in https://github.com/strawberry-graphql/strawberry/pull/3017 that we're analyzing coverage of the test files themselves. This both takes unnecessary execution time/cost, and reduces signal quality.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,7 @@
 # .coveragerc to control coverage.py
 [run]
 branch = True
+omit = tests/*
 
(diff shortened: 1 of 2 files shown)


patrick91:
I don't think we should skip tests, in the past we had tests that weren't running, and this helps with finding them 😊

Hope that's ok!

mattalbr:
If you're finding value in them, sure thing!

PR #3035, opened 2023-08-14
Fix missed pydantic v2 tests with error ConstrainedList removed in v2

tjeerddie:
- Mark missed pydantic constrained list tests with need_pydantic_v1 since it is removed in pydantic V2

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

### DIFF
--- a/tests/experimental/pydantic/schema/test_basic.py
+++ b/tests/experimental/pydantic/schema/test_basic.py
@@ -1,10 +1,13 @@
+import sys
 import textwrap
 from enum import Enum
 from typing import List, Optional, Union
 
 import pydantic
+import pytest
 
 import strawberry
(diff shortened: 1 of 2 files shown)


RELEASE.md
```diff
@@ -0,0 +1,4 @@
+Release type: patch
+
+Mark missed tests that are incompatible with pydantic v2 with need_pydantic_v1.
+Mark ConstrainedList tests with skipif python version under 3.9 because they break in 3.8
```
patrick91, line 4: I think we can skip the release note if we're only touching tests 😊

patrick91 (approved):
Thanks!

PR #3050, opened 2023-08-23
Handle UNSET values in the strawberry schema.

mgilson:
The codegen currently chokes on `strawberry.UNSET` default values.  This PR updates the codegen to stop raising an exception and instead pass a `GraphQLNullValue(value=UNSET)` to the plugins.  The default python plugin writes a `None` as the default value because there really isn't a well understood concept of `UNSET` in the python ecosystem.  User defined plugins have all the information that they need to handle this differently if they choose to do so.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/tests/codegen/snapshots/typescript/variables.ts
+++ b/tests/codegen/snapshots/typescript/variables.ts
@@ -4,6 +4,7 @@ type OperationNameResult = {
 
 type PersonInput = {
     name: string
+    age: number | undefined
 }
 
 type ExampleInput = {
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,14 @@
+Release type: minor
+
+`strawberry codegen` previously choked for inputs that used the
+`strawberry.UNSET` sentinal singleton value as a default.  The intent
+here is to say that if a variable is not part of the request payload,
+then the `UNSET` default value will not be modified and the service
+code can then treat an unset value differently from a default value,
+etc.
+
+For codegen, we treat the `UNSET` default value as a `GraphQLNullValue`.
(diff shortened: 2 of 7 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

patrick91 (commented):
Looks good! We only need the release file 😊 and I'm adding a note

tests/codegen/snapshots/typescript/variables.ts
```diff
@@ -4,6 +4,7 @@ type OperationNameResult = {
 
 type PersonInput = {
     name: string
+    age: number | undefined
```
patrick91, line 7: should we also do `null` here?

mgilson: My typescript isn't that great, but I think that the intent is that `personInput.age` would be unset (`undefined`) in the request payload (rather than having it be `null`).  My typescript isn't that great -- but at least `JS` has a better way to express that "this key was never set" (`undefined`) vs. "this key was set to some default 'missing' value" (`null`).  I think that `UNSET` most closely maps to the "this key was never set" semantics -- so `undefined` seems like the right choice to me here...

patrick91: you're right 😊

patrick91 (approved):

wlaub:
I'm suddenly getting snake case argument names instead of camel case in the kwargs passed to has_permissions. Is this an intentional change? I don't see it listed on the breaking changes page, but it's breaking some thing.

PR #3055, opened 2023-08-25
feat: support partial errors and multiple errors per field resolver

bradleyoesch:
_**EDIT:** this has been updated to no longer use an extension, but to provide native support instead_

Create a new `PartialResultsExtension` to allow adding exceptions to `info.context.partial_errors` to get added to the result after execution, allowing users to add errors to the errors array from a field resolver while still resolving that field.

## Description

_NOTE: It's entirely possible/likely that I misunderstand how partial/multi-error handling works in graphql-python and that there is a simpler, already supported way to do this._

One powerful feature of GraphQL is returning both data AND errors. The typical way to return an error in python graphql is to simply raise it, and [let the graphql library catch it](https://github.com/graphql-python/graphql-core/blob/0c93b8452eed38d4f800c7e71cf6f3f3758cd1c6/src/graphql/execution/execute.py#L979-L982) and add to the Execution Context errors that ultimately map to the `errors` array in the response.

So this means in a field resolver we can either return data OR raise an error. But we can't raise multiple errors (possibly incorrect) and we can't return data AND "raise" an error (also possibly incorrect). I have not been able to find documentation or examples supporting either case.

To work around this, this extension creates an interface so developers can manually populate errors from resolvers that get added to the result after execution.

## Caveats

I'll admit a couple things here:
1. As noted above, my understanding may be lacking and there is already support for this kind of behavior
2. Even if my understanding is correct and this extension works, this may not be the way this library would want to support this kind of thing
3. There's a little hackiness as noted in the copied block of the extension code that I don't love
4. Unclear how best to get the `nodes` arg for creating your own `GraphQLErrors` to ensure location and path fields exist in the error object in the response. See https://github.com/strawberry-graphql/strawberry/pull/874 for information around not exposing `Nodes` directly, in favor of strawberry's own wrapper type `Selection`.

## Usage example:

```python
import strawberry
from strawberry.extensions import PartialResultsExtension

schema = strawberry.Schema(Query, extensions=[PartialResultsExtension])

# ...

@strawberry.field
def query(self, info) -> bool:
    info.context.partial_errors.append(Exception("Partial failure"))
    return True
```

Results:

```json
{
  "data": {
    "query": true
  },
  "errors": [
    {
      "message": "Partial failure"
    }
  ]
}
```

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* None

## Checklist


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Bradley Oesch for the PR 👏

```

bradleyoesch:
Some example use cases:
- Query for data matching ids [0, 1, 2] and there's no id 1, so you return data: `[data0, null, data2]` with errors array with some message and the path pointing to the index of the missing data
- Plural mutation where some fail and some don't, similar kind of response
- Operation with multiple business-logic-like validation errors, e.g. user tries to do some action on "item id 123" but gets multiple errors, e.g. "User does not have enough coins for this action", whatever, something that is not like a graphql-schema-validation type of error

RELEASE.md
```diff
@@ -0,0 +1,73 @@
...
+
+Basic usage:
+
+```python
+import strawberry
+
+
+@strawberry.field
+def query(self, info) -> bool:
+    info.context.partial_errors.append(Exception("Partial failure"))
```
bradleyoesch, line 15: I think moving this off of context and into `info` would be preferred, since context can be anything by implementing frameworks/clients.

I'll look into it

docs/guides/errors.md
```diff
@@ -136,7 +136,14 @@ schema = strawberry.Schema(query=Query)
...
 
-If an error is expected then it is often best to express it in the schema. This allows the client to deal with the error in a robust way.
+Handling expected errors is typically approached in one or both of two ways:
+
+1. Return the error as part of the return type
+2. Append the error to the `errors` array
+
+### Returning as part of the return type
+
+It is often best to express expected errors in the schema. This allows the client to deal with the error in a robust way.
```
bradleyoesch, line 146: No change to this section, just reorganization.

This may be a personal preference, but I don't think this section with full examples really makes sense to live in Strawberry documentation. This is more of a schema design recommendation with nothing specific to strawberry itself. As long as users understand interfaces and unions in graphql and strawberry, implementation is straightforward, and, it's rather opinionated (as seen by the 6-7 examples in the references at the bottom).

I think it makes more sense, if we want to keep this section, to instead point to the references already linked and optionally point to other pages in strawberry documentation that would help readers understand how the core concepts of the error approaches could be implemented in strawberry.

docs/guides/errors.md
```diff
@@ -215,6 +222,82 @@ mutation RegisterUser($username: String!, $password: String!) {
 
 This approach allows you to express the possible error states in the schema and so provide a robust interface for your client to account for all the potential outcomes from a mutation.
 
+### Appending to `errors` array
+
+As seen in the [Unhandled execution errors][#unhandled-execution-errors] section above, errors are automatically added to the errors array by raising the error in execution.
```
bradleyoesch, line 227: Not sure if this is how to refer to another heading on this page

patrick91:
I'll take a proper look at this probably next week, but some notes that pop into my mind:

1. I wonder if we could somehow [use exception groups](https://peps.python.org/pep-0654/), granted we find a way to backport them (or maybe we can allow returning exceptions in resolvers? I wonder if other frameworks do this 
2. can you add a test with the exact use case you have? that should make it easier to understand
3. Do you think it is important to make sure we return the right path for the error?

patrick91:
also, thank you so much for working on this!

bradleyoesch:
@patrick91 

> 1. I wonder if we could somehow [use exception groups](https://peps.python.org/pep-0654/), granted we find a way to backport them (or maybe we can allow returning exceptions in resolvers? I wonder if other frameworks do this

I think this could be cool. I'm not sure what would happen since [the graphql-core library only catches exceptions, not groups](https://github.com/graphql-python/graphql-core/blob/0c93b8452eed38d4f800c7e71cf6f3f3758cd1c6/src/graphql/execution/execute.py#L979-L1007). Note that if we got this to work (with no other changes), this would enable multiple errors per resolver, but not an error _and_ data in a resolver.

I think having the option of the resolver returning either the `data` type OR something like an `ExecutionResult` (that models data and errors) would be cool, and then that context type could add any errors to the errors array, mapping any to `GraphQL` errors with path/location data, etc (which as I describe down in #3), I think should be more easily accessible per resolver. DGS (netflix java graphql library) defines a [`DataFetcherResult`](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/execution/DataFetcherResult.java) that you can return.

> 2. can you add a test with the exact use case you have? that should make it easier to understand

I think [this `test_partial_errors_yes_data_yes_errors` test](https://github.com/strawberry-graphql/strawberry/pull/3055/files#diff-2d9e49eeaff80ff54483f868c16c6ca48bdb2100cf86548fd28666eee94bda84R559) is pretty straightforward, but if you're asking for a less contrived test, yeah I can add a test e.g. resolving a list and some are null and you want some errors to point to the null ones
 
> 3. Do you think it is important to make sure we return the right path for the error?

I think it's important to give the users the ability to easily add the right path/location for the error (and if we can do it automatically, great). I'll note that [this example with a located error](https://github.com/strawberry-graphql/strawberry/pull/3055/files#diff-2b1b69303b927a484e02c7fad9fc87d0d3ff0dc22ae1da0ecd0dc935d922a23cR44) test shows that the easiest way (that I could find) to get the nodes for a resolver is to use `info.field_nodes`, which was deprecated in https://github.com/strawberry-graphql/strawberry/pull/874 when abstracting out some of the raw types.

I think it could be worth a separate issue to add in some functionality for easily getting the current resolver's located node and path, and then accessing that in here.

bradleyoesch:
@patrick91 Updated with an example use case, a mutation with a partial failure

https://github.com/strawberry-graphql/strawberry/pull/3055/commits/948dc657f682e2b419abdb0dfb6664c8667a9efb

cadlagtrader:
Hello @patrick91 Hello - do you plan to merge this PR soon ? 
That would be very useful for my use case too (market data partially missing)

patrick91:
I have some time off friday, so I should be able to review and merge! thanks for the ping!

bradleyoesch:
@patrick91 bump on this for review whenever you have time

RELEASE.md
bradleyoesch: I've gone with an iffy route of trying to support attribute setting and indexing generically, though it's kind of gross.

Since `info` comes from the graphql core library, I'm not so sure moving it into `info` makes much sense.

Since I don't love either approach, it kind of implies that maybe this isn't a great idea 😅

tests/schema/test_execution.py
```diff
@@ -474,3 +490,283 @@ def this_fails(self) -> Optional[str]:
...
+    assert error3.extensions == {"foo": "bar"}
+    assert error3.original_error is None
+    # the actual json in the response
+    assert error3.formatted == {
+        "message": "Raw GraphQLError with extensions",
+        "extensions": {"foo": "bar"},
+    }
+
+
+@pytest.mark.parametrize("context_type", ("class", "dict"))
```
bradleyoesch, line 658: This is a lil gross too 🤷

docs/guides/errors.md
bradleyoesch: I don't feel strongly about this months later :)

bradleyoesch:
@patrick91 I totally lost this one! Rebased main and tried to be a little more generically supportive of context types, but tbh I don't love the approach too much atm. Curious if you have any thoughts coming back to this now?

PR #3076, opened 2023-09-06
Add support for multipart subscriptions

patrick91:
This is a new protocol created by Apollo (and supported by Apollo Client)

# TODO

- [x] Add support for this in all HTTP integrations that support subscriptions
- [x] Send a ping every 5 seconds (an empty `{}` is enough)
- [x] Add a proper way to check if an operation is a subscription in our Async HTTP View
- [x] Make sure that the returned responses is actually being streamed
- [x] Return proper error when trying to run a subscription and it's not supported
- [x] Add docs (using a proxy breaks the disconnection)

Protocol docs: https://www.apollographql.com/docs/router/executing-operations/subscription-multipart-protocol/


strawberry/http/async_base_view.py
```diff
@@ -112,6 +112,14 @@ async def execute_operation(
 
         assert self.schema
 
+        return await self.schema.subscribe(
```
patrick91, line 115: here we need to check if we are doing a subscription or not

strawberry/http/async_base_view.py
```diff
@@ -185,10 +193,18 @@ async def run(
 
         response_data = await self.process_result(request=request, result=result)
 
+        # only if is a multipart subscription
+        return self.create_multipart_response(response_data, sub_response)
```
patrick91, line 197: this should be streamed (response data should be a generator or async iterator)

strawberry/http/async_base_view.py
patrick91: also I've noticed that if I try to do a subscription using schema.execute it doesn't error, but it does something weird, worth adding a check for this case.

strawberry/http/async_base_view.py
```diff
@@ -198,6 +214,9 @@ async def parse_http_body(
             data = self.parse_json(await request.get_body())
         elif content_type.startswith("multipart/form-data"):
             data = await self.parse_multipart(request)
+        elif content_type.startswith("multipart/mixed"):
+            # TODO: do a check that checks if this is a multipart subscription
+            data = await self.parse_multipart_subscriptions(request)
```
patrick91, line 219: the content type I get with apollo client is 

```
"multipart/mixed;boundary=graphql;subscriptionSpec=1.0,application/json"
```

tests/views/schema.py
```diff
@@ -151,7 +151,7 @@ def match_text(self, text_file: Upload, pattern: str) -> str:
 @strawberry.type
 class Subscription:
     @strawberry.subscription
-    async def echo(self, message: str, delay: float = 0) -> AsyncGenerator[str, None]:
+    async def echo2(self, message: str, delay: float = 0) -> AsyncGenerator[str, None]:
```
patrick91, line 154: I need to revert this

strawberry/http/async_base_view.py
patrick91: also the response should be something like this:

```
--graphql
{"payload":{"data":{"codegenStatus":{"__typename":"CodegenResolverChunk","index":88,"text":";"}}}}
--graphql
content-type: application/json

{"payload":{"data":{"codegenStatus":{"__typename":"CodegenMessage","text":"Done! Now I'll generate the template ✍️\n"}}}}
--graphql
content-type: application/json

{"payload":{"data":{"codegenStatus":{"__typename":"CodegenResult","sandboxUrl":"https://codesandbox.io/p/sandbox/asd"}}}}
--graphql
content-type: application/json

{"payload":null}
--graphql--
```

patrick91:
one thing I'm not sure about this is handling client closing the connections 🤔

strawberry/channels/handlers/http_handler.py
```diff
@@ -170,19 +187,37 @@ def create_response(
...
         )
 
+    async def run(
+        self,
+        request: Any,
+        context: Optional[Any] = UNSET,
+        root_value: Optional[Any] = UNSET,
+    ) -> Union[ChannelsResponse, MultipartChannelsResponse]:
+        # putting this here just for type checking
+        raise NotImplementedError()
```
cdroege, line 197: This line is the cause of the `TimeoutError` errors for the Channels integrations tests as this is called from the `handle` method. I believe that a `Protocol` definition for this would be the correct way to define this type information? But Channels itself has no type information and pyright and mypy both will complain about the other unknown methods (e.g. `send_headers`) when I tested such a definition (without adding even more type definitions for the other unknown methods).

I believe this is also why I defined the return type directly in line [202](https://github.com/strawberry-graphql/strawberry/blob/fc7ecd8dfc7ca9a41da52d1183a5827b38d151fe/strawberry/channels/handlers/http_handler.py#L202) before.

patrick91: ouch! I'll remove it for now, thanks! I wish I we got an exception though :D

botberry:
# Pre-release
:wave:

Releasing commit [52b2f2d467660baa99cc0b5f55ad1a83f740b8b9] to PyPi as pre-release! :package:

botberry:
# Pre-release
:wave:

Pre-release **0.210.0.dev.1697796691** [52b2f2d467660baa99cc0b5f55ad1a83f740b8b9] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.210.0.dev.1697796691
```

strawberry/http/async_base_view.py
```diff
@@ -201,15 +220,99 @@ async def run(
...
+                f"\r\n--{separator}\r\n",
+                "Content-Type: application/json\r\n\r\n",
+                self.encode_json(data),
+                "\n",
+            ]
+        )
+
+    def _stream_with_heartbeat(
+        self, stream: Callable[[], AsyncGenerator[str, None]]
+    ) -> Callable[[], AsyncGenerator[str, None]]:
```
patrick91, line 235: @bellini666

strawberry/channels/handlers/http_handler.py
```diff
@@ -301,5 +335,5 @@ def run(
         request: ChannelsRequest,
         context: Optional[Context] = UNSET,
         root_value: Optional[RootValue] = UNSET,
-    ) -> ChannelsResponse:
+    ) -> ChannelsResponse:  # pyright: ignore
```
bellini666, line 338: Should this be `ChannelsResponse | MultipartChannelsResponse`?

strawberry/channels/handlers/http_handler.py
```diff
@@ -185,16 +202,25 @@ def create_response(
...
-            )
+            if isinstance(response, MultipartChannelsResponse):
+                response.headers[b"Transfer-Encoding"] = b"chunked"
+                await self.send_headers(headers=response.headers)
+
+                async for chunk in response.stream():
+                    # TODO: we should change more body
+                    await self.send_body(chunk.encode("utf-8"), more_body=True)
+
+            else:
```
bellini666, line 218: Maybe do `elif isinstance(response, ChannelsResponse)` with `else: assert_never(response)` to make sure that we are handling every possibility in typing and runtime

strawberry/http/parse_content_type.py
```diff
@@ -0,0 +1,17 @@
+from email.message import Message
```
bellini666, line 1: Maybe just call this module `parsers.py` to allow for more parsers in it in the future?

tests/http/clients/async_django.py
```diff
@@ -55,6 +55,13 @@ async def _do_request(self, request: RequestFactory) -> Response:
             return Response(
                 status_code=400, data=e.args[0].encode(), headers=response.headers
             )
+
+        if isinstance(response, StreamingHttpResponse):
```
bellini666, line 59: This if/else can be rewritten to avoid some code duplication:

```python
return Response(
    status_code=response.status_code,
    data=(
        response.streaming_content
        if isinstance(response, StreamingHttpResponse) else
        response.content
    ),
    headers=response.headers,
)
```

strawberry/http/async_base_view.py
```diff
@@ -201,15 +220,99 @@ async def run(
...
+
+        async def merged() -> AsyncGenerator[str, None]:
+            heartbeat_task = asyncio.create_task(heartbeat())
+            task = asyncio.create_task(drain())
+
+            def cancel_tasks():
+                nonlocal cancelling
+                cancelling = True
+                task.cancel()
+                heartbeat_task.cancel()
```
bellini666, line 269: If I remember correctly, those needs to be awaited after cancelling:

```suggestion
                task.cancel()
                with contextlib.suppress(asyncio.CancelledError):
                    await task

                heartbeat_task.cancel()
                with contextlib.suppress(asyncio.CancelledError):
                    await heartbeat_task
```

strawberry/http/async_base_view.py
```diff
@@ -201,15 +220,99 @@ async def run(
...
+                else:
+                    raise
+
+        async def heartbeat():
+            while True:
+                await queue.put((False, self.encode_multipart_data({}, "graphql")))
+
+                await asyncio.sleep(5)
+
+        async def merged() -> AsyncGenerator[str, None]:
```
bellini666, line 261: Nice implementation :)

I usually use https://aiostream.readthedocs.io/en/latest/operators.html#aiostream.stream.concat for those kind of things, as it handles some extra corner cases, but I would not suggest it here as we would need an extra dependency for this.

botberry:

<table>
<thead>
</thead>
<tbody>
</tbody>
</table>

Learn more:

botberry:



---

This release adds support for multipart subscriptions in almost all\*
integrations!

[Multipart subcriptions](https://www.apollographql.com/docs/router/executing-operations/subscription-multipart-protocol/)
are a new protocol from Apollo GraphQL, built on the
[Incremental Delivery over HTTP spec](https://github.com/graphql/graphql-over-http/blob/main/rfcs/IncrementalDelivery.md),
which is also used for `@defer` and `@stream`.

The main advantage of this protocol is that when using the Apollo Client
libraries you don't need to install any additional dependency, but in future
this feature will make it easier for us to implement `@defer` and `@stream`

Also, this means that you don't need to use Django Channels for subscription,
since this protocol is based on HTTP we don't need to use websockets.

- Flask, Chalice and the sync Django integration don't support this yet.



Here's the tweet text: 
```

```

strawberry/channels/handlers/http_handler.py
```diff
@@ -247,6 +277,15 @@ async def get_context(
...
         return TemporalResponse()
 
+    async def create_multipart_response(
+        self,
+        request: ChannelsRequest,
+        stream: Callable[[], AsyncGenerator[str, None]],
+        sub_response: TemporalResponse,
+    ) -> MultipartChannelsResponse:
+        # TODO: sub response
+        return MultipartChannelsResponse(stream=stream)
```
bellini666, line 287: Taking a look at how `sub_response` is used today, shouldn't we be using it for status code and headers?

strawberry/django/views.py
```diff
@@ -177,14 +185,28 @@ def create_response(
 
         return response
 
+    async def create_multipart_response(
+        self,
+        request: HttpRequest,
+        stream: Callable[[], AsyncIterator[Any]],
+        sub_response: HttpResponse,
```
bellini666, line 192: Ditto here about `sub_response` (which applies basically to all `create_multipart_response`

bellini666 (approved):
LGTM! :)

Really excited about this! Just raised some concerns about `sub_response` not being used anywhere, other than that it is good!

strawberry/channels/handlers/http_handler.py
patrick91: yup, I'll try to add this

bellini666 (approved):

patrick91:
🔥

strawberry/schema/execute.py
```diff
@@ -124,16 +125,28 @@ async def execute(
...
+                        # TODO: should we process errors here?
+                        # TODO: make our own wrapper?
+                        return await subscribe(  # type: ignore
+                            schema,
+                            execution_context.graphql_document,
+                            root_value=execution_context.root_value,
+                            context_value=execution_context.context,
+                            variable_values=execution_context.variables,
+                            operation_name=execution_context.operation_name,
+                        )
```
DoctorJohn, line 138: This extends the `Schema.execute` method to handle subscriptions. Im wondering whether this was intentional since we already have a separate `Schema.subscribe` method? :eyes: @patrick91 @bellini666 

I noticed this change while upgrading to the latest version. In my codebase mypy now complains everywhere that `Schema.execute` can now return both `ExecutionResult` and `SubscriptionExecutionResult`. Before `Schema.execute` could only return an `ExecutionResult` while subscriptions were handled by `Schema.subscribe`.

patrick91: that was poor design on my part, it will be fixed in #3554 😊

PR #3124, opened 2023-09-30
fix: use markcoroutinefunction on django async view to support python 3.12

bellini666:
Trying to mark strawberry-django as compatible with Python 3.12 and discovered the tests were failing because of this.

Python 3.12 removed `_is_coroutine` check. `asgiref.sync.markcoroutinefunction` will use the new `inspect.markcoroutinefunction` on python 3.12 and fallback to the older behaviour for older python versions

Related: https://github.com/strawberry-graphql/strawberry-graphql-django/pull/359

### DIFF
--- a/strawberry/django/views.py
+++ b/strawberry/django/views.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import asyncio
 import json
 from typing import (
     TYPE_CHECKING,
@@ -12,6 +11,7 @@ from typing import (
     cast,
 )
 
+from asgiref.sync import markcoroutinefunction
 from django.http import HttpRequest, HttpResponseNotAllowed, JsonResponse
 from django.http.response import HttpResponse
 from django.template import RequestContext, Template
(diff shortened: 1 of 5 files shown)


botberry:



---

Mark Django's asyncview as a coroutine using `asgiref.sync.markcoroutinefunction`
to support using it with Python 3.12.



Here's the tweet text: 
```

```

strawberry/django/views.py
```diff
@@ -231,7 +231,8 @@ def as_view(cls, **initkwargs: Any) -> Callable[..., HttpResponse]:
         # https://docs.djangoproject.com/en/3.1/topics/async/#async-views
 
         view = super().as_view(**initkwargs)
-        view._is_coroutine = asyncio.coroutines._is_coroutine  # type: ignore[attr-defined]
+        markcoroutinefunction(view)
```
patrick91, line 234: do we need to add tests for this?

bellini666: Hrm, TBH it would be nice to have one

I was just trying to fix this with minimum effort, specially since strawberry-graphql-django tests this extensively =P, but I'll try to add one here :)

patrick91:
@bellini666 can you also pin flask in dev at <3? pytest-flask doesn't support it just yet

patrick91 (approved):

PR #3127, opened 2023-10-01
Add `_handle_errors` hook on http views

patrick91:
This should make this PR for sentry a bit better: https://github.com/getsentry/sentry-python/pull/2393

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,6 @@
+Release type: patch
+
+This release adds a new private hook in our HTTP views, it is called
+`_handle_errors` and it is meant to be used by Sentry (or other integrations)
+to handle errors without having to patch methods that could be overridden
+by the users
--- a/strawberry/http/async_base_view.py
+++ b/strawberry/http/async_base_view.py
@@ -4,10 +4,13 @@ from typing import (
     Callable,
     Dict,
     Generic,
+    List,
     Mapping,
     Optional,
     Union,
 )
+
+from graphql import GraphQLError
 
 from strawberry import UNSET
 from strawberry.exceptions import MissingQueryError
@@ -143,6 +146,11 @@ class AsyncBaseHTTPView(
         except KeyError as e:
             raise HTTPException(400, "File(s) missing in form data") from e
 
+    def _handle_errors(self, errors: List[GraphQLError]) -> None:
+        """
+        Hook to allow custom handling of errors, used by the Sentry Integration
+        """
+
     async def run(
         self,
         request: Request,
@@ -182,6 +190,9 @@ class AsyncBaseHTTPView(
             ) from e
         except MissingQueryError as e:
             raise HTTPException(400, "No GraphQL query found in the request") from e
+
+        if result.errors:
+            self._handle_errors(result.errors)
 
         response_data = await self.process_result(request=request, result=result)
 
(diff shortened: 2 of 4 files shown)


botberry:



---

This release adds a new private hook in our HTTP views, it is called
`_handle_errors` and it is meant to be used by Sentry (or other integrations)
to handle errors without having to patch methods that could be overridden
by the users



Here's the tweet text: 
```

```

sentrivana:
@patrick91 Thanks for adding the hook!

Would it be possible to also get access to the result as a whole after `process_result`? We'd like to be able to attach the response body to error events to help folks debug. Here's how it's done currently: https://github.com/getsentry/sentry-python/pull/2393/files#diff-3f467610968b3aebbd2cafa234cc12789d210cb1d82482bbf01d5b9d1ba9a891R315-R320

Due to how event processors work in the SDK it'd be awesome if it could just be one hook for both capturing the errors and the response. (TL;DR: we have to make sure we have the event processor with the response set up before we actually capture any errors. In other words, we need access to the response before or at the same time as the errors, which makes this a bit annoying.)

Rough idea how this could look in the async base view:

```python
        try:
            result = await self.execute_operation(
                request=request, context=context, root_value=root_value
            )
        except InvalidOperationTypeError as e:
            (...)
        except MissingQueryError as e:
            (...)

        errors = []  # added
        if result.errors:  # added
            errors = list(result.errors)  # added, copy the original errors to have access to them later

        response_data = await self.process_result(request=request, result=result)

        if errors:  # added
            self._handle_errors(errors, response_data)  # added

        return self.create_response(
            response_data=response_data, sub_response=sub_response
        )
```

What do you think? It's not the prettiest but probably least invasive as far as accommodating this need-response-at-the-same-time-as-errors requirement goes.

Also open to other ideas!

patrick91:
@sentrivana that's fine to me, just one question, why do you make a copy of the errors there?

sentrivana:
@patrick91 Just so that if someone modifies `result.errors` in their custom `process_result`, we can still work with the original errors. I don't see why anyone would do that though, so maybe I'm being overly cautious. We could also just use `result.errors` directly.

sentrivana:
@patrick91 In that case I'd work with `result.errors` as they come out of `process_result`. If folks are doing any anonymization/data scrubbing, either in `process_result` or even earlier, we should honor that before sending anything to Sentry.

With that I guess the hook just needs to be moved after `process_result` and have the `response_data` added to the args.

```python
        response_data = await self.process_result(request=request, result=result)

        if result.errors:  # added
            self._handle_errors(result.errors, response_data)  # added

        return self.create_response(
            response_data=response_data, sub_response=sub_response
        )
```

patrick91:
@sentrivana done, let me know if that looks ok!

sentrivana:
Looks great, I'll give it a try tomorrow and report back!

botberry:
# Pre-release
:wave:

Releasing commit [e80c58b509d0a8c7f30596e9da1ece5a5495586b] to PyPi as pre-release! :package:

patrick91:
@sentrivana I'll merge and release this now, please ping me when you make the release, I'll deprecate and update the docs for the sentry integration!

sentrivana:
@patrick91 Will do! Probably tomorrow or Thursday.

PR #3131, opened 2023-10-02
fix: exclude "self" from ConflictingArgumentsError check

bellini666:
Change how we check for conflicting resolver arguments to exclude `self` from those checks, which were introduced on version 0.208.0.

It is a common pattern among integrations, such as the Django one, to use `root: Model` in the resolvers for better typing inference.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,7 @@
+Release type: patch
+
+This release changes how we check for conflicting resolver arguments to
+exclude `self` from those checks, which were introduced on version 0.208.0.
+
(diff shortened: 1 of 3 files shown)


botberry:



---

This release changes how we check for conflicting resolver arguments to
exclude `self` from those checks, which were introduced on version 0.208.0.

It is a common pattern among integrations, such as the Django one, to
use `root: Model` in the resolvers for better typing inference.



Here's the tweet text: 
```

```

patrick91 (approved):

Issue #3125, opened 2023-09-30
Operation directive not working on QUERY


mritzerfeld:
Operation directive is not called for location QUERY.

## Describe the Bug

I have written an operation directive which should be available for use on QUERY, similar to this reduced example:

```
@strawberry.directive(locations=[DirectiveLocation.QUERY])
async def my_directive(value: Any):
    """
    Adds contextual information to node like content language.
    """
    
    print('called')
    return value

schema = strawberry.Schema(
    directives=[my_directive],
    query=RootQuery,
)
```

When running a query like

```
query @myDirective {
  someData {
    id
  }
}
```

the directive method isn't called at all. If running the directive on a field like `someData` or `id` however, the function gets called, event though FIELD is not in the list of locations. I tried defining a root value but this didn't make any difference.

## System Information

 - Operating system: macOS 14.0
 - Strawberry version (if applicable): 0.209.2

patrick91:
interesting, I can definitely reproduce this, I'll update one I found the cause :)

patrick91:
interesting, I can definitely reproduce this, I'll update one I found the cause :)

patrick91:
@mritzerfeld what's your use case for operation directives on query? I have a POC to make them work here: https://github.com/strawberry-graphql/strawberry/pull/3140

mritzerfeld:
@patrick91 Thanks for your response and the quick changes. I'm planning on using such a directive for providing query-wide meta information like localization.

I can confirm that the directive handler now gets called correctly. I did however get an exception if the handler function expects the info parameter. Line 136 of the changed `strawberry/extensions/directives.py` throws

```
AttributeError: 'NoneType' object has no attribute 'field_name'
```

in such case.

patrick91:
@mritzerfeld yes, the code is quite bad tbh :D

we don't have the info parameter for operations, so I might need to just pass None, for now at least :)

> I'm planning on using such a directive for providing query-wide meta information like localization.

ah interesting use case, I usually would do that via headers, but a directive sounds nice too.

How will you pass this information down to the resolvers?

mritzerfeld:
The advantage over headers is IMO that you can also enable such a directive on fields which allows for optionally providing different meta information on a nested field in a single query.

My initial plan was using the info.context for passing the information down the chain after having them validated in the directive first. But without a working info parameter I might have to think of a different way.

patrick91:
that's good to know, I think we can definitely have the context working, I'll let you know :)

[...]

PR #3140, opened 2023-10-07
add support for directives on documents

patrick91:
Context: https://github.com/strawberry-graphql/strawberry/issues/3125


patrick91:
@mritzerfeld would this work for you? at the moment you can't run code before any resolver, which might not be what you want

mritzerfeld:
@patrick91 The status quo works for me. I was able to build a workaround by combining a schema directive, a operation directive and a schema extension on a field on the the root level allowing me to pass information from the operation directive before the field resolver is executed.

patrick91:
I'll close this PR, if we need this in future we can re-implement it 😊

PR #3144, opened 2023-10-10
feat: propagate entity resolver original error details to response

### LINKED ISSUE #3135: Messages from errors thrown in entity resolution are swallowed by a generic failure message
In entity resolvers, if you throw an error, the library catches the error and wraps it in a GraphQL Error:
https://github.com/strawberry-graphql/strawberry/blob/628e4980b50e317cf99999aef052b8bb095370df/strawberry/federation/schema.py#L210-L216

* The message is replaced with a generic "Unable to resolve reference for {definition.origin}"
* The definition origin is the full python classpath, revealing underlying implementation details to the client
   * e.g. "Unable to resolve reference for <class 'somedjangoapp.graphql.types.MyEntity'>"
* I haven't double checked this but this might also remove all GraphQLError details like path/location if the thrown error is itself already a GraphQLError

The test itself obscures this issue:
https://github.com/strawberry-graphql/strawberry/blob/628e4980b50e317cf99999aef052b8bb095370df/tests/federation/test_entities.py#L215-L217

Dumbest potential solution:
```python
try:
    result = get_result()
except Exception as e:  # noqa: PERF203
    if isinstance(e, GraphQLError):
        result = e
    else:
        message = f"Unable to resolve reference for {definition.origin}"
        result = GraphQLError(
            e.message or message,
            original_error=e,
        )
```

I'm willing to make a PR if this seems reasonable!

## Upvote & Fund

<a href="https://polar.sh/strawberry-graphql/strawberry/issues/3135">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/3135/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

bradleyoesch:
Error details from errors thrown in entity resolution are swallowed by a generic failure message that also leaks internal code paths, so instead we should propagate the original error.

## Description

When an entity resolver throws an exception, add the thrown exception to the result. Turns out graphql-core handles mapping it to a `GraphQLError` with location/path info for us ([link](https://github.com/graphql-python/graphql-core/blob/0c93b8452eed38d4f800c7e71cf6f3f3758cd1c6/src/graphql/execution/execute.py#L1372)) so we don't want to raise our own `GraphQLError`.

If the exception is a `TypeError`, we assume the resolver was incorrectly implemented, which throws a `TypeError` under the hood (we check the type name explicitly so that clients can raise their own custom `TypeErrors` without losing data). The generic message now just outputs the type name rather than the full path, which includes file directory structure that shouldn't (imo) be exposed to the response.

To be honest, I don't _love_ this type error approach, so I'm up for alternatives!

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* Closes #3135 

## Checklist


botberry:



---

Update entity resolver exception handling to set the result to the original error instead of a `GraphQLError`, which obscured the original message and meta-fields.
There is no need to wrap the exception in a `GraphQLError` since the `graphql-core` library handles that for us in its
[execute code](https://github.com/graphql-python/graphql-core/blob/0c93b8452eed38d4f800c7e71cf6f3f3758cd1c6/src/graphql/execution/execute.py#L1372).
However, if the original it is a `TypeError`, it indicates improper use of resolver and may contain code-level error messages,
so it is replaced with a generic exception with a more client-friendly string: `"Unable to resolve reference for {EntityType}"`.



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Bradley Oesch for the PR 👏

```

strawberry/federation/schema.py
```diff
@@ -210,10 +209,14 @@ def entities_resolver(
...
-                    f"Unable to resolve reference for {definition.origin}",
-                    original_error=e,
-                )
+                if type(e).__name__ == "TypeError":
+                    # check explicitly for type name instead of `isinstance` so
+                    # clients can raise custom TypeErrors to avoid this wrapper
+                    result = Exception(
+                        f"Unable to resolve reference for {type_.definition.name}"
+                    )
+                else:
```
DoctorJohn, line 218: Hi, could you elaborate on why `TypeError` should still be wrapped in a new `Exception`?

bradleyoesch: Yes, probably worth a comment. It's for when the query is made with incorrect data, e.g. the `test_fails_properly_when_wrong_data_is_passed` test
https://github.com/strawberry-graphql/strawberry/pull/3144/files#diff-7e4c37801e2121b1d2d8f4cafddcf32731d47271b9097beeb23808ac9ca42196R174

I wrapped it in the Exception to preserve what I considered to be existing behavior from that test that already existed. I can't remember off the top of my head though what the message would be if it weren't wrapped...

I could look into that if you'd like

bradleyoesch: @DoctorJohn I've updated this a bit, I hope it's a little more clear now.

The failures seem to be around code coverage within my tests so they don't need important?

bradleyoesch: @DoctorJohn bumping one more time, update is in this commit: https://github.com/strawberry-graphql/strawberry/pull/3144/commits/56e89556220fa3b60d8a3bfe5c4cd1bd754f79ca

DoctorJohn: Thank you @bradleyoesch, the only reason this is not merged yet is still this bit here. You mentioned:

> I wrapped it in the Exception to preserve what I considered to be existing behavior from that test that already existed.

I want to take a closer look and see whether we need this bit or whether we can avoid it. Sorry this is taking way to long to land. I was a way for quite a while and now picking up all the unfinished work I left.

bradleyoesch: @DoctorJohn yeah no problem! I've forgotten about this several times myself :)

I would like to avoid it too, or at least have a cleaner solution than the type name checking going on, without breaking existing behavior too much

DoctorJohn: Hey @bradleyoesch, I just spend some time on this and refactor this part, finally bringing this over the finish line. I hope it's okay I pushed to your branch.

DoctorJohn (approved):
This looks good to me now. This PR makes it easier for devs to spot code errors and adds additional well-writiten tests to the federation integration, which is awesome. Thanks again @bradleyoesch

RELEASE.md
```diff
@@ -0,0 +1,3 @@
+Release type: patch
+
+Update entity resolver exception handling to set the result to the original error instead of a `GraphQLError`, which obscured the original message and meta-fields.
```
```suggestion
Update federation entity resolver exception handling to set the result to the original error instead of a `GraphQLError`, which obscured the original message and meta-fields.
```

patrick91 (approved):

strawberry/federation/schema.py
bradleyoesch: Love it thank you!!

PR #3162, opened 2023-10-23
Add a Quart view

catwell:
## Description

This PR adds a Quart view, as originally asked in [this issue](https://github.com/strawberry-graphql/strawberry/issues/149#issuecomment-671613711) but never actually implemented.

Quart 0.19 being a major release now depending on Flask 3 and Werkzeug 3, I had to bump the Flask support and remove pytest-flask which is incompatible (but AFAIK not required by Strawberry).

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Enhancement/optimization
- [x] Documentation

## Checklist

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the CONTRIBUTING document.
- [x] I have added tests to cover my changes.
- [x] I have tested the changes and verified that they work and don't break anything (as well as I can manage).

PR #3165, opened 2023-10-23
fix: upgrade pytest-flask to 1.3.0 to fix some unit testing issues

bellini666:


patrick91:
From #3162 looks like we can remove pytest-flask :D

botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

bellini666:
> From #3162 looks like we can remove pytest-flask :D

Oh, even better!

Well, let me know if you want to drop this or not. It is probably good to merge, I didn't add any `RELEASE.md` because it only affects tests, so I don't think it is worth producing a new release for it.

patrick91:
I think we can close it, so we don't cause merge conflicts 🗡️

PR #3180, opened 2023-10-30
fix: Remove GlobalID type from relay schema.

dusty-phillips:
## Description

Having this type in the schema violates the relay spec:

https://graphql.org/learn/global-object-identification/
https://relay.dev/docs/guides/graphql-server-specification/#schema

The GlobalID is technically a specialization of the existing strawberry.ID type, and the schema name should be "ID". However, we still want the ability to extract the type and internal ID from the GlobalID.

I don't really think this is the "correct" solution, but I wanted to put it up in case it helps with the discussion I started here: https://github.com/strawberry-graphql/strawberry/discussions/3177

I am open to discussion/suggestion if this is "close" to how @patrick91 wants to do it; otherwise happy to close it without merging once your more correct solution comes out.

This is a breaking change, as anyone who is relying on GlobalID in their relay schema is going to be affected, especially on client code.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* I never filed an issue, but discussion is https://github.com/strawberry-graphql/strawberry/discussions/3177

## Checklist

- [ ] I have updated the documentation accordingly. (I forgot this, looking into it now.)
- [x] I have updated tests to cover my changes.
- [x] I have tested the changes and verified that they work and don't break anything (**as well as I can manage** (I really want to emphasize that part; I'm not very confident in my testing so far.)).


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Dusty Phillips for the PR 👏

```

botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Dusty Phillips for the PR 👏

```

bellini666:
This is something that a lot of users already asked for: a way to have the scalar named `ID` instead of `GlobalID`

There're just 2 issues that I see with this PR:

1) It is a "breaking change" in the API, which makes me worried. Maybe we should do it either way and add a _huge_ `BREAKING CHANGE` notice in the release notes, but @patrick91 might prefer this to be done differently?

2) What to do with users that use both? Like, in your schema if you use both `relay.ID` and strawberry.ID` it will raise issues because you are using different scalars with the same name

patrick91:
> This is something that a lot of users already asked for: a way to have the scalar named `ID` instead of `GlobalID`
> 
> There're just 2 issues that I see with this PR:
> 
> 1. It is a "breaking change" in the API, which makes me worried. Maybe we should do it either way and add a _huge_ `BREAKING CHANGE` notice in the release notes, but @patrick91 might prefer this to be done differently?

Could we make it configurable somehow? It might be painful to maintain though, but at least it gives people more time to upgrade


> 2. What to do with users that use both? Like, in your schema if you use both `relay.ID` and strawberry.ID` it will raise issues because you are using different scalars with the same name

We could throw an error? or make relay.ID an alias?

DoctorJohn:
> Could we make it configurable somehow?

users could do the reverse of this comment (https://github.com/strawberry-graphql/strawberry/issues/3551#issuecomment-2196138923) to rename relay.ID back to GloabalID until they find time to properly migrate.

> We could throw an error?

I think we should. Using both scalars is violating the relay specs anyways

aryaniyaps:
+1. I think this should be merged soon! :)

bellini666:
I think a file inside `docs/breaking-changes` and also a note in the release itself should be enough to workaround the breaking change

Mostly, we need to make sure that the users know that:

`GlobalID` will start to be exposed as `ID` in the schema. It will still behave the same, it is just a change in the name inside the schema itself.

Also add an example on how to expose it as `GlobalID` back for anyone relying on that (I don't think almost anymore will be, unless they are using it together with `ID` itself)

aryaniyaps (approved):

diesieben07:
> Also add an example on how to expose it as GlobalID back for anyone relying on that (I don't think almost anymore will be, unless they are using it together with ID itself)

This assumes people are using Strawberry for internal APIs only. We are using it to create a public-facing API that clients rely on. We cannot just change the schema like that without breaking _all_ of them. And it seems to be impossible (through no fault of Strawberry) to migrate the name of a scalar gracefully, that is, by having an intermediary version of the API where _both_ names are supported. If that's not true I am happy to hear how.

But take this query:
```graphql
query Node($nodeId: GlobalID!) {
  node(id: $nodeId) {
    __typename
    id
  }
}
```
This breaks as soon as we rename `GlobalID` to `ID` and there is no way to have the query accept _both_ a `GlobalID` _and_ a `ID` for the same parameter, because you can't have unions of scalars.

PR #3182, opened 2023-10-30
Add metadata (the graphql.core `Node`) to `GraphQLField` and `GraphQLObject`. 

mgilson:
Clients can use this metadata do specializilation during the creation of client types.

## Description

GraphQL is a very flexible language where custom directives can be added for a large number of purposes.  It's possible that in some cases, a custom directive on a query object or field could be used to impact what types are emitted on a client.  e.g. a client might want to make fields nullable if they have a`@skip` or `@include` directive.  Since directives can be as diverse as the people who make them, we should allow that information to be passed to the client generator plugin which can then be used or ignored as necessary.

Since it's unclear what these custom directives would do or how someone might use them in the general case, for now, I'm just passing the whole graphql node for the field or type through.  In my concrete use-case, I have reasons to want to supply default values to the generated types from the client -- e.g.:

```py
@dataclass
class Foo:
     field: str = "Hello world"
```

Rather than

```py
@dataclass
class Foo:
     field: str
```

which is what gets generated by my plugin currently.

My graphql would look something like:

```graphql
query MyQuery {
    getFoo {
        field @defaultValue(value: "Hello world")
    }
}
```

And this is fine as long as the server is taught to ignore the directive (or the directive is stripped before the query actually gets issued which is how my plugin _actually_ works).

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/strawberry/codegen/query_codegen.py
+++ b/strawberry/codegen/query_codegen.py
@@ -33,6 +33,7 @@ from graphql import (
     ListTypeNode,
     ListValueNode,
     NamedTypeNode,
+    Node,
     NonNullTypeNode,
     NullValueNode,
     ObjectValueNode,
@@ -362,6 +363,8 @@ class QueryCodegen:
                 GraphQLFragmentType,
                 on=typename,
                 graphql_typename=typename,
+                graphql_node=fd,
+                graphql_type=query_type,
             )
 
             self._collect_types(
@@ -487,6 +490,28 @@ class QueryCodegen:
         variables, variables_type = self._convert_variable_definitions(
             operation_definition.variable_definitions, operation_name=operation_name
         )
+        if variables_type:
+            # There are multiple nodes in the `operation_definition.variable_definitions`
+            # list and collectively they are used to make the `variables_type`, so the
+            # closest node to the variable definition is the full operation definition node.
+            if variables_type.graphql_node is not None:  # pragma: no cover
+                raise ValueError(
+                    "Internal codegen error.  graphql_node should NOT be set."
+                )
+            variables_type.graphql_node = operation_definition
+
+            # There isn't actually a class in the strawberry schema that represents the
+            # variables in a query (it's _usually_ a wrapper-class around some
+            # `@strawberry.input` types that are referenced in the resolver function's
+            # arguments, however, the query variable could also be an input to some
+            # directive).  To this end, we currently just use the containing operation type
+            # (e.g. the ``Query`` or ``Mutation`` class and then we'll let consumers sort
+            # out how they want to use this metadata).
+            if variables_type.graphql_type is not None:  # pragma: no cover
+                raise ValueError(
+                    "Internal codegen error.  graphql_type should NOT be set."
+                )
+            variables_type.graphql_type = query_type
 
         return GraphQLOperation(
             operation_definition.name.value,
@@ -519,7 +544,27 @@ class QueryCodegen:
                 variable_type,
             )
 
-            type_.fields.append(GraphQLField(variable.name, None, variable_type))
+            # This field generally represents arguments to resolver functions.  e.g.:
+            #
+            # class Query:
+            #     @strawberry.field
+            #     def some_resolver(self, id: int) -> ReturnType:
+            #         ...
+            #
+            # In this case, there is no `strawberry.type` with `StrawberryField` members
+            # that represent the `id` argument to the query.  Because of that, we do not
+            # set the `strawberry_field` member of the `GraphQLField`. We could potentially
+            # also allow the `GraphQLField` to have an additional field that would hold
+            # an `inspect.Parameter` field that could be used in this case if we feel like
+            # that would be valuable to support in the future.
+            type_.fields.append(
+                GraphQLField(
+                    variable.name,
+                    None,
+                    variable_type,
+                    graphql_node=variable_definition,
+                )
+            )
             variables.append(variable)
 
         return variables, type_
@@ -584,6 +629,7 @@ class QueryCodegen:
             type_ = GraphQLObjectType(
                 strawberry_type.name,
                 [],
+                graphql_type=strawberry_type,
             )
 
             for field in strawberry_type.fields:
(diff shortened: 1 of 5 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

strawberry/codegen/types.py
```diff
@@ -32,6 +34,7 @@ class GraphQLField:
     alias: Optional[str]
     type: GraphQLType
     default_value: Optional[GraphQLArgumentValue] = None
+    definition_node: Optional[Node] = None
```
patrick91, line 37: do you think it would be possible to use the strawberry type here instead?

mgilson: We could do both...

The query is the part that is controlled by the client (they may not control the upstream strawberry type).  I actually have a branch that I haven't quite gotten working that tries to build a client from *only* a graphql schema (then the Python client could be generated against things that don't even use strawberry as the graphql provider).

But... It would be nice to have access to the strawberry type where available.  That's probably easier to work with than digging through the graphql AST...

pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [tool.poetry]
 name = "strawberry-graphql"
 packages = [ { include = "strawberry" } ]
-version = "0.211.1"
+version = "0.211.1.dev1"
```
patrick91, line 4:
```suggestion
version = "0.211.1"
```

strawberry/codegen/query_codegen.py
```diff
@@ -519,7 +522,14 @@ def _convert_variable_definitions(
                 variable_type,
             )
 
-            type_.fields.append(GraphQLField(variable.name, None, variable_type))
+            type_.fields.append(
+                GraphQLField(
+                    variable.name,
+                    None,
+                    variable_type,
+                    definition_node=variable_definition,
```
patrick91, line 530: should this have `strawberry_field` as well?

strawberry/codegen/types.py
```diff
@@ -55,6 +63,8 @@ class GraphQLFragmentType(GraphQLObjectType):
     fields: List[GraphQLField] = field(default_factory=list)
     graphql_typename: Optional[str] = None
     on: str = ""
+    definition_node: Optional[Node] = None
+    strawberry_type: Optional[StrawberryType] = None
```
patrick91, line 67: I wonder about the naming here :) I think I'd make something like this:

```suggestion
    graphql_node: Optional[Node] = None
    graphql_type: Optional[StrawberryType] = None
```

what do you think?

pyproject.toml
mgilson: Oops, that wasn't supposed to make it into the PR 😅

strawberry/codegen/query_codegen.py
mgilson: Hmmm ... originally, I didn't think so ... but now that I think about it a little more, possibly ...

This represents the input variables to the operation.  I guess this is represented by the query/mutation operation object?

Let me see what I can cook up.

patrick91: just for more context, I'd like to keep as abstracted as possible 😊 we might change the internal representation (ie moving away from GraphQL) in future

strawberry/codegen/types.py
mgilson: Sure, I'm not picky about the naming here.

strawberry/codegen/query_codegen.py
mgilson: So ... the more I dig into this, the more that this feels weird (or at least really complex).  For example, you could have a query like:

```graphql
query ($skipTitle: Boolean!, $val: Int!) {
  queryPost(val: $val) {
    id
    title @skip(if: $skipTitle)
    text
  }
}
```

In this case, the `$skipTitle` variable field really has nothing to do with any of the types in the backend schema definition.  The `val` field presumably _does_, but it's buried in the `Query.query_post` field's base_resolver (I guess you'd introspect the wrapped_func on that resolver to understand the backend metadata about this thing?)

For now, I think that I'll make the `graphql_type` on the `{type}Variables` point at the `Query` or `Mutation` class and then users can crawl that class however they want in their plugins.  If we ever see more concrete use-cases, we can make better decisions on how we want to set these pieces of metadata to support those use-cases at that point.

Does that sound reasonable to you?

mgilson:
It _looks_ like pyright released a new version yesterday -- I'm _guessing_ that's why all the pyright tests are now failing :).

womp womp.

strawberry/codegen/types.py
patrick91: let's do the change then 😊

tests/codegen/conftest.py
```diff
@@ -120,6 +130,14 @@ def get_person_or_animal(self) -> Union[Person, Animal]:
         p_or_a.age = 7
         return p_or_a
 
+    @strawberry.field
+    def get_person_with_inputs(self, name: str, age: int = 72) -> Person:
```
patrick91, line 134:
```suggestion
    def get_person_with_inputs(self, name: str, age: int = 72) -> Person:  # pragma: no cover
```

patrick91 (commented):
almost ready to go :)

strawberry/codegen/query_codegen.py
```diff
@@ -487,6 +490,28 @@ def _convert_operation(
...
+
+            # There isn't actually a class in the strawberry schema that represents the
+            # variables in a query (it's _usually_ a wrapper-class around some
+            # `@strawberry.input` types that are referenced in the resolver function's
+            # arguments, however, the query variable could also be an input to some
+            # directive).  To this end, we currently just use the containing operation type
+            # (e.g. the ``Query`` or ``Mutation`` class and then we'll let consumers sort
+            # out how they want to use this metadata).
+            if variables_type.graphql_type is not None:
+                raise ValueError(
```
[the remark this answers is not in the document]
mgilson: Yes, this is defensive to make sure we didn't mess something up.

These are the `{Type}Variables` objects that I mentioned in the other comments.  Basically, at the client level, it's just a dataclass that describes the inputs to a strawberry resolver (rather than being an actual _type_ on the strawberry/python side).

In that case, I made the decision to just set the graphql node to the `Operation` node and I set the graphql type to the query/mutation type.

PR #3188, opened 2023-10-31
Add support for generic interfaces

### LINKED ISSUE #3160: Support GraphQL interfaces which are generic Python types
## Feature Request Type

## Description

This is a bit similar to #1938, but the feature request is to completely support generic Python types as GraphQL interface types. 

The example below is just a toy one, but a practical use case is to e.g. map GraphQL types to database entities that back them.

See this repro schema:

```python
from typing import Any, Self

import strawberry
from graphql import GraphQLResolveInfo, GraphQLType


@strawberry.interface
class MyGenericInterface[T]:
    data: strawberry.Private[T]

    # I figured out the below by trial & error, plus reading this:
    #
    # https://github.com/graphql-python/graphql-core/blob/0c93b8452eed38d4f800c7e71cf6f3f3758cd1c6/src/graphql/execution/execute.py#L1470-L1528
    @classmethod
    def resolve_type(cls, result: Self, info: GraphQLResolveInfo, return_type: GraphQLType) -> str:
        return type(result).__name__

    @strawberry.field
    def value(self) -> str:
        raise NotImplementedError()


@strawberry.type
class ImplementationOne(MyGenericInterface[str]):
    @strawberry.field
    def value(self) -> str:
        return self.data


@strawberry.type
class ImplementationTwo(MyGenericInterface[bool]):
    @strawberry.field
    def value(self) -> str:
        if self.data is True:
            return "true"
        return "false"


@strawberry.type
class ReproQuery:
    # Typing the return value as just `MyGenericInterface` raises this error:
    #
    # TypeError: ReproQuery fields cannot be resolved. The type "<class 'backend.api.graphql.repro.MyGenericInterface'>" is generic, but no type has been passed
    @strawberry.field
    @staticmethod
    async def generic_field() -> MyGenericInterface[Any]:
        return ImplementationOne(data="foo")


repro_schema = strawberry.Schema(
    query=ReproQuery,
    types=[
        # Including `MyGenericInterface`` raises strawberry.exceptions.MissingTypesForGenericError
        ImplementationOne,
        ImplementationTwo,
    ],
)
```

Which generates the following GraphQL schema:

```graphql
schema {
  query: ReproQuery
}

interface AnyMyGenericInterface {
  value: String!
}

type ImplementationOne implements MyGenericInterface {
  value: String!
}

type ImplementationTwo implements MyGenericInterface {
  value: String!
}

interface MyGenericInterface {
  value: String!
}

type ReproQuery {
  genericField: AnyMyGenericInterface!
}
```

It's nice that `MyGenericInterface` exists in the GraphQL schema (that's what I want), but I can't type the `genericField` field as returning it (or else `The type "<class 'backend.api.graphql.repro.MyGenericInterface'>" is generic, but no type has been passed` exception gets raised).

The only thing I could make work was returning `MyGenericInterface[Any]`, but Strawberry creates a new GraphQL (interface) type for it: `AnyMyGenericInterface`. This doesn't work at runtime, raising this error when trying to resolve the field:

```
graphql.error.graphql_error.GraphQLError: Runtime Object type 'ImplementationOne' is not a possible type for 'AnyMyGenericInterface'.
```

https://github.com/graphql-python/graphql-core/blob/0c93b8452eed38d4f800c7e71cf6f3f3758cd1c6/src/graphql/execution/execute.py#L1520-L1525

This makes sense, because in GraphQL-land `ImplementationOne` doesn't implement the `AnyMyGenericInterface` interface.

Would it be possible for tell Strawberry that I want the return type of this field to be just the `MyGenericInterface` interface?

PS: I can think of a few different ways to do it. (1) Use just the generic type without type parameters. (2) Interpret `Any` / `object` type parameters as "I just want the generic interface class". (3) Do something special for GraphQL interface types, i.e. never create concrete versions of them when using Python generics.

patrick91:
Closes #3160

Adds support for generic interfaces. This is very work in progress, the actual implementation
is in a test :"D

The general idea is that we can add a `__class_getitem__` to generic interfaces, so that when specialized,
we can create a concrete copy based on the types passed (like when doing `MyInterface[str]`), this
might break in some cases, but should work for most of them.

We could this to standard types, but I think we handle these cases already, so it might
not be worth doing it (?)

One thing I needed to do to make this work is to replace annotations and dataclasses fields' types
with the actual types, I don't think this should break anything, in fact maybe we should have done this
earlier :)

### DIFF
--- a/strawberry/exceptions/__init__.py
+++ b/strawberry/exceptions/__init__.py
@@ -68,7 +68,6 @@ class InvalidTypeInputForUnion(Exception):
         super().__init__(message)
 
 
-# TODO: this doesn't seem to be tested
 class MissingTypesForGenericError(Exception):
     """Raised when a generic types was used without passing any type."""
 
--- a/strawberry/object_type.py
+++ b/strawberry/object_type.py
@@ -161,13 +161,7 @@ def _process_type(
         resolve_type=resolve_type,
     )
 
-    should_add_class_getitem = is_interface and cls.__strawberry_definition__.is_generic
-
-    # check if any field is generic
-    if not any(getattr(field_.type, "is_generic", None) for field_ in fields):
-        should_add_class_getitem = False
-
-    if should_add_class_getitem:
+    if is_interface and cls.__strawberry_definition__.is_generic:
 
         def __class_getitem__(cls: Type[Any], params: Any) -> Type[Any]:
             params = params if isinstance(params, tuple) else (params,)
--- a/strawberry/object_type.py
+++ b/strawberry/object_type.py
@@ -131,7 +131,7 @@ def _wrap_dataclass(cls: Type[Any]):
 
 
 def _process_type(
-    cls: Type,
+    cls: Type[Any],
     *,
     name: Optional[str] = None,
     is_input: bool = False,
@@ -160,6 +160,19 @@ def _process_type(
         is_type_of=is_type_of,
         resolve_type=resolve_type,
     )
+
+    if is_interface and cls.__strawberry_definition__.is_generic:
+
+        def __class_getitem__(cls: Type[Any], params: Any) -> Type[Any]:
+            params = params if isinstance(params, tuple) else (params,)
+            type_vars = (p.__name__ for p in cls.__parameters__)
+
+            type_var_map = dict(zip(type_vars, params))
+
+            return cls.__strawberry_definition__.copy_with(type_var_map=type_var_map)
+
+        cls.__class_getitem__ = classmethod(__class_getitem__)
+
     # TODO: remove when deprecating _type_definition
     DeprecatedDescriptor(
         DEPRECATION_MESSAGES._TYPE_DEFINITION,
--- a/strawberry/object_type.py
+++ b/strawberry/object_type.py
@@ -161,7 +161,13 @@ def _process_type(
         resolve_type=resolve_type,
     )
 
-    if is_interface and cls.__strawberry_definition__.is_generic:
+    should_add_class_getitem = is_interface and cls.__strawberry_definition__.is_generic
+
+    # check if any field is generic
+    if not any(getattr(field_.type, "is_generic", None) for field_ in fields):
+        should_add_class_getitem = False
+
+    if should_add_class_getitem:
 
         def __class_getitem__(cls: Type[Any], params: Any) -> Type[Any]:
             params = params if isinstance(params, tuple) else (params,)
--- a/strawberry/object_type.py
+++ b/strawberry/object_type.py
@@ -35,13 +35,19 @@ T = TypeVar("T", bound=Type)
 
 
 def _get_interfaces(cls: Type[Any]) -> List[StrawberryObjectDefinition]:
-    interfaces: List[StrawberryObjectDefinition] = []
+    interfaces: Dict[str, StrawberryObjectDefinition] = {}
+
     for base in cls.__mro__[1:]:  # Exclude current class
         type_definition = get_object_definition(base)
         if type_definition and type_definition.is_interface:
-            interfaces.append(type_definition)
+            # TODO: make this better, and add comment on why it is needed
+            if existing_definition := interfaces.get(type_definition.name):
+                if existing_definition.concrete_of == type_definition:
+                    break
 
-    return interfaces
+            interfaces[type_definition.name] = type_definition
+
+    return list(interfaces.values())
 
 
 def _check_field_annotations(cls: Type[Any]):
--- a/strawberry/types/types.py
+++ b/strawberry/types/types.py
@@ -146,7 +146,19 @@ class StrawberryObjectDefinition(StrawberryType):
 
     @property
     def is_generic(self) -> bool:
-        return is_type_generic(self.origin)
+        is_generic = is_type_generic(self.origin)
+
+        # we only treat types that have at least one field with a generic type
+        # as generic, this makes it possible to create generic types that don't
+        # have any generic fields (for example when using `strawberry.Private[T]`)
+        # without affecting the generated schema
+
+        if is_generic:
+            return any(
+                getattr(field_.type, "is_generic", None) for field_ in self.fields
+            )
+
+        return False
 
     @property
     def is_specialized_generic(self) -> bool:
--- a/strawberry/types/types.py
+++ b/strawberry/types/types.py
@@ -26,6 +26,9 @@ from strawberry.utils.inspect import get_specialized_type_var_map
 from strawberry.utils.typing import (
     is_generic as is_type_generic,
 )
+from strawberry.utils.typing import (
+    is_type_var,
+)
 
 if TYPE_CHECKING:
     from graphql import GraphQLAbstractType, GraphQLResolveInfo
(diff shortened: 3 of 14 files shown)


botberry:

<table>
<thead>
</thead>
<tbody>
</tbody>
</table>

Learn more:

botberry:
# Pre-release
:wave:

Releasing commit [6d2773c8bfa479182fdcedfe98f662b89736552b] to PyPi as pre-release! :package:

botberry:
# Pre-release
:wave:

Pre-release **0.212.0.dev.1698770659** [6d2773c8bfa479182fdcedfe98f662b89736552b] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.212.0.dev.1698770659
```

strawberry/exceptions/__init__.py
```diff
@@ -68,7 +68,6 @@ def __init__(self, annotation: GraphQLInputObjectType):
         super().__init__(message)
 
 
-# TODO: this doesn't seem to be tested
```
patrick91, line 71: maybe we should add a test for this, but it is working as expected 😊

strawberry/object_type.py
```diff
@@ -154,6 +160,19 @@ def _process_type(
         is_type_of=is_type_of,
         resolve_type=resolve_type,
     )
+
+    if is_interface and cls.__strawberry_definition__.is_generic:
```
patrick91, line 164: add comments on why this is needed :)

strawberry/types/types.py
```diff
@@ -26,6 +26,9 @@
 from strawberry.utils.typing import (
     is_generic as is_type_generic,
 )
+from strawberry.utils.typing import (
+    is_type_var,
```
patrick91, line 30: remove comma here

strawberry/types/types.py
```diff
@@ -111,6 +114,15 @@ def copy_with(
...
             {"__strawberry_definition__": new_type_definition},
         )
+
+        # update all annotations to point to the new type
+        for name, type_ in self.origin.__annotations__.items():
+            if is_type_var(type_):
+                actual_type = type_var_map[type_.__name__]
+
+                new_type.__annotations__[name] = actual_type
+                new_type.__dataclass_fields__[name].type = actual_type
```
patrick91, line 124: add comments here too, `new_type.__annotations__[name]` is not needed for this use case but it makes sense to be consistent

tests/objects/generics/test_generic_interface.py
```diff
@@ -0,0 +1,42 @@
...
+import pytest
+
+import strawberry
+from strawberry.annotation import StrawberryTypeVar
+from strawberry.type import get_object_definition
+
+T = TypeVar("T")
+
+
+@pytest.mark.xfail(reason="This should be fixed")
```
patrick91, line 12: TODO

tests/schema/test_generics_interface.py
```diff
@@ -0,0 +1,60 @@
+import textwrap
+from typing import Generic, TypeVar
+
+import strawberry
+
+
+def test_supports_generic_interface():
```
patrick91, line 7: I need to add more tests here (?)

tests/types/cross_module_resolvers/c_mod.py
```diff
@@ -1,13 +1,11 @@
...
 
-import a_mod
-import b_mod
-import x_mod
-from a_mod import AObject as C_AObject
-from b_mod import BObject as C_BObject
-
 import strawberry
 
+from . import a_mod, b_mod, x_mod
```
patrick91, line 5: I've added a bunch of missing `__init__.py` files, not needed for this PR, but still :)

tests/schema/test_generics_interface.py
```diff
@@ -0,0 +1,60 @@
...
+        field: T
+
+    @strawberry.type
+    class User(Example[str]):
+        ...
+
+    @strawberry.type
+    class Query:
+        @strawberry.field
+        def example(self) -> User:
```
kkom, line 22: so the use case I had in mind was doing something like this:

```suggestion
        def example(self) -> Example:
```

so that the generated schema would be:

```graphql
        type Query {
          example: StrExample!
        }

        interface StrExample {
          cursor: ID!
          field: String!
        }

        type User implements StrExample {
          cursor: ID!
          field: String!
        }
```

I tested it out and it doesn't work for my schema – raises an exception like this:

```
TypeError: ... fields cannot be resolved. The type "..." is generic, but no type has 
been passed
```

I wonder if what I'm after is even valid GraphQL? 🤔 I hope so!

kkom: Also, in my case the generic field would be private:

```
field: strawberry.Private[T]
```

 so there is no need to expose that type and create an instantiation of the generic

patrick91: thanks! having just private be generic makes things easier in a way, because we don't really need to make a GraphQL copy of the type, I actually just pushed a comment for similar case, but let me add a test for what you need :)

kkom: Amazing, thank you!

patrick91:
@kkom can you test this new pre-release (the [comment here](https://github.com/strawberry-graphql/strawberry/pull/3188#issuecomment-1787594508) should be updated in a few minutes)

this should support your use case, if that's the case I can split this PR in two so we add initial generic support for "generic" interfaces

patrick91:
@kkom apologies, I think it is still not working properly, trying to find a better implementation :)

kkom:
No worries, thanks so much for working on this so actively @patrick91 :)

If I don't get a chance to test it out today, I'll definitely do it tomorrow - or whenever you have an updated version!

tests/schema/test_generics_interface.py
kkom: > thanks! having just private be generic makes things easier in a way, because we don't really need to make a GraphQL copy of the type

Yup! I also think that it's reasonable to just fail in the non-private case.

Looking at the example below IMO it's impossible to tell what the desired GraphQL type should be. We don't know the type of the `field` field.

```python
    @strawberry.interface
    class Example(Generic[T]):
        cursor: strawberry.ID
        field: T

    @strawberry.type
    class Query:
        @strawberry.field
        def example(self) -> Example:
            ...
```

kkom: Similarly, I think it's okay for this to result in a failure as well:

```python
    @strawberry.interface
    class Example(Generic[T]):
        cursor: strawberry.ID
        field: T

    @strawberry.type
    class Query:
        @strawberry.field
        def example(self) -> Example[Any]:
            ...
```

because I don't think it's possible to generate a GraphQL type like this:

```graphql
        interface AnyExample {
          cursor: ID!
          field: Any! # there is no Any type in GraphQL
        }
```

patrick91:
ok, so the check I was doing is a bit more complex than I wanted it to be, I was trying to skip any generic type that didn't expose GraphQL fields from the generic machinery, but it's not only fields that can be generic, also arguments. I need to think about this a bit more. 

I'm off tomorrow, but will be back on it as soon as possible 😊

tests/schema/test_generics_interface.py
patrick91: I think both cases would already fail (at least when not used with interfaces) 😊

patrick91:
going to close this for now, we fixed the main issue in another PR 😊

kkom:
For sure! What's the use that #3202 doesn't address that you have on your mind?

patrick91:
> For sure! What's the use case that #3202 doesn't address which you have on your mind?

GraphQL Generic interfaces, like this

```python
@strawberry.interface
class GenericInterface[T]:
    this_is_exposed_on_graphql: T


@strawberry.type
class MyType(GenericInterface[int]):  # this would create a IntGenericInterface at runtime
    ...
```

PR #3202, opened 2023-11-03
Distinguish between python and graphql generics

patrick91:
So at the moment we treat any generic as a GraphQL generic, this is usually fine,
but there are some cases where this is not ideal, for example in something like this:

```python
@strawberry.type
class Edge[T]:
    cursor: strawberry.ID
    node_field: strawberry.Private[T]
```

where no GraphQL field is actually generic.

This PR changes how we check for generics by checking field types and argument types.

Not 100% sure yet if the implementation is fine, we might have some weird edge cases

This should be a better approach for fixing @kkom use case in #3160


strawberry/field.py
```diff
@@ -340,6 +344,7 @@ def resolve_type(
 
         # If this is a generic field, try to resolve it using its origin's
         # specialized type_var_map
+        # TODO: should we check arguments here too?
```
patrick91, line 347: I'd leave this for the future 😊

tests/python_312/test_python_generics.py
```diff
@@ -0,0 +1,69 @@
+"""
+These tests are for Generics that don't expose any generic parts to the schema.
+"""
```
patrick91, line 3: I've only done these for Python 3.12, it should be fine :D

botberry:



---

WIP



Here's the tweet text: 
```

```

botberry:
# Pre-release
:wave:

Releasing commit [a1b4d16acfd35f6aee009407154f6f0f4d9ef3b8] to PyPi as pre-release! :package:

patrick91:
this broke lazy generics, will check in the next few days :)

botberry:
# Pre-release
:wave:

Pre-release **0.212.0.dev.1699050277** [a1b4d16acfd35f6aee009407154f6f0f4d9ef3b8] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.212.0.dev.1699050277
```

kkom:
Beautiful! I can confirm it works for my use case. Thank you! 🙏 

As expected, I have to add `# pyright: ignore[reportMissingTypeArgument]` comments, since I'm using generic classes without providing type arguments, but this is fair fame. (I'm adding it to the proposed test case below for clarity, but you may not need it give your type checker config.)

Worth adding a test case like this, if you agree this is legit usage?

```python
@strawberry.interface
class GenericInterface[T]:
    data: strawberry.Private[T]

    @strawberry.field
    def value(self) -> str:
        raise NotImplementedError()

@strawberry.type
class ImplementationOne(GenericInterface[str]):
    @strawberry.field
    def value(self) -> str:
        return self.data

@strawberry.type
class ImplementationTwo(GenericInterface[bool]):
    @strawberry.field
    def value(self) -> str:
        if self.data is True:
            return "true"
        return "false"

@strawberry.type
class TestQuery:
    @strawberry.field
    @staticmethod
    async def generic_field() -> GenericInterface:  # pyright: ignore[reportMissingTypeArgument]
        return ImplementationOne(data="foo")
```

```graphql
interface GenericInterface {
  value: String!
}

type ImplementationOne implements GenericInterface {
  value: String!
}

type ImplementationTwo implements GenericInterface {
  value: String!
}

type TestQuery {
  genericField: GenericInterface!
}
```

kkom:
Awesome, will do later in the evening! But the test case looks exactly like what I'm after, so it should be all good :) Thanks!

kkom:
Just tested it, it works exactly as I wanted it to – thanks!

It's been a pretty simple use case so far, I'll let you know if anything more complex doesn't work – but I don't expect any issues really.

patrick91:
I'll merge this later today! So I can get @bellini666 to do a release for Strawberry Django 😊

AntonOfTheWoods:
@patrick91  Obviously my fault but... what is the expected way to force strawberry to generate these? Previously I was getting:

```
LanguageclassesInputCheckpoint
```

Generated when I had only:

```
    @strawberry.field
    async def pull_languageclasses(  # pylint: disable=E0213
        info: Info[Context, Any],
        limit: int,
        checkpoint: Optional[InputCheckpoint[Languageclasses]] = None,
    ) -> Return[Languageclasses]:
 ...
```

And now they aren't. My client (`rxdb`) relies on these existing - do I need to explicitly declare them somehow? If so, any pointers? Thanks!

patrick91:
@AntonOfTheWoods you'd need a field on the generic type to also use the type var

Other than that I don't a good workaround idea 🤔

AntonOfTheWoods:
Thanks... so I guess I'll just pin to < 0.212.0 or recreate the 30 or so types that this got rid of... sniff...

PR #3263, opened 2023-11-29
fix(docs): Channels examples are now working properly.

### LINKED ISSUE #3261: Outdated documentation or a bug: `KeyError: 'ws'`?
I was following the [Channels documentation](https://strawberry.rocks/docs/integrations/channels) trying to implement some Subscription behaviour using Django.  
I went through all the examples but -at the end- I'm still not able to send a chat message as described.

The problem -which cause the server to raise an Exception- is located in the `send_chat_message` mutation example.
In fact, the line `ws = info.context["ws"]` raise the `KeyError` as described in the title.

```py
class Mutation:
    @strawberry.mutation
    async def send_chat_message(
        self,
        info: Info,
        room: ChatRoom,
        message: str,
    ) -> None:
        ws = info.context["ws"]
        channel_layer = ws.channel_layer
 
        await channel_layer.group_send(
            f"chat_{room.room_name}",
            {
                "type": "chat.message",
                "room_id": f"chat_{room.room_name}",
                "message": message,
            },
        )
```

> [!NOTE]
> The `@strawberry.type` decorator is also missing from the example in the documentation.

---

In first place, I thought I was doing something wrong but then I did a little investigating into the two `Consumer` classes.

I started from the [`GraphQLWSConsumer`](https://github.com/strawberry-graphql/strawberry/blob/a90dea2a3053e857aa2adf685d5c4f5d39c527c4/strawberry/channels/handlers/ws_handler.py#L17) which works fine and can access the `ws` key object without any problems.  
It actually provides a [`ChannelsConsumer`](https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/channels/handlers/base.py#L64) object (which has the reference to the `channel_layer`) in the context.

https://github.com/strawberry-graphql/strawberry/blob/a90dea2a3053e857aa2adf685d5c4f5d39c527c4/strawberry/channels/handlers/ws_handler.py#L121-L128

---

I then moved on the [`GraphQLHTTPConsumer`](https://github.com/strawberry-graphql/strawberry/blob/a90dea2a3053e857aa2adf685d5c4f5d39c527c4/strawberry/channels/handlers/http_handler.py#L202), the one that gave me the initial exception.
It seems to provide no `ws` key in the context.

https://github.com/strawberry-graphql/strawberry/blob/a90dea2a3053e857aa2adf685d5c4f5d39c527c4/strawberry/channels/handlers/http_handler.py#L239-L245

In this case, unlike the previous class where `request` and `ws` keys were a reference to the same object, even using `request` instead of `ws` doesn't provide proper access to `channel_layer` since the [`ChannelsRequest`](https://github.com/strawberry-graphql/strawberry/blob/a90dea2a3053e857aa2adf685d5c4f5d39c527c4/strawberry/channels/handlers/http_handler.py#L47) class appears to not have the `channel_layer` property.

---

And here we return to the initial question...  
Is it a bug and the `channel_layer` property is actually missing or is the documentation simply outdated?

If so... How can I properly access the `channel_layer` object within a standard mutation?  
Should I use something like [this](https://channels.readthedocs.io/en/stable/topics/channel_layers.html#using-outside-of-consumers)?

```py
from channels.layers import get_channel_layer

channel_layer = get_channel_layer()
```

Byloth:
This PR fix some inaccuracies with the examples provided on the [Channels documentation](https://strawberry.rocks/docs/integrations/channels) page.  
All specific information is described in the attached issue.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation

* Closes #3261

## Checklist

### DIFF
--- a/docs/integrations/channels.md
+++ b/docs/integrations/channels.md
@@ -66,6 +66,8 @@ First, let's create some Strawberry-types for the chat.
 ```python
 # mysite/gqlchat/subscription.py
 
+import strawberry
+
 
 @strawberry.input
 class ChatRoom:
@@ -104,6 +106,13 @@ Now we will create the chat [subscription](../general/subscriptions.md).
 
 ```python
 # mysite/gqlchat/subscription.py
+
+import os
+import threading
+
+from typing import AsyncGenerator, List
+
+from strawberry.types import Info
 
 
 @strawberry.type
@@ -184,7 +193,10 @@ have to create a mutation for sending messages via the `channel_layer`
 ```python
 # mysite/gqlchat/subscription.py
 
+from channels.layers import get_channel_layer
 
+
+@strawberry.type
 class Mutation:
     @strawberry.mutation
     async def send_chat_message(
@@ -193,8 +205,7 @@ class Mutation:
         room: ChatRoom,
         message: str,
     ) -> None:
-        ws = info.context["ws"]
-        channel_layer = ws.channel_layer
+        channel_layer = get_channel_layer()
 
         await channel_layer.group_send(
             f"chat_{room.room_name}",
@@ -248,6 +259,7 @@ An example of this (continuing from channels tutorial) would be:
 
 ```python
 import os
+
 from channels.auth import AuthMiddlewareStack
 from channels.routing import ProtocolTypeRouter, URLRouter
 from django.core.asgi import get_asgi_application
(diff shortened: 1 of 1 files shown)


patrick91 (approved):
Awesome, thanks!

botberry:

Issue #2413, opened 2022-12-22
Support dependency injection in resolvers


german-glob:
Support dependency injection in resolvers

## Feature Request Type

- [ ] Core functionality
- [x] Alteration (enhancement/optimization) of existing feature(s)
- [ ] New behavior

## Description

In my use case, I have a public _GraphQL_ API and private _FastAPI_ API, both running on the same webserver and sharing some internal logic, so I'd like to be able to reuse my dependencies between them. 

Ideally, I'd like to do something like this:

```
from fastapi import Depends

def some_resolver(
    session_maker = Depends(session_maker),
    store = Depends(store)
):
    async with session_maker() as session:
        return store.search(session=session)
```
or this:
```
from dependency_injector.wiring import Provide, inject

@inject
def some_resolver(
    session_maker = Provide[Container.session_maker],
    store = Provide[Container.store]
):
    async with session_maker() as session:
        return store.search(session=session)
```

Currently, when trying any of those dependency injection methods, `graphql/type/definition.py` raises `TypeError: Query fields cannot be resolved. Unexpected type...`. 

I know I can pass my custom dependencies using GraphQLRouter's `context_getter` param, but that solution sacrifices typehinting.

Is it feasible for Strawberry to stop treating function params as graph query params when they are not `strawberry` types? Or approaching it from other direction, is it possible to introduce a way to 'mark' some params, so they are not treated as query params?

mayteio:
+1, this would be awesome!

jgadling:
I have a rough demo of using Strawberry's FieldExtension to do this. It uses FastAPI internals though, so ymmv if they do a refactor:
https://gist.github.com/jgadling/1971426b0075073ea6d13d64cade1310

patrick91:
@jgadling that's pretty cool! would be interested in having this into Strawberry?

jgadling:
@patrick91 absolutely, if you think it would be a useful addition!

patrick91:
judging from the reaction on this issue I think it is!

I guess ideally we could have a `strawberry.fastapi.field` in future, but I think just having the extension for now it is fine 😊

also /cc @erikwrede, I'm sure you'd like to see this!

mayteio:
> I guess ideally we could have a strawberry.fastapi.field in future, but I think just having the extension for now it is fine 😊

[...]

PR #3274, opened 2023-12-05
Add a way to filter fields

patrick91:
This is a POC for allowing to filter/extend fields from the GraphQL schema, it came from this discord conversation: https://discord.com/channels/689806334337482765/1180117357943853116/1180117357943853116

I think it would be cool to allow filtering, but not sure if this is the nicest way (I don't like having to pass that function around), maybe we could think about using the schema config?

/cc @erikwrede @bellini666 @jkimbo for opinions 😊

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,33 @@
+Release type: minor
+
+This release adds a new method `get_fields` on the `Schema` class.
+You can use `get_fields` to hide certain field based on some conditions,
+for example:
+
+```python
+@strawberry.type
+class User:
+    name: str
+    email: str = strawberry.field(metadata={"tags": ["internal"]})
+
+
+@strawberry.type
+class Query:
+    user: User
+
+
+def public_field_filter(field: StrawberryField) -> bool:
+    return "internal" not in field.metadata.get("tags", [])
+
+
+class PublicSchema(strawberry.Schema):
+    def get_fields(
+        self, type_definition: StrawberryObjectDefinition
+    ) -> List[StrawberryField]:
+        return list(filter(public_field_filter, type_definition.fields))
+
+
+schema = PublicSchema(query=Query)
+```
+
+The schema here would only have the `name` field on the `User` type.
--- a/docs/types/schema.md
+++ b/docs/types/schema.md
@@ -233,3 +233,43 @@ class StrawberryLogger:
 
         cls.logger.error(error, exc_info=error.original_error, **logger_kwargs)
 ```
+
+## Filtering/customising fields
+
+You can customise the fields that are exposed on a schema by subclassing the
+`Schema` class and overriding the `get_fields` method, for example you can use
+this to create different GraphQL APIs, such as a public and an internal API.
+Here's an example of this:
+
+```python
+@strawberry.type
+class User:
+    name: str
+    email: str = strawberry.field(metadata={"tags": ["internal"]})
+
+
+@strawberry.type
+class Query:
+    user: User
+
+
+def public_field_filter(field: StrawberryField) -> bool:
+    return "internal" not in field.metadata.get("tags", [])
+
+
+class PublicSchema(strawberry.Schema):
+    def get_fields(
+        self, type_definition: StrawberryObjectDefinition
+    ) -> List[StrawberryField]:
+        return list(filter(public_field_filter, type_definition.fields))
+
+
+schema = PublicSchema(query=Query)
+```
+
+<Note>
+
+The `get_fields` method is only called once when creating the schema, this is
+not intended to be used to dynamically customise the schema.
+
+</Note>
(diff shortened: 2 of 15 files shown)


bellini666:
I like the idea of using the field's `metadata`. We probably should try to explore that more! :)

> I think it would be cool to allow filtering, but not sure if this is the nicest way (I don't like having to pass that function around), maybe we could think about using the schema config?

I think a schema config for that doesn't exclude the possibility of having that `get_fields`, as it also enables the user to do some extra customization.

Another option (not sure how good it is) for this specifically would be to have a `FieldExtension` for that, like what @erikwrede was doing for the [PermissionExtension](https://github.com/erikwrede/strawberry/blob/2168-permission-field-extension/strawberry/permission.py#L92). That extension can receive a `is_visible_callback` (or a better name) which gets called to check if that field should be included in the schema or not.

skilkis:
Great idea! Defining filters / overrides I think is a great way to make Schema generation more customizable. See this comment here for a related aspect on when to apply FieldExtensions [[1](https://github.com/strawberry-graphql/strawberry/issues/2413#issuecomment-1656584066)]. I like @bellini666's remark on using the field metadata, perhaps we could have multiple levels of filtering (Schema level, field level) where field level takes precedence over global filtering to allow overriding global rules on a per field basis.

patrick91:
> This is a great PoC on schema filtering. Ultimately, I believe the final filtering should not happen inside the schema generator itself, but rather in some extension, following @bellini666's hint on FieldExtensions. 

How do you see that working?

> Dynamically building schemas from the same type source is not an approach I'd recommend, there 
should rather be a possibility to dynamically filter the schemas for your different consumers. 

I'm not sure about the difference about the two approaches?

erikwrede:
> > Dynamically building schemas from the same type source is not an approach I'd recommend, there
> > should rather be a possibility to dynamically filter the schemas for your different consumers.
> 
> I'm not sure about the difference about the two approaches?

The key difference is that you are operating on the schema and the same API endpoint, but introspection and access to types and fields depends on your role during access of the schema.

patrick91:
> > > Dynamically building schemas from the same type source is not an approach I'd recommend, there
> > > should rather be a possibility to dynamically filter the schemas for your different consumers.
> > 
> > 
> > I'm not sure about the difference about the two approaches?
> 
> The key difference is that you are operating on the schema and the same API endpoint, but introspection and access to types and fields depends on your role during access of the schema.

Right, but that's more difficult to implement right, no? We either need to create the schema dynamically per request (which could be expensive, but maybe we can cache it), or we need to customise the introspection to not show these fields, plus also prevent them from being called :)

jonfinerty:
Hi there, I'm original requester of this functionality in the discord thread, so I thought I'd share some of our use-case context, particularly around the dynamic building versus filtering. 

We're attempting to show a subset of our API/types to external integrators, with just the data we're happy to show them (and so generally have a smaller integration surface area). We then broadly extend those graphql types for our own front-end use so we don't have to maintain duplicate python classes. In-fact, we actually have _another_ superset schema on top of that of data/mutations we show/use in internal admin tooling.

Being able to compose an entirely different schema feels quite useful because we can (and I think prefer to) expose it on a different endpoint - in the case of the internal admin tooling we limit access to that endpoint to only our own tooling. The schema itself being more static (it's likely 80% of our schema for example will appear/disappear based on this) feels like it would play with other tooling a bit more neatly. For example we use `strawberry export-schema` a fair bit to do static type checking and codegen in our frontend, and other integrations. 

Hopefully that's useful context, and do I appreciate it's our specific use-case you're trying to adapt to the general case.

patrick91:
@jonfinerty thanks for the context!

I'll add another reference for inspiration: https://graphql-ruby.org/authorization/visibility.html

I'll probably take a day off next week to think about this more 😊

erikwrede:
@patrick91 great source from the ruby docs, this one is more aligned with my suggestion of modifying the field visibility instead of managing two separate schemas. You would still be able to host your schema on separate endpoints, just injecting some endpoint information into the visibility-determining extension.

I prefer this approach for several reasons: first, it leaves schema generation untouched. Secondly, and most importantly, we avoid having future issues asking for swapping out certain types in some of the different sub-schema. This is an antipattern by design and we should avoid enabling users to build towards this kind of scenario. 

I think the main concept we need to extract from this "user story" is the ability to modify the visibility of your schema for different audiences. And to achieve this, my favored approach is to have a single schema underneath that can be re-used, similar to what is described in the ruby docs.

Of course it's more expensive, so I'm opening to discuss.

botberry:



---

This release adds a new method `get_fields` on the `Schema` class.
You can use `get_fields` to hide certain field based on some conditions,
for example:

```python
@strawberry.type
class User:
    name: str
    email: str = strawberry.field(metadata={"tags": ["internal"]})


@strawberry.type
class Query:
    user: User


def public_field_filter(field: StrawberryField) -> bool:
    return "internal" not in field.metadata.get("tags", [])


class PublicSchema(strawberry.Schema):
    def get_fields(
        self, type_definition: StrawberryObjectDefinition
    ) -> List[StrawberryField]:
        return list(filter(public_field_filter, type_definition.fields))


schema = PublicSchema(query=Query)
```

The schema here would only have the `name` field on the `User` type.



Here's the tweet text: 
```

```

bellini666 (approved):
LGTM :)

PR #3309, opened 2023-12-28
docs: use BigInt instead of Int64 as recommendation for large integer scalars

erikwrede:
The most common type for `BigInt`-ish scalars in the GraphQL ecosystem is `BigInt`, not `Int64` (cf. `graphql-scalars` by the Guild & others). I adjusted the documentation to reflect this. Additionally I refactored the recommendation of replacing the default scalar with `BigInt`, as it turns all returned numbers into strings, decreasing schema usability in the average case.

/cc @patrick91 @bellini666

### DIFF
--- a/docs/types/scalars.md
+++ b/docs/types/scalars.md
@@ -282,7 +282,7 @@ BigInt = strawberry.scalar(
 You can adapt your schema to automatically use this scalar for all integers by using the `scalar_overrides` parameter
 
 <Tip>
-Only use this override, if you expect most of your integers to be 64-bit. Since most GraphQL schemas
+Only use this override if you expect most of your integers to be 64-bit. Since most GraphQL schemas
 follow standardized design patterns and most clients require additional effort to handle all numbers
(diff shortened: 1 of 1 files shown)


docs/types/scalars.md
```diff
@@ -262,30 +262,37 @@ date_time = strawberry.scalar(
...
-    description="Int64 field",
+    description="BigInt field",
 )
 ```
 
-Remember to override these changes in your schema instance:
+You can adapt your schema to automatically use this scalar for all integers by using the `scalar_overrides` parameter
+
+<Tip>
+Only use this override, if you expect most of your integers to be 64-bit. Since most GraphQL schemas
```
patrick91, line 285:
```suggestion
Only use this override if you expect most of your integers to be 64-bit. Since most GraphQL schemas
```

patrick91 (approved):
only left a minor comma comment :D

bellini666 (approved):
LGTM :)

patrick91 (approved):

PR #3315, opened 2024-01-03
feat: support union operator on BasePermission

### LINKED ISSUE #2350: Combining permissions with -OR-
In my project I have resolvers that should be acessible if at least one of multiple conditions is met -- aka "OR".

I'm currently implementing it in a single permission class, but it would be ideal if I could combine them. A few examples:

```
@strawberry.type
class Mutation:
    @strawberry.mutation(permission=Permissions.SuperAdmin | Permission.ObjectCreator)
    def update_object(self, object_id: ID, changes: Something) -> MutationResult:
        ...
```

```
@strawberry.type
class ParentChildRelationship:
    @strawberry.field(permission=Permissions.IsParent | Permissions.IsChild)
    def relationship_details(self) -> Something:
        ...
```

## Feature Request Type

- [ ] Core functionality
- [ ] New behavior

replygirl:
## Description

 _**👉 tl;dr:**_ Lets you use `|` between permission instances, e.g. `is_cool_color = IsBlue() | IsGreen()`

- Added: `BasePermission.__or__` instance method, which returns an `OrPermission` instance
- Added: `OrPermission`, inheriting from `BasePermission`
  - `__init__` takes a tuple of `BasePermission` instances to be evaluated
  - `has_permission` returns `True` as long as `has_permission` returns `True` from any of its input permissions
  - `message` and `on_unauthorized` refer to the last failed permission

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

- closes #2350
  - Per the discussion I am also into [DRF's OperandHolder pattern](https://github.com/encode/django-rest-framework/blob/0f39e0124d358b0098261f070175fa8e0359b739/rest_framework/permissions.py#L11). If folks are strongly opinionated I can take a stab at that instead

## Checklist

  - Also fully tested on the job 👍

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,22 @@
+Release type: minor
+
+Permissions can now be combined with the `|` operator to require the user pass _either_ of the `has_permission` checks:
+
+```python
+import strawberry
+from strawberry.permission import PermissionExtension
+
+
+@strawberry.type
+class User:
+    @strawberry.field(
+        extensions=[
+            PermissionExtension(
+                # require auth AND (node is current user OR current user is staff)
+                permissions=[IsAuthenticated(), IsExactUser() | IsStaff()]
+            )
+        ]
+    )
+    def ssn(self) -> str:
+        return "555-55-5555"
+```
--- a/docs/guides/permissions.md
+++ b/docs/guides/permissions.md
@@ -204,6 +204,29 @@ To customize the error handling, the `on_unauthorized` method on
 the `BasePermission` class can be used. Further changes can be implemented by
 subclassing the `PermissionExtension` class.
 
+## `|` operator support
+
+Permissions can be combined with the `|` operator to require the user pass _either_ of the `has_permission` checks:
+
+```python
+import strawberry
+from strawberry.permission import PermissionExtension
+
+
+@strawberry.type
+class User:
+    @strawberry.field(
+        extensions=[
+            PermissionExtension(
+                # require auth AND (node is current user OR current user is staff)
+                permissions=[IsAuthenticated(), IsExactUser() | IsStaff()]
+            )
+        ]
+    )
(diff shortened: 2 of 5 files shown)


botberry:



---

Permissions can now be combined with the `|` operator to require the user pass _either_ of the `has_permission` checks:

```python
import strawberry
from strawberry.permission import PermissionExtension


@strawberry.type
class User:
    @strawberry.field(
        extensions=[
            PermissionExtension(
                # require auth AND (node is current user OR current user is staff)
                permissions=[IsAuthenticated(), IsExactUser() | IsStaff()]
            )
        ]
    )
    def ssn(self) -> str:
        return "555-55-5555"
```



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to replygirl for the PR 👏

```

erikwrede (changes requested):
This looks like a great usability addition. I will conduct more thorough testing later, but for now, can you please add the following test cases so we don't accidentally break something later:

- two resolvers use the same OR filter -> directive duplication doesn't cause an error
- permutation of entries in the OR filter in two separate resolve (My preference is to use the same directive in this case and not create two separate directives, i.e. always `AorB` instead of `AorB` and `BorA`. We can sort this out by ordering the permission names lexicographically using sort)

How should we handle multiple ORs that are chained? Should the or-filter flat map them into a single filter ,or should there be nested OR-filters? In this case, there should also be test cases on schema generation 

- More than 2 permissions disjuncted, (`AorBorC`)

replygirl:
all sounds good! i should have time to update in the next few days

erikwrede:
@replygirl the only problem I see with nesting is the handling of schema directives. Maybe they need to be flatMapped in this case, but this should be easy to handle since only the outermost OR-extension will be respected. I'm open to any solutions 😊

PR #3333, opened 2024-01-10
Fix usage with default and default_factory arguments

devkral:
## Description

Currently it is not always possible to use the default and default_factory argument on strawberry.field.
It fails randomly and especially on root Query
This PR fixes the issue.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/tests/schema/test_basic.py
+++ b/tests/schema/test_basic.py
@@ -580,3 +580,43 @@ def test_kw_only():
             FooBar(foo=1)
         FooBar(bar=2)
         FooBar(foo=1, bar=2)
+
+
+def test_use_default_factory_on_root():
+    @strawberry.type
+    class Sub:
+        foo: str = strawberry.field(default="bar")
+
+    @strawberry.type
+    class Query:
+        sub: Sub = strawberry.field(default_factory=Sub)
+
+    schema = strawberry.Schema(query=Query)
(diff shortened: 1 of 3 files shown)


botberry:



---

Fix using strawberry with default or default_factory (dataclasses basics)



Here's the tweet text: 
```

```

tests/schema/test_basic.py
```diff
@@ -580,3 +580,43 @@ class FooBar2:
...
+
+    result = schema.execute_sync(
+        """
+        query TestQuery{
+            sub {
+                foo
+            }
+        }
+        """
+    )
```
patrick91, line 604: isn't this fixed by passing a root value to schema.execute? 😊

devkral: sounds good. There should be a default root value in execute instead of None

patrick91: I get you, but I think that would make us quite opinionated, and also we can't guess how all types will be instantiated, this would only work if types don't require any argument 😊

devkral: This task isn't easy as you have to provide a mutation or query instance to execute.
this is a big task

PR #3363, opened 2024-01-28
Fix custom fruit pagination example

eudago:
I was reading the relay pagination and found this little bug, indeed the code works, but not as expected (you can find more information in this discussion [thread](https://github.com/strawberry-graphql/strawberry/discussions/3061)).

## Description

Fixed Fruit Custom Pagination to understand the relay pagination 

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation

### DIFF
--- a/docs/guides/relay.md
+++ b/docs/guides/relay.md
@@ -296,7 +296,10 @@ class FruitCustomPaginationConnection(relay.Connection[Fruit]):
 
 @strawberry.type
 class Query:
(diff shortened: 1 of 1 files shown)


DoctorJohn (approved):
Thanks for the PR @eudago! Looks like the docs only outlined how to use a custom paginated connection, your changes make the example behave like expected.

Issue #2942, opened 2023-07-12
Move StrawberryConfig to a typed dict


patrick91:
as suggested in https://github.com/strawberry-graphql/strawberry/pull/2698#issuecomment-1631030171

we should change strawberry's schema config to be a type dictionary, I think this will make the dx a bit nicer, since you can do this:

```
import strawberry

...

schema = strawberry.Schema(..., config={"auto_camel_case": False})
```

which is a reduction of one import (and bit less characters) 😊

The main disadvantage for users it that they might not get autocompletion if their editor is not setup well.

The main disadvantage for us is that it might get a bit cumbersome to get values (or fallback to default values).

In any case if we do this, we should deprecate the old config, and write a codemod, which should be easy to do 😊

skilkis:
@patrick91 https://peps.python.org/pep-0705/ might be useful

jgadling:
As an alternative to TypedDict, which is pretty basic - the pydantic_settings module has an *excellent* configuration interface: https://docs.pydantic.dev/latest/usage/pydantic_settings/

It allows setting default values, enforces type/value validation, can take overrides from env vars by default, and be extended to read settings from a variety of different sources (config files, etc).

patrick91:
@jgadling thanks for the suggestion! I think that would be neat, but I'd prefer to keep the base dependencies at a minimum if possible 😊

Checho3388:
Hi @patrick91 ,
I'd like to contribute on this one. If you're Ok I'll create a MR with a proposal.

Changing `straweery.Schema` to expect a dict instead of a dataclass, would change the signature of the constructor in an incompatible manner. Have I understood this properly? Would this be a major change in `RELEASE.md` then?

youngjaekwon:
Hi @patrick91, I'm interested in tackling this one. Had a few questions before diving in:

1. For the existing `StrawberryConfig` dataclass — are we deprecating it with a warning, or keeping both the dataclass and dict approaches as permanent options?

2. What about `_unsafe_disable_same_type_validation`? It's prefixed with `_` but users are the ones setting it. Should it show up in the new TypedDict or stay exclusive to the dataclass?

3. The issue mentions a codemod — mind if I split that into a follow-up PR? I'd like to keep the first one focused on just the dict support.

Here's what I'm thinking for the approach:
- Create a `StrawberryConfigDict` (TypedDict, `total=False`)
[...]

PR #3365, opened 2024-01-29
Move Strawberry config to a typed dict

Checho3388:
## Description

This MR aims to change Strawberry's schema config to be a typed dictionary. Then only one import would be required to instantiate the schema.
```python
import strawberry
...
schema = strawberry.Schema(..., config={"auto_camel_case": False})
```
In the `Schema` class initialization the former `dataclass` is used to handle default values and validations. Then, my approach for this fix was not to remove the old dataclass. 

UTs were updated accordingly.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* #2942 

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: major
+
+Make schema config parameter a TypedDict
--- a/docs/types/schema-configurations.md
+++ b/docs/types/schema-configurations.md
@@ -13,15 +13,13 @@ example below:
 ```python
 import strawberry
 
-from strawberry.schema.config import StrawberryConfig
-
 
 @strawberry.type
 class Query:
     example_field: str
 
 
-schema = strawberry.Schema(query=Query, config=StrawberryConfig(auto_camel_case=False))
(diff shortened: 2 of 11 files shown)


botberry:



---

Make schema config parameter a TypedDict



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to @cheche3388 for the PR 👏

```

patrick91 (commented):
nice! I'll take a proper look soon, but I was wonder if you're interested in writing a codemod for this too? 😊

We have one here: https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/codemods/annotated_unions.py

Also I think we could keep the name of the dict the same, since it is mostly backwards compatible (`dict(a=1)` is almost the same as `{"a": 1}`)

Checho3388:
Hi @patrick91 ,
Sure, I just wrote the codemod (at least an attempt to do it!). Please let me know your thoughts.
Regards

PR #3379, opened 2024-02-11
prototype: object type extending

Kitefiko:
Prototype for https://github.com/strawberry-graphql/strawberry/issues/2605


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Kitefiko for the PR 👏

```

strawberry/types/type_extension.py
```diff
@@ -0,0 +1,56 @@
...
+
+    from graphql import GraphQLAbstractType, GraphQLResolveInfo
+
+    from strawberry.field import StrawberryField
+    from strawberry.type import WithStrawberryObjectDefinition
+
+
+class TypeExtension:
+    def before_wrap_dataclass(self, cls: Type) -> None:
+        pass
```
bellini666, line 18: Maybe we should use something similar to our extensions api and define this as `on_wrap_dataclass`, in which the user can `yield` inside, meaning he can call "before" code before it, and "after" code after it. That also allows they to except exceptions

What do you think?

strawberry/types/type_extension.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def on_field(self, field: Field | StrawberryField) -> Any:
+        return field
```
bellini666, line 21: Maybe the same here

Kitefiko:
Just rebased and adjusted so solution is no longer outdated.

bellini666:
I like where this is going...

Checking it today I'm wondering if we should alllow a list of type extensions instead of just one.

Then when doing, `field = extension.on_field` you just replace it with

```python
for extension in extensions:
    field = extension.on_field(extension)
```

And so on.

This would allow for much more customization.

What do you think about this @patrick91 ?

patrick91:
> Checking it today I'm wondering if we should alllow a list of type extensions instead of just one.

yup, we do this for fields already 😊

Kitefiko:
> Then when doing, `field = extension.on_field` you just replace it with
> 
> ```python
> for extension in extensions:
>     field = extension.on_field(field)
> ```

> This would allow for much more customization.

There are 2 reasons why I didn't do that. 
- Reduce performance hit to minimum - no looping (this _might_ not matter tho)
- Keep it more generic. Since you already have access to lower level and thus can modify entire object (including fields), it seemed redundant to enforce this via library

You could do something like
```python
class MyTypeExtension(TypeExtension):
    def __init__(self, field_extensions: Iterable[...]):
        self._field_extensions = field_extensions

    def on_field(self, field: dataclasses.Field[Any]):
        for ext in self._field_extensions:
            field = ext.apply(field)
        return field
```

What do you guys think?

Also could you explain a little more how you envision the usage of context manager styled hooks @bellini666 ?

bellini666:
> > Then when doing, `field = extension.on_field` you just replace it with
> > ```python
> > for extension in extensions:
> >     field = extension.on_field(field)
> > ```
> 
> > This would allow for much more customization.
> 
> There are 2 reasons why I didn't do that.
> 
> * Reduce performance hit to minimum - no looping (this _might_ not matter tho)
> * Keep it more generic. Since you already have access to lower level and thus can modify entire object (including fields), it seemed redundant to enforce this via library

IMO I think it is still worth it. This only runs once, and the performance hit here should be insignificant.

> Also could you explain a little more how you envision the usage of context manager styled hooks @bellini666 ?

Hrm, I think the code that I mentioned changed now, but I was thinking in something like this: https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/extensions/base_extension.py#L27 . With that you can have some code that executes "before" and "after" without having to define 2 functions for that.

But I don't think that makes sense for the `on_field` for example, as it is supposed to be able to actually _change_ the field.

strawberry/experimental/pydantic/object_type.py
```diff
@@ -177,7 +178,7 @@ def wrap(cls: Any) -> Type[StrawberryTypeFromPydantic[PydanticModel]]:
         )
 
         wrapped = _wrap_dataclass(cls)
-        extra_strawberry_fields = _get_fields(wrapped, {})
+        extra_strawberry_fields = _get_fields(wrapped, TypeExtension(), {})
```
erikwrede, line 181: Instantiating an empty type extension feels wrong to me.

strawberry/object_type.py
```diff
@@ -135,17 +135,19 @@ def _process_type(
...
     original_type_annotations = original_type_annotations or {}
 
     interfaces = _get_interfaces(cls)
-    fields = _get_fields(cls, original_type_annotations)
+    fields = _get_fields(cls, extension, original_type_annotations)
     is_type_of = getattr(cls, "is_type_of", None)
     resolve_type = getattr(cls, "resolve_type", None)
 
-    cls.__strawberry_definition__ = StrawberryObjectDefinition(
+    cls.__strawberry_definition__ = extension.create_object_definition(
```
erikwrede, line 150: I'm not sure about letting the extension influence the instantiation of the strawberry object definition. This severely limits our ability to add new parameters to the constructor later on without breaking the public API. Addtionally, this implies that only one type extension is supported on a type.

However, having the extension change all of the fields manually after creation could also lead to worse DevX. Open for other opinions on this /cc @strawberry-graphql/core

erikwrede (commented):
I'm a bit late to the party, but thanks for picking this up again! I see that it picks up some ideas from my initial concept, but also opens up to some changes. I've added a few comments that came to my mind immediately. 

The main open point I see is wether we want to support one or multiple type extensions on the same type. My initial draft aimed for the support of multiple extensions, however I am open to other opinions and curious to hear why you chose to take this approach.

strawberry/experimental/pydantic/object_type.py
Kitefiko: This I currently take as a temp code, I am planning on implementing `PydanticTypeExtension` once API is agreed upon.

Kitefiko:
> The main open point I see is wether we want to support one or multiple type extensions on the same type. 

Yes. This is currently blocker right now. I would really appreciate  for more people to weight in. 
> My initial draft aimed for the support of multiple extensions, however I am open to other opinions and curious to hear why you chose to take this approach.

I have touched on this in previous comment but the main points IMO are:

__Single__
- Ability to hook into objects' field creation
- Ability to directly inherit and modify `StrawberryObjectDefinition` used for `__strawberry_definition__`. This allows for custom attrs, methods, `has_MY_object_definition`, `get_MY_object_definition`, ...

__Multiple__
- Ability to hook into objects' field creation
- You can use multiple `TypeExtension` out of the box
- Ability to add extra info via `metadata`? field on `StrawberryObjectDefinition`
- Strawberry internals are less exposed
- Non-official (current approach) ability to add custom attrs, methods etc. via `setattr` e.g. `__strawberry_django_definition__`

You could create `TypeExtension` that would act as __Multiple__ `TypeExtension`, but not vice versa.
I think that __Multiple__ kinda clashes with and takes responsibility from `strawberry.field(extensions=[...]` and `FieldExtension` concept.

Kitefiko:
Hey,
got some time to look into this again and reevaluated previous approach such that:
- none or multiple extensions may be used
- the API is the same as for `field`
- `TypeExtension` hooks are (mostly unchanged)
  - `on_wrap_dataclass` - before `_wrap_dataclass` and after `_process_type`, context manager approach (e.g. `SchemaExtension.on_operation`)
  - `on_field` - callback function that requires return of Field
  - `on_object_definition` - called after `StrawberryObjectDefinition` is created and is passed in as param (this is the biggest change)
  
 
Should `StrawberryObjectDefinition` have `extensions` (same as `StrawberryField`)?
Should `StrawberryObjectDefinition` have `metadata` for custom data from extensions?

I think that this approach addresses most suggestions above and is less intrusive. 
Please take a look, I am looking forward to feedback.

Issue #3377, opened 2024-02-09
Fields defined with `... = field(resolver=...)` are treated like dataclass attributes by type checkers


lukaspiatkowski:
## Describe the Bug
The code below doesn't give any warnings using `mypy` with Strawberry plugin or `PyRight`:
```
from strawberry import field, type


def resolver() -> int:
    return 42


@type
class Foo:
    field1: str
    field2: int = field(resolver=resolver)

    @field
    def field3(self) -> str:
        return "Hello, world!"


reveal_type(Foo.__init__)  # def (self: interface.graphql.schema.Foo, *, field1: builtins.str)
reveal_type(Foo.field1)  # builtins.str
reveal_type(Foo.field2)  # builtins.int
reveal_type(Foo.field3)  # strawberry.field.StrawberryField

print(Foo(field1="test").field2)
# In runtime the above prints
# <bound method resolver of Foo(field1='test')>


def bar(foo: Foo) -> int:
    # This should not work, foo.field2 is a method
    return foo.field2


def baz(foo: Foo, val: int) -> None:
    # This should not work, foo.field2 is a method
    foo.field2 = val
```

So in short: when you define `field2: int = field(resolver=resolver)` the type checkers will treat `field2` as an attribute on the instances of your class and allow you to read from or write to that attribute as if it was plain `int` and not a resolver. It leads to runtime errors.

## System Information

 - Python: 3.11.5
 - Mypy: 1.7.0
 - PyRight: 1.1.331
 - Strawberry: 0.219.2

## Additional Context

First of all I wanted to thank you for this amazing project. I've spend 5 months on rewriting our codebase with ~1000 objects and ~1000 resolvers from Graphene to Strawberry and the type improvements we got from it are amazing. There were few issues on the way, mostly due to our customizations, but this one bothered me as Strawberry didn't help to catch it.

The original Graphene code I had was something like this:
```
def resolve_field2(self: Foo) -> int:
    # do some auth checks
	return self.field2

class Foo(graphene.ObjectType):
	field2: int

Foo.resolve_field2 = resolve_field2
```
Due to rewriting a lot of objects at once I've used a codemod, so I haven't caught a bug in this version:
```
def resolve_field2(self: Foo) -> int:
    # do some auth checks
	return self.field2

@strawberry.type
class Foo:
	field2: int = strawberry.field(resolver=resolve_field2)
```
Running this code results in a runtime error, because `self.field2` is a method, not an integer.
[...]

PR #3383, opened 2024-02-15
Add Resolver type, use it as a return type for field and fix tests

lukaspiatkowski:
## Description

Change one of the overrides of `strawberry.field` to return `-> Resolver[T]` if `resolver = ...` is provided. This forces you to annotate your fields as `field: Resolver[int] = ...` and since `Resolver` is an abstract final class you cannot create an instance of it. This will prevent you from mistakenly writing or reading that field.

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* #3377

## Checklist


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Lukasz Piatkowski for the PR 👏

```

patrick91:
couldn't we use Callable instead of Resolver? 😊

lukaspiatkowski:
I will fill up the documentation and release notes as soon as someone tells me there is a chance for merging this PR in :) Also my approach to writing the `Resolver` class itself is a bit hacky. I want to maintain my own fork with this change if it doesn't get merged, so I wanted to avoid changing core schema converting functions. I am fine fixing this part.

lukaspiatkowski:
> couldn't we use Callable instead of Resolver? 😊

Hmm, yes now that you say that we probably could do it, although it would be a bit of pain. Note that `Callable` is parametrized over 2 generic (`Callable[Param, Ret]`), in this case we don't care about `Param` part. But AFAIK Python doesn't let you yet write type aliases that parametrize part of the Generic, like so:
```
Resolver: TypeAlias = Callable[Any, T]

    field: Resolver[int] = strawberry.field(resolver=...)
```
So you would have to write the whole thing every time:
```
    field: Callable[Any, Int] = strawberry.field(resolver=...)`
```
which is IMO a mouthfull and not necessary correct - people could try and call this function, but without proper typing they wouldn't get the arguments right.

lukaspiatkowski:
Oh, I see that you are using libcst for codemoding backward incompatible changes. I was about to write one for myself, but I can add it here in this PR. Stay tuned! :)

lukaspiatkowski:
Codemod added as promised

lukaspiatkowski:
One thing that I've noticed when using this new type on my codebase was that if you have an interface it might get difficult to get the types right. Consider this code:
```
@interface
class Base:
    a: int

@type
class A(Base):
    a: int = field(resolver=...)

@type
class B(Base):
    a: int
```
With the new `Resolver` type you won't be able to write:
```
@interface
class Base:
    a: int

@type
class A(Base):
    a: Resolver[int] = field(resolver=...)

@type
class B(Base):
    a: int
```
because the child class `A` changes the type of the base class `Base`. And for good reason! You wouldn't want this code to work without warnings:
```
def foo(base: Base) -> int:
    return base.a

def bar(a: A) -> int:
    return foo(a)
```

So for future reference if you stumble on this problem while changing your codebase my suggestion would be to write something like this:
```
@interface
class Base:
    a: Resolver[int]

@type
class A(Base):
    a: Resolver[int] = field(resolver=...)

@type
class B(Base):
	_a: Private[int]
    a: Resolver[int] = field(resolver=lambda self: self._a)
```

patrick91:
@lukaspiatkowski not sure, sorry I didn't have time to take a proper look at this, I have set some time to look at PRs today, hopefully I get to this one

lukaspiatkowski:
Just friendly pinging @patrick91  here :)

patrick91:
@lukaspiatkowski sorry, still haven't had the time to properly think about this, I'll add it to my things for next week 😊

erikwrede:
Hey, I just went through your PR but I'm not fully into the details yet. Couldn't we also solve this with `@overload`, respecting wether a resolver was passed or not, as well? This would remove the need for additional resolver types or "Faking" object definitions. Need to try this out soon, but feel free to lmk what you think 😊

lukaspiatkowski:
@erikwrede not sure what you are proposing as this solution is changing the overload case for when the resolver is passed. Introducing the Resolver type is the core of this design, so removing the need for it would be missing the point here :)
The "Fake" part isn't pretty though, I agree. It was the fastest non-intrusive way of handling the addition of Resolver type (I had a feeling that you won't be merging this PR soon, so I kept it simple to be able to easily maintain my fork). Probably adding handling of Resolver somewhere deeper in the code would look better.

PR #3394, opened 2024-02-24
Add tests for python code in documentation

Sirusblk:
## Description

Add tests for python code in documentation. Adds a new dev dependency: `pytest-examples` and with that package is a single parameterized pytest in `tests/docs/test_docs.py` which will automatically search the docs folder for any python code examples in 

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: minor
+
+Adds documentation tests for python code blocks in any markdown files in the `docs` folder. The tests are marked as flaky and allowed to fail. When testing the examples, imports for `strawberry` or the `typing` packages are automatically included.
--- a/poetry.lock
+++ b/poetry.lock
@@ -148,13 +148,13 @@ files = [
 
 [[package]]
 name = "anyio"
-version = "4.2.0"
+version = "4.3.0"
 description = "High level compatibility layer for multiple asynchronous event loop implementations"
 optional = false
 python-versions = ">=3.8"
 files = [
-    {file = "anyio-4.2.0-py3-none-any.whl", hash = "sha256:745843b39e829e108e518c489b31dc757de7d2131d53fac32bd8df268227bfee"},
-    {file = "anyio-4.2.0.tar.gz", hash = "sha256:e1875bb4b4e2de1669f4bc7869b6d3f54231cdced71605e6e64c9be77e3be50f"},
+    {file = "anyio-4.3.0-py3-none-any.whl", hash = "sha256:048e05d0f6caeed70d731f3db756d35dcc1f35747c8c403364a8332c630441b8"},
+    {file = "anyio-4.3.0.tar.gz", hash = "sha256:f75253795a87df48568485fd18cdd2a3fa5c4f7c5be8e5e36637733fce06fed6"},
 ]
 
 [package.dependencies]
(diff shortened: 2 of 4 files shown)


botberry:



---

Adds documentation tests for python code blocks in any markdown files in the `docs` folder. The tests are marked as flaky and allowed to fail. When testing the examples, imports for `strawberry` or the `typing` packages are automatically included.



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to David McLaren for the PR 👏

```

botberry:



---

Adds documentation tests for python code blocks in any markdown files in the `docs` folder. The tests are marked as flaky and allowed to fail. When testing the examples, imports for `strawberry` or the `typing` packages are automatically included.



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to David McLaren for the PR 👏

```

DoctorJohn (approved):
Awesome proposal @Sirusblk! Your changes look good to me, I cross-checked them with the [pytest-examples docs](https://github.com/pydantic/pytest-examples).

The CI failures seem to be present on the main branch as well. I'll approve this PR, but will wait for our CI to go green again before merging :)

Sirusblk:
I'll see if I can find a solution to making sure the schema matches the python code snippet. I'll also look into seeing if we can test the response as well.

> 2. We also need to figure out how to test the examples, prints might be good, but I think they make the snippets quite verbose, but I really like that pytest-examples has a way to print them. We could also do something custom custom for testing schemas, like here: ...

[pytest-examples docs](https://github.com/pydantic/pytest-examples?tab=readme-ov-file#check-print-statements) do show we can handle prints like so:

```python
print(1 + 2)
#> 3
```
 
If you like this I can go ahead and try this out where appropriate. If the syntax doesn't look good we could try assert statements or roll our own syntax like you suggested.

We do reuse a lot of the same structures we could inject some prebuilt data structures into the module namespace when testing to avoid repeat code in each example snippet. For testing schemas, we might be able to similarly use the extracted code examples and be able to compare afterward. I'll poke around and see.

> Also I think it is worth spending some time to make this a test that doesn't fail. We could introduce this for just one file (maybe the getting started) or we could try to fix as many things as possible.

I marked them as flaky so it wouldn't block pytest runs but I can remove that and limiting the scope to possibly a file or two with the documentation updated to get started I figured trying to rewrite all the examples all at once might be too much for a single MR. I could tackle this with a string of MRs once we have a solid base to work from.

Thanks for the feedback!

patrick91:
I think for now it would be fine to just go with this:

```
print(1 + 2)
#> 3
```

we can make it better in future, I'm afraid I'd do to much bike shedding :D 

For the imports, I think it's fine to leave them in, I'll test if we can hide them from the docs side 😊 maybe we do a flag like this:

```python hide-imports
import strawberry

@strawerry.type
class P:
    a: str
```

Sirusblk:
@patrick91 I figured out a way to test `python+schema` code snippets. I disabled the flaky flag and got all the general docs working. I added a way to keep the global namespace around for code snippets in the same doc file but it relies on global variables which I don't love.

I'm still trying to find a good way to handle `graphql+response` code snippets.

Sirusblk:
Never mind I adapted some of the `pytest_examples` module code to work with other code snippet types. I added another file to demonstrate those checks. It caught some good minor issues already.

PR #3408, opened 2024-03-14
Boolean expression permissions

### LINKED ISSUE #3407: Allow Boolean operators for Permissions
Currently, permissions are provided as a list of `BasePermission` instances, treated implicitly as ANDs.  It would be grand to instead be able to use AND and OR operators when passing in permissions.

## Feature Request Type

- [ ] Core functionality
- [ X] Alteration (enhancement/optimization) of existing feature(s)
- [ ] New behavior

## Description

This has been discussed on discord.  Proposed approach is overriding the `&` and `|` operators for `BasePermission`,  which would allow permissions to be passed in as follows:

`IsAuthed() & (ControlsAccount() | IsAdmin())`

This may need some changes to the `on_unauthorized` and `has_permission` interface, as the separation / statelessness between `has_permission` and `on_unauthorized` means that a parent permission's `on_unauthorized` wouldn't know WHICH of its child permissions failed.

Additionally, some thought will be needed to handling the combination of sync and async combined permissions

vethan:
Adding the ability to perform boolean expressions on permissions using the permission extension

## Description

Allows permissions to be supplied in the form `IsAuthed() & (ControlsAccount() | IsAdmin())`, by overriding the implementation of permissions on `BasePermission`

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* Closes #3407 
* Closes #3315

## Checklist


botberry:



---

Adds the ability to use the `&` and `|` operators on permissions to form boolean logic. For example, if you want
a field to be accessible with either the `IsAdmin` or `IsOwner` permission you
could define the field as follows:

```python
import strawberry
from strawberry.permission import PermissionExtension, BasePermission


@strawberry.type
class Query:
    @strawberry.field(
        extensions=[
            PermissionExtension(permissions=[IsAdmin() | IsOwner()], fail_silently=True)
        ]
    )
    def name(self) -> str:
        return "ABC"
```



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to @vethan4 for the PR 👏

```

ThirVondukr:
Do we need to have new public methods `resolve_permission_async` and `resolve_permission_sync`?
You can just return a `bool | Awaitable[bool]` from has_permission based on if any of the underlying permissions are actually async, and I don't think you need to have `is_async` property either
It would be a good thing to not extend public API without a need for that 🤔

vethan:
> Do we need to have new public methods `resolve_permission_async` and `resolve_permission_sync`? 

This may just be my lack of pythonic syntax knowledge creeping up, but I'm not sure if I can return the Union from the parent class without the knowledge of the children's return type. (Aka, does the AND need to await to figure out it's own permission or not)

strawberry/permission.py
```diff
@@ -87,7 +100,81 @@ class AutoDirective:
...
+
+    def __or__(self, other: BasePermission):
+        return OrPermission(self, other)
+
+
+class BoolPermission(BasePermission, abc.ABC):
+    left: BasePermission
+    right: BasePermission
+
+    def __init__(self, left: BasePermission, right: BasePermission):
```
erikwrede, line 120: This should accept a list of conditions to support `X & Y & Z` without nesting

strawberry/permission.py
```diff
@@ -151,9 +238,14 @@ def resolve(
...
-            if not permission.has_permission(source, info, **kwargs):
-                return self._on_unauthorized(permission)
+        try:
+            for permission in self.permissions:
+                permission.resolve_permission_sync(source, info, **kwargs)
+        except BaseException as e:
+            if self.fail_silently:
+                return [] if self.return_empty_list else None
+            else:
+                raise e
```
erikwrede, line 248: Why did you remove the on_onauthorized call here? We also want to avoid catching all errors here. There may be some unexpected exceptions unrelated to permissions resulting from the has_permission call.

strawberry/permission.py
```diff
@@ -163,13 +255,14 @@ async def resolve_async(
...
-            if not has_permission:
-                return self._on_unauthorized(permission)
+        try:
+            for permission in self.permissions:
+                await permission.resolve_permission_async(source, info, **kwargs)
+        except BaseException as e:
+            if self.fail_silently:
+                return [] if self.return_empty_list else None
+            else:
+                raise e
```
erikwrede, line 265: see above

strawberry/permission.py
```diff
@@ -56,12 +56,25 @@ def has_permission(
...
+            return self.on_unauthorized()
+
+    async def resolve_permission_async(
+        self, source: Any, info: Info, **kwargs: Any
+    ) -> None:
+        if await await_maybe(self.has_permission(source, info, **kwargs)):
+            return None
+        else:
+            return self.on_unauthorized()
+
```
erikwrede, line 72: Please keep this in the permission extension. The extension has all the logic to handle what to do on failure (e.g. fail silently). The control flow is less clear like this and we need to catch exceptions and rethrow them.

strawberry/permission.py
```diff
@@ -74,7 +87,7 @@ def on_unauthorized(self) -> None:
         raise error
 
     @property
-    def schema_directive(self) -> object:
+    def schema_directive(self) -> List[object]:
```
erikwrede, line 90: This is a Breaking change, can we instead introduce some new directive logic for @And and @Or? Let's discuss this in the PR

strawberry/permission.py
```diff
@@ -56,12 +56,25 @@ def has_permission(
             "Permission classes should override has_permission method"
         )
 
+    def resolve_permission_sync(self, source: Any, info: Info, **kwargs: Any) -> None:
+        if self.has_permission(source, info, **kwargs):
+            return None
+        else:
+            return self.on_unauthorized()
+
+    async def resolve_permission_async(
```
erikwrede, line 65: Splitting into sync and async is a breaking change, I'd like to avoid this. We can think about a different solution :)

erikwrede (changes requested):
Thanks for the initial version! I have a few comments but am short on time, thats why theyre very short. I'll revisit this in a few days 🙂

strawberry/permission.py
vethan: This is such an obvious improvement, cheers!

vethan: (Halfway through this change, probably finish it off next week)

strawberry/permission.py
vethan: Yeah, it seemed a bit gross, *very* open to a better solution. As mentioned, might be my lack of Python Syntax knowledge here!

strawberry/permission.py
vethan: The biggest issue is when it comes to composite permissions, the extension probably won't have knowledge of which exact permission failed.  The only solutions I could think of at the time would be a different breaking change (returning an object with information or something similar from has_permission), or having special logic for handling has_permission for the composite permissions (which in my mind seemed confusing?  But might be the way to do it).  Other views/advice eagerly recieved!

strawberry/permission.py
vethan: I probably should have put a comment in the code for this:
BaseException was a placeholder, wasn't sure what exception type to catch, and the try catch is a bit of a performance hit anyway. If we can find a way to implement the composites while keeping the on_unauthorized call in the extension, this would naturally fall out nicely!

strawberry/permission.py
ThirVondukr: @vethan Sorry for that, but could you probably adapt some of the stuff from my implementation on discord? 🤔 
https://discord.com/channels/689806334337482765/741575032257511485/1217820662245101609

If I remember correctly it should work with both async and sync by inspecting whenever any of child permissions are async

ThirVondukr: Also if applied operator is the same as the one in `PermissionHolder` I think you could avoid nesting them too by returning something like
```py
PermissionHolder(operator=..., permissions=[*self.permissions, other])
```

vethan: Your implementation was great for has_permission, but then there was no way to work through on_unauthorised correctly unfortunately :/  Need to split the logic a bit for that

ThirVondukr: Yeah, `on_unauthorised` is a bit hard to implement because we don't know which permission didn't succeed without checking them again (and we can't even do that)

strawberry/permission.py
```diff
@@ -89,6 +90,106 @@ class AutoDirective:
...
+
+
+class AndPermission(CompositePermission):
+    def has_composite_permission(
+        self, source: Any, info: Info, **kwargs: Any
+    ) -> Tuple[bool, int]:
+        for index, permission in enumerate(self.child_permissions):
+            if not permission.has_permission(source, info, **kwargs):
+                return False, index
+        return True, 0
```
ThirVondukr, line 163: Maybe it would be a better idea to return filed permission and not it's index?

strawberry/permission.py
```diff
@@ -180,8 +317,6 @@ def supports_sync(self) -> bool:
         """The Permission extension always supports async checking using await_maybe,
         but only supports sync checking if there are no async permissions"""
         async_permissions = [
-            True
-            for permission in self.permissions
-            if iscoroutinefunction(permission.has_permission)
+            True for permission in self.permissions if permission.is_async
         ]
         return len(async_permissions) == 0
```
ThirVondukr, line 322:
```suggestion
        return all(not permission.is_async for permission in self.permissions)
```

erikwrede:
Hey everyone, I worked on streamlining this with @patrick91  over the last days. New version is in #3451. Please comment there! @vethan thanks for the solid ground work. I just reduced the complexity a bit in some places but essentially it's your PR. The only missing thing is some schema directive support. We have decided to deprecate the old way of handling Schema Directives in favor of a new `Permission` directive containing custom info listing all the required permissions for a field.

patrick91:
@erikwrede maybe you can push those commits on this PR? do you have permission? 😊

botberry:
# Pre-release
:wave:

Releasing commit [901abf00defd1d5f99274a4d0718d4df09456cbe] to PyPi as pre-release! :package:

botberry:
# Pre-release
:wave:

Pre-release **0.229.2.dev.1715873118** [901abf00defd1d5f99274a4d0718d4df09456cbe] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.229.2.dev.1715873118
```

erikwrede:
Just pushed some additional fixes from local

vethan:
Hey gang: Will this get merged in at some point?  It's falling further and further behind main which is going to cause issues

patrick91:
@vethan I'll add this to the tuesday PRs to review :D /cc @bellini666

zionts:
Hi there — has this merged in / will it? 👀

patrick91:
@zionts hey! no, but we probably should

I'll try to get this review by end of next week, if that's ok 😊

strawberry/permission.py
```diff
@@ -89,6 +127,111 @@ class AutoDirective:
...
+            permission_response = await await_maybe(
+                permission.has_permission(source, info, **kwargs)
+            )
+            has_permission, context = unpack_maybe(permission_response, {})
+            if not has_permission:
+                return False, {"failed_permissions": [(permission, context)]}
+        return True
+
+    def __and__(self, other: BasePermission):
+        return AndPermission([*self.child_permissions, other])
```
bellini666: thought: What if I would like to do `IsAdmin() | (IsFoo() & IsBar())`? From what I can see here, I don't think this would be possible, as `IsFoo() & IsBar()` would be an `AndPermission` that has no `__or__`

Maybe this `__and__` and the `__or__` should be defined in `BasePermission`?

or, just brainstorming =P, have just one `CompositePermission` with both methods, and solve the expressions recursively, taking advantage of `any`/`all`, which should short-circuit when the condition becomes false.

Wdyt?

Something like this:

```python
class Operation(enum.Enum):
    AND = 'and'
    OR = 'or'

class Permission:
    def __and__(self, other):
        return CompositePermission(Operation.AND, self, other)

    def __or__(self, other):
        return CompositePermission(Operation.OR, self, other)

    def has_permission(self):
        raise NotImplementedError("Subclasses should implement this method")


class CompositePermission:
    def __init__(self, operation: Operation, *permissions: Permission | CompositePermission):
        self.operation = operation
        self.permissions = permissions  # this is a sequence of Permission or CompositePermission instances

    def __and__(self, other):
        return CompositePermission(Operation.AND, *self, other)

    def __or__(self, other):
        return CompositePermission(Operation.OR, *self, other)

    def has_permission(self):
        if self.operation == Operation.AND:
            has_permission = all(p.has_permission() for p in self.permissions)
        elif self.operation == Operation.OR:
            has_permission = any(p.has_permission() for p in self.permissions)
        else:
            raise ValueError(f"Unknown operation: {self.operation}")

        return has_permission
```

cc @patrick91

patrick91: let's chat about this on Tuesday!

vethan: @bellini666  Since _and_ and _or_ are defined on the BasePermission, and CompositePermission inherits from BasePermission, this would work perfectly fine as is!  The test `test_raises_graphql_error_when_nested` is an example of these composites!

The direct implementation on AndPermission and OrPermission is a bit of an optimisation.  Without it, a chain of would be: (((IsFoo AND IsBar) AND IsAnother) AND IsYetAnother).  It optimises it down to (IsFoo AND isBar AND IsAnother AND isYetAnother).

I hope that makes sense?

patrick91: I think @bellini666 suggestion is good, we can remove AndPermission and OrPermssion and just use the BasePermission, similar to what Django does with the Q object

https://github.com/django/django/blob/stable/5.2.x/django/db/models/query_utils.py#L39

strawberry/permission.py
```diff
@@ -74,6 +109,9 @@ def on_unauthorized(self) -> None:
         raise error
 
     @property
+    @deprecated(
+        "@schema_directive is deprecated and will be disabled by default on 31.12.2024 with future removal planned. Use the new @permissions directive instead."
+    )
```
patrick91: let's maybe do this in another PR if need it 😊

strawberry/permission.py
```diff
@@ -60,17 +71,36 @@ def has_permission(self, source, info, **kwargs):
...
+        If there's a need to pass additional information to the `on_unauthorized` method, return a tuple. The first element should be False, and the second element should be a dictionary containing the additional information.
+
+        Args:
+            source (Any): The source field that the permission check is being performed on.
+            info (Info): The GraphQL resolve info associated with the field.
+            **kwargs (Any): Additional arguments that are typically passed to the field resolver.
 
-        This method should be overridden by the subclasses.
+        Returns:
+            bool or tuple: Returns True if the user has the necessary permissions. Returns False or a tuple (False, additional_info) if the user does not have the necessary permissions. In the latter case, the `on_unauthorized` method will be invoked.
```
patrick91, line 97: I think I'd prefer to remove this change for now, I'll do that when I get back to this PR 😊

PR #3416, opened 2024-03-20
Schema extensions execution order should be LIFO

### LINKED ISSUE #3413: Surprising FIFO behaviour of lifecycle hooks
## Describe the (maybe) Bug

I'm surprised that various lifecycle hooks (`on_operation`, `on_parse`, `on_validate`, `on_execute`) are completed in a FIFO fashion, rather than LIFO.

I would expect that if we're wrapping an operation with `on_operation` with 2 extension the following will happen (LIFO):

* First extension starts (before `yield` part)
* Second extension starts (before `yield` part)
* Second extension completes (after `yield` part)
* First extension completes (after `yield` part)

However, the order I'm _actually_ seeing is the following (FIFO):

* First extension starts (before `yield` part)
* Second extension starts (before `yield` part)
* First extension completes (after `yield` part)
* Second extension completes (after `yield` part)

I'm concerned about it because extension can mutate state, so it would be good for them to behave like a context manager. [Example of state mutation.](https://strawberry.rocks/docs/guides/custom-extensions#execution-context)

In fact, I do find it surprising that this is how things work out. Notably, overriding `resolve` doesn't have the same problem – but it also happens in a slightly different way.

## Repro details

Here're some toy extensions I built to investigate things:

```python
class MyCustomExtension(SchemaExtension):
    id = "?"

    @override
    async def on_validate(self) -> AsyncGenerator[None, None]:
        print(f"GraphQL validation start ({self.id})")
        yield
        print(f"GraphQL validation end ({self.id})")

    @override
    def on_parse(self) -> Generator[None, None, None]:
        print(f"GraphQL parsing start ({self.id})")
        yield
        print(f"GraphQL parsing end ({self.id})")

    @override
    def on_execute(self) -> Generator[None, None, None]:
        print(f"GraphQL execution start ({self.id})")
        yield
        print(f"GraphQL execution end ({self.id})")

    @override
    def on_operation(self) -> Generator[None, None, None]:
        print(f"GraphQL operation start ({self.id})")
        yield
        print(f"GraphQL operation end ({self.id})")

    @override
    async def resolve(
        self,
        _next: Callable[..., object],
        root: object,
        info: GraphQLResolveInfo,
        *args,
        **kwargs,
    ) -> AwaitableOrValue[object]:
        random_id = randint(0, 1000)

        print(f"GraphQL resolver {random_id} start ({self.id})")
        result = await await_maybe(_next(root, info, *args, **kwargs))
        print(f"GraphQL resolver {random_id} end ({self.id})")
        return result


class MyCustomExtensionA(MyCustomExtension):
    id = "A"


class MyCustomExtensionB(MyCustomExtension):
    id = "B"
```

I'm testing it by running a simple query against a GraphQL Schema:

```python
@strawberry.type
class Me:
    id: str

@strawberry.type
class Query:
    @strawberry.field
    @staticmethod
    async def me() -> Me:
        return Me(id="foo")

schema = MySchema(
    query=Query,
    extensions=[MyCustomExtensionA, MyCustomExtensionB],
)
```

When running a simple GraphQL query against this schema:

```graphql
query {
  me { id }
}
```

I see the following lines being printed:

```
GraphQL operation start (A)
GraphQL operation start (B)
GraphQL parsing start (A)
GraphQL parsing start (B)
GraphQL parsing end (A)
GraphQL parsing end (B)
GraphQL validation start (A)
GraphQL validation start (B)
GraphQL validation end (A)
GraphQL validation end (B)
GraphQL execution start (A)
GraphQL execution start (B)
GraphQL resolver 598 start (B)
GraphQL resolver 975 start (A)
GraphQL resolver 975 end (A)
GraphQL resolver 598 end (B)
GraphQL resolver 196 start (B)
GraphQL resolver 638 start (A)
GraphQL resolver 638 end (A)
GraphQL resolver 196 end (B)
GraphQL execution end (A)
GraphQL execution end (B)
GraphQL operation end (A)
GraphQL operation end (B)
```

## System Information

 - Strawberry version (if applicable): `0.220.0`

## Upvote & Fund

<a href="https://polar.sh/strawberry-graphql/strawberry/issues/3413">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/3413/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

nrbnlulu:
## Description
Fix schema extensions execution order to be LIFO

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

fix #3413

## Checklist


botberry:



---

Hot fix for #3413.
With this release schema extensions are guaranteed
to be run in LIFO order.



Here's the tweet text: 
```

```

botberry:



---

Hot fix for #3413.
With this release schema extensions are guaranteed
to be run in LIFO order.



Here's the tweet text: 
```

```

kkom:
Thank you for addressing it so quickly! Really appreciate it!

ThirVondukr:
@nrbnlulu Maybe it would be better to use `AsyncExitStack` to achieve the same instead? 🤔
Though if an exception occurs all exits would be called, so I'm not sure if that's desired behavior

patrick91:
> @nrbnlulu Maybe it would be better to use `AsyncExitStack` to achieve the same instead? 🤔 Though if an exception occurs all exits would be called, so I'm not sure if that's desired behavior

Do you have an example of how that would look like?

Maybe calling all exits could be actually good?

ThirVondukr:
@patrick91 Basically exit stack can guarantee that all exits are called and exceptions are propagated into them, using it should be pretty simple:
```py
import asyncio
import contextlib


@contextlib.asynccontextmanager
async def my_context(number: int):
    print(number)
    yield
    print(f"Exit {number}")


async def main() -> None:
    async with contextlib.AsyncExitStack() as exit_stack:
        for i in range(5):
            await exit_stack.enter_async_context(my_context(i))


if __name__ == "__main__":
    asyncio.run(main())

```
```
0
1
2
3
4
Exit 4
Exit 3
Exit 2
Exit 1
Exit 0
```

nrbnlulu:
@patrick91 
I think it would be possible to wrap the extensions with `@(async)contextmanager`...

Though I think to avoid overhead I would consider caching the parsed hooks and it would require to
move the hooks parsing logic upwards (Maybe schema creation level).
 
This would also be good for extensions would want to add / modify types in the schema.
i.e
```py
class JwtExtension:
	def on_types_gather(self, types: dict[str, type[) -> None:
		types.update(jwt_types)
```

ThirVondukr:
In my opinion if a generator only has one `yield` and you *effectively* use it as a context manager - it should be a context manager, so, maybe using (async)generator should be deprecated and users should be encouraged to decorate these methods.

strawberry/extensions/context.py
```diff
@@ -193,15 +203,25 @@ def __exit__(
         self.run_hooks_sync(is_exit=True)
 
     async def __aenter__(self):
-        await self.run_hooks_async()
+        self.async_exit_stack = contextlib.AsyncExitStack()
```
ThirVondukr, line 179: Is this safe? Would there be any concurrent calls to `__aenter__/__aexit__`?

strawberry/extensions/context.py
```diff
@@ -125,78 +138,63 @@ def from_callable(
...
-
-        The ``is_exit`` flag is required as a `StopIteration` cannot be raised from
-        within a coroutine.
-        """
-        ctx = (
-            contextlib.suppress(StopIteration, StopAsyncIteration)
-            if is_exit
-            else contextlib.nullcontext()
-        )
+    def __enter__(self):
```
ThirVondukr, line 156:
```suggestion
    def __enter__(self) -> None:
```

strawberry/extensions/context.py
```diff
@@ -125,78 +138,63 @@ def from_callable(
...
     def __exit__(
         self,
         exc_type: Optional[Type[BaseException]],
         exc_val: Optional[BaseException],
         exc_tb: Optional[TracebackType],
     ):
-        self.run_hooks_sync(is_exit=True)
+        self.exit_stack.__exit__(exc_type, exc_val, exc_tb)
 
     async def __aenter__(self):
```
ThirVondukr, line 178:
```suggestion
    async def __aenter__(self) -> None:
```

strawberry/extensions/context.py
```diff
@@ -73,10 +80,16 @@ def get_hook(self, extension: SchemaExtension) -> Optional[WrappedHook]:
 
         if hook_fn:
             if inspect.isgeneratorfunction(hook_fn):
-                return WrappedHook(extension, hook_fn(extension), False)
+                hook_fn = types.MethodType(
+                    contextlib.contextmanager(hook_fn), extension
+                )
+                return WrappedHook(extension, hook_fn, False)
```
ThirVondukr, line 86: Maybe it would be better to pass `True/False` as kwarg, it's a bit more readable

patrick91: yup, or all of them as kwargs!

strawberry/extensions/context.py
ThirVondukr: Nvm, reading through the code it should be fine to assign to self here

ThirVondukr:
By the way, from what I see each time a query is ran you create new instance of `ExtensionContextManagerBase` or it's subclass, running `get_hook`, which could get expensive 🤔

patrick91:
I need to add a test case to see how it works when an error is raised

patrick91 (approved):
Thanks @erikwrede, @ThirVondukr, and @nrbnlulu for this!

PR #3419, opened 2024-03-21
Fix info by adding a custom `__class_getitem__`

patrick91:
I should have tested this at runtime, woops 😅

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,22 @@
+Release type: patch
+
+This release properly allows passing one argument to the `Info` class.
+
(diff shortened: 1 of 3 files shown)


botberry:



---

This release properly allows passing one argument to the `Info` class.

This is now fully supported:

```python
import strawberry

from typing import TypedDict


class Context(TypedDict):
    user_id: str


@strawberry.type
class Query:
    @strawberry.field
    def info(self, info: strawberry.Info[Context]) -> str:
        return info.context["user_id"]
```



Here's the tweet text: 
```

```

PR #3451, opened 2024-04-14
feat: add boolean permission support

erikwrede:
Supersedes #3408

- First pass looking at boolean operators on permissions
- Boolean permissions tests
- Update docs
- [pre-commit.ci] auto fixes from pre-commit.com hooks
- Formatting fix
- [pre-commit.ci] auto fixes from pre-commit.com hooks
- Move try catch out of loop!
- [pre-commit.ci] auto fixes from pre-commit.com hooks
- Few fixes from AI review
- [pre-commit.ci] auto fixes from pre-commit.com hooks
- Revert some breaking changes.  Move from boolean to composite permissions as base class.  Branch on composite permissions to handle them differently!
- [pre-commit.ci] auto fixes from pre-commit.com hooks
- Absolutely hadn't finished my changes!
- [pre-commit.ci] auto fixes from pre-commit.com hooks
- Fix to undefined Info type
- [pre-commit.ci] auto fixes from pre-commit.com hooks
- Test the async versions of composite permissions, and some linting fixes
- [pre-commit.ci] auto fixes from pre-commit.com hooks
- Final linting fixes
- [pre-commit.ci] auto fixes from pre-commit.com hooks
- Make sure to test asyncs
- [pre-commit.ci] auto fixes from pre-commit.com hooks
- Of course I missed a single _async
- refactor: use default has permission syntax for boolean permissions
- chore: adjust type annotation for kwargs


botberry:



---

Adds the ability to use the `&` and `|` operators on permissions to form boolean logic. For example, if you want
a field to be accessible with either the `IsAdmin` or `IsOwner` permission you
could define the field as follows:

```python
import strawberry
from strawberry.permission import PermissionExtension, BasePermission


@strawberry.type
class Query:
    @strawberry.field(
        extensions=[
            PermissionExtension(
                permissions=[(IsAdmin() | IsOwner())], fail_silently=True
            )
        ]
    )
    def name(self) -> str:
        return "ABC"
```



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Erik Wrede for the PR 👏

```

strawberry/permission.py
```diff
@@ -89,6 +111,85 @@ class AutoDirective:
...
+    async def _has_permission_async(
+        self, source: Any, info: Info, **kwargs: dict
+    ) -> Union[bool, (False, dict)]:
+        for permission in self.child_permissions:
+            if not await await_maybe(permission.has_permission(source, info, **kwargs)):
+                return False, {"failed_permissions": [permission]}
+        return True
+
+    def __and__(self, other: BasePermission):
+        self.child_permissions.append(other)
```
ThirVondukr, line 160: I think `__and__` and `__or__` should return a new instance, because it shouldn't mutate the object itself

strawberry/permission.py
```diff
@@ -89,6 +111,85 @@ class AutoDirective:
...
+    ) -> Union[bool, Awaitable[bool], (False, dict), Awaitable[(False, dict)]]:
+        if self.is_async:
+            return self._has_permission_async(source, info, **kwargs)
+        failed_permissions = []
+        for permission in self.child_permissions:
+            if permission.has_permission(source, info, **kwargs):
+                return True
+            failed_permissions.append(permission)
+
+        return False, {"failed_permissions": failed_permissions}
```
ThirVondukr, line 176: Probably get rid of dictionaries here, maybe make a dataclass instead 🤔

strawberry/permission.py
```diff
@@ -132,13 +237,16 @@ def apply(self, field: StrawberryField) -> None:
             elif isinstance(field.type, StrawberryList):
                 self.return_empty_list = True
             else:
-                errror = PermissionFailSilentlyRequiresOptionalError(field)
-                raise errror
+                error = PermissionFailSilentlyRequiresOptionalError(field)
+                raise error
```
ThirVondukr, line 241:
```suggestion
                raise PermissionFailSilentlyRequiresOptionalError(field)
```

strawberry/permission.py
```diff
@@ -132,13 +237,16 @@ def apply(self, field: StrawberryField) -> None:
...
 
-    def _on_unauthorized(self, permission: BasePermission) -> Any:
+    def _on_unauthorized(self, permission: BasePermission, **kwargs: dict) -> Any:
```
ThirVondukr, line 243: This is probably a mistake - with kwargs you annotate value type (e.g. `int` instead of `dict[str, int]`, not sure what was intended here 🙂

ThirVondukr: If there are no particular operations performed on kwargs values you could annotate them as `object`, as it's the most narrow type, `Any` being worse for that use case

strawberry/permission.py
erikwrede: The idea here was to remain as flexible as possible, allowing users to pass context in the future as well. With a dataclass we would be again very static about handling additional permission error context.

strawberry/permission.py
erikwrede: I feel like that would create a lot of heap waste if you have `A or B or C or D` in a lot of fields. Is it actually in the python spec that and / or shouldn't mutate the object itself? I only found this https://docs.python.org/3/reference/datamodel.html#emulating-numeric-types

strawberry/permission.py
erikwrede: I wonder how that happened, thanks 🙂

strawberry/permission.py
ThirVondukr: Since there's `|=` operator I think it shouldn't 🤔 numeric types are kind of special because they're never mutated

strawberry/permission.py
ThirVondukr: Hm, I'm not sure how context would be used here, is it in the release noted (that I didn't read)? 😅

strawberry/permission.py
vethan: Agreed it seems weird to me to mutate objects type.  It's a side-effect that you wouldn't really expect to have to plan around

strawberry/permission.py
```diff
@@ -51,17 +52,35 @@ class BasePermission(abc.ABC):
     @abc.abstractmethod
     def has_permission(
         self, source: Any, info: Info, **kwargs: Any
-    ) -> Union[bool, Awaitable[bool]]:
+    ) -> Union[bool, Awaitable[bool], (False, dict), Awaitable[(False, dict)]]:
```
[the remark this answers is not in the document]
ThirVondukr: By the way, this looks like a mistake `(a, b)` is not a valid type 🤔 Perhaps you need `Tuple[Literal[False], dict]`?

strawberry/permission.py
erikwrede: You are 100 % right, i only thought of ad hoc use cases when using the operators on a field extension, not about pre-composed permissions

strawberry/permission.py
erikwrede: It's internal for now, but it allows us to pass information to raise_exception from has_permission. We opted to separate exception raising from permission checking back when implementing this extension due to the fail silently feature.

erikwrede: So basically the context contains "this permission failed out of the list of and / or permissions"

patrick91:
Let's keep working in #3408 😊

PR #3453, opened 2024-04-15
Add docs about self/parent/root

patrick91:
still wip


patrick91:
I need to figure how and where to link this 😊

docs/types/resolvers-self.md
```diff
@@ -0,0 +1,218 @@
...
+Historically Strawberry only supported passing the parent value by adding a
+parameter called `root`:
+
+```python
+def get_full_name(root: User) -> str:
+    return f"{root.first_name} {root.last_name}"
+```
+
+This is still supported, but we recommend using `strawberry.Parent`, since it
+follows Strawberry's philosofy of using type annotations. Also, with
```
bellini666, line 75:
```suggestion
follows Strawberry's philosophy of using type annotations. Also, with
```

docs/types/resolvers-self.md
```diff
@@ -0,0 +1,218 @@
...
+still work:
+
+```python
+def get_full_name(user: strawberry.Parent[User]) -> str:
+    return f"{user.first_name} {user.last_name}"
+```
+
+## Accessing parent's data in a method resolver
+
+Both method also work when defining a method resolver, so we can still use
```
bellini666, line 86: `Both methods` maybe?

But also try to avoid `method` to describe the "method of doing something", as it can be ambiguous with an object method as defined here

docs/types/resolvers-self.md
```diff
@@ -0,0 +1,218 @@
...
+class User:
+    first_name: str
+    last_name: str
+
+    @strawberry.field
+    def full_name(self, parent: strawberry.Parent[User]) -> str:
+        return f"{parent.first_name} {parent.last_name}"
+```
+
+But, here's where things get more interesting. If this were a pure Python class,
```
bellini666, line 103: It sounds to me that this should be `was` but I might be wrong 😅 

```suggestion
But, here's where things get more interesting. If this was a pure Python class,
```

docs/types/resolvers-self.md
```diff
@@ -0,0 +1,218 @@
...
+we would use `self` directly. Well, turns out we can!
+
+Strawberry also passes the parent value to the `self` parameter, this is purely
+provided as a shortcut, since there's a few things happening under the hood that
+might make this not fully correct. (NOT SURE ABOUT THIS WORDING)
```
bellini666, line 108: I always find it difficult to explain to people why "self" is not actually self in some situations, specially when working with django types.

docs/types/resolvers-self.md
```diff
@@ -0,0 +1,218 @@
...
+class User:
+    first_name: str
+    last_name: str
+
+    @strawberry.field
+    def full_name(self) -> str:
+        return f"{self.first_name} {self.last_name}"
+```
+
+Much better no? `self` on resolver methods is pretty convient, and it works like
```
bellini666, line 126:
```suggestion
Much better, no? `self` on resolver methods is pretty convenient, and it works like
```

docs/types/resolvers-self.md
```diff
@@ -0,0 +1,218 @@
...
+it should in Python, but there might be cases where it doesn't follow Python's
+semantics. This is because under the hood resolvers are actually called if they
```
bellini666, line 128:
```suggestion
semantics. This is because under the hood resolvers are actually called as if they
```

bellini666 (commented):
Really like this!

Left some suggestions in the code, but take those with a grain of salt as my english skills are not the best hahaha

docs/types/resolvers-self.md
```diff
@@ -0,0 +1,218 @@
...
+still work:
+
+```python
+def get_full_name(user: strawberry.Parent[User]) -> str:
+    return f"{user.first_name} {user.last_name}"
+```
+
+## Accessing parent's data in a method resolver
+
+Both method also work when defining a method resolver, so we can still use
```
patrick91, line 86:
```suggestion
Both options also work when defining a method resolver, so we can still use
```

Issue #3224, opened 2023-11-14
Support the `extensions` Request Parameter for the HTTP protocol


omarzouk:
Hello! The GQL spec specifies a field called `extensions` as part of the [request parameters spec](https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#request-parameters) in the HTTP protocol. We have recently come into a usecase where we would like to make use of this field to provide extra metadata to our apis in a federated graph world.

The `extensions` field is supported in the [subscriptions protocol](https://github.com/strawberry-graphql/strawberry/blob/4130a75f23a7fec7b4a6e9fbba913c4607437f37/strawberry/subscriptions/protocols/graphql_transport_ws/types.py#L67), however it is not supported in the HTTP protocol.

Up until now we had been using HTTP headers to achieve the passing of metadata, however, in order to reuse the same API on subscriptions, we now have to put the metadata in the extensions field. This means our clients would need to send the data in the header when using HTTP and in the extensions field when using subscriptions. Our APIs also have to be able to handle both. It would be nice to be able to use the extensions field everywhere.

## Feature Request Type
- [X] Alteration (enhancement/optimization) of existing feature(s)

omarzouk:
it is a `map` object similar to the `variables` field, so would look something like this:
```{"query":"query getAccount($id: ID!) {\n  getAccount(id: $id) {\n    id\n    name\n\n  }\n","variables":{ "id": "XXX" }, "extensions":{"some_property": "some_value"}}```

We would then like to have access to it in the `get_context` function somehow. Perhaps under the `Request` object?

patrick91:
thanks! 

this might be a good time to refactor some parts of accessing the context. Ref #1258 #2687 


How urgent is this?

omarzouk:
Would be nice to have it in the upcoming couple months. We would love if we can start using it early next year

omarzouk:
@patrick91 hello! do u think there is bandwidth to prioritise the refactor you mentioned during January? if not, could we add support for the variable to the existing setup?

We would love to start using extensions soon. As opposed to headers, extensions are meant to work on both HTTP and Websockets which make them much better at carrying soft metadata, like device timezone or locale

patrick91:
@omarzouk I think we can make it happen, just to clarify, you want to be able to update value of extensions from a resolver, something like:

```python
@strawberry.type
class Mutation:
    @strawberry.mutation
    def login(self, info: Info) -> bool:
[...]

PR #3461, opened 2024-04-19
Add support for extensions to the HTTP protocol

omarzouk:
## Description
The GQL spec specifies a field called extensions as part of the [request parameters spec](https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#request-parameters) in the HTTP protocol. We are missing support for this

Adding the field to be parsed and propagating it as part of the `Info` object. This is done by wrapping the entire context object in a temporary container, and then unwrapping it inside the `Info` context getter. A new function is also added to the `Info` class called `input_extensions`. It exposes the extensions data.

Example usage by the end-user:
```python
@strawberry.field
def value_from_extensions(self, key: str, info: strawberry.Info) -> str:
    return info.input_extensions[key]
```

## Types of Changes

- [X] Core
- [ ] Bugfix
- [X] New feature
- [X] Documentation

* fixes https://github.com/strawberry-graphql/strawberry/issues/3224

## Checklist

## Summary by Sourcery

This pull request introduces support for the 'extensions' field in HTTP requests, allowing it to be parsed and included in the ExecutionContext. This change is accompanied by updates to the documentation and new tests to ensure proper functionality.

- **New Features**:
    - Added support for the 'extensions' field in HTTP requests, allowing it to be parsed and propagated as part of the ExecutionContext.
- **Documentation**:
    - Updated documentation to reflect the new support for the 'extensions' field in HTTP requests.
- **Tests**:
    - Added tests to verify the correct handling and propagation of the 'extensions' field in both GET and POST HTTP requests.


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Omar Marzouk for the PR 👏

```

botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Omar Marzouk for the PR 👏

```

erikwrede (commented):
Thanks for the PR! Can you add a test to ensure the exension dict is actually correctly passed down the chain? 🙂

omarzouk:
@erikwrede Hi! yea I think I need to do a fresh take on this, I think I hit a dead end with my current approach, I've been investigating how to do it, and I think we have a couple options:

1- Wrap the context field in another object that holds the extensions and the context, i.e. wrap the field before [this line](https://github.com/strawberry-graphql/strawberry/pull/3461), then unwrap it inside the [Info class](https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/types/info.py#L82). We would then introduce an `input_extensions` or `protocol_extensions` field to Info that reads the extensions from the wrapper.

2- Extend the definition of the [get_context](https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/http/async_base_view.py#L72) function, add a 3rd parameter called `extensions`. Usually http users use the `Request` param to get headers, extensions can be extracted by the user here as well. However, this would require moving the body parsing and data fetching to happen before the `get_context` is called, i.e. [this operation](https://github.com/strawberry-graphql/strawberry/blob/39ccffc65f6fc3dc7af69596e85d2aaf69b070af/strawberry/http/async_base_view.py#L91) would have to move above [this line](https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/http/async_base_view.py#L161)

The benefit of doing Option 1 would be that we can add this right away to http, without having to modify other protocols, and just include a type check for the wrapper in the `info.context` getter. It can then be added to other protocols over time, in other PRs.

What do u think? should I go ahead and do Option 1?

omarzouk:
hi @erikwrede, I went ahead and implemented Option 1 mentioned above, which is to wrap the context object and unwrap it in the Info getter. I also added a new test that is passing successfully. Please let me know what you think. 🙏

omarzouk:
pinging @patrick91 as well, since we discussed this originally in https://github.com/strawberry-graphql/strawberry/issues/3224

strawberry/http/__init__.py
```diff
@@ -33,6 +33,7 @@ class GraphQLRequestData:
     query: Optional[str]
     variables: Optional[Dict[str, Any]]
     operation_name: Optional[str]
+    extensions: Optional[Dict[str, Any]]
```
bellini666, line 36: Wondering if this should default to `None`?

```suggestion
    extensions: Optional[Dict[str, Any]] = None
```

strawberry/http/async_base_view.py
```diff
@@ -22,6 +22,7 @@
 from strawberry.types import ExecutionResult
 from strawberry.types.graphql import OperationType
 
+from ..types.context_wrapper import ContextWrapper
```
bellini666, line 25: I think it is ok (and usually nice) to use relative imports for something already inside it (i.e. using `from .<something>`), but when importing from a parent (i.e. `from ..<something>) I think it starts to get messy.

In this case I would suggest importing from `strawberry.types.context_wrapper`

strawberry/http/sync_base_view.py
```diff
@@ -23,6 +23,7 @@
 from strawberry.types import ExecutionResult
 from strawberry.types.graphql import OperationType
 
+from ..types.context_wrapper import ContextWrapper
```
bellini666, line 26: Ditto

strawberry/types/info.py
```diff
@@ -79,8 +80,16 @@ def selected_fields(self) -> List[Selection]:
 
     @property
     def context(self) -> ContextType:
+        if type(self._raw_info.context) is ContextWrapper:
```
bellini666, line 83: better?

```suggestion
        if isinstance(self._raw_info.context, ContextWrapper):
```

strawberry/types/info.py
```diff
@@ -79,8 +80,16 @@ def selected_fields(self) -> List[Selection]:
...
+            return self._raw_info.context.context
         return self._raw_info.context
 
+    @property
+    def input_extensions(self) -> Dict[str, Any]:
+        if type(self._raw_info.context) is ContextWrapper:
```
bellini666, line 89:
```suggestion
        if isinstance(self._raw_info.context, ContextWrapper):
```

strawberry/types/context_wrapper.py
```diff
@@ -0,0 +1,8 @@
+from dataclasses import dataclass
+from typing import Any, Dict, Optional
+
+
+@dataclass
+class ContextWrapper:
+    context: Optional[Any]
+    extensions: Optional[Dict[str, Any]]
```
bellini666, line 8: Same suggestion as above

```suggestion
    extensions: Optional[Dict[str, Any]] = None
```

tests/views/schema.py
```diff
@@ -108,6 +108,10 @@ def root_name(self) -> str:
     def value_from_context(self, info: strawberry.Info) -> str:
         return info.context["custom_value"]
 
+    @strawberry.field
+    def value_from_extensions(self, key: str, info: strawberry.Info) -> str:
```
bellini666, line 112: nit: info first, to follow the pattern of others that receive `info` as the first one, but this one receives an extra `key`

Maybe also adding a `, *, ` in the middle to enforce `key` to be passed as a keyword would be good?

```suggestion
    def value_from_extensions(self, info: strawberry.Info, key: str) -> str:
```

strawberry/http/base.py
```diff
@@ -67,6 +67,15 @@ def parse_query_params(
             if variables:
                 params["variables"] = self.parse_json(variables)
 
+        if "extensions" in params:
+            extensions = params["extensions"]
+
+            if isinstance(extensions, list):
+                extensions = extensions[0]
```
bellini666, line 74: Why only getting the first on the list?

From the rest of the code `extensions` should be a dict. When can it be a list? If it is, is there a possibility of it having more than one value inside it? If yes, what does that mean? Should we warn/error out about that?

bellini666 (commented):
Some comments, but I'm not totally aware of this, so other @strawberry-graphql/core will be able to do a better review :)

strawberry/http/base.py
omarzouk: Since it has the exact same type definition as the `variables` parameter, I just copy pasted how that is being parsed. I think it might be a list in GET requests?

strawberry/http/__init__.py
omarzouk: @bellini666 Since the `variables` parameter in the same object is also optional, and didn't have a default value, I wanted to keep this consistent so I don't introduce unexpected behaviour in case it gets initialised automatically some other way? 🤷‍♂️

strawberry/http/async_base_view.py
omarzouk: done

strawberry/http/sync_base_view.py
omarzouk: done

strawberry/http/__init__.py
bellini666: Fair point!

Maybe all of them should? =P
I'm fine in keeping the consistency or changing everything, lets see what others think

strawberry/http/base.py
bellini666: Hrmm I see. @strawberry-graphql/core do you know why we do that?

DoctorJohn: I just dug around a little bit. Before #2681 was merged there used to be the following comment in the Sanic integration:

https://github.com/strawberry-graphql/strawberry/blob/860e5f236b936440d4d28b8cea3f7bc9316a1612/strawberry/sanic/views.py#L101-L108

This could explain why we parse variables in that way.

DoctorJohn: Update: I'm working on a PR for this. Looks like it might be Sanics fault after all.

DoctorJohn: #3558

omarzouk:
Hello! apart from the extra check for list in the parameter parsing which is potentially redundant and might be removed, as I understand it, that change is also not really blocking for this PR, right? https://github.com/strawberry-graphql/strawberry/pull/3558 could be updated to remove the list parsing for `extensions` after this one is merged?

Is there any other comments regarding this PR? the approach in general around wrapping context and exposing the extra field in the `Info` object? how would you like me to proceed?

@bellini666 @DoctorJohn @erikwrede @patrick91

omarzouk:
Hi! I have merged main to include https://github.com/strawberry-graphql/strawberry/pull/3558 and removed the extra list type check for the extensions param. Anything else you guys would like me to change about this PR? if not I can should I create a release file?

@bellini666 @DoctorJohn @erikwrede @patrick91

strawberry/types/info.py
```diff
@@ -79,8 +80,16 @@ def selected_fields(self) -> List[Selection]:
...
     def context(self) -> ContextType:
+        if isinstance(self._raw_info.context, ContextWrapper):
+            return self._raw_info.context.context
         return self._raw_info.context
 
+    @property
+    def input_extensions(self) -> Dict[str, Any]:
+        if isinstance(self._raw_info.context, ContextWrapper):
+            return self._raw_info.context.extensions
+        return {}
```
patrick91, line 91: how come this is called input_extensions? 😊

strawberry/http/sync_base_view.py
```diff
@@ -112,11 +113,15 @@ def execute_operation(
 
         assert self.schema
 
+        context_wrapper = ContextWrapper(
+            context=context, extensions=request_data.extensions
+        )
```
patrick91, line 118: is this context wrapper only to pass request data?

strawberry/types/info.py
omarzouk: honestly, I just didn't have a better name 😅  I just felt `extensions` is used in so many contexts in strawberry and people might be confused as to which one this is. But I'm happy to change it to whatever you think makes most sense

strawberry/http/sync_base_view.py
omarzouk: yes exactly. Since the `context` data comes from `get_context` and that can be any type of object the user decides to return, this felt like a safe way to attach more data

patrick91: I'll try to review and merge this this week 😊 thanks for the patience!

omarzouk:
hello! sorry to keep pinging, we are really eager to use this in our APIs, do u think its possible to get some traction on it? I'm happy to help any way I can @patrick91 🙏

patrick91:
@omarzouk sure, I think I'll have time this week(end) (I'm currently travelling) 😊

And please consider sponsoring us, so we can prioritise issues better ;)

patrick91:
@omarzouk an update on this, I think I found a nice way to pass the request extensions to info, but it is quite different from this, so I might try it in another PR

omarzouk:
> And please consider sponsoring us, so we can prioritise issues better ;)

@patrick91 Hi Patrick! I talked with our management about  it, it took a while because github's invoice system for sponsorships is a big mess, but we [Soundtrack Technologies](https://github.com/soundtrackyourbrand) is now a 🍓 Sponsor! you guys built an awesome library, we use it for a lot of our apps, so it makes all the sense in the world to help you maintain it!

patrick91:
@omarzouk looks there's an alpha version of GraphQl-core which makes passing data around much easier, would you be ok with using their alpha?

omarzouk:
hi @patrick91 what is the status of this now? I see graphql-core have implemented some support for it in the [alpha](https://github.com/graphql-python/graphql-core/issues/226#issuecomment-2629309490) version. Are we working on something in Strawberry so we can make use of the new capability?

patrick91:
@omarzouk I'll take a look at this tomorrow! sorry!

patrick91:
@omarzouk I think we (@bellini666) got a nice implementation, would be interested in trying the pre-release? 😊

botberry:
# Pre-release
:wave:

Releasing commit [832ad16db4024690c2bf1e27220c0ca2f1a1374a] to PyPi as pre-release! :package:

botberry:
# Pre-release
:wave:

Pre-release **0.269.0.dev.1747164009** [832ad16db4024690c2bf1e27220c0ca2f1a1374a] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.269.0.dev.1747164009
```

strawberry/types/info.py
```diff
@@ -113,6 +113,11 @@ def context(self) -> ContextType:
         """The context passed to the query execution."""
         return self._raw_info.context
 
+    @property
+    def input_extensions(self) -> dict[str, Any]:
```
omarzouk, line 117: @patrick91 First of all, thanks a lot for taking the time to do it! I tested it in one of our APIs, the data does go through all the way! However it is returned as a string and not as a dict, so we need to `json.loads` the data before we return it from this function. The [spec](https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#request-parameters) states that the parameter is expected to be a map so parsing it should be safe to do.

omarzouk: @patrick91 actually nvm I was wrong, I was over-escaping the input, totally my bad. The implementation is actually correct!

omarzouk (commented):
The pre-release works perfectly, thanks a lot guys! ❤️

strawberry/types/info.py
patrick91: @omarzouk awesome! I'll release this after we fix the unrelated issue with GlobalID :D

botberry:

mdgilene:
Input extension information only being available from the `Info` object seems pretty limiting. It would make sense that someone would want to implement a custom extension and have access to the input extension information from the `ExecutionContext`. But currently from what I can tell this is not the case currently.

Are there any plans to make this info available to the various extension lifecycle hook methods?

patrick91:
@mdgilene can you open an issue with an example use case? We'll try to make it happen 😊

PR #3474, opened 2024-04-26
chore(bench): update CodSpeed/action to v2

adriencaccia:
## Description

Upgrading to the v2 of https://github.com/CodSpeedHQ/action will bring a better base run selection algorithm, better logging, and continued support.


erikwrede (approved):
Thanks for pushing the upgrade!

PR #3499, opened 2024-05-14
add support for default values with a field that is a pydantic model

### LINKED ISSUE #3285: Pydantic models as a default values in inputs
When using `strawberry.experimental.pydantic.type`, default values of Pydantic models are not considered, if they are Pydantic models themselves.

## Describe the Bug

Let's have an example:

```python
import strawberry


class Test(BaseModel):
    id: int = 10


class Test2(BaseModel):
    test: Test = Test()
    id: int = 10


@strawberry.experimental.pydantic.type(model=Test, all_fields=True)
class StrawberryTest:
    ...


@strawberry.experimental.pydantic.type(model=Test2, all_fields=True)
class StrawberryTest2:
    ...


@strawberry.experimental.pydantic.type(model=Test, all_fields=True, is_input=True)
class StrawberryTestInput:
    ...


@strawberry.experimental.pydantic.type(model=Test2, all_fields=True, is_input=True)
class StrawberryTest2Input:
    ...


@strawberry.type(is_input=True)
class StrawberryTestPlainInput:
    id: int = 10


@strawberry.type(is_input=True)
class StrawberryTestPlain2Input:
    test: StrawberryTestPlainInput = StrawberryTestPlainInput()
    id: int = 10


@strawberry.type
class Query:
    @strawberry.field()
    def test(self, x: StrawberryTestInput) -> int:
        return 42

    @strawberry.field()
    def test2(self, x: StrawberryTest2Input) -> int:
        return 42

    @strawberry.field()
    def testplain2(self, x: StrawberryTestPlain2Input) -> int:
        return 42


print(schema)
```

The generated schema for `StrawberryTestPlain2Input` is 

```
input StrawberryTestPlain2Input {
  test: StrawberryTestPlainInput! = {id: 10}
  id: Int! = 10
}
```

Which is correct. But for `StrawberryTest2Input`, the schema is 

```
input StrawberryTest2Input {
  test: StrawberryTestInput!
  id: Int! = 10
}
```

This is incorrect because it says that the `test` field is mandatory, although it has a default value and therefore isn't mandatory.

I would expect that the `test` field of `StrawberryTest2Input` will be either nullable or that the default value will be provided in the same way as in the `StrawberryTestPlain2Input`.

## System Information

 - Operating system: macOS Ventura 
 - Strawberry version (if applicable): 0.215.2
 - Python version: 3.10.10

## Additional Context

I wonder if this is expected behaviour and can be somehow fixed on my side or a bug.

## Upvote & Fund

<a href="https://polar.sh/strawberry-graphql/strawberry/issues/3285">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/3285/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

ppease:
In this PR, I am adding support for having a pydantic model be a default value for a field on an input.  Previously this would not generate the proper schema.

## Description

My original fix was modifying code in `from_input_field` inside of the schema converter.  This is the point where we are instantiating a graphql input type, and I was checking to see if the default value had a model dump method on it.  This seems a little better since it is still within the pydantic specific code.  The issue I'm trying to solve is that when we print the schema from within `print_input_value` we use `ast_from_value` on the default value.  We fall into the `is_input_object_type` case and then we check if there is an object definition for the value.  If `is_input_object_type` evaluates to `True` then we call `strawberry.asdict`.  I think we'd want to do something similar with a pydantic instance.  Today we fail the check and end up returning `None` which causes the default value to not be printed in the schema.  What I'm doing in this MR is detecting if our default value is a pydantic instance, and if it is then I am storing the serialized value for the pydantic model as the default value.  

I'm curious to know if there is a better way to plug into the printing process because that feels more inline with what we are doing with dataclasses today.  I could add code that I had previously into either `print_input_value` or `ast_from_value`, but that felt a little more hacky since we are putting pydantic specific stuff in higher level code.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

This fixes #3285

## Checklist


strawberry/experimental/pydantic/_compat.py
```diff
@@ -17,6 +17,9 @@
 IS_PYDANTIC_V2: bool = PYDANTIC_VERSION.startswith("2.")
 IS_PYDANTIC_V1: bool = not IS_PYDANTIC_V2
 
+if IS_PYDANTIC_V2:
```
ppease, line 20: I'm not sure this is absolutely necessary, but when I was running the tests locally with pydantic v2 all of the mypy automated tests were failing without this import.

strawberry/schema/schema_converter.py
```diff
@@ -362,9 +362,16 @@ def from_input_field(
             ),
         )
         default_value: object
-
+        model_dump = getattr(field.default_value, "model_dump", None)
+        is_not_type = not isinstance(field.default_value, type)
         if field.default_value is UNSET or field.default_value is dataclasses.MISSING:
             default_value = Undefined
+        elif dataclasses.is_dataclass(field.default_value) and is_not_type:
+            dataclass_instance = cast(Any, field.default_value)
```
ppease, line 370: I was fighting with mypy here.  The dataclasses.asdict method only takes dataclass instances.  Mypy thought we still had a dataclass instance or a dataclass type, however, I'm specifically checking to see if it's a type or not with `is_not_type`.

ppease: I'm guessing I will have to pull this out anyway.  I was hoping to solve this for dataclasses and pydantic models, but it looks like newer versions of python are more strict on what values can be set as the default value for a field on a dataclass.  The new tests I added are failing in python 3.11+.

strawberry/schema/schema_converter.py
```diff
@@ -362,9 +362,16 @@
...
+            default_value = dataclasses.asdict(dataclass_instance)
+        elif model_dump and callable(model_dump) and is_not_type:
```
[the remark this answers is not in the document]
ppease: This will only be called once for each input, and I don't know of a better way to dump the values.  I suppose we could try and cache the values if this became a problem?

tests/schema/test_input.py
```diff
@@ -27,3 +29,72 @@ def filter(self, input: FilterInput) -> str:
     assert not result.errors
     assert result.data
     assert result.data["filter"] == "Hello nope"
+
+
+def test_input_with_nonscalar_field_default():
```
ppease, line 34: I originally thought this had issues, but as long as you specify the default factory, this works as intended.  I didn't see any tests explicitly testing this so I added one.

strawberry/experimental/pydantic/_compat.py
```diff
@@ -17,6 +17,9 @@
 IS_PYDANTIC_V2: bool = PYDANTIC_VERSION.startswith("2.")
 IS_PYDANTIC_V1: bool = not IS_PYDANTIC_V2
 
+if IS_PYDANTIC_V2 and TYPE_CHECKING:
+    import pydantic.v1
```
patrick91, line 21: I think we can remove this 😊

tests/schema/test_generics_nested.py
```diff
@@ -259,9 +259,9 @@ def blocks(self) -> List[BlockInterface]:
         }
 
         """
-        The `JSON` scalar type represents JSON values as specified by [ECMA-404](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf).
+        The `JSON` scalar type represents JSON values as specified by [ECMA-404](https://ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf).
```
patrick91, line 262: can you rebase? I think this was fixed recently

patrick91 (approved):
Sorry for the late reply, I think this looks good!

We still have to figure if we want to remove the default values from the schema or not 🤔  See #3517

But we can merge this either way 😊

tests/schema/test_generics_nested.py
ppease: You were right.  I merged main and this change went away.

patrick91:
@ppease is there any reason why this is draft? just double checking!

strawberry/experimental/pydantic/_compat.py
patrick91: interesting, can you remove it anyway? I'll double check what's wrong

ppease:
> @ppease is there any reason why this is draft? just double checking!

This fix resolves my original issue.  The only reason I made it a draft is I just wanted to make sure the fix made sense architecturally, and there wasn't a more preferred way to fix the issue.

patrick91:
@ppease I think the fix is fine! 😊

ppease:
@patrick91 Sounds good!  Thanks for reviewing it!  I went ahead and marked it as ready.

strawberry/experimental/pydantic/_compat.py
ppease: done!

patrick91:
@ppease thanks! looks like the tests all passed here, so I merged it 

Mypy sometimes does weird things locally with pydantic :D

PR #3522, opened 2024-05-28
GH-3521: Provide a more powerful Info class to resolve_reference

### LINKED ISSUE #3521: Wrong info type passed to resolve_reference classmethod
## Describe the Bug

Take the following example:

```py
class User:
    id: strawberry.ID
    email: str

    @classmethod
    def resolve_reference(cls, id: strawberry.ID, info: strawberry.Info) -> Self:
      print(info.selected_fields)
```

This will result in an error saying the attribute doesn't exist, even though it should. The problem appears to stem from the fact that `GraphQLResolveInfo` is passed through instead in this instance rather than `Info` which provides the attribute.

(This is a federated example.)

## System Information

 - Operating system: MacOS (14.4)
 - Strawberry version (if applicable): 0.232.0

## Additional Context

https://discord.com/channels/689806334337482765/1244954381326090331

parafoxia:
## Description

This PR modifies the `Info` class such that instances are able to be injected into `resolve_reference` classmethods instead of `GraphQLResolveInfo` instances.

This provides more functionality, primarily enabling users to get the selected fields of the query where they weren't able to before. Some attributes (anything related to `self._field`) are unavailable in this state and just return `None`.

It doesn't attempt to port over the converter logic from the unfederated `Schema` class as it was not necessary for this fix, and is more suitable for a follow-up.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* Closes #3521.

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,32 @@
+Release type: patch
+
+Strawberry will now inject `strawberry.Info` instances to `resolve_reference` classmethods instead of `GraphQLResolveInfo` instances.
+
+This allows for the use of the additional functionality provided by `strawberry.Info`, such as filtering what's resolved based on the selected fields:
+
+```py
+@strawberry.federation.type(keys=["id"])
+class User:
+    id: strawberry.ID
+    email: str
+    orders: list[Order] = strawberry.field(default_factory=list)
+
+    @classmethod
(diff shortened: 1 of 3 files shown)


parafoxia:
Obseleted by #3525.

PR #3525, opened 2024-05-29
Implement federation fields and types using strawberry

### LINKED ISSUE #3521: Wrong info type passed to resolve_reference classmethod

patrick91:
- **Add tests for federation schema**
- **Implement federation fields and types using strawberry**

Closes #3521 #3522

### DIFF
--- a/strawberry/federation/schema.py
+++ b/strawberry/federation/schema.py
@@ -113,10 +113,6 @@ class Schema(BaseSchema):
         added if the schema contains an entity type.
         """
 
-        # note we don't add the _entities field here, as we need to know if the
-        # schema contains an entity type first and we do that by leveraging
-        # the schema converter type map, so we don't have to do that twice
-
         import strawberry
         from strawberry.tools.create_type import create_type
         from strawberry.tools.merge_types import merge_types
--- a/strawberry/federation/schema.py
+++ b/strawberry/federation/schema.py
@@ -348,10 +348,10 @@ def _get_entity_type(
         while isinstance(type_, StrawberryContainer):
             type_ = type_.of_type
 
-        if not hasattr(type_, "__strawberry_definition__"):
-            continue
+        type_definition = get_object_definition(type_, strict=False)
 
-        type_definition = get_object_definition(type_, strict=True)
+        if type_definition is None:
+            continue
 
         if type_definition.is_object_type and _has_federation_keys(type_definition):
             entity_types.add(type_)
--- a/strawberry/federation/schema.py
+++ b/strawberry/federation/schema.py
@@ -46,7 +46,7 @@ if TYPE_CHECKING:
     from strawberry.schema_directive import StrawberrySchemaDirective
 
 
-FederationAny = scalar(NewType("_Any", object), name="_Any")
+FederationAny = scalar(NewType("_Any", object), name="_Any")  # type: ignore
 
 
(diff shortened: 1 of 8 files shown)


botberry:



---

This release refactors our Federation integration to create types using
Strawberry directly, instead of using low level types from GraphQL-core.

The only user facing change is that now the `info` object passed to the
`resolve_reference` function is the `strawberry.Info` object instead of the one
coming coming from GraphQL-core. This is a **breaking change** for users that
were using the `info` object directly.

If you need to access the original `info` object you can do so by accessing the
`_raw_info` attribute.

```python
import strawberry


@strawberry.federation.type(keys=["upc"])
class Product:
    upc: str

    @classmethod
    def resolve_reference(cls, info: strawberry.Info, upc: str) -> "Product":
        # Access the original info object
        original_info = info._raw_info

        return Product(upc=upc)
```



Here's the tweet text: 
```

```

parafoxia (approved):
Approving wrt #3521, obvs the rest would need another review lmao. Thanks for this!

strawberry/federation/schema.py
```diff
@@ -354,33 +322,54 @@ def _warn_for_federation_directives(self) -> None:
...
+        while isinstance(type_, StrawberryContainer):
+            type_ = type_.of_type
 
-    entity_type = GraphQLUnionType("_Entity", federation_key_types)  # type: ignore
+        if not hasattr(type_, "__strawberry_definition__"):
+            continue
 
-    def _resolve_type(self: Any, value: Any, info: Any) -> str:
-        return self.__strawberry_definition__.name
+        type_definition = get_object_definition(type_, strict=True)
```
bellini666, line 354: nit: Github will not allow me to comment on all lines, but you can avoid the `hasattr` by using `get_object_definition` directly with `strict=False`

```python
type_definition = get_object_definition(type_, strict=False)
if type_definition is None:
    continue
```

strawberry/federation/schema.py
```diff
@@ -354,33 +322,54 @@ def _warn_for_federation_directives(self) -> None:
...
+
+        if type_definition.is_object_type and _has_federation_keys(type_definition):
+            entity_types.add(type_)
+
+        for field in type_definition.fields:
+            if field.type and field.type not in seen:
+                seen.add(field.type)
+                stack.append(field.type)
```
bellini666, line 362: nit: reduce one indentation here (but ignore if you prefer the other way)

```suggestion
            if field.type and field.type in seen:
                continue

            seen.add(field.type)
            stack.append(field.type)
```

bellini666 (approved):
LGTM 👍🏼 

Left some nits, but ignore if you prefer the other way :)

PR #3529, opened 2024-05-31
fix: Fix an issue preventing typing `root` with `strawberry.Parent`

bellini666:
Fixes https://github.com/strawberry-graphql/strawberry-django/issues/538

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -6,6 +6,7 @@ This release fixes a typing issue where trying to type a `root` argument with
 ```python
 import strawberry
 
+
 @strawberry.type
 class SomeType:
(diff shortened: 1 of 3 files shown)


botberry:



---

This release fixes a typing issue where trying to type a `root` argument with
`strawberry.Parent` would fail, like in the following example:

```python
import strawberry


@strawberry.type
class SomeType:
    @strawberry.field
    def hello(self, root: strawberry.Parent[str]) -> str:
        return "world"
```

This should now work as intended.



Here's the tweet text: 
```

```

patrick91 (approved):

PR #3549, opened 2024-06-24
Add caching for `resolve`

### LINKED ISSUE #3544: Slow performance for queries that return many items
Hello, I'm trying to improve the performance of a graphql query that looks like:

```graphql
query GetSampleEidMapQuery($project: String!) {
  project(name: $project) {
    samples {
      assays {
        id
        externalIds
        meta
      }
    }
  }
}
```

This returns a result that has around 3000 samples, and each sample has between 1 and 4 assays. So less than 10,000 objects in total. The query takes between 3 and 5 seconds and returns around 600kB of json. So it's not a small amount of data but also not exactly huge. I initially thought this might be slow SQL queries but it turns out around 85% of the query time is in strawberry processing the results. Here's the pyinstrument profiling that shows this [pyinstrument.html.zip](https://github.com/user-attachments/files/15923872/pyinstrument.html.zip)

Is there anything that can be done to reduce the time that it takes for strawberry to handle results? I've tried both the `ParserCache` and `ValidationCache` as well as disabling validation entirely but unfortunately that made very little difference. 

Not sure if it helps but this is our graphql schema: https://github.com/populationgenomics/metamist/blob/dev/api/graphql/schema.py

Thank you!

## Upvote & Fund

<a href="https://polar.sh/strawberry-graphql/strawberry/issues/3544">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/3544/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

patrick91:
Closes #3544

This PR adds cache to the `resolve` method to make sure we don't do the same work multiple times.

### DIFF
--- a/strawberry/annotation.py
+++ b/strawberry/annotation.py
@@ -54,7 +54,7 @@ ASYNC_TYPES = (
 
 
 class StrawberryAnnotation:
-    __slots__ = "raw_annotation", "namespace", "__resolved_cached__"
+    __slots__ = "raw_annotation", "namespace", "__resolve_cache__"
 
     def __init__(
         self,
@@ -65,7 +65,7 @@ class StrawberryAnnotation:
         self.raw_annotation = annotation
         self.namespace = namespace
 
-        self.__resolved_cached__: Optional[Union[StrawberryType, type]] = None
+        self.__resolve_cache__: Optional[Union[StrawberryType, type]] = None
 
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, StrawberryAnnotation):
@@ -101,7 +101,7 @@ class StrawberryAnnotation:
     def annotation(self, value: Union[object, str]) -> None:
         self.raw_annotation = value
 
-        self.__resolved_cached__ = None
+        self.__resolve_cache__ = None
 
     def evaluate(self) -> type:
         """Return evaluated annotation using `strawberry.util.typing.eval_type`."""
@@ -129,10 +129,10 @@ class StrawberryAnnotation:
 
     def resolve(self) -> Union[StrawberryType, type]:
         """Return resolved (transformed) annotation."""
-        if self.__resolved_cached__ is None:
-            self.__resolved_cached__ = self._resolve()
+        if self.__resolve_cache__ is None:
+            self.__resolve_cache__ = self._resolve()
 
-        return self.__resolved_cached__
+        return self.__resolve_cache__
 
     def _resolve(self) -> Union[StrawberryType, type]:
         # TODO: I wonder if this resolve should be creating types?
@@ -179,7 +179,7 @@ class StrawberryAnnotation:
         module = sys.modules[field.origin.__module__]
         self.namespace = module.__dict__
 
-        self.__resolved_cached__ = None  # Invalidate cache to allow re-evaluation
+        self.__resolve_cache__ = None  # Invalidate cache to allow re-evaluation
 
     def create_concrete_type(self, evaled_type: type) -> type:
         if has_object_definition(evaled_type):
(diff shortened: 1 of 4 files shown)


strawberry/arguments.py
```diff
@@ -123,7 +124,7 @@ def __init__(
                             arg.resolve_forward_ref(first)
                         )
 
-    @property
+    @cached_property
```
skilkis, line 127: Might be safer to delegate the caching to `StrawberryAnnotation.resolve()` as currently we allow mutability of the underlying `raw_annotation`. Not sure what the use-case would be, but it might confuse some that want to modify annotations at runtime 🧐

patrick91: thanks! I'll update it later or tomorrow 😊

patrick91:
reminder to make a pre-release and test on strawberry django (thanks @skilkis)

strawberry/arguments.py
erikwrede: @skilkis was just about to comment the same :D

erikwrede:
image
We might be able to add additional caching here

patrick91:
@erikwrede yeah, but I'd like to check if we are doing too much useless work first :D

strawberry/annotation.py
```diff
@@ -101,19 +101,17 @@ def annotation(self) -> Union[object, str]:
     def annotation(self, value: Union[object, str]) -> None:
         self.raw_annotation = value
 
+        self.__resolved_cached__ = None
+
     def evaluate(self) -> type:
```
patrick91, line 106: this is used in a couple of places directly, so we might want to reintroduce the cache, but I need to double check why we use it directly

botberry:
# Pre-release
:wave:

Releasing commit [522442125986d5e27fa89ef98ee3b2a5b262344a] to PyPi as pre-release! :package:

botberry:
# Pre-release
:wave:

Pre-release **0.235.1.dev.1719337273** [522442125986d5e27fa89ef98ee3b2a5b262344a] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.235.1.dev.1719337273
```

patrick91:
just tested this on Strawberry Django and all the tests passed 😊

botberry:



---

This release improves the performance when returning a lot of data, especially
when using generic inputs (where we got a 7x speedup in our benchmark!).



Here's the tweet text: 
```

This release improves the performance when returning a lot of data, especially
when using generic inputs (where we got a 7x speedup in our benchmark!).

```

strawberry/annotation.py
```diff
@@ -131,6 +129,15 @@ def _get_type_with_args(
 
     def resolve(self) -> Union[StrawberryType, type]:
         """Return resolved (transformed) annotation."""
+        if self.__resolve_cache__ is None:
+            self.__resolve_cache__ = self._resolve()
+
+        return self.__resolve_cache__
+
+    def _resolve(self) -> Union[StrawberryType, type]:
+        # TODO: I wonder if this resolve should be creating types?
```
patrick91: Leaving this here for the future 😊

    `resolve` is called every time we get the field type, which, turns out,
    is called a lot of times, even when creating the dataclass, maybe
    there's some opportunity for improvements here, but we would need
    to call `resolve` anyway, so it's probably not worth changing how this
    works

strawberry/annotation.py
patrick91: I think we only use it here: https://github.com/strawberry-graphql/strawberry/blob/095463d4d182bfb94c26b08f333b9be31eab359a/strawberry/types/fields/resolver.py#L121

@skilkis do you think it is worth caching the result here too? 😊

bellini666 (approved):

PR #3554, opened 2024-07-02
Support subscriptions extensions

### LINKED ISSUE #2097: Extenstions support for subscriptions.

### LINKED ISSUE #3598: GraphQL Core 3.3 tests appear to be running 3.2 instead
## Describe the Bug

Within the `nox -r -t tests -s "Tests-3.12(gql_core='3.3.0')"` tests (it happens on all Python versions):
```
Package operations: 0 installs, 1 update, 0 removals

  - Downgrading graphql-core (3.3.0a5 https://github.com/graphql-python/graphql-core/archive/876aef67b6f1e1f21b3b5db94c7ff03726cb6bdf.zip -> 3.2.3)
```

This also appears in the logs for the Core 3.2 tests as well. The Pydantic tests have different logs, but they're also installing 3.2 rather than 3.3.

https://github.com/strawberry-graphql/strawberry/actions/runs/10384390719/job/28751301930?pr=3592

## System Information

N/A -- This is a CI bug.

## Upvote & Fund

<a href="https://polar.sh/strawberry-graphql/strawberry/issues/3598">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/3598/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

nrbnlulu:
## Description
Support extensions in subscriptions operations.
Thing to note is that GraphQL core ~3.2 doesn't support middlewares (which are used by our schema extension `resolve`) on subscription. Support for this was added in [876aef6](https://github.com/graphql-python/graphql-core/commit/876aef67b6f1e1f21b3b5db94c7ff03726cb6bdf).
We can initially remove `resolve` support until we upgrade to 3.3 Please LMK what you guys think.

PS: There are few tests that are not passing since I upgraded GraphQL-core (for the midleware support) and there are some incompatibilities (mostly due to spaces in the SDL).  
 

## Types of Changes

- [x] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

fix #2097
closes #3598 

## TODO
- [x] fix integrations tests
- [x] test extensions hooks for many yields
- [x] test first yield is an error
- [x] test that extensions results are cleaned between yields
- [x] remove deprecation's about execution context. we'll do that in another pr
- [x] resolve optimization `TODO`'s after review is accepted. 
### Things I am not certain about
1. `on_execute` was chosen the mitigate the normal `on_execute` per subscription message.
The thing is after the last yield of the subscription we still wrap with `on_execution` causing a last on_execution lifetime although the async gen raised `StopAsyncIteration` I am not sure this is the wanted behavior.
2. Currently I clear `ExecutionContext.extensions_result` on every yield do you think this is the wanted behaviour?

## Summary by Sourcery

This pull request introduces support for schema extensions in subscription operations, allowing hooks to be executed for each yield in a subscription. It also includes significant refactoring to improve error handling and performance, along with comprehensive tests and updated documentation.

- **New Features**:
    - Added support for schema extensions in subscription operations, allowing hooks like `on_execute` to be called for each subscription yield.
- **Enhancements**:
    - Refactored the execution and subscription handling to use internal types instead of graphql-core result types.
    - Improved error handling in subscription operations to differentiate between immediate errors and execution errors.
    - Introduced caching for extensions and middleware manager in the schema to optimize performance.
- **Documentation**:
    - Updated documentation to include examples and explanations for using schema extensions with subscriptions.
- **Tests**:
    - Added comprehensive tests for subscription extensions, including tests for immediate errors, errors after yields, and extension results clearing between yields.
    - Refactored existing tests to align with the new subscription handling and error processing logic.


botberry:



---

Support for schema-extensions in subscriptions.

i.e:
```python

    async def on_operation(self, execution_context):
        #  This part is called before the subscription started
        yield
        #  This part is called after the subscription ended

    async def on_execute(self, execution_context):
        #  This part is called before a new result is comming from the subscription
        yield
        #  This part is called after the subscription yielded a result
```



Here's the tweet text: 
```

```

nrbnlulu:
This pr can be considered ready for review. There are few tests that are not passing since I upgraded GraphQL-core (for the midleware support). 

We can initially remove `resolve` support until my pr there would be accepted (or not :upside_down_face: ) Please LMK what you guys think.

RELEASE.md
```diff
@@ -0,0 +1,18 @@
+Release type: minor
+
+Support for schema-extensions in subscriptions.
+
```
DoctorJohn, line 4: We should mention that this is a breaking release, requiring all extensions to be instanciated and possibly all extension hooks to be updated.

We could possible use our favorite trick of making the execution_context argument requestable to stay backwards compatible. And when extensions try to access self.execution_context we could raise a deprecation warning.

pyproject.toml
```diff
@@ -35,7 +35,8 @@ build-backend = "poetry.masonry.api"
 
 [tool.poetry.dependencies]
 python = "^3.8"
-graphql-core = "~3.2.0"
+# TOOD: revert this
+graphql-core = {git = "https://github.com/graphql-python/graphql-core.git", branch = "main"}
```
DoctorJohn, line 39: In case we want to merge this before an official graphql-core relase, we should pin the commit in which your changes landed.

DoctorJohn (changes requested):
Thanks for the PR @nrbnlulu let's hope this attempt will be the one. I started reviewing this and realized I need to play with it a bit to leave meaningful feedback. So here are just two (likely obvious) comments to start with before I forget to post them. Will take a closer look at the rest now and the following days.

RELEASE.md
nrbnlulu: Actually I intentionally leaved it for another PR. the current extensions behavior remains unchained.
Do you think it should be done here?

strawberry/extensions/context.py
```diff
@@ -56,10 +58,13 @@ def __init_subclass__(cls) -> None:
     LEGACY_ENTER: str
     LEGACY_EXIT: str
 
-    def __init__(self, extensions: List[SchemaExtension]) -> None:
+    def __init__(
+        self, extensions: List[SchemaExtension], execution_ctx: ExecutionContext
+    ) -> None:
         self.hooks: List[WrappedHook] = []
         self.default_hook: Hook = getattr(SchemaExtension, self.HOOK_NAME)
         for extension in extensions:
```
nrbnlulu, line 66: maybe this could be done earlier

strawberry/extensions/context.py
```diff
@@ -56,10 +58,13 @@ def __init_subclass__(cls) -> None:
...
+            extension.execution_context = execution_ctx
```
nrbnlulu, line 67: maybe this could be done earlier

RELEASE.md
DoctorJohn: > the current extensions behavior remains unchained.

I'm sorry, I got tricked by GitHub's review UI. I thought the code instantiating extensions if needed was gone, but it has just been moved.

strawberry/schema/subscribe.py
```diff
@@ -0,0 +1,154 @@
...
+    )
+    # GrapQL-core might return an initial error result instead of an async iterator.
+    # This happens when "there was an immediate error" i.e resolver is not an async iterator.
+    # To overcome this while maintaining the extension contexts we do this trick.
+    first = await asyncgen.__anext__()
+    if isinstance(first, ExecutionResultError):
+        # close the async generator (calling `.aclose()` on it won't close the context managers)
+        with contextlib.suppress(StopAsyncIteration):
+            await asyncgen.__anext__()
+        return first
```
bellini666, line 146: Why `.aclose()` will not close it?

Also, maybe use [aclosing](https://docs.python.org/3/library/contextlib.html#contextlib.aclosing) (it is only available to python 3.10+ but we can backport it) to make sure we always close it no matter what?

```python
# Copied from cpython's code for contextlib.aclosing
class aclosing(Generic[_I], contextlib.AbstractAsyncContextManager):
    def __init__(self, thing: _I):
        super().__init__()
        self.thing = thing

    async def __aenter__(self) -> _I:
        return self.thing

    async def __aexit__(self, *exc_info):
        await self.thing.aclose()
```

If we really need the extra `__anext__` we can pass an extra `await_next_on_exit: bool = False` to it and do that before the `.aclose` itself

update: I also checked some docs here and `.aclose` should make the context close, unless it was using `yield` without a `try: finally` clause. But in this case it is using a class with `__exit__`/`__aexit__`

strawberry/schema/subscribe.py
```diff
@@ -0,0 +1,154 @@
...
+            # Handle immediate errors.
+            if isinstance(agen_or_result, OriginalExecutionResult):
+                yield await _handle_execution_result(
+                    execution_context,
+                    ExecutionResultError(data=None, errors=agen_or_result.errors),
+                    extensions_runner,
+                    process_errors,
+                )
+            else:
+                aiterator = agen_or_result.__aiter__()
```
bellini666, line 83: I comment below about using `aclosing`, this is another good place to use it

bellini666 (commented):
Left some comments, will take a better look later in the weekend :)

tests/schema/test_subscription.py
```diff
@@ -236,3 +238,51 @@ async def example(
...
+
+    @strawberry.type
+    class Subscription:
+        @strawberry.subscription()
+        async def example(self) -> AsyncGenerator[str, None]:
+            return "fds"
+
+    schema = strawberry.Schema(query=Query, subscription=Subscription)
+
+    query = """#graphql
```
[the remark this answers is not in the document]
nrbnlulu: Yep, syntax highlighting

tests/websockets/test_graphql_transport_ws.py
```diff
@@ -54,6 +54,20 @@ async def ws(ws_raw: WebSocketClient) -> WebSocketClient:
...
+    Assert that the NextMessage payload contains the provided data.
+    If extensions is provided, it will also assert that the
+    extensions are present
+    """
+    assert response["type"] == "next"
+    assert response["id"] == id
+    assert set(response["payload"].keys()) <= {"data", "errors", "extensions"}
+    assert response["payload"]["data"] == data
+    if extensions is not None:
+        assert response["payload"]["extensions"] == extensions
```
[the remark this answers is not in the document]
nrbnlulu: Should it be `[]` or not included at all?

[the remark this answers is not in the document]
bellini666: Hrmm good question! I would say `[]` but maybe the team prefers the other way

RELEASE.md
```diff
@@ -0,0 +1,18 @@
+Release type: minor
+
+Support for schema-extensions in subscriptions.
```
patrick91, line 3: can we add an example of how this could be used? so people know what kind of functionality this enables?

RELEASE.md
```diff
@@ -0,0 +1,18 @@
...
+
+    async def on_operation(self, execution_context):
+        #  This part is called before the subscription starts
+        yield
+        #  This part is called after the subscription ends
+
+    async def on_execute(self, execution_context):
+        #  This part is called before a new result is coming from the subscription
+        yield
+        #  This part is called after the subscription yields a result
```
patrick91, line 17: this runs for all operations, right? I wonder if we should explain how to differentiate between queries/mutations and subscriptions

pyproject.toml
patrick91: could we conditionally support it based on the version of GraphQL core?

strawberry/extensions/context.py
```diff
@@ -56,10 +58,13 @@ def __init_subclass__(cls) -> None:
     LEGACY_ENTER: str
     LEGACY_EXIT: str
 
-    def __init__(self, extensions: List[SchemaExtension]) -> None:
+    def __init__(
+        self, extensions: List[SchemaExtension], execution_ctx: ExecutionContext
```
patrick91, line 62:
```suggestion
        self, extensions: List[SchemaExtension], execution_context: ExecutionContext
```

let's keep the long name, since we use that in other places too 😊

strawberry/extensions/runner.py
```diff
@@ -25,62 +23,41 @@ class SchemaExtensionsRunner:
...
     def get_extensions_results_sync(self) -> Dict[str, Any]:
         data: Dict[str, Any] = {}
-
         for extension in self.extensions:
-            if inspect.iscoroutinefunction(extension.get_results):
-                msg = "Cannot use async extension hook during sync execution"
-                raise RuntimeError(msg)
-
-            data.update(extension.get_results())  # type: ignore
+            if extension.get_results is not SchemaExtension.get_results:
```
patrick91, line 46: why this change?

strawberry/schema/execute.py
```diff
@@ -67,83 +72,86 @@ def validate_document(
     )
 
 
-def _run_validation(execution_context: ExecutionContext) -> None:
+def _run_validation(context: ExecutionContext) -> None:
```
patrick91, line 75: do we need this change? 🤔

strawberry/schema/execute.py
```diff
@@ -67,83 +72,86 @@ def validate_document(
...
+        assert context.graphql_document
+        context.errors = validate_document(
+            context.schema._schema,
+            context.graphql_document,
+            context.validation_rules,
         )
 
 
-async def execute(
+def execute_sync(
```
patrick91, line 87: could you move the function definitions to be in the same order that were before? :D 

it should make it easier to review this file 😊

strawberry/schema/schema.py
```diff
@@ -67,6 +66,13 @@
 }
 
 
+def _implements_resolve(obj: SchemaExtension | type[SchemaExtension]) -> bool:
+    """Return whether the extension implements the resolve method."""
+    return bool(
+        (ret := getattr(obj, "resolve", None)) and ret is not SchemaExtension.resolve
+    )
```
patrick91, line 73: this seems similar to a check above, can move it to a class property? 😊

strawberry/schema/schema.py
```diff
@@ -179,15 +185,39 @@ def __init__(
...
-            extensions.append(DirectivesExtensionSync if sync else DirectivesExtension)
+    def get_extensions(self, sync: bool = False) -> List[SchemaExtension]:
+        init_extensions = []
+        extensions = (
+            *self.extensions,
+            DirectivesExtensionSync if sync else DirectivesExtension,
+        )
+        for extension in extensions:
+            # If the extension has already been instantiated then set the
+            # `execution_context` attribute
```
patrick91, line 198: we are not doing this here, right?

strawberry/schema/schema.py
```diff
@@ -241,40 +271,62 @@ def get_fields(
     ) -> List[StrawberryField]:
         return type_definition.fields
 
-    async def execute(
```
patrick91, line 244: let's reorder the methods here too 😊

strawberry/types/execution.py
```diff
@@ -92,5 +97,14 @@ class ExecutionResult:
     extensions: Optional[Dict[str, Any]] = None
 
 
+class ExecutionResultError(ExecutionResult):
+    """
+    Used to diffrentiate between a successful subscription result and an immediate error.
```
patrick91, line 102:
```suggestion
    Used to differentiate between a successful subscription result and an immediate error.
```

patrick91 (commented):
did an initial review, thanks so much for doing this 😊

RELEASE.md
nrbnlulu: Yeah I'll add a section in the docs

strawberry/extensions/runner.py
nrbnlulu: no need to call this for extensions that not implementing this function.

strawberry/schema/execute.py
nrbnlulu: ah yeah sorry.

pyproject.toml
nrbnlulu: Strawberry tests won't pass 3.3
https://github.com/strawberry-graphql/strawberry/actions/runs/9928799671/job/27425551808?pr=3570#step:6:7838

I think it is possible to support it based on the version but I am not sure it would be the place to add support for 3.3. maybe in another PR.

tests/websockets/test_graphql_transport_ws.py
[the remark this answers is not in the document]
nrbnlulu: Note that as per the spec it is completely optional
https://spec.graphql.org/October2021/#sel-FAPHJCBCBGyoB

strawberry/types/execution.py
```diff
@@ -51,6 +54,8 @@ class ExecutionContext:
     graphql_document: Optional[DocumentNode] = None
     errors: Optional[List[GraphQLError]] = None
     result: Optional[GraphQLExecutionResult] = None
+    extensions_results: Dict[str, Any] = dataclasses.field(default_factory=dict)
+    context_class: Optional[Type[GraphQLExecutionContext]] = None
```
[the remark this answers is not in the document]
nrbnlulu: Actually I don't think I use it anymore :upsidedown:

strawberry/schema/subscribe.py
nrbnlulu: ah this was because of try finally thing. I think it is solved by this https://github.com/nrbnlulu/strawberry/blob/61154dcfab222051b56a77d13c49a40fe203fda8/strawberry/extensions/context.py#L206 though

strawberry/schema/execute.py
nrbnlulu: nope :monkey:

strawberry/subscriptions/protocols/graphql_ws/handlers.py
```diff
@@ -160,22 +162,26 @@ async def handle_keep_alive(self) -> None:
 
     async def handle_async_results(
         self,
-        result_source: AsyncGenerator,
+        result_source: AsyncIterator[ExecutionResult],
```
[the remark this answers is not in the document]
nrbnlulu: > an AsyncIterator is also an AsyncGenerator
Are you sure? I think its the other way around :thinking:

strawberry/schema/subscribe.py
nrbnlulu: how so? AsyncIterator has no `.aclose`?

strawberry/extensions/runner.py
```diff
@@ -25,65 +23,45 @@ class SchemaExtensionsRunner:
...
+        return ParsingContextManager(self.extensions, self.execution_context)
 
     def executing(self) -> ExecutingContextManager:
-        return ExecutingContextManager(self.extensions)
+        return ExecutingContextManager(self.extensions, self.execution_context)
+
+    @classmethod
+    def _implments_get_rseults(cls, extension: SchemaExtension) -> bool:
+        """Whether the extension implements get_results."""
+        return type(extension).get_results is not SchemaExtension.get_results
```
DoctorJohn, line 46: Does anyone know whether these explicit checks might be slower than just calling the method and catching possible `NotImplementedError` exceptions (which we could let the unimplemented method throw by default).

RELEASE.md
```diff
@@ -0,0 +1,56 @@
...
+        yield
+        # the subscription has yielded a new result
+        self.monitor_end(self.execution_context.operation_name, f"after yield {count}")
+        count += 1
+
+    # other hooks can are the same as per normal execution.
+
+
+schema = strawberry.Schema(
+    query=Query, subscription=Subscription, extensions=[MyExtension()]
```
DoctorJohn, line 54: As of right now this code would not work, because the extension can only be initialized with a `execution_context` argument. This would work:

```suggestion
    query=Query, subscription=Subscription, extensions=[MyExtension]
```

DoctorJohn (changes requested):
I left some more comments while pondering over the overall design.

> Currently I clear ExecutionContext.extensions_result on every operation do you think this is the wanted behaviour?

I think having `extensions_result` reflect the current value of the subscription event is indeed the behaviour we want. Thought asking whether the current implementation of this behaviour is well implement this way is a good question :thinking:

RELEASE.md
```diff
@@ -0,0 +1,56 @@
+Release type: minor
+
+Support for schema-extensions in subscriptions.
+
+i.e:
+```python
```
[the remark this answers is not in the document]
nrbnlulu: Do you have an idea for something like this?

strawberry/extensions/runner.py
nrbnlulu: Actually seems thats a downgrade.
https://perfpy.com/743

Maybe we can cache them though.

strawberry/extensions/runner.py
```diff
@@ -25,65 +23,45 @@ class SchemaExtensionsRunner:
...
     def get_extensions_results_sync(self) -> Dict[str, Any]:
         data: Dict[str, Any] = {}
-
         for extension in self.extensions:
-            if inspect.iscoroutinefunction(extension.get_results):
-                msg = "Cannot use async extension hook during sync execution"
-                raise RuntimeError(msg)
-
-            data.update(extension.get_results())  # type: ignore
+            if type(self)._implments_get_rseults(extension):
```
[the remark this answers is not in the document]
nrbnlulu: I thought that because its a classmethod it is more clear... but I don't have much against just using self.

RELEASE.md
```diff
@@ -0,0 +1,56 @@
...
+
+    async def on_operation(self):
+        # This would run when the subscription starts
+        self.monitor_start(self.execution_context.operation_name)
+        yield
+        # The subscription has ended
+        self.monitor_end(self.execution_context.operation_name)
+
+    async def on_execute(self):
+        count = 0
```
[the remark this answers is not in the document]
nrbnlulu: Yeah, shame I missed that :upside_down_face: I think using self is not good as well due to #3571 
maybe injecting that on execution context would be better.

strawberry/channels/testing.py
```diff
@@ -144,9 +144,9 @@ async def subscribe(
             message_type = response["type"]
             if message_type == NextMessage.type:
                 payload = NextMessage(**response).payload
-                ret = ExecutionResult(payload["data"], None)
+                ret = ExecutionResult(payload.get("data"), None)
                 if "errors" in payload:
-                    ret.errors = self.process_errors(payload["errors"])
+                    ret.errors = self.process_errors(payload.get("errors") or [])
```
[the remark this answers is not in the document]
Speedy1991: > In case you want to use .get, wouldn't .get("errors", []) be shorter than .get("errors") or []

A littlebit offtopic, but there are cases where a `or` makes more sense than a default value on `get`, e.g. to get a robust `False` value but not `None`:

```
d = dict(a=1, b=None)
value = d.get('b', False)  # value is None
value = d.get('b') or False  # value is False
```

nrbnlulu:
@DoctorJohn 
> Thought asking whether the current implementation of this behaviour is well implement this way is a good question 🤔
wym?

nrbnlulu:
Why are the tests passing on 3.2.3
https://ih1.redbubble.net/image.875111905.4798/flat,750x,075,f-pad,750x1000,f8f8f8.jpg

nrbnlulu:
Draftrd since I have found a bug when using contextvars

tests/schema/extensions/schema_extensions/conftest.py
```diff
@@ -0,0 +1,123 @@
...
+    return request.param
+
+
+@contextlib.contextmanager
+def hook_wrap(list_: List[str], hook_name: str):
+    list_.append(f"{hook_name} Entered")
+    try:
+        yield
+    finally:
+        list_.append(f"{hook_name} Exited")
```
bellini666, line 93: nit: maybe a "category" style would be better than a human readable one in this case?

no strong feelings though

```suggestion
    list_.append(f"enter: {hook_name}")
    try:
        yield
    finally:
        list_.append(f"exit: {hook_name}")
```

tests/benchmarks/test_subscriptions.py
```diff
@@ -24,3 +25,23 @@ async def _run():
             assert value.data["something"] == "Hello World!"
 
     benchmark(lambda: asyncio.run(_run()))
```
bellini666, line 27: This did regress a lot. Is that expected? Any way to improve that?

strawberry/utils/__init__.py
```diff
@@ -0,0 +1,3 @@
+import graphql
+
+IS_GQL_32 = "3.3" not in graphql.__version__
```
bellini666, line 3: Maybe better to try to parse this? I'm almost sure we have a dependency which can transform this into a tuple like `(3, 3, ...)`. Then this could be a comparison like `version < (3, 3)`, meaning it would not break even if 3.4/4.0 gets released

strawberry/subscriptions/protocols/graphql_ws/handlers.py
```diff
@@ -122,29 +130,14 @@ async def handle_start(self, message: OperationMessage) -> None:
 
         if self.debug:
             pretty_print_graphql_operation(operation_name, query, variables)
+        result_source = self.schema.subscribe(
```
bellini666, line 133:
```suggestion

        result_source = self.schema.subscribe(
```

strawberry/schema/schema.py
```diff
@@ -214,15 +214,64 @@ class Query:
...
+            provided_operation_name=operation_name,
+        )
 
-        return extensions
+    @classmethod
+    def _implements_resolve(cls, obj: SchemaExtension | type[SchemaExtension]) -> bool:
+        """Whether the extension implements the resolve method."""
+        if isinstance(obj, SchemaExtension):
+            return type(obj).resolve is not SchemaExtension.resolve
+        return obj.resolve is not SchemaExtension.resolve
```
bellini666, line 274: nit: This could be a `SchemaExtension` classmethod

also, `return obj.resolve is not SchemaExtension.resolve` would fail with `AttributeError` if the obj doesn't have a resolve key in its dict

strawberry/schema/execute.py
```diff
@@ -76,117 +85,132 @@ def _run_validation(execution_context: ExecutionContext) -> None:
         )
 
 
+async def _parse_and_validate_async(
+    context: ExecutionContext, extensions_runner: SchemaExtensionsRunner
+) -> Optional[PreExecutionError]:
+    if not context.query:
+        raise MissingQueryError()
+    async with extensions_runner.parsing():
```
bellini666, line 93:
```suggestion

    async with extensions_runner.parsing():
```

bellini666 (commented):
Left some nitpicks, nothing major. LGTM in general after both my reviews

I think either @DoctorJohn and/or @patrick91 are the ones who will be able to actually tell if this is good to go

strawberry/extensions/base_extension.py
```diff
@@ -21,9 +21,9 @@ class LifecycleStep(Enum):
 class SchemaExtension:
     execution_context: ExecutionContext
 
-    def __init__(self, *, execution_context: ExecutionContext) -> None:
-        self.execution_context = execution_context
-
+    # to support extensions that still use the old signature
+    # we have an optional argument here for ease of initialization.
+    def __init__(self, execution_context: ExecutionContext | None = None) -> None: ...
```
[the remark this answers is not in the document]
nrbnlulu: yeah for sure though I though to do it in #3571

tests/benchmarks/test_subscriptions.py
nrbnlulu: Yeah probably expected, because it tests the lifespan part (which got significantly more complicated).
It is expected to be much closer in the new benchmark I added.

tests/schema/extensions/schema_extensions/conftest.py
nrbnlulu: This would change many tests. I would not do this because it would trigger file changed for who already reviewed them. LMK if its critical for you.

strawberry/schema/subscribe.py
```diff
@@ -0,0 +1,166 @@
...
+        process_errors,
+        middleware_manager,
+        execution_context_class,
+    )
+    # GrapQL-core might return an initial error result instead of an async iterator.
+    # This happens when "there was an immediate error" i.e resolver is not an async iterator.
+    # To overcome this while maintaining the extension contexts we do this trick.
+    first = await asyncgen.__anext__()
+    if isinstance(first, PreExecutionError):
+        await asyncgen.aclose()
```
patrick91, line 158: couldn't we just check if there's an error to the response, isn't that enough?

tests/benchmarks/test_subscriptions.py
patrick91: can we move that benchmark to another PR? so we can see the new difference :)

noxfile.py
```diff
@@ -47,9 +47,10 @@ def _install_gql_core(session: Session, version: str) -> None:
     # hack for better workflow names  # noqa: FIX004
     if version == "3.2.3":
         session._session.install(f"graphql-core=={version}")  # type: ignore
-    session._session.install(
-        "https://github.com/graphql-python/graphql-core/archive/876aef67b6f1e1f21b3b5db94c7ff03726cb6bdf.zip"
-    )  # type: ignore
+    else:
+        session._session.install(
+            "https://github.com/graphql-python/graphql-core/archive/876aef67b6f1e1f21b3b5db94c7ff03726cb6bdf.zip"
+        )  # type: ignore
```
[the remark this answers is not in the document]
nrbnlulu: My change only landed in 876aef67b6f1e1f21b3b5db94c7ff03726cb6bdf
image

strawberry/extensions/runner.py
```diff
@@ -13,39 +11,22 @@
...
-        init_extensions: List[SchemaExtension] = []
-
-        for extension in extensions:
-            # If the extension has already been instantiated then set the
-            # `execution_context` attribute
-            if isinstance(extension, SchemaExtension):
-                extension.execution_context = execution_context
-                init_extensions.append(extension)
-            else:
-                init_extensions.append(extension(execution_context=execution_context))
```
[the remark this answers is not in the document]
nrbnlulu: yep

strawberry/schema/execute.py
```diff
@@ -249,23 +279,17 @@ def execute_sync(
...
         raise e
     except Exception as exc:
-        error = (
-            exc
-            if isinstance(exc, GraphQLError)
-            else GraphQLError(str(exc), original_error=exc)
-        )
-        execution_context.errors = [error]
-        process_errors([error], execution_context)
+        errors = [_coerce_error(exc)]
```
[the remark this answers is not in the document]
nrbnlulu: I think not. I just wrapped it in a function. it makes sure that arbitrary exceptions are wrapped in `GraphQLError`

strawberry/schema/schema.py
```diff
@@ -214,15 +214,64 @@ class Query:
...
+                extensions = [
+                    *self.extensions,
+                    DirectivesExtensionSync if sync else DirectivesExtension,
+                ]
+            extensions.extend(self.extensions)
+            self._cached_extensions = [
+                ext if isinstance(ext, SchemaExtension) else ext(execution_context=None)
+                for ext in extensions
+            ]
+        return self._cached_extensions
```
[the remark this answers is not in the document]
nrbnlulu: I am actually worried that hashing `self` would consume...

strawberry/schema/subscribe.py
nrbnlulu: hmm might be. though this is more explicit IMO. Websockets integrations rely on this on protocol level.

docs/guides/custom-extensions.mdx
```diff
@@ -150,7 +150,8 @@ class MyExtension(SchemaExtension):
 
 - Execution
 
-`on_execute` can be used to run code on the execution step of the GraphQL
+`on_execute` can be used to run code on the execution step of the GraphQL. When
+using subscriptions, `on_execute` will be called for each subscription yield.
```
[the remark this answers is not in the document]
nrbnlulu: That's a heavy question. I tried to look some references in other implementations though it seems like we are pioneers :upside_down_face: 
https://github.com/graphql-go/graphql/blob/f2b39caf7c92cab3f3f92726b55f59c74f54cfbc/subscription.go#L26

IMO a new hook might be better assuming `stream` and `defer` can benefit from it. otherwise I think `execute`
should suffice. 
Also I can't think of something you can do with the current `on_execute` that you can't do with `on_operation`  and no ready made extension uses it as well.
image

strawberry/schema/schema.py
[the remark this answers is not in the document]
patrick91: not sure I understand :D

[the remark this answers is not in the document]
nrbnlulu: oh I thought that `__hash__` in schema hashes the schema as a string... ig we can do 
```py
def __hash__(self) -> int:
	id(self)
```
(it is required that `self` would be hashable for cached_prop)

strawberry/channels/testing.py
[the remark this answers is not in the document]
nrbnlulu: It causes mypy error for some reason

tests/benchmarks/test_subscriptions.py
nrbnlulu: [#3593](https://github.com/strawberry-graphql/strawberry/pull/3593)

nrbnlulu: image
VS
image

Speedy1991:
I have a question about this MR:
As far i understand the extensions are wrapped around the yield of the async generator?
If I have dataloaders on the info.context object, what will happen to the dataloaders?

```
    @strawberry.subscription
    async def example(self, info: Info, argument: strawberry.ID) -> MyType:
        # dataloaders are injected via extension
        obj = await info.context['my_dataloader'].load(argument)
        while True:
            await asyncio.sleep(0.5)
            value = pubsub.get_message()
            if not value:
                continue
            # What happens with my dataloaders injected via extension? the old ones are destroyed/overwritten?
            yield MyType(obj=obj)
```

patrick91:
@Speedy1991 do you want to try with a pre-release?

nrbnlulu:
> I have a question about this MR: As far as I understand the extensions are wrapped around the yield of the async generator? If I have dataloaders on the info.context object, what will happen to the dataloaders?
> 
> ```
>     @strawberry.subscription
>     async def example(self, info: Info, argument: strawberry.ID) -> MyType:
>         obj = await info.context['my_dataloader'].load(argument)
>         while True:
>             await asyncio.sleep(0.5)
>             value = pubsub.get_message()
>             if not value:
>                 continue
>             # What happens with my dataloaders injected via extension? the old ones are destroyed/overwritten?
>             yield MyType(obj=obj)
> ```

They should stay in the context for as long as the subscription is running unless you do some explicit logic that clears them.

strawberry/schema/execute.py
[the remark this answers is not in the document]
DoctorJohn: https://github.com/nrbnlulu/strawberry/blob/1fc8c5467c13026a502a78cb862b4e28f9f34d07/strawberry/schema/execute.py#L145-L148

Looks like it still behaves as before.

tests/schema/test_get_extensions.py
```diff
@@ -29,32 +29,36 @@ def test_returns_empty_list_when_no_custom_directives():
 def test_returns_extension_passed_by_user():
     schema = strawberry.Schema(query=Query, extensions=[MyExtension])
 
-    assert schema.get_extensions() == [MyExtension]
+    assert isinstance(schema.get_extensions()[0], MyExtension)
```
DoctorJohn, line 32: Previously the test also made sure the correct amount of extensions are returned.

```suggestion
    assert len(schema.get_extensions()) == 1
    assert isinstance(schema.get_extensions()[0], MyExtension)
```

tests/schema/test_get_extensions.py
```diff
@@ -29,32 +29,36 @@ def test_returns_empty_list_when_no_custom_directives():
...
 
 
 def test_returns_directives_extension_when_passing_directives():
     schema = strawberry.Schema(query=Query, directives=[uppercase])
 
-    assert schema.get_extensions() == [DirectivesExtension]
+    assert isinstance(schema.get_extensions()[0], DirectivesExtension)
```
DoctorJohn, line 38:
```suggestion
    assert len(schema.get_extensions()) == 1
    assert isinstance(schema.get_extensions()[0], DirectivesExtension)
```

tests/schema/test_get_extensions.py
```diff
@@ -29,32 +29,36 @@ def test_returns_empty_list_when_no_custom_directives():
...
+        schema.get_extensions(), [MyExtension, DirectivesExtension]
+    ):
+        assert isinstance(ext, ext_cls)
 
 
 def test_returns_directives_extension_when_passing_directives_sync():
     schema = strawberry.Schema(query=Query, directives=[uppercase])
 
-    assert schema.get_extensions(sync=True) == [DirectivesExtensionSync]
+    assert isinstance(schema.get_extensions(sync=True)[0], DirectivesExtensionSync)
```
DoctorJohn, line 54:
```suggestion
    assert len(schema.get_extensions(sync=True)) == 1
    assert isinstance(schema.get_extensions(sync=True)[0], DirectivesExtensionSync)
```

DoctorJohn (changes requested):
Looks like the PR mature quite a bit. I got halfway through reviewing it today. I'll review the changes to the websocket protocol implementations in the next 2 days.

RELEASE.md
```diff
@@ -0,0 +1,116 @@
...
+after yield 0
+before yield 1
+after yield 1
+{'notifications': 'Hello, world!'}
+before yield 2
+after yield 2
+{'notifications': 'Hello, world!'}
+before yield 3
+after yield 3
+{'notifications': 'Hello, world!'}
```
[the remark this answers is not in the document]
nrbnlulu: Have you ran it with the latests graphql-core?

RELEASE.md
```diff
@@ -0,0 +1,116 @@
...
+    def hello(self) -> str:
+        return "Hello, world!"
+
+
+@strawberry.type
+class Subscription:
+    @strawberry.subscription
+    async def notifications(self, info: strawberry.Info) -> AsyncIterator[str]:
+        for _ in range(3):
+            yield "!dlrow ,olleH"
```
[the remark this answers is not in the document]
nrbnlulu: Not sure wym

RELEASE.md
```diff
@@ -0,0 +1,116 @@
...
+after yield 4
+Subscription ended
+```
+### Breaking changes
+This release also updates the signature of `Schema.subscribe`.
+From:
+```py
+async def subscribe(
+    self,
+    # TODO: make this optional when we support extensions
```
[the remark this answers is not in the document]
nrbnlulu: Oops

strawberry/extensions/base_extension.py
```diff
@@ -21,9 +21,11 @@ class LifecycleStep(Enum):
 class SchemaExtension:
     execution_context: ExecutionContext
 
-    def __init__(self, *, execution_context: ExecutionContext) -> None:
-        self.execution_context = execution_context
-
+    # to support extensions that still use the old signature
+    # we have an optional argument here for ease of initialization.
```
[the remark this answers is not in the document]
nrbnlulu: no, its the `execution_context` argument which is useless by now.

RELEASE.md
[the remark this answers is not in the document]
DoctorJohn: Very good point, I was on their pre-release version which does not contain your middleware realted changes yet. Sorry.

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -243,106 +244,66 @@ async def handle_subscribe(self, message: SubscribeMessage) -> None:
         if isinstance(context, dict):
             context["connection_params"] = self.connection_params
         root_value = await self.get_root_value()
-
+        result_source: Awaitable[ExecutionResult] | Awaitable[SubscriptionResult]
```
DoctorJohn, line 247: Nice :100:

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -243,106 +244,66 @@ async def handle_subscribe(self, message: SubscribeMessage) -> None:
...
+                assert first_res_or_agen.errors
+                await operation.send_initial_errors(first_res_or_agen.errors)
+            # that's a mutation result
+            elif isinstance(first_res_or_agen, ExecutionResult):
+                await operation.send_next(first_res_or_agen)
+                await operation.send_message(CompleteMessage(id=operation.id))
+            else:
+                async for result in first_res_or_agen:
+                    await operation.send_next(result)
+                await operation.send_message(CompleteMessage(id=operation.id))
```
DoctorJohn, line 296: I appreciate the increased readability! LGTM in regards to protocol-compliance too

tests/websockets/test_graphql_transport_ws.py
```diff
@@ -962,3 +911,20 @@ async def test_subscription_errors_continue(ws: WebSocketClient):
         response = await ws.receive_json()
         assert response["type"] == CompleteMessage.type
         assert response["id"] == "sub1"
+
+
+@patch.object(MyExtension, MyExtension.get_results.__name__, return_value={})
+async def test_no_extenions(mock: Mock, ws: WebSocketClient):
```
DoctorJohn, line 917: Small typo in the tests name. IMHO it would hurt if this test and the corresponding test below for `graphql_ws` would have a slightly more descriptive name making it more obvious what's being tested.

```suggestion
async def test_no_extensions(mock: Mock, ws: WebSocketClient):
```

DoctorJohn (approved):
Just finished reviewing the websocket protocol parts. Looks like readability went up in some places and we still follow the protocol specs :tada: I left some minor comments and my approval because IMHO this PR is pretty much good to go.

strawberry/schema/schema.py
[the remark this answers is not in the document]
bellini666: I think what @patrick91 means in here would be to do:

```python
@cached_property
def extensions(self):
    extensions = []
    ...
    return extensions
```

But having said that, there's a problem in here. You have a `sync: bool = False` argument that is not used in the cache, meaning that it would cache different stuff depending if the first call was with `sync=True` or `sync=False`

If this continues to be a method, I think the cache need to take that flag into consideration and cache different stuff

If we go the `cached_property` way (which I personally like`, this can be solved by defining both `extensions` and `sync_extensions`.

What do you think?

bellini666 (commented):
One last review round. After this one I'm ready to actually approve this

nrbnlulu:
> What you have right now will not check what response["payload"]["extensions"] is at all

That is intended. I don't want to check it for all the tests that are using this function. 
only when I test extensions.

> My suggestion (and I think @DoctorJohn is saying the same) is to also check it when extensions is None, to make sure > that it contains an empty list/dict/whatever would be returned in that case

That is actually tested [here](https://github.com/nrbnlulu/strawberry/blob/ee1ee1c5299c49ca3503017b4d6b28e651530003/tests/websockets/test_graphql_ws.py#L567)

strawberry/schema/subscribe.py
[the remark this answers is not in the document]
nrbnlulu: Why would it be though?

tests/websockets/test_graphql_transport_ws.py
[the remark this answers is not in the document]
nrbnlulu: > What you have right now will not check what response["payload"]["extensions"] is at all

That is intended. I don't want to check it for all the tests that are using this function. 
only when I test extensions.

> My suggestion (and I think @DoctorJohn is saying the same) is to also check it when extensions is None, to make sure > that it contains an empty list/dict/whatever would be returned in that case

That is actually tested [here](https://github.com/nrbnlulu/strawberry/blob/ee1ee1c5299c49ca3503017b4d6b28e651530003/tests/websockets/test_graphql_ws.py#L567)

strawberry/schema/subscribe.py
```diff
@@ -0,0 +1,166 @@
...
+        return first
+
+    async def _wrapper() -> AsyncGenerator[ExecutionResult, None]:
+        yield first
+        async for result in asyncgen:
+            yield result
```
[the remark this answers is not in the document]
nrbnlulu: Can you think of a cenario that this would happend? I'll add a test for it.

strawberry/schema/schema.py
```diff
@@ -214,15 +214,57 @@ class Query:
...
+        self, execution_context: ExecutionContext, extensions: list[SchemaExtension]
+    ) -> SchemaExtensionsRunner:
+        return SchemaExtensionsRunner(
+            execution_context=execution_context,
+            extensions=extensions,
+        )
 
-        if self.directives:
-            extensions.append(DirectivesExtensionSync if sync else DirectivesExtension)
+    def _get_middleware_manager(
```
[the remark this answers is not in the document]
nrbnlulu: This one is kinda trickier because `extensions` are not just a boolean argument. I think the current approuch suffice.
LMK if thats a blocker, I'll think of something better.

patrick91 (approved):
let's go!

patrick91 (approved):
let's go!

PR #3556, opened 2024-07-04
Add issue manager workflow

DoctorJohn:
## Description

This is the issue manager Erik proposed. I configured it with one of our existing labels "info-needed". Issues which have the "info-needed" label will automatically be closed with a message after two weeks of inactivity.

We can easily extend the delay (it's in seconds), I went with 14 days for nor particular reason. We could also configure this for more labels, but I guess this is the most useful one.

### DIFF
--- a/.github/workflows/issue-manager.yml
+++ b/.github/workflows/issue-manager.yml
@@ -24,7 +24,7 @@ jobs:
           config: >
             {
               "info-needed": {
-                "delay": 1209600,
+                "delay": P14D,
                 "message": "Hi, this issue requires extra info to be actionable. We're closing this issue because it has not been actionable for a while now. Feel free to provide the requested information and we'll happily open it again! 😊"
               }
             }
--- a/.github/workflows/issue-manager.yml
+++ b/.github/workflows/issue-manager.yml
@@ -1,0 +1,30 @@
+name: Issue Manager
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+  issue_comment:
+    types:
+      - created
+  issues:
+    types:
+      - labeled
+  pull_request_target:
+    types:
+      - labeled
+  workflow_dispatch:
+
+jobs:
+  issue-manager:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: tiangolo/issue-manager@0.5.0
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          config: >
+            {
+              "info-needed": {
+                "delay": 1209600,
+                "message": "Hi, this issue requires extra info to be actionable. We're closing this issue because it has not been actionable for a while now. Feel free to provide the requested information and we'll happily open it again! 😊"
+              }
+            }


.github/workflows/issue-manager.yml
```diff
@@ -0,0 +1,30 @@
...
+  issue-manager:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: tiangolo/issue-manager@0.5.0
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          config: >
+            {
+              "info-needed": {
+                "delay": 1209600,
```
bellini666, line 27: Might be worth adding a comment here specifying that this is 14 days =P

.github/workflows/issue-manager.yml
```diff
@@ -0,0 +1,30 @@
...
+                "message": "Hi, this issue requires extra info to be actionable. We're closing this issue because it has not been actionable for a while now. Feel free to provide the requested information and we'll happily open it again."
```
bellini666, line 28: Maybe add a 😊 at the end to sound more friendly? hahaha

bellini666 (approved):

.github/workflows/issue-manager.yml
DoctorJohn: Done 😊

.github/workflows/issue-manager.yml
DoctorJohn: Good point. Unfortunately the config is raw JSON and JSON does not really support comments. However, delay also accepts ISO 8601 according to the [issue manager docs](https://github.com/tiangolo/issue-manager/tree/0.5.0?tab=readme-ov-file#delay). So I used that instead. Now it reads `P14D` which means it's a period of 14 days 😊

.github/workflows/issue-manager.yml
```diff
@@ -0,0 +1,30 @@
...
+      - labeled
+  workflow_dispatch:
+
+jobs:
+  issue-manager:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: tiangolo/issue-manager@0.5.0
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
patrick91, line 23: Shall we use the bot token? :D

```suggestion
          token: ${{ secrets.BOT_TOKEN }}
```

erikwrede (approved):

.github/workflows/issue-manager.yml
erikwrede: github token is usually the token created specifically for this pipeline in this repository, so I believe it's safer to use that.

patrick91: what makes using another token less safe? 😊

bellini666: > what makes using another token less safe? 😊

I think `GITHUB_TOKEN` will only have permissions for stuff in the current pipeline, while `BOT_TOKEN` has a lot more permissions, which could give too much power to a 3rd party lib (the action itself) that we don't control.

Having said that, this comes from a very trusted source, so I'm not against using `BOT_TOKEN` if we prefer it :)

Issue #2464, opened 2023-01-17
DataLoader with a DefaultCache throws key error when attempting to remove key that does not exist


vjrasane:
Caused by:

```python
    def delete(self, key: K) -> None:
        del self.cache_map[self.cache_key_fn(key)]
```

https://github.com/strawberry-graphql/strawberry/blame/8498682d0c192e5b940f93928fef0f230918c8c7/strawberry/dataloader.py#L83

This can be worked around by providing a custom cache to the dataloader that checks existence before removal.

PR #3569, opened 2024-07-11
Fix default cache dataloader raise key error on non-existing key

Dartt0n:
## Description

Right now, calling the `.clean(key)` method on a non-cached key will raise a `KeyError` exception. There are several ways to solve this problem:
1. Create a custom `CustomDefaultCache` class and override the `delete` method to first check if the key exists.
2. Check if the key is cached before deleting it each time the `.clean` method is called.
3. Cache the key before calling `.clean`.
4. Set the default behavior to be exception-free.
Among these options, the fourth solution seems to be the best, so these changes are proposed.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #2464

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,12 @@
+Release type: patch
+
+Calling `.clean(key)` on default dataloader with non-existing `key` will not throw `KeyError` error anymore. Example:
+```python
+async def load_data(keys):
+    return [str(key) for key in keys]
+
+dataloader = DataLoader(load_fn=load_data)
+dataloader.clean(42) # does not throw KeyError anymore
+```
+
+This is a patch release, so no breaking changes.
--- a/tests/test_dataloaders.py
+++ b/tests/test_dataloaders.py
@@ -301,6 +301,8 @@ async def test_clear_nocache():
 
     loader = DataLoader(load_fn=idx, cache=False)
 
+    loader.clean(1) # no effect on non-cached values
+
     assert await loader.load_many([1, 2, 3]) == [(1, 1), (2, 1), (3, 1)]
 
     loader.clear(1)
--- a/strawberry/dataloader.py
+++ b/strawberry/dataloader.py
@@ -84,7 +84,7 @@ class DefaultCache(AbstractCache[K, T]):
         self.cache_map[self.cache_key_fn(key)] = value
 
     def delete(self, key: K) -> None:
-        del self.cache_map[self.cache_key_fn(key)]
+        self.cache_map.pop(self.cache_key_fn(key), default=None)
 
     def clear(self) -> None:
         self.cache_map.clear()


botberry:



---

Calling `.clean(key)` on default dataloader with non-existing `key` will not throw `KeyError` error anymore. Example:
```python
async def load_data(keys):
    return [str(key) for key in keys]


dataloader = DataLoader(load_fn=load_data)
dataloader.clean(42)  # does not throw KeyError anymore
```

This is a patch release, so no breaking changes.



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Anton Kudryavtsev for the PR 👏

```

botberry:



---

Calling `.clean(key)` on default dataloader with non-existing `key` will not throw `KeyError` error anymore. Example:
```python
async def load_data(keys):
    return [str(key) for key in keys]


dataloader = DataLoader(load_fn=load_data)
dataloader.clean(42)  # does not throw KeyError anymore
```

This is a patch release, so no breaking changes.



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Anton Kudryavtsev for the PR 👏

```

Dartt0n:
Hmm, seems like type linting fails to resolve:
```
strawberry/dataloader.py:87: error: No overload variant of "pop" of "dict" matches argument types "Hashable", "None"  [call-overload]
```

However, a overload exist:
```
strawberry/dataloader.py:87: note: Possible overload variants:
strawberry/dataloader.py:87: note:     def pop(self, Hashable, /) -> Future[T]
strawberry/dataloader.py:87: note:     def pop(self, Hashable, Future[T], /) -> Future[T]
strawberry/dataloader.py:87: note:     def [_T] pop(self, Hashable, _T, /) -> Future[T] | _T # this one
```
The type is `def [None] pop(self, Hashable, NoneType, /) -> Future[T] | None`

Will change code to
```python
if key in self.cache_map:
    del self.cache_map[key]
```
to satisfy linter

Dartt0n:
Seems thats all check passed, waiting for the review

patrick91:
@Dartt0n in what cases would this be useful? 😊

wouldn't an error be a useful indication that something was wrong somewhere else?

Dartt0n:
> @Dartt0n in what cases would this be useful? 😊
> 
> wouldn't an error be a useful indication that something was wrong somewhere else?

For example, some services modify an object, and this service needs to invalidate the cache for the modified object. However, the service does not know whether the object has already been loaded into the cache. This is a common problem for larger applications.

I think most developers who face this issue either use try-catch blocks or create their own wrapper (like original issue suggest). Therefore, I believe that making this behavior non-exceptional would be beneficial for the developer's user experience.

Additionally, as an example, the Redis DEL command does not throw an error when invalidating a non-existent cache entry.
https://redis.io/docs/latest/commands/del/ 

If you think that this explicit behavior would be more appropriate for the project, I suggest including a clear statement about it in the documentation in the "Cache Invalidation" section.
https://strawberry.rocks/docs/guides/dataloaders#cache-invalidation

garyd203:
> @Dartt0n in what cases would this be useful? 😊
> 
> wouldn't an error be a useful indication that something was wrong somewhere else?

Hey @patrick91 , I would find this useful too (I came here to do a similar PR but found this instead), for similar reasons to those described by @Dartt0n .

My reasoning is that cache invalidation is usually about ensuring an object doesn't exist in the cache. Strict remove-or-error semantics don't make sense because the nature of a data-loader is that the cache entries could be added from anywhere, and with arbitrary concurrency.

For example, my current use-case is that an internal helper (`set_site_users()`) modifies the list of objects attached to a user via foreign key. The list of objects is loaded for resolvers via a data-loader (`load_users_at_site()`), and hence is cached. After modifying the list of objects in `set_site_users()`, we should invalidate the cache for the `load_users_at_site()` data-loader for that key - but depending which mutation calls that helper (and the input to that mutation), the data-loader may or may not have been called already, and hence `.clear()` may succeed or may error. There's no benefit in having the `set_site_users()` helper track whether the `load_users_at_site()` data-loader has been called and then handle the error specially (because both scenarios are legitimate), we just want to unilaterally clear that key from the data-loader cache.

RELEASE.md
```diff
@@ -0,0 +1,16 @@
+Release type: patch
+
+Calling `.clean(key)` on default dataloader with non-existing `key` will not throw `KeyError` anymore. Example:
```
garyd203, line 3:
```suggestion
Calling `.clear(key)` on default dataloader with non-existing `key` will not throw `KeyError` anymore. Example:
```

RELEASE.md
```diff
@@ -0,0 +1,16 @@
...
+```python
+from strawberry.dataloader import DataLoader
+
+
+async def load_data(keys):
+    return [str(key) for key in keys]
+
+
+dataloader = DataLoader(load_fn=load_data)
+dataloader.clean(42)  # does not throw KeyError anymore
```
garyd203, line 13:
```suggestion
dataloader.clear(42)  # does not throw KeyError anymore
```

tests/test_dataloaders.py
```diff
@@ -301,6 +301,11 @@ async def idx(keys: List[int]) -> List[Tuple[int, int]]:
 
     loader = DataLoader(load_fn=idx, cache=False)
 
+    try:
+        loader.clean(1)  # no effect on non-cached values
+    except Exception as e:
+        raise AssertionError("clean non-cached values does not raise KeyError")
```
garyd203, line 307: I'm not sure what your intention is here. Normally a test will implement desired behaviour, and if it works without errors then (that part of) the test passes. So we could simplify to this ...

```suggestion
    loader.clear(1)  # no effect on non-cached values
```

...which is similar to the other calls below to `loader.clear()` / `loader.clear_many()` / etc

garyd203 (commented):
Thanks for doing this.

I believe you made a typo, the function is called `.clear()` rather than `.clean()`

tests/test_dataloaders.py
garyd203: You should also add this to the `test_clear()` test function above, so that we can exercise the `.clear()` functionality with an underlying cache

See https://github.com/strawberry-graphql/strawberry/pull/3569/files#diff-9e2920b06c9fb1a3a76b9f675328e820421756fa44783ebdd45ca2c7a047c8a7R266-R279

Dartt0n:
Hi @garyd203! Thank you for bringing this up! The pull request remained incomplete because I eventually implemented my class, which inherits from AbstractCacheMap and used it for the project (failed to implement Redis cache due to other issues related to strawberry dataloaders, so ended up storing everything in app memory :( ). I have updated the code based on your comments, if you have free time, please review

garyd203 (approved):
LGTM, thanks @Dartt0n 

@patrick91 Are you happy with this change?

PR #3576, opened 2024-07-21
fix: Fix optional lazy types resolution when using future annotations

### LINKED ISSUE #3572: TypeError: MyCustomType fields cannot be resolved. unsupported operand type(s) for |: 'LazyType' and 'NoneType'
## Describe the Bug
I had an error generating types with the next values and the error thrown was 

> TypeError: MyCustomType fields cannot be resolved. unsupported operand type(s) for |: 'LazyType' and 'NoneType'

This was the status the packages and the error happened on the change from:
```
  Django= 4.2.11
  strawberry-graphql = "^0.235.2"
  strawberry-graphql-django = "^0.44.2"
  python= "3.11.9"
```

To

```
  Django= 4.2.11
  strawberry-graphql = "^0.235.2"
  strawberry-graphql-django = "^0.45.0"
  python= "3.11.9"
```

The code was like this when the error raised:
```
class MyCustomType:
    my_custom_field: Annotated[AType, strawberry.lazy("xx.graphql.types")] | None

class AnotherMyCustomType:
    my_another_custom_field: AType | None = None

```

Changed to this to get fixed:

```
class MyCustomType:
    my_custom_field: AType | None = None

class AnotherMyCustomType:
    my_another_custom_field: AType | None = None
```

I did not change it both to be lazy that would be the other solution

The error at the end was how the types were generated because I had in the same file a type that was declared as lazy and also as not lazy, I did not have the error before and when I upgraded to those values the error came up.

I fixed the error changing this to not lazy but I mention here @bellini666 that suggested me to fill up a bug in case is related to how the types are generated.

I hope this helps

Thanks!

## System Information

 - Operating system: Apple M1 Pro Sonoma 14.4.1 -> But the system is on Docker python:3.11.9-bullseye (debian)
 - Strawberry version (if applicable): "0.235.2"

## Additional Context
[error_traceback.txt](https://github.com/user-attachments/files/16231460/error_traceback.txt)
[discord thread](https://discord.com/channels/689806334337482765/689861980776955948/1261325861597089802)

## Upvote & Fund

<a href="https://polar.sh/strawberry-graphql/strawberry/issues/3572">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/3572/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

bellini666:
When using future annotations, our `eval_type` converts `Annotated[foo, strawberry.lazy("xxx.foo")]`
into `LazyType` already, as can be seen in the following places:

- https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/utils/typing.py#L369
- https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/utils/typing.py#L308

Because of that, a code that works without future annotations like `Annotated[foo, strawberry.lazy("xxx.foo")] | None`
will be converted to `LazyType | None`, and would fail because `LazyType` doesn't implement `__or__`.

This is adding `__or__` to it, allowing the code to work as expected.

Fix #3572

## Summary by Sourcery

This pull request fixes the resolution of optional lazy types using the `| None` operator when future annotations are enabled. It introduces the `__or__` method to `LazyType` to support this functionality and includes new tests to ensure the fix works as expected. Additionally, a release note has been added to document this patch.

- **Bug Fixes**:
    - Fixed an issue where optional lazy types using `| None` were failing to resolve correctly when using future annotations by adding `__or__` method to `LazyType`.
- **Documentation**:
    - Added a release note documenting the patch fix for optional lazy types using `| None` with future annotations.
- **Tests**:
    - Added tests to verify the correct resolution of optional lazy types using the `| None` operator with and without future annotations.

### DIFF
--- a/strawberry/types/lazy_type.py
+++ b/strawberry/types/lazy_type.py
@@ -5,7 +5,6 @@ import warnings
 from dataclasses import dataclass
 from pathlib import Path
 from typing import (
-    TYPE_CHECKING,
     Any,
     ForwardRef,
     Generic,
@@ -13,14 +12,14 @@ from typing import (
     Tuple,
     Type,
     TypeVar,
+    Union,
     cast,
 )
-
-if TYPE_CHECKING:
-    from typing_extensions import Self
+from typing_extensions import Self
 
 TypeName = TypeVar("TypeName")
 Module = TypeVar("Module")
+Other = TypeVar("Other")
 
 
 @dataclass(frozen=True)
@@ -58,6 +57,9 @@ class LazyType(Generic[TypeName, Module]):
             package = current_frame.f_back.f_globals["__package__"]
 
         return cls(type_name, module, package)
+
+    def __or__(self, other: Other) -> Union[Self, Other]:
+        return Union[self, other]
 
     def resolve_type(self) -> Type[Any]:
         module = importlib.import_module(self.module, self.package)
--- a/tests/types/test_lazy_types.py
+++ b/tests/types/test_lazy_types.py
@@ -1,7 +1,11 @@
 # type: ignore
 import enum
+import sys
+import textwrap
 from typing import Generic, TypeVar
 from typing_extensions import Annotated, TypeAlias
+
+import pytest
 
 import strawberry
 from strawberry.annotation import StrawberryAnnotation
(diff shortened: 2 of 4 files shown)


tests/types/test_lazy_types.py
```diff
@@ -195,3 +196,52 @@ def test_lazy_function_in_union():
     [type1, type2] = resolved.types
     assert type1.resolve_type() is LaziestType
     assert type2.resolve_type() is LazyEnum
+
+
+def test_optional_lazy_type_using_or_operator():
```
bellini666, line 201: This was already working fine because this file doesn't import future annotations

botberry:



---

This release fixes an issue where optional lazy types using `| None` were
failing to be correctly resolved inside modules using future annotations. e.g.

```python

from typing import Annotated, TYPE_CHECKING

import strawberry

if TYPE_CHECKING:
    from types import Group


@strawberry.type
class Person:
    group: Annotated["Group", strawberry.lazy("types.group")] | None
```

This should now work as expected.



Here's the tweet text: 
```

```

tests/types/test_lazy_types_future_annotations.py
```diff
@@ -0,0 +1,64 @@
+from __future__ import annotations
+
+import textwrap
+from typing_extensions import Annotated
+
+import strawberry
+
+
+def test_optional_lazy_type_using_or_operator():
```
bellini666, line 9: this is the one that breaks without the `__or__`

tests/types/test_lazy_types_future_annotations.py
```diff
@@ -0,0 +1,64 @@
...
+          typeCList: [TypeC!]!
+        }
+
+        type TypeC {
+          name: String!
+        }
+        """
+        assert str(schema).strip() == textwrap.dedent(expected).strip()
+    finally:
+        del SomeType, AnotherType
```
DoctorJohn, line 64: Just out of curiosity: why do these types have to be globals and need to be cleaned up like this at the end?

strawberry/types/lazy_type.py
```diff
@@ -59,6 +58,9 @@ def __class_getitem__(cls, params: Tuple[str, str]) -> "Self":
 
         return cls(type_name, module, package)
 
+    def __or__(self, other: Other) -> Union[Self, Other]:
+        return Union[self, other]
```
DoctorJohn, line 62: Learned something new today from these two lines. I was wondering how they behave at runtime, but looks like they just do what we expect them to do.

DoctorJohn (approved):
Looks good to me: well tested and pretty straight forward fix. Awesome @bellini666

tests/types/test_lazy_types_future_annotations.py
bellini666: They need to be global because the lazy importer will try to import it from the module, and in here we are defining those inside the function.

Another option would be to define the types themselves at the module level, but then it becomes difficult to change them for other tests :(

The cleanup at the end is just to make sure that the global namespace doesn't get polluted and possibly affects other tests

DoctorJohn: Thanks for the explanation Thiago, this makes a lot of sense.

patrick91 (approved):
Nice! 🤗

PR #3592, opened 2024-08-07
Add ability to set a custom info class for a schema

parafoxia:
## Description

This PR adds a option to `StrawberryConfig` that allows the user to provide a subclass of `strawberry.Info` to use in types and queries instead of the default.

An example can be found in the RELEASE.md.

## Rationale

Being able to extend the `Info` class allows for the introduction of special utilities that use information within the instance.

These are too specialised to be contributed to Strawberry generally, but custom `Info` support would allow these use-cases, and others like them, to be used where individuals and organisations see fit.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

* None

## Checklist

## Summary by Sourcery

Introduce the ability to set a custom info class in `StrawberryConfig`, allowing users to provide a subclass of `strawberry.Info` for use in types and queries. Update documentation with an example of this new feature.

New Features:
- Add option to `StrawberryConfig` to allow users to provide a custom subclass of `strawberry.Info` for use in types and queries.

Documentation:
- Add example usage of the new `info_class` option in `RELEASE.md`.


botberry:



---

You can now configure your schemas to provide a custom subclass of
`strawberry.types.Info` to your types and queries.

```py
import strawberry

from .models import ProductModel


class CustomInfo(strawberry.Info):
    def is_selected(field: str) -> bool:
        """Check if the field is selected on the top-level of the query."""
        return field in [sel.name for sel in info.selected_fields]


@strawberry.type
class Product:
    id: strawberry.ID
    orders: list[Order]


@strawberry.type
class Query:
    @strawberry.field
    def product(self, id: strawberry.ID, info: CustomInfo) -> Product:
        kwargs = {"id": id}

        if info.is_selected("orders"):
            # Do some expensive operation here that we wouldn't want to
            # do if the field wasn't selected.
            kwargs["orders"] = ...

        return Product(**kwargs)


schema = strawberry.Schema(
    Query,
    config=StrawberryConfig(info_class=CustomInfo),
)
```



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Ethan Henderson for the PR 👏

```

RELEASE.md
```diff
@@ -0,0 +1,43 @@
...
+```py
+import strawberry
+from strawberry.schema.config import StrawberryConfig
+
+from .models import ProductModel
+
+
+class CustomInfo(strawberry.Info):
+    def is_selected(field: str) -> bool:
+        """Check if the field is selected on the top-level of the query."""
```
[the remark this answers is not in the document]
parafoxia: Is already is...?

strawberry/schema/config.py
```diff
@@ -13,6 +15,7 @@ class StrawberryConfig:
     default_resolver: Callable[[Any, str], object] = getattr
     relay_max_results: int = 100
     disable_field_suggestions: bool = False
+    info_class: type[Info] = field(default_factory=lambda: Info)
```
parafoxia, line 18: Looking at this again, the default can just be the class -- the mutability trap doesn't affect this.

parafoxia:
I've just rebased this on master so it should now pass the Pyright test stuff.

parafoxia:
@patrick91 Would it be possible for you to look at this today?

RELEASE.md
```diff
@@ -0,0 +1,47 @@
...
+import strawberry
+from strawberry.schema.config import StrawberryConfig
+
+from .models import ProductModel
+
+
+class CustomInfo(strawberry.Info):
+    def is_selected(self, field: str) -> bool:
+        """Check if the field is selected on the top-level of the query."""
+        return field in [sel.name for sel in info.selected_fields]
```
patrick91, line 16: I feel like this should in Strawberry :D 

Also I'm worried people will copy this without knowing how it filly works, could we make a different example? 😊

strawberry/schema/config.py
```diff
@@ -21,5 +24,8 @@ def __post_init__(
         if auto_camel_case is not None:
             self.name_converter.auto_camel_case = auto_camel_case
 
+        if not issubclass(self.info_class, Info):
+            raise ValueError("`info_class` must be a subclass of strawberry.types.Info")
```
patrick91, line 28: can we add a `# pragma: no cover`? here? 😊

patrick91 (commented):
Looks solid, I left a probably annoying comment on the release, but other than that I think it's ok to merge :)

Ah, maybe we want to include this in the docs too :)

RELEASE.md
parafoxia: I have considered it but the implementation we currently have (which is recursive) needs refactoring. I have since got a slightly different utility for a slightly different use-case which would probably be better here anyways, so I'll use that.

If I ever feel the `is_selected` is in a state to contribute in, I definitely will! 😄 Just a case of figuring out how best to make it work in as many cases as possible (and appropriate).

strawberry/schema/config.py
parafoxia: Sure!

parafoxia:
> Looks solid, I left a probably annoying comment on the release, but other than that I think it's ok to merge :)
> 
> Ah, maybe we want to include this in the docs too :)

Happy to put this in the docs, I did not realise the config system had docs 😅

RELEASE.md
parafoxia: So I've got a completely different example from that (currently only in local, will push it all once I've done the docs), hopefully it's a better example. All our use cases for this are quite complex so it's tricky to think of something simple 😅

docs/types/schema-configurations.mdx
```diff
@@ -105,3 +105,23 @@ schema = strawberry.Schema(
     query=Query, config=StrawberryConfig(disable_field_suggestions=True)
 )
 ```
+
+### info_class
+
+By default Strawberry will create an object of type `strawberry.types.Info` when
+the user defines `info: Info` as a parameter to a type or query. You can change
+this behaviour by setting `info_class` to a subclass of `strawberry.types.Info`.
```
DoctorJohn, line 113: While `strawberry.type.info` is technically correct, IMHO naming `strawberry.Info` in the docs whill encourage users to subclass the top-level exported `Info` type just like in your example in the release file.

```suggestion
By default Strawberry will create an object of type `strawberry.Info` when
the user defines `info: Info` as a parameter to a type or query. You can change
this behaviour by setting `info_class` to a subclass of `strawberry.Info`.
```

strawberry/schema/config.py
```diff
@@ -21,5 +24,8 @@ def __post_init__(
         if auto_camel_case is not None:
             self.name_converter.auto_camel_case = auto_camel_case
 
+        if not issubclass(self.info_class, Info):  # pragma: no cover
+            raise TypeError("`info_class` must be a subclass of strawberry.types.Info")
+
```
DoctorJohn, line 29: Adding a test for this wouldn't hurt :) Looks like we don't have a fitting test file for it yet (I'd suggest creating `tests/schema/test_config.py`).

docs/types/schema-configurations.mdx
parafoxia: Yeah, sounds reasonable, will commit the suggestion.

strawberry/schema/config.py
parafoxia: Cool cool, will tackle this tomorrow.

DoctorJohn (approved):
Excellent @parafoxia, thanks for the changes. This looks good to me now. Let's have @patrick91 take a final look at the example in the RELEASE file. IMHO just showing how to use a custom Info class would have been enough there, but I'm also not against the current example, especially given the extra afford you already put into it.

parafoxia:
Just rebased onto the changes in the doc system.

parafoxia:
> Excellent @parafoxia, thanks for the changes. This looks good to me now. Let's have @patrick91 take a final look at the example in the RELEASE file. IMHO just showing how to use a custom Info class would have been enough there, but I'm also not against the current example, especially given the extra afford you already put into it.

Thanks! I wasn't sure how full an example it needed to be lmao, so I just went hog on it 😆

parafoxia:
@patrick91 Just bumping this 😄

parafoxia:
The one problem I'm running into is that custom info objects don't appear to be loading into extensions (in the case I'm testing, permission extensions, though I've seen it in field extensions as well). That's what that previous commit was trying to resolve, but it seemingly hasn't worked.

parafoxia:
Just updated this to custom info classes get injected into extensions as well.

DoctorJohn:
> Just updated this to custom info classes get injected into extensions as well.

Oh, great catch! It would be great to have a test case checking that the custom Info class is used in directives and extensions. What do you think? :)

parafoxia:
> > Just updated this to custom info classes get injected into extensions as well.
> 
> Oh, great catch! It would be great to have a test case checking that the custom Info class is used in directives and extensions. What do you think? :)

Done!

parafoxia:
Hi @patrick91, just bumping this for review again.

patrick91 (approved):
Thank you! And sorry for the late review, this looks great!

botberry:

PR #3612, opened 2024-09-02
Improve support for multipart subscriptions

patrick91:
This should allow for SSE to be used in the future.


patrick91:
Handled in #3554

PR #3646, opened 2024-09-25
fix: show 405 error if request is GET and queries are not allowed

alexei:
## Description

When `allow_queries_via_get` is `False`, `GET` requests are processed despite the fact that queries are disallowed. Currently the response is "400: No GraphQL query found in the request" which is raised at a deeper level. I believe such requests should be rejected immediately as they needlessly consume resources.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

## Summary by Sourcery

Fix the handling of GET requests by returning a 404 error when queries are disallowed, instead of processing them and returning a 400 error at a deeper level.

Bug Fixes:
- Return a 404 error for GET requests when queries are not allowed, preventing unnecessary resource consumption.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+This release fixes an issue where a GET request is processed despite it being disallowed.
--- a/strawberry/http/async_base_view.py
+++ b/strawberry/http/async_base_view.py
@@ -184,6 +184,9 @@ class AsyncBaseHTTPView(
             else:
                 raise HTTPException(404, "Not Found")
 
+        if request_adapter.method == "GET" and not self.allow_queries_via_get:
+            raise HTTPException(404, "Not Found")
+
         sub_response = await self.get_sub_response(request)
         context = (
             await self.get_context(request, response=sub_response)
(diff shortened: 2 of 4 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Alexandru Mărășteanu for the PR 👏

```

RELEASE.md
```diff
@@ -0,0 +1,3 @@
+Release type: patch
+
+This release fixes an issue where a GET request is processed despite it being disallowed.
```
erikwrede, line 3: Let's be more specific here
```suggestion
This release changes the default behavior on queries via GET requests to a 405 error. To enable queries via get, please use `self.allow_queries_via_get`.
```

alexei: Hey @erikwrede I appreciate your reviewing this. I agree the message can be improved, but let's be clear about the change. Queries via GET are still allowed by default just like before - I did not touch that, though I believe the default should be more conservative. I changed the behavior when it's turned off: 

- before this change you get a GraphQL query not found error;
- after this change you get a HTTP method not allowed error.

bellini666: @alexei so probably we can go with an explanation like that, saying that before, even though queries via GET are allowed by default, when disallowing them you would only get a GraphQL query not found error, but now you should get a 405 error instead.

PR #3652, opened 2024-09-28
Take oneOf directive into account in codegen module

enoua5:
## Description

Adds support for `oneOf` to the `codegen` module

- The codegen manager now passes `is_one_of` to `GraphQLObjectType` making it part of the codegen plugin api
- Python plugin has been updated to output oneOf types as a union of classes with one field each
- Typescript plugin has been updated to output oneOf types as a union of objects where in each all but one field has a type of `never`

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* None

## Checklist

## Summary by Sourcery

Implement support for the `oneOf` directive in the code generation module, enhancing the Python and TypeScript plugins to handle `oneOf` types as unions. Add corresponding tests and snapshots to verify the new functionality.

New Features:
- Add support for the `oneOf` directive in the code generation module, allowing the generation of union types for `oneOf` fields in both Python and TypeScript plugins.

Enhancements:
- Update the Python plugin to output `oneOf` types as a union of classes, each with one field.
- Update the TypeScript plugin to output `oneOf` types as a union of objects, where each object has all but one field typed as `never`.

Tests:
- Add new test cases and snapshots for the `oneOf` directive in both Python and TypeScript code generation to ensure correct functionality.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,53 @@
+Release type: minor
+
+The Query Codegen system now supports the `@oneOf` directive.
+
+When writing plugins, you can now access `GraphQLObjectType.is_one_of` to determine if the object being worked with has a `@oneOf` directive.
+
+The default plugins have been updated to take advantage of the new attribute.
+
+For example, given this schema:
+
+```python
+@strawberry.input(one_of=True)
+class OneOfInput:
+    a: Optional[str] = strawberry.UNSET
+    b: Optional[str] = strawberry.UNSET
+
+
+@strawberry.type
+class Query:
+    @strawberry.field
+    def one_of(self, value: OneOfInput) -> str: ...
+
+
+schema = strawberry.Schema(Query)
+```
+
+And this query:
+
+```graphql
+query OneOfTest($value: OneOfInput!) {
+  oneOf(value: $value)
+}
+```
+
+The query codegen can now generate this Typescript file:
+
+```typescript
+type OneOfTestResult = {
+    one_of: string
+}
+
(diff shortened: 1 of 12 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Jacob Allen for the PR 👏

```

enoua5:
I unchecked "My change requires a change to the documentation". I would add a note about this to the documentation on codegen plugins, but the docs already don't go over any of the other types that are passed into the plugin.

enoua5:
If anyone has opinions on how those Typescript unions should be formatted, I'd be happy to take the feedback :smile:

patrick91:
@enoua5 I'll check this on the weekend! thank you so much 😊

enoua5:
> If anyone has opinions on how those Typescript unions should be formatted, I'd be happy to take the feedback 😄

Looking at [`dotansimha/graphql-code-generator`](https://github.com/dotansimha/graphql-code-generator), it looks like it would generate code like this:

```typescript
type OneOfInput =
  { a: string, b?: never }
  |   { a?: never, b: string };
```

I'm not a big fan of the hanging `=`, but, thoughts?

patrick91:
@enoua5 I think that's fine, we could maybe in future run prettier/biome to the output if the user has one of them installed :)

enoua5:
@patrick91 That's a fair point. Do you think its not worth changing it from how it is now, then?

Issue #3299, opened 2023-12-18
Maximum recursion depth exceeded when running codegen


emosenkis:
When trying to generate Python client code using `strawberry codegen`, I get `RecursionError: maximum recursion depth exceeded`:
```
│ .pyenv/versions/venv/lib/python3.10/site-packages/strawberry/codegen/query_codegen.py:657 in     │
│ <lambda>                                                                                         │
│                                                                                                  │
│   654 │   │   │   wrapper = (                                                                    │
│   655 │   │   │   │   GraphQLOptional                                                            │
│   656 │   │   │   │   if wrapper is None                                                         │
│ ❱ 657 │   │   │   │   else lambda t: GraphQLOptional(wrapper(t))  # type: ignore[misc]           │
│   658 │   │   │   )                                                                              │
│   659 │   │                                                                                      │
│   660 │   │   elif isinstance(type_, StrawberryList):                                            │
│                                                                                                  │
│ ╭────────────────────────────────────── locals ──────────────────────────────────────╮           │
│ │       t = GraphQLObjectType(                                                       │           │
│ │           │   name='UpdateRecipientResultUpdateRecipientErrors',                   │           │
│ │           │   fields=[                                                             │           │
│ │           │   │   GraphQLField(                                                    │           │
│ │           │   │   │   name='message',                                              │           │
│ │           │   │   │   alias=None,                                                  │           │
│ │           │   │   │   type=GraphQLScalar(name='String', python_type=None),         │           │
│ │           │   │   │   default_value=None                                           │           │
│ │           │   │   )                                                                │           │
│ │           │   ],                                                                   │           │
│ │           │   graphql_typename=None                                                │           │
│ │           )                                                                        │           │
│ │ wrapper = <function QueryCodegen._unwrap_type.<locals>.<lambda> at 0x7fba74bfad40> │           │
│ ╰────────────────────────────────────────────────────────────────────────────────────╯           │
```

Using Python 3.10 and strawberry 0.217.0

jacobmoshipco:
I am also having this issue on the same line. The `wrapper` function is getting rebound to call itself. I'm having the issue with a nullable list, but I'm unsure if that's the root cause, because there's already a test for it. I'm drafting a PR to fix this, but I can't figure out a test case that the current implementation is failing. Do you still have the code that was causing the issue?

jacobmoshipco:
[...]

PR #3653, opened 2024-09-30
Fix codegen crash on nullable lists of non-scalars

enoua5:
## Description

Codegen would previously crash when working with a nullable list of non-scalars. This was due to a lambda referencing a value that would get rewritten with that same lambda causing an endless recursion.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #3299

## Checklist

- [ ] ~~My change requires a change to the documentation.~~
- [ ] ~~I have updated the documentation accordingly.~~

## Summary by Sourcery

Fix the code generation crash when processing nullable lists of non-scalar types by adjusting the lambda function handling and add corresponding tests.

Bug Fixes:
- Fix a crash in the code generation tool when handling nullable lists of non-scalar types by correcting the lambda function handling.

Tests:
- Add tests to cover the scenario of nullable lists of non-scalar types to ensure the code generation tool handles them correctly.

Chores:
- Add a release note indicating a patch release to fix the codegen crash issue.

### DIFF
--- a/strawberry/codegen/query_codegen.py
+++ b/strawberry/codegen/query_codegen.py
@@ -645,7 +645,7 @@ class QueryCodegen:
     ) -> Tuple[
         Union[type, StrawberryType], Optional[Callable[[GraphQLType], GraphQLType]]
     ]:
-        wrapper = None
+        wrapper: Optional[Callable[[GraphQLType], GraphQLType]] = None
 
         if isinstance(type_, StrawberryOptional):
             type_, previous_wrapper = self._unwrap_type(type_.of_type)
--- a/strawberry/codegen/query_codegen.py
+++ b/strawberry/codegen/query_codegen.py
@@ -648,17 +648,19 @@ class QueryCodegen:
         wrapper = None
 
         if isinstance(type_, StrawberryOptional):
-            type_, wrapper = self._unwrap_type(type_.of_type)
+            type_, previous_wrapper = self._unwrap_type(type_.of_type)
             wrapper = (
                 GraphQLOptional
-                if wrapper is None
-                else lambda t: GraphQLOptional(wrapper(t))  # type: ignore[misc]
+                if previous_wrapper is None
+                else lambda t: GraphQLOptional(previous_wrapper(t))  # type: ignore[misc]
             )
 
         elif isinstance(type_, StrawberryList):
-            type_, wrapper = self._unwrap_type(type_.of_type)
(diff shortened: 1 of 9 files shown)


botberry:



---

Fixes an issue where the codegen tool would crash when working with a nullable list of types.



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Jacob Allen for the PR 👏

```

strawberry/codegen/query_codegen.py
```diff
@@ -648,17 +648,19 @@ def _unwrap_type(
...
             )
 
         elif isinstance(type_, StrawberryList):
-            type_, wrapper = self._unwrap_type(type_.of_type)
+            type_, previous_wrapper = self._unwrap_type(type_.of_type)
             wrapper = (
-                GraphQLList if wrapper is None else lambda t: GraphQLList(wrapper(t))
+                GraphQLList
+                if previous_wrapper is None
+                else lambda t: GraphQLList(previous_wrapper(t))
```
jacobmoshipco, line 663: I'm seeing a MyPy error on the second `wrapper = ...` assignment here. I'm guessing that's why the first one has the `#type: ignore[misc]` annotations? If so, would it be safe to add those annotations to the second one?

(ps, I'm the same person as the author, just on a different account)

tests/codegen/queries/optional_list_of_people.graphql
```diff
@@ -0,0 +1,6 @@
+query OptionalListOfPeople {
```
DoctorJohn, line 1: IMHO, it would be beneficial to use an operation name here that differs from the name of the queried field to make it easier to verify the generated code is correct. We could go with just `OperationName`.

DoctorJohn, line 1: Let's also rename this file (and the corresponding snapshot files) to better describe the test case. All other files in this folder are named after what they are testing in GraphQL terms. From the current name, it might not be clear why we need to test an "optional list of people." The file could be named `nullable_list_of_non_scalars.graphql` instead, matching your PR's description.

DoctorJohn (changes requested):
Hi @enoua5, thanks for the PR. I had to read the corresponding issue to understand the changes and what was going on. Thanks for fixing the recursion. I left two minor suggestions regarding the test's readability/verifiability.

DoctorJohn (approved):
Awesome, thanks again @enoua5. This looks good to me now

patrick91 (approved):

PR #3661, opened 2024-10-07
Rename misleading test client argument name

### LINKED ISSUE #3580: TestClient `asserts_errors` flag behaviour is counter intuitive
## Feature Request Type

- [ ] Core functionality
- [ ] New behavior

## Description

In the `BaseGraphQLTestClient` (`/test/client.py`), there is an option `asserts_errors` which is by default `True`, and asserts that `response.errors is None`.

I'm finding this to be counterintuitive for two reasons:

1. It's not the job of a query client to decide what should be in the response (especially by default), this is the job of my test code ("separation of concerns").
2. It's named incorrectly. "Asserts Errors" implies it asserts that there ARE errors, not that there AREN'T.

I think this check should be removed. If it is kept, the name should be made explicit (`assert_no_errors`) and ideally made opt-in rather than opt-out. 

### Special concern

Is it typical for cases to arise in which errors would be returned *alongside* an otherwise valid response? Because if that's the case then I kind of understand why it's opt-out by default. Otherwise, testing for valid data response should be sufficient, AFAICT?

## Contributing

Probably much quicker for a regular maintainer but I'm happy to make a PR for this if accepted!

## Upvote & Fund

<a href="https://polar.sh/strawberry-graphql/strawberry/issues/3580">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/strawberry-graphql/strawberry/issues/3580/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

DoctorJohn:
## Description

This PR renames the `asserts_errors` of our user-facing test clients to `assert_no_errors` as proposed in #3580. The old option is still available but will raise a deprecation warning.

Also, before this PR, only the Django test client had tests. Now, all test clients share the tests. (That's why this PR is a little bigger than one might expect).

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* Fixes #3580

## Summary by Sourcery

Rename the `asserts_errors` argument to `assert_no_errors` in test clients for clarity, add shared tests for all clients, and document the change with a release note.

Enhancements:
- Rename the `asserts_errors` argument to `assert_no_errors` in test clients to better reflect its purpose, while maintaining backward compatibility with a deprecation warning for the old argument.

Documentation:
- Add a release note documenting the renaming of the `asserts_errors` option to `assert_no_errors` in test clients, highlighting the backward compatibility and deprecation warning.

Tests:
- Add shared tests for all test clients, including AIHOTTP, ASGI, and Django, to ensure consistent behavior across different environments.
- Introduce tests to verify the deprecation warning for the `asserts_errors` argument and the functionality of the `assert_no_errors` option.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,4 @@
+Release type: minor
+
+The AIOHTTP, ASGI, and Django test clients' `asserts_errors` option has been renamed to `assert_no_errors` to better reflect its purpose.
+This change is backwards-compatible, but the old option name will raise a deprecation warning.
--- a/strawberry/aiohttp/test/client.py
+++ b/strawberry/aiohttp/test/client.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import warnings
 from typing import (
     Any,
     Dict,
@@ -16,8 +17,9 @@ class GraphQLTestClient(BaseGraphQLTestClient):
         query: str,
         variables: Optional[Dict[str, Mapping]] = None,
         headers: Optional[Dict[str, object]] = None,
-        asserts_errors: Optional[bool] = True,
+        asserts_errors: Optional[bool] = None,
         files: Optional[Dict[str, object]] = None,
+        assert_no_errors: Optional[bool] = True,
     ) -> Response:
         body = self._build_body(query, variables, files)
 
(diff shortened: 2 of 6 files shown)


botberry:



---

The AIHOTTP, ASGI, and Django test clients' `asserts_errors` option has been renamed to `assert_no_errors` to better reflect its purpose.
This change is backwards-compatible, but the old option name will raise a deprecation warning.



Here's the tweet text: 
```

```

patrick91 (approved):
Nice, thanks! 

Maybe we should consolidate the test clients in future (we have private ones in the tests folder :D)

DoctorJohn:
> Maybe we should consolidate the test clients in the future (we have private ones in the tests folder :D)

Yeah, I think that's a great idea. I didn't even know we had a public test client because it doesn't appear to be documented and did not really have tests. Our internal test client is much more capable and could be a good basis for a unified test client.

PR #3639, opened 2024-09-23
Add support for using raw Python enum types in schema

nrbnlulu:
## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement/optimization
- [ ] Documentation

## Issues Fixed or Closed by This PR

fix #3543 

## Checklist

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the CONTRIBUTING document.
- [ ] I have added tests to cover my changes.
- [ ] I have tested the changes and verified that they work and don't break anything (as well as I can manage).

## Summary by Sourcery

Add support for using raw Python enum types in schemas, enabling the use of enums without `@strawberry.enum` decoration. Fix issue #3543 and update tests to cover the new functionality. Document the changes in RELEASE.md.

New Features:
- Add support for using raw Python enum types in schemas, allowing enums that are not decorated with `@strawberry.enum` to be used.

Bug Fixes:
- Fix issue #3543 by allowing the use of non-decorated enum types in schemas.

Documentation:
- Add a new section in RELEASE.md documenting the support for using raw Python enum types in schemas.

Tests:
- Add tests to verify the default implementation of enum, str enum, and int enum types in schemas.

PR #3665, opened 2024-10-08
Bump 3.13 to non dev

patrick91:
Doing this PR to also check why tests in #3639 are broken


bellini666:
I think we can't run Django 4.0 and 3.2 tests with 3.13 as they use the `cgi` module, which was deprecated and got removed.

PR #3670, opened 2024-10-13
Fix graphql ws did not ignore parsing errors

DoctorJohn:
## Description

The legacy WS protocol dictates that servers should ignore text messages sent by clients that are not valid JSON. Here's the [relevant section](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md#gql_connection_error) from the legacy protocol specification:

> GQL_CONNECTION_ERROR
> 
> [...]
> 
> It server also respond with this message in case of a parsing errors of the message (which does not disconnect the client, just ignore the message).

I accidentally changed this behavior in a recent refactor to match the newer protocol, which requires servers to close the WebSocket in this case.

This PR restores the intended behavior and updates the relevant tests.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Summary by Sourcery

Fix the handling of non-JSON messages in the legacy GraphQL WebSocket protocol to ignore parsing errors instead of closing the connection, and update tests to reflect this behavior.

Bug Fixes:
- Restore the behavior of ignoring non-JSON messages in the legacy GraphQL WebSocket protocol, aligning with the legacy protocol specification.

Documentation:
- Add a RELEASE.md file documenting the fix for the regression in the legacy GraphQL WebSocket protocol.

Tests:
- Update tests to ensure non-JSON WebSocket messages are ignored rather than causing a connection closure in the legacy protocol.

### DIFF
--- a/strawberry/aiohttp/views.py
+++ b/strawberry/aiohttp/views.py
@@ -91,7 +91,7 @@ class AioHTTPWebSocketAdapter(AsyncWebSocketAdapter):
         self.ws = ws
 
     async def iter_json(
-        self, ignore_parsing_errors: bool
+        self, *, ignore_parsing_errors: bool = False
     ) -> AsyncGenerator[Dict[str, object], None]:
         async for ws_message in self.ws:
             if ws_message.type == http.WSMsgType.TEXT:
--- a/strawberry/aiohttp/views.py
+++ b/strawberry/aiohttp/views.py
@@ -26,7 +26,11 @@ from strawberry.http.async_base_view import (
     AsyncHTTPRequestAdapter,
     AsyncWebSocketAdapter,
 )
-from strawberry.http.exceptions import HTTPException, NonJsonMessageReceived
+from strawberry.http.exceptions import (
+    HTTPException,
+    NonJsonMessageReceived,
+    NonTextMessageReceived,
+)
 from strawberry.http.types import FormData, HTTPMethod, QueryParams
 from strawberry.http.typevars import (
     Context,
@@ -86,16 +90,19 @@ class AioHTTPWebSocketAdapter(AsyncWebSocketAdapter):
         self.request = request
         self.ws = ws
 
-    async def iter_json(self) -> AsyncGenerator[Dict[str, object], None]:
+    async def iter_json(
+        self, ignore_parsing_errors: bool
+    ) -> AsyncGenerator[Dict[str, object], None]:
         async for ws_message in self.ws:
             if ws_message.type == http.WSMsgType.TEXT:
                 try:
                     yield ws_message.json()
                 except JSONDecodeError:
-                    raise NonJsonMessageReceived()
+                    if not ignore_parsing_errors:
+                        raise NonJsonMessageReceived()
 
             elif ws_message.type == http.WSMsgType.BINARY:
-                raise NonJsonMessageReceived()
+                raise NonTextMessageReceived()
 
     async def send_json(self, message: Mapping[str, object]) -> None:
         await self.ws.send_json(message)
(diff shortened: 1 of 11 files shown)


botberry:



---

This release fixes a regression in the legacy GraphQL over WebSocket protocol.
Legacy protocol implementations should ignore client message parsing errors.
During a recent refactor, Strawberry changed this behavior to match the new protocol, where parsing errors must close the WebSocket connection.
The expected behavior is restored and adequately tested in this release.



Here's the tweet text: 
```

```

strawberry/aiohttp/views.py
```diff
@@ -86,16 +90,19 @@ def __init__(self, request: web.Request, ws: web.WebSocketResponse) -> None:
         self.request = request
         self.ws = ws
 
-    async def iter_json(self) -> AsyncGenerator[Dict[str, object], None]:
+    async def iter_json(
+        self, ignore_parsing_errors: bool
```
bellini666, line 94: nit: ensure this needs to be passed as a keyword to avoid someone calling as `iter_json(True)` and the `True` there is hard to understand the meaning

```suggestion
        self,
        *,
        ignore_parsing_errors: bool,
```

bellini666: Also, should it have a default value? (not sure, just asking =P)

strawberry/channels/handlers/ws_handler.py
```diff
@@ -31,20 +31,23 @@ class ChannelsWebSocketAdapter(AsyncWebSocketAdapter):
     def __init__(self, request: GraphQLWSConsumer, response: GraphQLWSConsumer) -> None:
         self.ws_consumer = response
 
-    async def iter_json(self) -> AsyncGenerator[Dict[str, object], None]:
+    async def iter_json(
+        self, ignore_parsing_errors: bool
```
bellini666, line 35: Ditto

strawberry/http/async_base_view.py
```diff
@@ -81,7 +81,9 @@ async def get_form_data(self) -> FormData: ...
 
 class AsyncWebSocketAdapter(abc.ABC):
     @abc.abstractmethod
-    def iter_json(self) -> AsyncGenerator[Dict[str, object], None]: ...
+    def iter_json(
+        self, ignore_parsing_errors: bool
```
bellini666, line 85: Ditto

strawberry/litestar/controller.py
```diff
@@ -192,13 +196,22 @@ class LitestarWebSocketAdapter(AsyncWebSocketAdapter):
     def __init__(self, request: WebSocket, response: WebSocket) -> None:
         self.ws = response
 
-    async def iter_json(self) -> AsyncGenerator[Dict[str, object], None]:
+    async def iter_json(
+        self, ignore_parsing_errors: bool
```
bellini666, line 200: Ditto

strawberry/litestar/controller.py
```diff
@@ -192,13 +196,22 @@ class LitestarWebSocketAdapter(AsyncWebSocketAdapter):
...
-                raise NonJsonMessageReceived()
+            while self.ws.connection_state != "disconnect":
+                text = await self.ws.receive_text()
+
+                # Litestar internally defaults to an empty string for non-text messages
+                if text == "":
+                    raise NonTextMessageReceived()
+
+                try:
+                    yield json.loads(text)
```
bellini666, line 211: Not related to this PR (so nothing to change here), but this got me thinking if we should allow for different json libs to be used in places like this, like `orjson`.

bellini666 (approved):
Left some nits, but LGTM :)

strawberry/litestar/controller.py
DoctorJohn: Very good point, I'll create a PR for it!

strawberry/aiohttp/views.py
DoctorJohn: Whoops. For some reason, I always thought keyword-only args were too new for the oldest Python version Strawberry supports. But they're not. I just made them keyword-only and also added a default.
I first didn't add a default to make it more "explicit," but `False` is a sane default here.

PR #3674, opened 2024-10-17
Fix create multiple directives

Speedy1991:
## Description

Directives are added multiple times leading to VSC errors like `The directive "@featuresRequired" can only be used once at this location.`

image

Another option to fix this would be to make field.directives a `set` but this is used on multiple places and I don't think it's worth the rework. If the fix is ok for you I'll add a test with e.g. [this sandbox setup](https://play.strawberry.rocks/?gist=b865d07a2ac9aa6211d4f6bd0d3ec291)

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* Check if the directive is already set on the field as directive

## Checklist

### DIFF
--- a/strawberry/permission.py
+++ b/strawberry/permission.py
@@ -158,9 +158,12 @@ class PermissionExtension(FieldExtension):
     def apply(self, field: StrawberryField) -> None:
         """Applies all of the permission directives to the schema and sets up silent permissions."""
         if self.use_directives:
-            field.directives.extend(
-                p.schema_directive for p in self.permissions if p.schema_directive
-            )
+            # Dedupe multiple directives
+            # https://github.com/strawberry-graphql/strawberry/issues/3596
+            permission_directives = {p.schema_directive for p in self.permissions if p.schema_directive}
+            existing_field_directives = set(field.directives)
+            extend_directives = permission_directives - existing_field_directives
+            field.directives.extend(extend_directives)
         # We can only fail silently if the field is optional or a list
         if self.fail_silently:
             if isinstance(field.type, StrawberryOptional):


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Arthur for the PR 👏

```

botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Arthur for the PR 👏

```

strawberry/permission.py
```diff
@@ -158,9 +158,14 @@ def __init__(
     def apply(self, field: StrawberryField) -> None:
         """Applies all of the permission directives to the schema and sets up silent permissions."""
         if self.use_directives:
-            field.directives.extend(
+            # Dedupe multiple directives
+            # https://github.com/strawberry-graphql/strawberry/issues/3596
```
bellini666, line 162: I don't think we need the issue linked here as it is easy to check the commit that changed this on git history

bellini666 (commented):
Hey @Speedy1991 ,

Thank you very much for this!

Could we have a test for this to avoid any future regressions?

strawberry/permission.py
Speedy1991: yep, ok

Speedy1991:
> Hey @Speedy1991 ,
> 
> Thank you very much for this!
> 
> Could we have a test for this to avoid any future regressions?

yea, i wasn't sure if refactoring the directives would be a more stable fix, see this comment:

> Another option to fix this would be to make field.directives a set but this is used on multiple places and I don't think it's worth the rework. If the fix is ok for you I'll add a test

bellini666:
> yea, i wasn't sure if refactoring the directives from a list to a set would be a more stable fix, see this comment:

Refactoring it into a `set` could possibly have unwanted consequences. Because they are not "ordered" you can't enforce permissions to run in a given order. So I think your fix is the way to go.

Speedy1991:
Hm I think I need to refactor this a littlebit if the order matters, because the current fix cast the lists to sets and calculates the difference. As far i know casting to/from a set does not guarantee to keep the order.

bellini666:
> Hm I think I need to refactor this a littlebit if the order matters, because the current fix cast the lists to sets and calculates the difference. As far i know casting to/from a set does not guarantee to keep the order.

Good point, I actually missed that (even though I commented about this same very issue lol)

Speedy1991:
@bellini666 I refactored the code to keep the order + added a test with a permission class.

Is it possible to add a reason to a permission class? Then i would add another test, confirming the same directive with different reasons will be kept.

bellini666 (approved):

PR #3689, opened 2024-11-07
Strict graphql ws typed dicts

DoctorJohn:
## Description

This PR refactors all types of the legacy graphql-ws protocol. I continued using typed dicts (vs dataclasses) for performance reasons, simplicity, and the ability to model null vs undefined easily. In the linked issue I complained about them, but now that we have `NotRequired` and `Required` available for typing, they're fun to work with and enable the modelling of super precise types.

Pylance is now pretty happy when I open the protocol implementation, some type casts were removed, and some minor type errors in tests surfaced and were fixed.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Enhancement/optimization
- [ ] Documentation

## Issues Fixed or Closed by This PR

* Fixes #1669
* Closes #2771

## Summary by Sourcery

Refactor the graphql-ws protocol types to use stricter typed dicts, enhancing type safety and clarity. Update tests to reflect these changes and add a release note to document the improvements.

Enhancements:
- Refactor all types of the legacy graphql-ws protocol to use stricter typed dicts, improving type safety and clarity.

Documentation:
- Add a release note detailing the refactoring of the graphql-ws protocol types and the improvements in type strictness.

Tests:
- Update tests to align with the new strict typing of the graphql-ws protocol, ensuring compatibility and correctness.

PR #3701, opened 2024-11-17
Migrate graphql-transport-ws types to TypedDicts

DoctorJohn:
## Description

I recently migrated the legacy ws protocols types to typed dicts (#3689), in this PR the modern ws protocols types are migrated to typed dicts. This, again, comes with some nice benefits such as:

- precise modeling of `null` vs `undefined` values
- no more need for custom `.to_dict()` methods on dataclasses to avoid performance penalties
- no more need for `UNSET` conversions
- more precise test assertions (that's were all the extra code comes from)
- (the protocol implementation now also no longer uses `Any`)

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

## Summary by Sourcery

Enhancements:
- Migrate graphql-transport-ws types from data classes to TypedDicts for better type precision and code simplification.

### DIFF
--- a/strawberry/channels/testing.py
+++ b/strawberry/channels/testing.py
@@ -17,23 +17,10 @@ from graphql import GraphQLError, GraphQLFormattedError
 
 from channels.testing.websocket import WebsocketCommunicator
 from strawberry.subscriptions import GRAPHQL_TRANSPORT_WS_PROTOCOL, GRAPHQL_WS_PROTOCOL
-from strawberry.subscriptions.protocols.graphql_transport_ws.types import (
-    ConnectionAckMessage,
-    ConnectionInitMessage,
-    ErrorMessage,
-    NextMessage,
-    SubscribeMessage,
-    SubscribeMessagePayload,
+from strawberry.subscriptions.protocols.graphql_transport_ws import (
+    types as transport_ws_types,
 )
-from strawberry.subscriptions.protocols.graphql_ws.types import (
-    ConnectionAckMessage as GraphQLWSConnectionAckMessage,
-)
-from strawberry.subscriptions.protocols.graphql_ws.types import (
-    ConnectionInitMessage as GraphQLWSConnectionInitMessage,
-)
-from strawberry.subscriptions.protocols.graphql_ws.types import (
-    StartMessage as GraphQLWSStartMessage,
-)
+from strawberry.subscriptions.protocols.graphql_ws import types as ws_types
 from strawberry.types import ExecutionResult
 
 if TYPE_CHECKING:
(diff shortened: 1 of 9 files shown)


strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -134,39 +130,28 @@ async def handle_connection_init_timeout(self) -> None:
...
-    async def handle_message(self, message: dict) -> None:
+    async def handle_message(self, message: Message) -> None:
         try:
-            message_type = message.pop("type")
+            if message["type"] == "connection_init":
+                await self.handle_connection_init(message)
 
-            if message_type == ConnectionInitMessage.type:
-                await self.handle_connection_init(ConnectionInitMessage(**message))
+            elif message["type"] == "ping":
```
erikwrede, line 138: Why are we switching from constants to strings here? I get that these strings are obvious, but my intuition tells me a constant suits this better

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -198,11 +180,11 @@ async def handle_connection_init(self, message: ConnectionInitMessage) -> None:
             return
 
         self.connection_init_received = True
-        await self.send_message(ConnectionAckMessage())
+        await self.websocket.send_json(ConnectionAckMessage({"type": "connection_ack"}))
```
erikwrede, line 183: As much as I like the improvement stemming from the dicts above, I really dislike this redundancy of types. What about adding a new `create` `classmethod` on the `TypedDict` class that handles filling this value? Not very pythonic, but at least you don't need to pass the message types again

erikwrede (changes requested):

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
DoctorJohn: I want the new TypedDicts to be the single source of truth for the shape of graphql-transport-ws messages. However, unlike with dataclassses, it's not possible to store a constant value in a TypedDict. The only option is to use `typing.Literal` and let the type checker enforce that `message["typed"]` must always match one of the literal message types. (which is what we do here).

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
DoctorJohn: Similar story to the previous comment: At runtime TypedDicts are just regular dicts. When subclassing `TypedDict` we can't add any custom methods (or constants) to it.
While this new style looks more verbose, it's also more explicit since it's no longer hiding away any message fields. I honestly like it, especially since it makes pyright and mypy super happy.

Here's something we could do with the hightlighted code:
`self.websocket.send_json` takes any dict as argument, that's why we wrap our message dict in `ConnectionAckMessage` so that the type checker ensures it has a valid shape.
Instead we could add and call a (new) `send_message(self, message: Message)` method, which expects a valid `Message` as argument. In the example above, we would no longer need to wrap our typed dict in `ConnectionAckMessage`, because the type checker would ensure that the message's shape is valid based on it's `type` field.

So instead of:

`await self.websocket.send_json(COnnectionAckMessage({"type": "connection_ack"})`

we would write:

`await self.send_message({"type": "connection_ack"})`

and type checker would still provide autocompletion and enforce that messages have a valid shape.

DoctorJohn: (Just added two commits doing exactly that)

strawberry/channels/testing.py
```diff
@@ -109,19 +96,21 @@ async def gql_init(self) -> None:
         if self.protocol == GRAPHQL_TRANSPORT_WS_PROTOCOL:
             assert res == (True, GRAPHQL_TRANSPORT_WS_PROTOCOL)
             await self.send_json_to(
-                ConnectionInitMessage(payload=self.connection_params).as_dict()
+                transport_ws_types.ConnectionInitMessage(
+                    {"type": "connection_init", "payload": self.connection_params}
+                )
```
patrick91, line 101: I don't know if **I** prefer, but we can also do this:


```suggestion
                transport_ws_types.ConnectionInitMessage(
                    type="connection_init", payload= self.connection_params
                )
```

no? 😊

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
patrick91: I think with types, plain strings are better 😊

patrick91 (approved):

erikwrede (approved):

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
erikwrede: My issue was mainly with the redundancy and verbosity of `COnnectionAckMessage({"type": "connection_ack"})`, but when reviewing I forgot to touch the possibility of just sending a normal dict :D Happy with this solution

PR #3708, opened 2024-11-20
Let WS and HTTP use the same JSON encoder

DoctorJohn:
## Description

It was already possible and documented to override `encode_json` on views to customize the JSON encoder used for responses sent via HTTP. The same `encode_json` method is now also used to encode WebSocket messages.

This was proposed a few weeks ago by @bellini666 and it's super handy if someone wants to use `orjson` or otherwise customize JSON responses and messages.

Note that this change is kinda breaking since the `encode_json` methode previously only expected `GraphQLHTTPResponse` dicts, whereas now it must be able to handle any kind of json data.
In reality anyone overriding this method can probably already handle any json data.

(Will do the same for decoding in a separate PR).

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

## Summary by Sourcery

Unify the JSON encoding process for HTTP and WebSocket by using the same encode_json method, allowing for consistent customization across both protocols. Update documentation and add tests to support this change.

New Features:
- Enable the use of a single JSON encoder for both HTTP responses and WebSocket messages by allowing the override of the encode_json method.

Documentation:
- Update documentation to reflect the change in the encode_json method, which now applies to both HTTP and WebSocket JSON responses.
- Add a breaking changes document for version 0.251.0, detailing the changes to the encode_json method signature.

Tests:
- Add a test to verify that WebSocket handlers use the view's encode_json method for encoding JSON messages.

### DIFF
--- a/docs/breaking-changes/0.251.0.md
+++ b/docs/breaking-changes/0.251.0.md
@@ -1,0 +1,25 @@
+---
+title: 0.251.0 Breaking Changes
+slug: breaking-changes/0.251.0
+---
+
+# v0.251.0 Breaking Changes
+
+We slightly changed the signature of the `encode_json` method used to customize
+the JSON encoder used by our HTTP views.
+
+Originally, the method was only meant to encode HTTP response data. Starting
+with this release, it's also used to encode WebSocket messages.
+
+Previously, the method signature was:
+
+```python
+def encode_json(self, response_data: GraphQLHTTPResponse) -> str: ...
+```
+
+To upgrade your code, change the method signature to the following and make sure
+your method can handle the same inputs as the built-in `json.dumps` method:
+
+```python
+def encode_json(self, data: object) -> str: ...
+```
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,7 @@
+Release type: minor
+
+Starting with this release, the same JSON encoder is used to encode HTTP
+responses and WebSocket messages.
+
+This enables developers to override the `encode_json` method on their views to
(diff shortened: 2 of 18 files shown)


botberry:



---

Starting with this release, the same JSON encoder is used to encode HTTP
responses and WebSocket messages.

This enables developers to override the `encode_json` method on their views to
customize the JSON encoder used by all web protocols.



Here's the tweet text: 
```

```

docs/breaking-changes/0.251.0.md
```diff
@@ -0,0 +1,25 @@
...
+
+```python
+def encode_json(self, response_data: GraphQLHTTPResponse) -> str: ...
+```
+
+To upgrade your code, change the method signature to the following and make sure
+your method can handle the same inputs as the built-in `json.dumps` method:
+
+```python
+def encode_json(self, data: object) -> str: ...
```
patrick91, line 24: could we type this better? is it worth it?

DoctorJohn: `json.loads` expects `Any` so I went with `object`. Technically it's anything that can be serialized to JSON. But the `json` library doesn't provide a type for it. We would have to create our own, not sure whether it's worth it.

patrick91: @DoctorJohn mmh, is it because the ws integration can send pretty much anything, right?

patrick91: ok, sounds good to me 😊

patrick91 (approved):

Issue #3481, opened 2024-05-01
`strawberry.Parent` not supporting forward refs


andrewkruse:
I would like the strawberry documentation on accessing parent with function resolvers on this [page](https://strawberry.rocks/docs/guides/accessing-parent-data#accessing-parents-data-in-function-resolvers) tweaked to be more clear, or maybe corrected?

From what I understand in the docs, its suggesting you end up with the following. However, this doesn't even run? I have tried swapping the definitions both directions, they have the same issue. I had to resort to the `self` method on a method resolver, which seems less desirable to me since the docs specifically call out that it might not work quite right everywhere.

> and it works like it should in Python, but there might be cases where it doesn’t properly follow Python’s semantics

```
def get_full_name(parent: strawberry.Parent[User2]) -> str:
    return f"{parent.first_name} {parent.last_name}"

@strawberry.type
class User2:
    first_name: str
    last_name: str
    full_name: str = strawberry.field(resolver=get_full_name)
```

```
Traceback (most recent call last):
  File "/Users/.../Library/Application Support/JetBrains/IntelliJIdea2024.1/plugins/python/helpers/pydev/pydevd.py", line 1535, in _exec
    pydev_imports.execfile(file, globals, locals)  # execute the script
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/.../Library/Application Support/JetBrains/IntelliJIdea2024.1/plugins/python/helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "/Users/.../src/api/python-graphql-poc/src/main.py", line 21, in <module>
    def get_full_name(parent: strawberry.Parent[User2]) -> str:
                                                ^^^^^
NameError: name 'User2' is not defined
```

Perhaps there's a quirk in here where the structure of my file is part of the problem since everything is top level? I am using FastAPI, uvicorn and strawberry.

patrick91:
@andrewkruse I think you found a bug, this should work:

```python
from __future__ import annotations   # NOTE: this

import strawberry


def get_full_name(user: strawberry.Parent[User]) -> str:
    return f"{user.first_name} {user.last_name}"


@strawberry.type
class User:
    first_name: str
    last_name: str
    full_name: str = strawberry.field(resolver=get_full_name)


@strawberry.type
class Query:
    @strawberry.field
[...]

PR #3714, opened 2024-12-02
Update accessing parent data guide with lazy type declaration

mattalbr:
## Description

The "accessing parent data" guide did not actually work as written because the strawberry.Parents need to use lazy declarations if the class is not yet defined.

This was identified and discussed in https://github.com/strawberry-graphql/strawberry/issues/3481#issuecomment-2503867702

The strawberry lazy incantation is quite verbose, hence why "self" remains useful. 

I'm definitely not a doc writer, so just did my best. Feel free to suggest edits.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation

* https://github.com/strawberry-graphql/strawberry/issues/3481

## Checklist

### DIFF
--- a/docs/guides/accessing-parent-data.md
+++ b/docs/guides/accessing-parent-data.md
@@ -57,12 +57,14 @@ pass us the parent of the field, we need to add a new argument with type
 `strawberry.Parent[ParentType]`, like so:
 
 ```python
-def get_full_name(parent: strawberry.Parent[User]) -> str:
+def get_full_name(parent: strawberry.Parent[Annotated["User", strawberry.lazy(".")]]) -> str:
     return f"{parent.first_name} {parent.last_name}"
 ```
 
 `strawberry.Parent` tells Strawberry to pass the parent value of the field, in
-this case it would be the `User`.
+this case it would be the `User`. Because the `User` class has not been defined
+yet at the point that we define the resolver, we need to use a [lazy](
+https://strawberry.rocks/docs/types/lazy) type declaration.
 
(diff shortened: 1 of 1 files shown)


bellini666:
Checking the changes to the docs here I wonder if straightforward cases like this, where `Parent` is something in the same file where it is defined, could avoid requiring `lazy`

i.e. lazy is mostly useful for when we want to avoid circular import issues, and because of that, we need to import the other type inside a `if TYPE_CHECKING` statement.

For the cases documented here, `User` is defined in the same file so I would expect that it would work without requiring lazy.

Maybe we should try to fix the issue instead? What do you think @mattalbr ?

bellini666:
@mattalbr I'll try to play around with that during the weekend and see if I can come up with a solution :)

mattalbr:
@bellini666 I noticed this still in my PRs and just wanted to check in!

bellini666:
> @bellini666 I noticed this still in my PRs and just wanted to check in!

Hey @mattalbr , sorry about that, it has been very busy for me at this end/start of the year T_T

I'm planning on trying to tackle this in the next couple of weeks. Please ping me if I don't 😅

mattalbr:
> > @bellini666 I noticed this still in my PRs and just wanted to check in!
> 
> Hey @mattalbr , sorry about that, it has been very busy for me at this end/start of the year T_T
> 
> I'm planning on trying to tackle this in the next couple of weeks. Please ping me if I don't 😅

Just checking back in @bellini666 !

mattalbr:
No worries at all, I know how it goes sometimes! Thanks for taking that on!

Issue #2988, opened 2023-07-28
Support for query complexity


patrick91:
Hey folks, I while ago I wrote down some ideas on how we can add support for query complexity. For this we'll need two things:

1. An extensions that calculates the costs based on an operation
2. Support for customising the costs based on fields/arguments

For customising costs on fields and arguments I was thinking about something like this:

```python
import strawberry

@strawberry.type
class Query:
    @strawberry.field(cost=2)
    def expensive_list(self, limit: Annotated[int, strawberry.argument(cost_multiplier=1)]) -> list[str]:
        return ["Item"] * limit
```

Here we are saying that each individual item in the `expensive_list` field costs 2, and the total is a multiplication between the field cost and the limit, so the result would be:

```
field_cost × limit × 1
```

We can also have defaults like this:

```
class StrawberryConfig:
    # Existing attributes and methods ...
    
    default_argument_cost_multiplier: Dict[str, Union[int, float]] = {
        "limit": 1,
        "first": 1, # probably not needed, but worth showing as an example
        "last": 1
    }
```

`default_argument_cost_multiplier` is a map between argument name and their multiplier, though I don't like the name that much.

What do you think?

devkral:
nice idea. But maybe a proposal:

why not extending graphene-protector? This extension works for strawberry as well as graphene as well as plain graphql-python

devkral:
(side note: a project of mine)

patrick91:
@devkral oh I forgot about it!

I think we could use internally 😊 my reasoning around having this feature built-in is so that we have a great API (I quite the new argument on field/argument) and also we make sure it always works, especially if we change the core of Strawberry 😊

devkral:
I see your point. Assigning arguments with costs is cool idea.

bradleyoesch:
@patrick91 

I've been looking at this quite a bit as well.

Here are some resources I've found and things I like about them that we could maybe take advantage of

- https://graphql-ruby.org/queries/complexity_and_depth
   - Some way to log/see the complexity of a request (maybe only in graphiql? or configurable to decide when to allow clients to be able to see it?)
   - Default values for common types, e.g. simple types default to 1, connection types multiply
   - Complexity override functions, not just integers
- https://github.com/pa-bru/graphql-cost-analysis
   - Map of costs (as you've posted above)
   - Clear documentation on how unions and interfaces are handled
[...]

PR #3721, opened 2024-12-11
Add schema extension for estimating query complexity

serramatutu:
## Summary
This commit introduces a new `QueryComplexityEstimator` class. It's a `SchemaExtension` which traverses the query tree during validation time and estimates the complexity of executing of each node of the tree. Its intended use is primarily token-bucket rate-limiting based on query-complexity, but the API is flexible enough to allow users to do whatever they want with the results.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

* https://github.com/strawberry-graphql/strawberry/issues/2988

## Public API

I introduced the following types users can interact with:

### `QueryComplexityEstimator`
This is the `SchemaExtension` that should get added to the global Strawberry `Schema`. It has 3 possible arguments:
- `default_estimator: FieldComplexityEstimator | int`: The default `FieldComplexityEstimator` that should be used for fields which don't specify an override. This provides a nice default so users don't need to bother adding complexity estimation logic for every single field. Just throwing in a good default should solve 90% of your problems. If this is an `int`, we'll automatically use a `ConstantFieldComplexityEstimator`.
- `callback: Callable[[Dict[str, int], ExecutionContext], None]`: In this callback, users can implement their own logic that uses the complexity value, such as raising exceptions for rate-limiting, logging this complexity to Datadog or whatever.
- `response_key: str | None`: If it's provided, the extension will return the complexity map to the client under `result.extensions[response_key]`. If it's not provided, complexities will not be returned to clients.

### `FieldComplexityEstimator`
This is the base class that estimates the complexity of running a single field. Users can inherit from this class to implement their own complexity calculation logic. It has a `estimate_complexity()` method which accepts the following args:
- `child_complexities: Iterator[int]`: an iterator over the complexities of child nodes (i.e nodes in the selection set). If this is empty, it means the node we're currently evaluating is a leaf.
- `arguments: FieldArgumentsType`: This is a dictionary of all the arguments (parameters) for a field. This can be used to calculate the complexity based on a user-defined parameter like `page_size`, `document_length` or whatever.

The following 2 types are concrete implementations of `FieldComplexityEstimator` that users can use out of the box without bothering to implement them. I added these since I expect these patterns to be quite common.

### `ConstantFieldComplexityEstimator`
An implementation of `FieldComplexityEstimator` that takes in a single `complexity: int` parameter and always returns that, no matter the cost of its children or parameter values.

### `SimpleFieldComplexityEstimator`
An implementation of `FieldComplexityEstimator` that takes in a single `scalar_complexity: int` and returns that for scalar nodes. Object nodes return the sum of their children.

Admittedly, this name kinda sucks. I couldn't think of anything better though, so I'm happy to change if anyone has a better idea.

## Example

Here's an example that combines all of that to make it easier to digest:
```python
from typing import Iterator

from graphql.error import GraphQLError

import strawberry
from strawberry.extensions import FieldComplexityEstimator, QueryComplexityEstimator


class MyEstimator(FieldComplexityEstimator):
    def estimate_complexity(
        self, child_complexities: Iterator[int], arguments: dict[str, Any]
    ) -> int:
        children_sum = sum(child_complexities)
        # scalar fields cost 1
        if children_sum == 0:
            return 1

        # non-list object fields cost the sum of their children
        if "page_size" not in field_kwargs:
            return children_sum

        # paginated fields cost gets multiplied by page size
        return children_sum * field_kwargs["page_size"]


# initialize your rate-limiter somehow
rate_limiter = ...


def my_callback(
    complexities: dict[str, int], execution_context: ExecutionContext
) -> None:
    # add complexities to execution context
    execution_context.context["complexities"] = complexities

    # apply a token-bucket rate-limiter by user_id
    total_cost = sum(complexities.values())
    bucket = rate_limiter.get_bucket_for_key(execution_context.context["user_id"])
    tokens_left = bucket.take_tokens(total_cost)
    if tokens_left <= 0:
        raise GraphQLError(
            "Rate-limit exhausted. Please wait for some time before trying again."
        )


schema = strawberry.Schema(
    Query,
    extensions=[
        QueryComplexityEstimator(
            default_estimator=MyEstimator(),
            callback=my_callback,
        ),
    ],
)
```

## Caveats, possible issues and other solutions I considered
There are some caveats to this implementation, and I wanted to bring them to your attention. I'm happy to discuss and/or change the implementation if any of these are dealbreakers :)

### Traversing the tree twice
`QueryComplexityEstimator` runs on `on_validate` and it traverses the GraphQL document AST using its own logic. This means that if it gets used with another extension that does the same such as `QueryDepthLimiter`, it is possible that a significant chunk of the tree will get traversed twice since each extension traverses the tree on its own. This could introduce performance issues, especially for large queries.

The idea I had to optimize this was implement a generic `Visitor` class that gets called by the Strawberry core only once, and then all the "visiting" extensions would register themselves to this visitor. While it's probably faster since it traverses the tree only once, this other approach isn't as flexible since extensions wouldn't get to pick visiting order, and the state management in each extension probably gets a little more annoying. It would also require some significant changes to the Strawberry core, which I thought was out of scope for this PR.

### The estimator is not-eager
Differently from `MaxTokensLimiter` and `QueryDepthLimiter`, the `QueryComplexityEstimator` is not eager. In other words, there is no predefined "limit" that will stop tree traversal and simply return "you're over the limit". Instead, it always calculates the complexity of the _entire query_, and it is on the developer to use the `callback` and do whatever they want with the results.

The main reasoning behind this is that by the nature of rate-limiting, the limit fluctuates, and needs to be dynamically assessed, so we can't create a global `limit` threshold. I thought of solving this by making a `pre` hook that would return the current limit given the `ExecutionContext`. IMO this would make the API more convoluted and annoying than it needs to be, so in the end I discarded that idea and just went with `callback`.

This decision has one important consideration: users should be strongly encouraged to use `MaxTokensLimiter` and `QueryDepthLimiter` to limit the size of the GraphQL documents before passing them into `QueryComplexityEstimator`, otherwise huge queries could cause unbounded recursion and become an attack vector.

Would love to hear your thoughts on this, though!

## Checklist

Thank you for creating and maintaing Strawberry :D


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Lucas Valente for the PR 👏

```

strawberry/extensions/query_complexity_estimator.py
```diff
@@ -0,0 +1,375 @@
...
+                node=query,
+                fragments=fragments,
+            )
+            for name, query in queries.items()
+        }
+
+        self.results = query_complexities
+
+        if callable(self.callback):
+            self.callback(query_complexities, self.execution_context)
```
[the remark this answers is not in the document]
serramatutu: Hmmm not sure about this. If the user callback fails, we should just error IMO.

strawberry/extensions/query_complexity_estimator.py
```diff
@@ -0,0 +1,375 @@
...
+                )
+                for child in node.selection_set.selections
+            )
+
+        if isinstance(node, FieldNode):
+            node_type = self._get_type_info(parent_type.name, node.name.value)
+            variables = self.execution_context.variables or {}
+            node_body = node
+            args = {
+                to_snake_case(arg.name.value): variables.get(arg.value.name.value, None)
```
serramatutu, line 332: Is there an inverse function to `self.execution_context.schema.config.name_converter.apply_naming_config` that will turn camel into snake case without hardcoding it here. What we need is to get the `python_name` of a field from its `graphql_name`

strawberry/extensions/query_complexity_estimator.py
```diff
@@ -0,0 +1,375 @@
...
+    FieldNode,
+    FragmentDefinitionNode,
+    Node,
+    OperationDefinitionNode,
+)
+from graphql.language.ast import FragmentSpreadNode, VariableNode
+
+from strawberry.extensions.base_extension import SchemaExtension
+from strawberry.extensions.field_extension import FieldExtension
+from strawberry.extensions.query_depth_limiter import (
```
serramatutu, line 25: NOTE: I reused some of the pre-existing machinery from `QueryDepthLimiter`.

jacobmoshipco:
I'm noticing `estimate_complexity` doesn't seem to take information about what field is having its complexity estimated. Would the extension be able to take a `StrawberryField` instance like `strawberry.Info` contains, or is it too early in the GraphQL parsing flow to be able to instantiate that? Apologies if I'm just missing something.

serramatutu:
@jacobmoshipco Yea I think that's a good point. I can't really see any practical use cases right now that would require the field itself, but just having it there would make the API a lot more flexible in case users wanna do introspection etc

I will add that :)

hugobessa:
@serramatutu is this PR moving forward? Do you want any help? I could fix the CI build or something

serramatutu:
@hugobessa It's been a while since I last looked at any of this 😔 

I still gotta implement the suggestions mentioned above and possible rebase the branch since it's on a very very old version of Strawberry.

patrick91:
@serramatutu do you still want to work on this? just had a look with @bellini666 and @Ckk3 and the code looks sensible but we also want to allow users to add costs directly to the field (like in the mentioned issue)

We can update this PR if you're busy!

serramatutu:
@patrick91 Hey! Sorry I haven't been able to take this over the finish line. I'm kinda busy at the moment but feel free to modify this however you want to make it fit!

This branch is pretty old though so it might need some big refactors.

PR #3744, opened 2024-12-28
fix: make payload:null handling in connection_init consistent between protocols

Object905:
When client specifies `payload: null` in connection_init message graphql_transport_ws erored with `Invalid connection init payload`, while in `graphql_ws` this case was handled.

### DIFF
--- a/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
+++ b/strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
@@ -167,7 +167,8 @@ class BaseGraphQLTransportWSHandler(Generic[Context, RootValue]):
             self.connection_init_timeout_task.cancel()
 
         payload = message.get("payload", {})
-
+        if payload is None:
+            payload = {}
         if not isinstance(payload, dict):
             await self.websocket.close(
                 code=4400, reason="Invalid connection init payload"
--- a/strawberry/subscriptions/protocols/graphql_ws/handlers.py
+++ b/strawberry/subscriptions/protocols/graphql_ws/handlers.py
@@ -97,7 +97,9 @@ class BaseGraphQLWSHandler(Generic[Context, RootValue]):
 
     async def handle_connection_init(self, message: ConnectionInitMessage) -> None:
         payload = message.get("payload")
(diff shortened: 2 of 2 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Zverev Konstantin for the PR 👏

```

strawberry/subscriptions/protocols/graphql_ws/handlers.py
```diff
@@ -97,7 +97,9 @@ async def handle_message(
 
     async def handle_connection_init(self, message: ConnectionInitMessage) -> None:
         payload = message.get("payload")
-        if payload is not None and not isinstance(payload, dict):
+        if payload is None:
```
Object905, line 100: For graphql_ws before this PR connection_params could be null, if payload is null.
So this may be a slightly breaking change.

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -167,7 +167,8 @@ async def handle_connection_init(self, message: ConnectionInitMessage) -> None:
             self.connection_init_timeout_task.cancel()
 
         payload = message.get("payload", {})
-
+        if payload is None:
```
Object905, line 170: Not breaking for graphql_transport_ws, because it was required to be dict.

It's possible to make other way around and allow null in graphql_transport_ws, like in graphql_ws. 
But since it's "recommended" protocol I decided to change graphql_ws and not the other way around.

DoctorJohn:
Hi @Object905, I noticed this recently myself and planned to do two things:

1. Update these checks
2. No longer default to empty dictionaries at all, so that we can differentiate between
    - no payload
    - a `null` payload
    - an empty object as payload

Your PR essentially addresses (1.), that's great, but it might be worth it to do both things in the same PR so that we only have one breaking release. What do you think? You could also just do (1.) and make sure both handlers accept the payloads defined by their protocols (I recommend checking the `types.py` files next to the handlers).

Either way, we would need tests for these changes.

PR #3745, opened 2025-01-02
Type internal test clients stricter

DoctorJohn:
## Description

This PR improves the typing of our internal HTTP test clients.

The motivation behind this is to see what parts of them could be re/used for our public test clients and to make pylance and me much happier when I work in vscode lol

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

## Summary by Sourcery

Enhancements:
- Add generic type parameters to internal HTTP test clients to improve type checking.

### DIFF
--- a/tests/http/clients/asgi.py
+++ b/tests/http/clients/asgi.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import contextlib
 import json
-from collections.abc import AsyncGenerator
+from collections.abc import AsyncGenerator, Mapping
 from io import BytesIO
 from typing import Any, Optional, Union
 from typing_extensions import Literal
@@ -86,7 +86,7 @@ class AsgiHttpClient(HttpClient):
     async def _graphql_request(
         self,
         method: Literal["get", "post"],
-        query: Optional[str] = None,
+        query: str,
         variables: Optional[dict[str, object]] = None,
         files: Optional[dict[str, BytesIO]] = None,
         headers: Optional[dict[str, str]] = None,
@@ -152,7 +152,7 @@ class AsgiHttpClient(HttpClient):
         return Response(
             status_code=response.status_code,
             data=response.content,
-            headers=response.headers,
+            headers=dict(response.headers),
         )
 
     @contextlib.asynccontextmanager
@@ -162,13 +162,12 @@ class AsgiHttpClient(HttpClient):
         *,
         protocols: list[str],
     ) -> AsyncGenerator[WebSocketClient, None]:
-        try:
-            with self.client.websocket_connect(url, protocols) as ws:
-                yield AsgiWebSocketClient(ws)
-        except WebSocketDisconnect as error:
-            ws = AsgiWebSocketClient(None)
-            ws.handle_disconnect(error)
-            yield ws
+        with self.client.websocket_connect(url, protocols) as ws:
+            ws_client = AsgiWebSocketClient(ws)
+            try:
+                yield ws_client
+            except WebSocketDisconnect as error:
+                ws_client.handle_disconnect(error)
 
 
 class AsgiWebSocketClient(WebSocketClient):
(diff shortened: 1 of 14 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

tests/http/clients/asgi.py
```diff
@@ -162,13 +162,12 @@ async def ws_connect(
...
-        except WebSocketDisconnect as error:
-            ws = AsgiWebSocketClient(None)
-            ws.handle_disconnect(error)
-            yield ws
+        with self.client.websocket_connect(url, protocols) as ws:
+            ws_client = AsgiWebSocketClient(ws)
+            try:
+                yield ws_client
+            except WebSocketDisconnect as error:
+                ws_client.handle_disconnect(error)
```
erikwrede, line 170: this is not functionally equivalent, can Disconnect not be thrown during the context manager init?
Also, the except part doesn't seem to be covered by tests, maybe we should change that

tests/http/clients/async_flask.py
```diff
@@ -16,12 +16,12 @@
 from .flask import FlaskHttpClient
 
 
-class GraphQLView(BaseAsyncGraphQLView):
+class GraphQLView(BaseAsyncGraphQLView[dict[str, object], object]):
```
erikwrede, line 19: we are reusing `[dict[str, object], object]` quite often for `Context, RootValue` in the specific framework views. Do you think it makes sense to factor this out into a separate intermediate layer?

tests/http/clients/fastapi.py
```diff
@@ -179,10 +179,10 @@ async def ws_connect(
...
-            ws = AsgiWebSocketClient(None)
-            ws.handle_disconnect(error)
-            yield ws
+        with self.client.websocket_connect(url, protocols) as ws:
+            ws_client = AsgiWebSocketClient(ws)
+
+            try:
+                yield ws_client
+            except WebSocketDisconnect as error:
+                ws_client.handle_disconnect(error)
```
erikwrede, line 188: Same as above, are we sure about the DC errors?

erikwrede (commented):
minor thoughts, rest LGTM

tests/http/clients/asgi.py
DoctorJohn: The change is intentional: the init method can not throw.

Thanks for pointing out the test coverage. Before this change, the except block was invalid, telling us that we don't actually have tests relying on it. Since this is a client used solely for testing, we should probably just remove it instead of adding an artificial test just for coverage of code only used for testing.

tests/http/clients/async_flask.py
DoctorJohn: Fair enough! It would be nice to have a base class for test views that implements `get_context` and `get_root_value`. That would work decently well for `get_root_value`, every view that uses context vars (i.e., fastapi and litestar) would still override `get_context`. But yup, that could be done. I'll try and see whether it makes things better :)

erikwrede: We could also do that in a separate pr to keep things here typing-only

tests/http/clients/asgi.py
erikwrede: Fine with that, let's add a comment detailing this though.

tests/http/clients/async_flask.py
DoctorJohn: Good point, I'll do that instead.

tests/http/clients/asgi.py
DoctorJohn: It turned out the special cleanup code is a relic from before our HTTP & WS tests and test clients were shared. I have fully removed it now.

patrick91 (approved):

bellini666 (approved):

PR #3749, opened 2025-01-09
Add `strawberry.cast` 

bellini666:
The common `node: Node` used to resolve relay nodes means we will be relying on `is_type_of` to check if the returned object is in fact a subclass of the `Node` interface.

However, integrations such as Django, SQLAlchemy and Pydantic will not return the type itself, but instead an alike object that is later resolved to the expected type.

In case there are more than one possible type defined for that model that is being returned, the first one that replies `True` to `is_type_of` check would be used in the resolution, meaning that when asking for `PublicUser:123`, strawberry could end up returning `User:123`, which can lead to security issues (such as data leakage).

In here we are introducing a new `strawberry.cast`, which will be used to mark an object with the already known type by us, and when asking for `is_type_of` that mark will be used to check instead, ensuring we will return the correct type.

## Summary by Sourcery

Bug Fixes:
- Fix a security issue where resolving a relay node with multiple possibilities could return the wrong type, leading to potential data leakage.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,21 @@
+Release type: minor
+
+The common `node: Node` used to resolve relay nodes means we will be relying on
+is_type_of to check if the returned object is in fact a subclass of the Node
+interface.
+
+However, integrations such as Django, SQLAlchemy and Pydantic will not return
+the type itself, but instead an alike object that is later resolved to the
+expected type.
+
+In case there are more than one possible type defined for that model that is
+being returned, the first one that replies True to `is_type_of` check would be
+used in the resolution, meaning that when asking for `"PublicUser:123"`,
+strawberry could end up returning `"User:123"`, which can lead to security
+issues (such as data leakage).
+
+In here we are introducing a new `strawberry.cast`, which will be used to mark
+an object with the already known type by us, and when asking for is_type_of that
+mark will be used to check instead, ensuring we will return the correct type.
+
+That `cast` is already in place for the relay node resolution and pydantic.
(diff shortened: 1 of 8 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

strawberry/experimental/pydantic/object_type.py
```diff
@@ -29,6 +29,7 @@
     get_private_fields,
 )
 from strawberry.types.auto import StrawberryAuto
+from strawberry.types.cast import is_strawberry_cast_obj
```
patrick91, line 32:
```suggestion
from strawberry.types.cast import is_strawberry_type_cast
```

RELEASE.md
```diff
@@ -0,0 +1,21 @@
...
+being returned, the first one that replies True to `is_type_of` check would be
+used in the resolution, meaning that when asking for `"PublicUser:123"`,
+strawberry could end up returning `"User:123"`, which can lead to security
+issues (such as data leakage).
+
+In here we are introducing a new `strawberry.cast`, which will be used to mark
+an object with the already known type by us, and when asking for is_type_of that
+mark will be used to check instead, ensuring we will return the correct type.
+
+That `cast` is already in place for the relay node resolution and pydantic.
```
patrick91, line 21:
```suggestion
This release introduces `strawberry.cast`, which can be used when returning
a non Strawberry type to tell Strawberry what GraphQL type you wanted to return.

This is mostly useful when using unions or interfaces and returning non Strawberry 
types, like in integration like Strawberry Django, SQLAlchemy and Pydantic.
```

patrick91 (approved):

PR #3753, opened 2025-01-14
WIP defer support

patrick91:
Worked on this with @bellini666

Still need to check:

- [ ] extensions
- [ ] commented out code
- [ ] stream (Streamable type)
- [ ] multipart subscriptions are probably broken


patrick91:
Me and Thiago spent a bit of time on this again today 😊 Here's a report of what we need to think about for next time:

`handle_execution_results` (or similar) is currently called from multiple places in `execute` (and `subscribe`). We were thinking of moving this to the schema class so that we reduce the places where this functions is called, but also we move this function call where the results are actually returned (in the http and ws integration).

We also need to think about the `extensions` field on the response, apparently this can be on the initial response, on the subsequent response *and* on each increment result.

Not sure how can fill it in the subsequent response data, see this test for the shape:

```python
async def test_basic_defer(method: Literal["get", "post"], http_client: HttpClient):
    response = await http_client.query(
        method=method,
        query="""{
            hello

            ... @defer {
                asyncHello
            }
        }""",
    )

    async with contextlib.aclosing(response.streaming_json()) as stream:
        initial = await stream.__anext__()

        assert initial == {
            "data": {"hello": "Hello world"},
            "incremental": [],
            "hasNext": True,
            # TODO: why is this None?
            "extensions": None,
        }

        subsequent = await stream.__anext__()

        assert subsequent == {
            "incremental": [
                {
                    "data": {"asyncHello": "Hello world"},
                    "extensions": {"example": "example"},
                }
            ],
            "hasNext": False,
            # TODO: how do we fill these?
            "extensions": None,
        }

```

patrick91:
We updated to GraphQL-core alpha 7, we were discussing to remove our wrappers for the execution results because they make the code much worse now that we have more type

nrbnlulu:
Diogo we support you :heart: 
image

patrick91:
> we were discussing to remove our wrappers for the execution results because they make the code much worse now that we have more type

So, doing this is messy, we have a lot of code that uses our types, so I think it might be best to make new wrapper for incremental results too :'D

patrick91:
Defer was added via another PR!

Issue #1383, opened 2021-10-28
support batched queries


Speedy1991:
It would be awesome if strawberry would support [batched queries](https://www.apollographql.com/docs/react/api/link/apollo-link-batch-http/)

I already spend some time into this and this is my very first working example
The error handling is garbage but I hope you get the idea:

```
# strawberry/http.py
def process_result(result: ExecutionResult) -> Union[GraphQLHTTPResponse, List[GraphQLHTTPResponse]]:
    if isinstance(result, list):
        data: List[GraphQLHTTPResponse] = {"data": [r.data for r in result]}
        data["errors"] = [[format_graphql_error(err) for err in (getattr(r, 'errors', []) or [])] for r in result]
        if not data["errors"]:
            del data["errors"]
        data["extensions"] = [getattr(r, 'extensions', None) for r in result]
        if not data["extensions"]:
            del data["extensions"]
    else:
        data: GraphQLHTTPResponse = {"data": result.data}
        if result.errors:
            data["errors"] = [format_graphql_error(err) for err in result.errors]
        if result.extensions:
            data["extensions"] = result.extensions

    return data

def parse_request_data(data: Union[Dict, List[Dict]]) -> Union[GraphQLRequestData, List[GraphQLRequestData]]:
    if isinstance(data, list):
        if not all([ d.get('query', None) for d in data]):
            raise MissingQueryError()
        results = [
            GraphQLRequestData(query=d["query"], variables=d.get("variables"), operation_name=d.get("operationName"))
            for d in data
        ]
        return results

    if "query" not in data:
        raise MissingQueryError()

    result = GraphQLRequestData(
        query=data["query"],
        variables=data.get("variables"),
        operation_name=data.get("operationName"),
    )

    return result
```

```
# strawberry/django.views.py
if isinstance(request_data, list):
    result = [self.schema.execute_sync(
        rd.query,
        root_value=self.get_root_value(request),
        variable_values=rd.variables,
        context_value=context,
        operation_name=rd.operation_name,
    ) for rd in request_data]

else:
    result = self.schema.execute_sync(
        request_data.query,
        root_value=self.get_root_value(request),
        variable_values=request_data.variables,
        context_value=context,
        operation_name=request_data.operation_name,
    )
```

aryaniyaps:
@patrick91 should this be added to the roadmap? Graphene has this feature and it is simple to implement IMO.

cheradenine:
[...]

PR #3755, opened 2025-01-15
Add Query batching support

aryaniyaps:
# Add Query Batching Support

## Description

This PR adds Query batching support for GraphQL APIs using Strawberry GraphQL.

This makes Strawberry GraphQL compatible with batching features supported by various
client side libraries, such as [Apollo GraphQL](https://www.apollographql.com/docs/react/api/link/apollo-link-batch-http) and [Relay](https://github.com/relay-tools/react-relay-network-modern?tab=readme-ov-file#batching-several-requests-into-one).

Example (FastAPI):

```py
import strawberry

from fastapi import FastAPI
from strawberry.fastapi import GraphQLRouter


@strawberry.type
class Query:
    @strawberry.field
    def hello(self) -> str:
        return "Hello World"


graphql_app = GraphQLRouter(schema, batch=True)

app = FastAPI()
app.include_router(graphql_app, prefix="/graphql/batch")
```

Example (Flask):
```py
from flask import Flask
from strawberry.flask.views import GraphQLView

from api.schema import schema

app = Flask(__name__)

app.add_url_rule(
    "/graphql/batch",
    view_func=GraphQLView.as_view("graphql_view", schema=schema, batch=True),
)

if __name__ == "__main__":
    app.run()
```

Note: Query Batching is not supported for multipart subscriptions

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [ ] Documentation

* https://github.com/strawberry-graphql/strawberry/issues/1383

## Checklist

### DIFF
--- a/strawberry/http/async_base_view.py
+++ b/strawberry/http/async_base_view.py
@@ -533,9 +533,13 @@ class AsyncBaseHTTPView(
             raise HTTPException(400, "Unsupported content type")
 
         if isinstance(data, list):
-            if protocol == "multipart-subscription" or not self.batch:
+            if protocol == "multipart-subscription":
                 # note: multipart-subscriptions are not supported in batch requests
-                raise HTTPException(400, "Batch requests are not supported")
+                raise HTTPException(
+                    400, "Batching is not supported for multipart subscriptions"
+                )
+            if not self.batch:
+                raise HTTPException(400, "Batching is not enabled")
             return [
                 GraphQLRequestData(
                     query=item.get("query"),
--- a/strawberry/http/async_base_view.py
+++ b/strawberry/http/async_base_view.py
@@ -126,6 +126,8 @@ class AsyncBaseHTTPView(
         BaseGraphQLWSHandler[Context, RootValue]
     )
 
+    batch: bool
+
     @property
     @abc.abstractmethod
     def allow_queries_via_get(self) -> bool: ...
@@ -147,7 +149,9 @@ class AsyncBaseHTTPView(
 
     @abc.abstractmethod
     def create_response(
-        self, response_data: GraphQLHTTPResponse, sub_response: SubResponse
+        self,
+        response_data: Union[GraphQLHTTPResponse, list[GraphQLHTTPResponse]],
+        sub_response: SubResponse,
     ) -> Response: ...
 
     @abc.abstractmethod
@@ -178,8 +182,12 @@ class AsyncBaseHTTPView(
     ) -> WebSocketResponse: ...
 
     async def execute_operation(
-        self, request: Request, context: Context, root_value: Optional[RootValue]
-    ) -> Union[ExecutionResult, SubscriptionExecutionResult]:
+        self,
+        request: Request,
+        context: Context,
+        root_value: Optional[RootValue],
+        sub_response: SubResponse,
+    ) -> Union[ExecutionResult, list[ExecutionResult], SubscriptionExecutionResult]:
         request_adapter = self.request_adapter_class(request)
 
         try:
@@ -197,6 +205,22 @@ class AsyncBaseHTTPView(
 
         assert self.schema
 
+        if isinstance(request_data, list):
+            # batch GraphQL requests
+            tasks = [
+                self.execute_single(
+                    request=request,
+                    request_adapter=request_adapter,
+                    sub_response=sub_response,
+                    context=context,
+                    root_value=root_value,
+                    request_data=data,
+                )
+                for data in request_data
+            ]
+
+            return await asyncio.gather(*tasks)
+
         if request_data.protocol == "multipart-subscription":
             return await self.schema.subscribe(
                 request_data.query,  # type: ignore
@@ -206,14 +230,48 @@ class AsyncBaseHTTPView(
                 operation_name=request_data.operation_name,
             )
 
-        return await self.schema.execute(
-            request_data.query,
+        return await self.execute_single(
+            request=request,
+            request_adapter=request_adapter,
+            sub_response=sub_response,
+            context=context,
             root_value=root_value,
-            variable_values=request_data.variables,
-            context_value=context,
-            operation_name=request_data.operation_name,
-            allowed_operation_types=allowed_operation_types,
+            request_data=request_data,
         )
+
+    async def execute_single(
+        self,
+        request: Request,
+        request_adapter: AsyncHTTPRequestAdapter,
+        sub_response: SubResponse,
+        context: Context,
+        root_value: Optional[RootValue],
+        request_data: GraphQLRequestData,
+    ) -> ExecutionResult:
+        allowed_operation_types = OperationType.from_http(request_adapter.method)
+
+        if not self.allow_queries_via_get and request_adapter.method == "GET":
+            allowed_operation_types = allowed_operation_types - {OperationType.QUERY}
+
+        assert self.schema
+
+        try:
+            result = await self.schema.execute(
+                request_data.query,
+                root_value=root_value,
+                variable_values=request_data.variables,
+                context_value=context,
+                operation_name=request_data.operation_name,
+                allowed_operation_types=allowed_operation_types,
+            )
+        except InvalidOperationTypeError as e:
+            raise HTTPException(
+                400, e.as_http_error_reason(request_adapter.method)
+            ) from e
+        except MissingQueryError as e:
+            raise HTTPException(400, "No GraphQL query found in the request") from e
+
+        return result
 
     async def parse_multipart(self, request: AsyncHTTPRequestAdapter) -> dict[str, str]:
         try:
@@ -326,16 +384,12 @@ class AsyncBaseHTTPView(
                 return await self.render_graphql_ide(request)
             raise HTTPException(404, "Not Found")
 
-        try:
-            result = await self.execute_operation(
-                request=request, context=context, root_value=root_value
-            )
-        except InvalidOperationTypeError as e:
-            raise HTTPException(
-                400, e.as_http_error_reason(request_adapter.method)
-            ) from e
-        except MissingQueryError as e:
-            raise HTTPException(400, "No GraphQL query found in the request") from e
+        result = await self.execute_operation(
+            request=request,
+            context=context,
+            root_value=root_value,
+            sub_response=sub_response,
+        )
 
         if isinstance(result, SubscriptionExecutionResult):
             stream = self._get_stream(request, result)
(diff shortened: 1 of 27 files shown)


botberry:



---

## Add GraphQL Query batching support

GraphQL query batching is now supported across all frameworks (sync and async)
To enable query batching, add a valid `batching_config` to the schema configuration.

This makes your GraphQL API compatible with batching features supported by various
client side libraries, such as [Apollo GraphQL](https://www.apollographql.com/docs/react/api/link/apollo-link-batch-http) and [Relay](https://github.com/relay-tools/react-relay-network-modern?tab=readme-ov-file#batching-several-requests-into-one).

Example (FastAPI):

```py
import strawberry

from fastapi import FastAPI
from strawberry.fastapi import GraphQLRouter


@strawberry.type
class Query:
    @strawberry.field
    def hello(self) -> str:
        return "Hello World"


schema = strawberry.Schema(
    Query, config=StrawberryConfig(batching_config={"max_operations": 10})
)

graphql_app = GraphQLRouter(schema)

app = FastAPI()
```

Example (Flask):
```py
import strawberry

from flask import Flask
from strawberry.flask.views import GraphQLView

app = Flask(__name__)


@strawberry.type
class Query:
    @strawberry.field
    def hello(self) -> str:
        return "Hello World"


schema = strawberry.Schema(
    Query, config=StrawberryConfig(batching_config={"max_operations": 10})
)

app.add_url_rule(
    "/graphql/batch",
    view_func=GraphQLView.as_view("graphql_view", schema=schema),
)

if __name__ == "__main__":
    app.run()
```

Note: Query Batching is not supported for multipart subscriptions



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to @aryaniyaps for the PR 👏

GraphQL Query Batching support is here! 🍓 ✨ and it works across all frameworks (sync/async)!

```

tests/http/test_query_batching.py
```diff
@@ -0,0 +1,114 @@
...
+        ],
+        headers={"content-type": "application/json"},
+    )
+
+    assert response.status_code == 400
+
+    assert "Batching is not enabled" in response.text
+
+
+async def test_returns_error_for_multipart_subscriptions(
```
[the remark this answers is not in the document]
aryaniyaps: I've done that!

aryaniyaps:
sourcery review --diff "git diff main" .

strawberry/http/async_base_view.py
```diff
@@ -206,15 +228,49 @@ async def execute_operation(
...
-            allowed_operation_types=allowed_operation_types,
+            request_data=request_data,
         )
 
+    async def execute_single(
+        self,
+        request: Request,
+        request_adapter: AsyncHTTPRequestAdapter,
+        sub_response: SubResponse,
+        context: Context,
```
Object905, line 245: Seems like context is shared between all operations?
This may cause unexpected side effects. 
Although sometimes that may even be beneficial (for dataloaders etc).

There seems to be no way to isolate request, but it would be nice to have different context/root value per operation in my case.

aryaniyaps: I think that the context should be shared, reusing dataloaders and other resources must be one of the main factors behind adopting query batching, if we want separate context, we can use separate requests with batching disabled!

Object905: Well, maybe my use case not that common, because I use context not only for dataloaders, but also for some application logic and some tracing.
Since sharing is simpler - separate context might be a separate feature behind config flag.

aryaniyaps: > Well, maybe my use case not that common, because I use context not only for dataloaders, but also for some application logic and some tracing. Since sharing is simpler - separate context might be a separate feature behind config flag.

but Im curious @Object905 what benefits would batching bring if the context is not shared?

Object905: Reducing request count mostly to avoid network round-trips. 
For example one of my subgraphs is "external" and has ping of about 50ms. But even for "local" sub-graphs this might be noticeable.

Also there are other means to share global data (request itself, thread locals, etc.) besides context.

Object905: Also in one service I use context for application logic that's dependent on query input parameters - and different context for different queries might break this completely.

aryaniyaps: > Reducing request count mostly to avoid network round-trips. For example one of my subgraphs is "external" and has ping of about 50ms. But even for "local" sub-graphs this might be noticeable.
> 
> Also there are other means to share global data (request itself, thread locals, etc.) besides context.

Sharing user specific data using thread locals would be a real pain IMO.

Can't the problem you're describing be solved by reusing the same context and using a dataloader? That's the usual pattern while doing data fetching with GraphQL..

Object905: I am using dataloaders. And I intend to share them for batching (through request most likely and context will get shared dataloaders from request).

But application logic that I have in context - just can't be shared, because it's dependent on inputs, session, query itself, etc.
It can be separated manually (custom context, that hashes current operation/variables and stores different context for each one) - but this will be flaky and much better supported by strawberry itself.

aryaniyaps: I'm unsure about this, though open-minded, could we get opinions from the maintainers regarding this? What you're suggesting seems like an anti-pattern to me, but it's probably because I haven't used context in that way so far..

aryaniyaps:
@sourcery-ai review

strawberry/http/async_base_view.py
erikwrede: After thinking about it, I see advantages on both sides, so I'm slightly leaning towards making context sharing configurable for batching. Here's an example where the entire thing breaks though: 

**GraphQL Schema:**
```graphql
directive @IN_Context(language: String) on QUERY

type Query {
  product(id: ID!): Product
}

type Product {
  id: ID!
  name: String!
  description: String
}
```

**GraphQL Query Example:**
```graphql
query GetProduct @IN_Context(language: "en") {
  product(id: "1") {
    id
    name
    description
  }
}
```

Shopify uses this directive to provide localization. A common procedure would involve adding the language inside that directive into context. Now consider two batched queries with different languages and a shared context. One of those languages will be in context, hence, the other response will be broken. 

I'm sure we can come up with many other cases where this might cause unexpected behavior.

aryaniyaps: @erikwrede got it, creating a separate context makes more sense to me now. I'll work on making the context sharing configurable

aryaniyaps: I've been trying to implement this @erikwrede however I feel like I've hit a roadblock.

I cannot just get a new context like this for every request while using FastAPI:

```py
        if isinstance(request_data, list):
            # batch GraphQL requests
            if not self.schema.config.batching_config["share_context"]:
                tasks = [
                    self.execute_single(
                        request=request,
                        request_adapter=request_adapter,
                        sub_response=sub_response,
                        # create a new context for each request data
                        context=await self.get_context(request, response=sub_response),
                        root_value=root_value,
                        request_data=data,
                    )
                    for data in request_data
                ]
```

That's because the FastAPI context getter is deeply integrated with the FastAPI DI system, and is run only once per request. what should be done for this use case?

aryaniyaps: also, I was trying to create a test case, replicating shopify's context driven translation.

I tried to create QUERY level schema directives like this but it wasn't working.

```py
@strawberry.directive(locations=[DirectiveLocation.QUERY], name="in_context")
def in_context(language: str, info: strawberry.Info):
    # put the language in the context here
    print("in_context called")
    info.context.update({"language": language})
```

am I doing this correctly? I wasn't able to find any documentation/ test cases which were using this type of directive

erikwrede: let me tag @DoctorJohn for a second opinion on the FastAPI Issue, he's done quite a bit of web view work recently.

Regarding the shopify example, this is not something we have as a test case in strawberry. I was just referring to it as an example where multiple operations could cause clashes in context. To be fair, multiple queries in a single GraphQL Document are also a valid request that could cause clashes. In general, I still think this is a trade off worth making. As mentioned on Discord, I'll look further into the problem with directives and get back.

erikwrede: In case this is a blocker for too long, I'd be open to splitting this into 2 separate PRs as well. One that adds batching support, and the other adding an option to share or not share context.

Object905: Sure. I can't give the full code, but here is the gist of it.

```
class Query:
    @strawberry.field
    def location_terms(input: LocationInfoInput, info: StrawInfo) -> LocationTerms:
        input = info.context.reduced_location_input(input, fetch_terms=False)
        ...
```

This is an "entry point" into location api. It can't handle multiple addresses in a single query by design (due to context sharing, etc.).
`reduced_location_input` - might be quite heavy, because it goes through geocoding and a lot of db queries, so doing it twice in a single request is a no-no.

Also in context  there are other "heavy" objects that depend on location and building them a second time during query is heavy perf hit.

aryaniyaps:
hello everyone, since this has been a blocker for quite some time, I've updated the code to raise ValueErrors when context sharing is disabled. Could this PR be merged to add batching support with context sharing?

I'll open a new PR for disabling context sharing!

patrick91:
@bellini666 let's chat about this at PyCon!

strawberry/http/sync_base_view.py
```diff
@@ -152,6 +216,18 @@ def parse_http_body(self, request: SyncHTTPRequestAdapter) -> GraphQLRequestData
         else:
             raise HTTPException(400, "Unsupported content type")
 
+        if isinstance(data, list):
```
bellini666, line 220: Q: Shouldn't we be running the checks below for each item as well?

strawberry/http/async_base_view.py
patrick91: We discussed this with @DoctorJohn and the conclusion was that that's better to have the context based on the request/connection, not on the operation.

We might revisit this though, but it is how it works in subscriptions already 😊

strawberry/http/async_base_view.py
```diff
@@ -201,6 +207,22 @@ async def execute_operation(
         if not self.allow_queries_via_get and request_adapter.method == "GET":
             allowed_operation_types = allowed_operation_types - {OperationType.QUERY}
 
+        if isinstance(request_data, list):
+            # batch GraphQL requests
+            return await asyncio.gather(
+                *[
+                    self.execute_single(
```
bellini666, line 215: Maybe `asyncio.create_task` in here to make sure any [contextvar](https://docs.python.org/3/library/contextvars.html) set by those diverge if changed, and one operation doesn't affect the other?

patrick91 (approved):
Dope! Thank you @aryaniyaps and @DoctorJohn

strawberry/http/async_base_view.py
DoctorJohn: Isn't it sufficient that `gather` does automatically scheduling each coro as a task? :eyes: Ideally we would add a test for it. But I'm working from my super slow laptop and it's such a pain. I could do it back at home.

strawberry/http/sync_base_view.py
DoctorJohn: I thought about it. Technically it's not covered by the batching spec, but yes, it would make sense. (something something, laptop is super slow, I'll do it at home, because we want to merge this before the end of EuroPython). welp

PR #3766, opened 2025-02-03
Support aliases in get_specialized_type_var_map

alexey-pelykh:
## Description

Support aliases (TypeVar passthrough) in `get_specialized_type_var_map`:

```python
def test_get_specialized_type_var_map_double_generic_passthrough():
    @strawberry.type
    class Foo(Generic[_T]): ...

    @strawberry.type
    class Bar(Foo[_K], Generic[_K]): ...

    @strawberry.type
    class Bin(Bar[int]): ...

    assert get_specialized_type_var_map(Bin) == {
        "_T": int,
        "_K": int,
    }
```

Without this change, errors like

```
KeyError: '_T'
```

will appear

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

## Summary by Sourcery

Support TypeVar passthrough in `get_specialized_type_var_map`.

Enhancements:
- Resolve type variables correctly when a generic type inherits from another generic type, passing the type variable through.

Tests:
- Add a test case for TypeVar passthrough with double generics.

### DIFF
--- a/strawberry/utils/inspect.py
+++ b/strawberry/utils/inspect.py
@@ -4,13 +4,16 @@ from functools import lru_cache
 from typing import (
     Any,
     Callable,
+    Generic,
     Optional,
+    Protocol,
     TypeVar,
+    Union,
+    get_args,
     get_origin,
 )
-from typing_extensions import get_args
 
-from strawberry.utils.typing import is_generic_alias
+import strawberry
 
 
 def in_async_context() -> bool:
@@ -67,13 +70,13 @@ def get_specialized_type_var_map(cls: type) -> Optional[dict[str, type]]:
     # {}
 
     get_specialized_type_var_map(Bar)
-    # {~T: ~T}
+    # {}
 
     get_specialized_type_var_map(IntBar)
-    # {~T: int}
+    # {~T: int, ~K: int}
 
     get_specialized_type_var_map(IntBarSubclass)
-    # {~T: int}
+    # {~T: int, ~K: int}
 
     get_specialized_type_var_map(IntBarFoo)
     # {~T: int, ~K: str}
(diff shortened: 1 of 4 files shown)


botberry:



---

Support aliases (TypeVar passthrough) in `get_specialized_type_var_map`.



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Alexey Pelykh for the PR 👏

```

strawberry/utils/inspect.py
```diff
@@ -67,57 +70,60 @@ class IntBarFoo(IntBar, Foo[str]): ...
...
+                if type_param not in param_args:
+                    param_args[type_param] = arg
+
+        if orig_bases := getattr(origin, "__orig_bases__", None):
+            types.extend(orig_bases)
+    if not param_args:
+        return None
+
+    resolve = True
+    while resolve:
```
bellini666, line 111: **question:** why do we need this `resolve` and the while in here?

strawberry/utils/inspect.py
```diff
@@ -67,57 +70,60 @@ class IntBarFoo(IntBar, Foo[str]): ...
     # {}
 
     get_specialized_type_var_map(Bar)
-    # {~T: ~T}
```
bellini666, line 70: **question:** this was wrong in the documentation, right?

strawberry/utils/inspect.py
```diff
@@ -67,57 +70,60 @@ class IntBarFoo(IntBar, Foo[str]): ...
...
             continue
 
-        type_var_map.update(
-            {p.__name__: a for p, a in zip(params, args) if not isinstance(a, TypeVar)}
-        )
-
-    return type_var_map
+        if (type_params := getattr(origin, "__parameters__", None)) is not None:
+            args = get_args(tp)
+            for type_param, arg in zip_longest(type_params, args):
```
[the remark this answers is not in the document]
alexey-pelykh: It's more of the convenience for getting None
```python
from typing import TypeVar, Generic, get_args, get_origin
from itertools import zip_longest

OneT = TypeVar("OneT")
TwoT = TypeVar("TwoT")
ThreeT = TypeVar("TwoT")

class Base(Generic[TwoT, OneT]):
    pass

type_params = getattr(Base, "__parameters__")
print(f"type_params = {type_params}")

args = get_args(get_origin(Base))
print(f"args = {args}")

print(list(zip_longest(type_params, args)))
```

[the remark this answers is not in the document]
alexey-pelykh: As for the order, `__args__` and `__parameters__` indeed should have same ordering

strawberry/utils/inspect.py
alexey-pelykh: To resolve aliases as in example. On first pass it's:
```
_T: _K,
_K": int,
```
so that needs to be resolved further

strawberry/utils/inspect.py
alexey-pelykh: Yep

alexey-pelykh: The comment contradicts `test_get_specialized_type_var_map_generic_not_specialized` that covers exactly this case

strawberry/utils/inspect.py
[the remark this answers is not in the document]
bellini666: Ahh ok. Maybe a comment here would be worth for anyone checking this in the future?

strawberry/utils/inspect.py
bellini666: Maybe this could be rewritten to something like this:

```python
for type_param, arg in list(param_args.items()):
    while isinstance(arg, TypeVar):
        arg = param_args.get(arg) if arg is not type_param else None

    param_args[type_param] = resolved_arg
```

Haven't thought if there are corner cases in this simplified code, but wdyt?

alexey-pelykh: 👍 Updated with use of sentinel instead of None

strawberry/utils/inspect.py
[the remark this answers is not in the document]
alexey-pelykh: I've updated the code to be more explicit - I really don't think there can be a valid case when args are of different length that parameters - they are either there or not at all, as in - unspecified generic

bellini666 (changes requested):
Just one small adjustment and it should be good to go! 😊

strawberry/utils/inspect.py
```diff
@@ -67,57 +69,83 @@ class IntBarFoo(IntBar, Foo[str]): ...
...
 
-    orig_bases = getattr(cls, "__orig_bases__", None)
-    if orig_bases is None:
-        # Specialized generic aliases will not have __orig_bases__
-        if get_origin(cls) is not None and is_generic_alias(cls):
-            orig_bases = (cls,)
-        else:
-            # Not a specialized type
-            return None
+    class Unresolved:  # pragma: no cover
```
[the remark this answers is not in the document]
alexey-pelykh: done

bellini666 (approved):

PR #3773, opened 2025-02-10
Improve performance of `convert_argument` for list of primitives

blothmann:
Improve performance of `convert_argument` for list types that do not need any conversion of its arguments by just returning a copy of the list. 

## Description

This has a drastical improvement of the performance if the input list is large. In my use case under some circumstances I am handling lists of hundred thousands of elements. Instead of iterating through each element, checking some conditions and returning the element, just return a copy of the list is a huge performance improvement.

Since tests cover everything that I have changed already, I added a performance test for large lists of integers.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

## Summary by Sourcery

Optimize argument conversion for lists of primitive types by directly returning a copy of the list if no type conversion is required.

Enhancements:
- Improve performance of `convert_argument` for large lists of primitive types.

Tests:
- Add performance tests for converting large lists of integers.


botberry:



---

This release improves performance of argument conversion for lists of primitives.



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to blothmann for the PR 👏

```

strawberry/types/arguments.py
```diff
@@ -135,6 +135,31 @@ def is_graphql_generic(self) -> bool:
         return is_graphql_generic(self.type)
 
 
+def _convert_argument_value_can_be_returned(
```
patrick91, line 138: can we find a different name for this? :D I don't think it is easy to understand (maybe we can add a doc string too)?

Great stuff btw 😊

blothmann: We are looking for a scalar or enum or an optional version of those, which can simply be returned in convert_argument. So maybe something like `_is_primitive`?

blothmann:
Tried to make the function naming more clear to determine in which case a copy of the list can be returned: if item type is a leaf (`_is_leaf_type`) or an optional leaf (`_is_optional_leaf_type`). Same naming (`is_leaf_type`) is also used in graphql-core library.
Added test case for optional leaf elements.

strawberry/types/arguments.py
```diff
@@ -164,22 +204,27 @@ def convert_argument(
 
     if isinstance(type_, StrawberryList):
         value_list = cast("Iterable", value)
+
+        if _is_leaf_type(
+            type_.of_type, scalar_registry, skip_classes=(GlobalID,)
+        ) or _is_optional_leaf_type(
+            type_.of_type, scalar_registry, skip_classes=(GlobalID,)
+        ):
+            return list(value_list)
```
patrick91, line 213: @blothmann do you remember why we make a copy here? do we need it? 😊

blothmann: @patrick91 actually, I am not sure. It is a safer option, since the previous code also made a copy. And at least here we only know that `value_list` is an `Iterable`. However, I think it is provided as a list from the graphql-core library.

patrick91: I'll try removing it later! might not be needed at all 😊

tests/benchmarks/test_arguments.py
```diff
@@ -0,0 +1,21 @@
...
+    test_value = list(range(ntypes))
+    type_ = StrawberryList(int)
+
+    def run():
+        result = convert_argument(
+            test_value, type_, DEFAULT_SCALAR_REGISTRY, StrawberryConfig()
+        )
+        assert test_value == result
+
+    benchmark(run)
```
bellini666, line 21: **suggestion:** maybe we can add this in another PR and base, merge, and then base this one on top of it to check the actual performance improvements we are getting?

strawberry/types/arguments.py
bellini666: FYI checked the code here, doesn't seem to be needed. Also removed the `list`, ran the tests, all passed

bellini666 (commented):
Left a suggestion, in general LGTM. Will leave it up to @patrick91 to give the final approval

botberry:

Issue #3777, opened 2025-02-11
Memory leak in subscriptions


jakub-bacic:
Release 0.240.0 introduced a regression in GQL subscriptions. When WS connection is closed abruptly, subscription coroutine isn't properly finished causing memory leaks. The problem does NOT occur when the subscription ends on its own or when the connection is closed by the client-side.

## Reproduction steps

Please find the attached snippets to reproduce the issue.

```python
# test_server.py

import asyncio
import logging
import uuid
from typing import AsyncGenerator

import strawberry
from fastapi import FastAPI
from strawberry import Schema
from strawberry.fastapi import GraphQLRouter

logging.basicConfig(level=logging.INFO, format="[%(asctime)s] %(message)s")
logger = logging.getLogger()


@strawberry.type
class Query:
    """
    Just a stub because strawberry requires non-empty Query type.
    """
    @strawberry.field()
    def active_subscriptions(self) -> int:
        return len(active_subscriptions)


# some globally available set to count active subscriptions
active_subscriptions = set()


@strawberry.type
class Subscription:
    @strawberry.subscription()
    async def test(self) -> AsyncGenerator[str, None]:
        # add subscription to the set to track active connections
        subscription_id = uuid.uuid4()
        active_subscriptions.add(subscription_id)

        logger.info(f"[{subscription_id}] Started new subscription (active = {len(active_subscriptions)})")

        try:
            # return a single value after 5 seconds and finish the subscription
            await asyncio.sleep(5)
            yield "finished"
            logger.info(f"[{subscription_id}] Returning value after 5 seconds")
        finally:
            # remove subscription from the set
            active_subscriptions.remove(subscription_id)
            logger.info(f"[{subscription_id}] Subscription finished (active = {len(active_subscriptions)})")


schema = Schema(
    query=Query,
    subscription=Subscription,
)

app = FastAPI()
app.include_router(
    GraphQLRouter(
        schema,
    ),
    prefix="/graphql",
)

```

```python
# test_client.py

import asyncio
import sys
import time
from multiprocessing import Process

from gql import gql, Client
from gql.transport.websockets import WebsocketsTransport


transport = WebsocketsTransport(
    url="ws://localhost:8000/graphql",
)


async def start_listening():
    async with Client(
        transport=transport,
        fetch_schema_from_transport=False,
    ) as session:
        query = gql("""
            subscription onTest {
                test
            }
        """)

        print("Connecting")
        async for msg in session.subscribe(query):
[...]

PR #3778, opened 2025-02-11
Fix cleanup logic in graphql_ws protocol handler

jakub-bacic:
## Description

Ensure all tasks are canceled when performing cleanup in subscription handlers when `graphql_ws` protocol is used.

## Types of Changes

- [ ] Core
- [X] Bugfix
- [ ] New feature
- [ ] Documentation

* fixes https://github.com/strawberry-graphql/strawberry/issues/3777

## Checklist

## Summary by Sourcery

Fix cleanup logic for GraphQL subscriptions using the `graphql_ws` protocol to ensure all tasks are cancelled during cleanup.

Bug Fixes:
- Fixed a bug where subscription tasks were not cancelled when cleaning up in `graphql_ws` protocol handler.

Tests:
- Added tests to verify that all active subscriptions are cancelled when a client disconnects unexpectedly.

### DIFF
--- a/strawberry/subscriptions/protocols/graphql_ws/handlers.py
+++ b/strawberry/subscriptions/protocols/graphql_ws/handlers.py
@@ -76,8 +76,7 @@ class BaseGraphQLWSHandler(Generic[Context, RootValue]):
                 with suppress(BaseException):
                     await self.keep_alive_task
 
-            for operation_id in list(self.subscriptions.keys()):
-                await self.cleanup_operation(operation_id)
+            await self.cleanup()
 
     async def handle_message(
         self,
@@ -201,6 +200,10 @@ class BaseGraphQLWSHandler(Generic[Context, RootValue]):
         with suppress(BaseException):
             await self.tasks[operation_id]
         del self.tasks[operation_id]
+
+    async def cleanup(self) -> None:
+        for operation_id in list(self.tasks.keys()):
+            await self.cleanup_operation(operation_id)
 
     async def send_data_message(
         self, execution_result: ExecutionResult, operation_id: str
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3574,23 +3574,23 @@ the attributes of the `IgnoreContext` class.
 For example, the following query:
 ```python
 """
-    query {
-      matt: user(name: "matt") {
-        email
-      }
-      andy: user(name: "andy") {
-        email
-        address {
-          city
-        }
-        pets {
-          name
-          owner {
-            name
-          }
-        }
+query {
+  matt: user(name: "matt") {
+    email
+  }
+  andy: user(name: "andy") {
+    email
+    address {
+      city
+    }
+    pets {
+      name
+      owner {
+        name
       }
     }
+  }
+}
 """
 ```
 can have its depth limited by the following `should_ignore`:
@@ -3607,17 +3607,17 @@ query_depth_limiter = QueryDepthLimiter(should_ignore=should_ignore)
 so that it *effectively* becomes:
 ```python
 """
-    query {
-      andy: user(name: "andy") {
-        email
-        pets {
-          name
-          owner {
-            name
-          }
-        }
+query {
+  andy: user(name: "andy") {
+    email
+    pets {
+      name
+      owner {
+        name
       }
     }
+  }
+}
 """
 ```
 
@@ -10702,7 +10702,7 @@ from typing import Annotated
 class Query:
     @strawberry.field
     def user_by_id(
-        id: Annotated[str, strawberry.argument(description="The ID of the user")]
+        id: Annotated[str, strawberry.argument(description="The ID of the user")],
     ) -> User: ...
 ```
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+This release fixes the issue that some subscription resolvers were not canceled if a client unexpectedly disconnected.
--- a/docs/types/exceptions.md
+++ b/docs/types/exceptions.md
@@ -64,7 +64,7 @@ def name(
         str,
         strawberry.argument(description="This is a description"),
         strawberry.argument(description="Another description"),
-    ]
+    ],
 ) -> str:
     return "Name"
 
--- a/noxfile.py
+++ b/noxfile.py
@@ -150,7 +150,7 @@ def test_pydantic(session: Session, pydantic: str, gql_core: str) -> None:
 
 @session(python=PYTHON_VERSIONS, name="Type checkers tests", tags=["tests"])
 def tests_typecheckers(session: Session) -> None:
-    session.run_always("poetry", "install", external=True)
+    session.run_always("poetry", "install", "--with", "integrations", external=True)
 
     session.install("pyright")
     session.install("pydantic")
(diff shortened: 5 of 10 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Jakub Bacic for the PR 👏

```

nrbnlulu (commented):
Thanks for the fix. 
I don't think the tests really proves something tho.
you never assert
`assert Subscription.active_echo_subscriptions > 0` 

I think the way to go is the to add another query for subscription count, then query for that before and after the subscription.

jakub-bacic:
The test fails on that assertion if you revert the fix in cleanup logic so I think it does prove the changes are needed 😅. 

Adding additional GQL query just for checking that sounds a little bit extreme to me but it's definitely doable if we want to go that way. Alternatively, we can include two assertions (before and after `ws.close`) like this:

```diff
--- a/tests/websockets/test_graphql_ws.py
+++ b/tests/websockets/test_graphql_ws.py
@@ -806,12 +806,16 @@ async def test_unexpected_client_disconnects_are_gracefully_handled(
                 "type": "start",
                 "id": "sub1",
                 "payload": {
-                    "query": 'subscription { echo(message: "Hi", delay: 0.5) }',
+                    "query": 'subscription { echo(message: "Hi", delay: 2) }',
                 },
             }
         )
 
+        await asyncio.sleep(0.5)
+        assert Subscription.active_echo_subscriptions == 1
+
         await ws.close()
-        await asyncio.sleep(1)
+        await asyncio.sleep(0.5)
+
         assert not process_errors.called
         assert Subscription.active_echo_subscriptions == 0
```

Any thoughts?

jakub-bacic:
The original test also relies on `sleep` function but I see your point - that change won't make it better.

If I understood you correctly, your point was that if we assert `active_echo_subscription == 0` at the end of the test without checking if it's `> 0` in the middle, we risk false positive (cause the initial/default value is also `0`)? If so, the simplest solution to that problem is to replace "active subscription counter" with "full log of lifecycle events":

```python
class Subscription:
     echo_subscription_calls = []

    @strawberry.subscription
    async def echo(self, message: str, delay: float = 0) -> AsyncGenerator[str, None]:
        try:
            Subscription.echo_subscription_calls.append(("before", message, delay))
            await asyncio.sleep(delay)
            yield message
        finally:
            Subscription.echo_subscription_calls.append(("after", message, delay))
            
...
# in tests

assert Subscription.echo_subscription_calls == [
  ("before", "Hi", 0.5),
  ("after", "Hi", 0.5),
]
```

Other attempts I tried:
- patch `echo` subscription resolver so we don't have to modify `Subscription` class
  - `mock.patch.object("Subscription", "echo", func)` didn't have any effects when applied in the test
- use `on_operation` hook in extension to track subscription lifecycle
  - when I added that hook to `MyExtension`, it seems to be executed only for `ChannelsHttpClient` (is that a known issue?)

strawberry/subscriptions/protocols/graphql_ws/handlers.py
```diff
@@ -202,6 +201,10 @@ async def cleanup_operation(self, operation_id: str) -> None:
             await self.tasks[operation_id]
         del self.tasks[operation_id]
 
+    async def cleanup(self) -> None:
+        for operation_id in list(self.tasks.keys()):
```
bellini666, line 205: **question:** is it ok to just do this for `self.tasks` and not for `self.subscriptions` anymore? Wondering if the correct thing would be to do:

```suggestion
        for operation_id in set(self.tasks) | set(self.subscriptions):
```

jakub-bacic: As far as I can see, each subscription has a matching task with the same `operation_id`. However, the opposite isn't true - not every task has a matching subscription entry (that was the root cause of this issue and why a simple switch from `self.subscriptions` to `self.tasks` fixed that).

Task is added in `handle_start`:
```python
        result_handler = self.handle_async_results(
            operation_id, query, operation_name, variables
        )
        self.tasks[operation_id] = asyncio.create_task(result_handler)
```

Meanwhile, the subscription entry is added later in `handle_async_results`, only after receiving the first value from the asynchronous generator:

```python
            agen_or_err = await self.schema.subscribe(
                query=query,
                variable_values=variables,
                context_value=self.context,
                root_value=self.root_value,
            )
            if isinstance(agen_or_err, PreExecutionError):
                assert agen_or_err.errors
                await self.send_message(
                    {
                        "type": "error",
                        "id": operation_id,
                        "payload": agen_or_err.errors[0].formatted,
                    }
                )
            else:
                self.subscriptions[operation_id] = agen_or_err
```

That's why if cleanup is executed before the subscription yields any value, the task will have no matching subscription yet.

Additionally, it's safe to call `cleanup_operation` for each task because it checks internally if the given operation exists in `self.subscriptions` before attempting to delete it.

bellini666:
Other than https://github.com/strawberry-graphql/strawberry/pull/3778#discussion_r1953053676, which has already been answered, I think in general this LGTM

@nrbnlulu any other considerations?

@patrick91 @DoctorJohn, do you want to take a look as well?

nrbnlulu:
test case still doesn't prove anything I had my 2 cents on how the test should be implemented. 
bug ig how @jakub-bacic suggested is also possible but a bit hacky IMO.

jakub-bacic:
> test case still doesn't prove anything... I had my 2 cents on how the test should be implemented. bug ig how @jakub-bacic suggested is also possible but a bit hacky IMO.

I wouldn’t say it proves nothing - the test fails on the original code but passes now, meaning it at least reproduces the issue.

Adding a GQL query just to fetch the number of active subscriptions feels like overkill, and I’m not sure it will resolve the issue 🤔. Won’t it still be a race condition? Do we have a guarantee that the echo subscription increments the counter before the query fetches its value?

I can revisit this next Wednesday. I’ll explore a few alternatives and keep you posted :)

nrbnlulu:
> > test case still doesn't prove anything... I had my 2 cents on how the test should be implemented. bug ig how @jakub-bacic suggested is also possible but a bit hacky IMO.
> 
> I wouldn’t say it proves nothing - the test fails on the original code but passes now, meaning it at least reproduces the issue.
> 
> Adding a GQL query just to fetch the number of active subscriptions feels like overkill, and I’m not sure it will resolve the issue 🤔. Won’t it still be a race condition? Do we have a guarantee that the echo subscription increments the counter before the query fetches its value?
> 
> Is there anything other than the testing part you're concerned about?
> 
> I can revisit this next Wednesday. I’ll explore a few alternatives and keep you posted :)

Thats just my personal opinion.. I prefer query since the schema could (*theoretically*) be served in another process / thread).. but as I said thats just my non-obligating opinion...

jakub-bacic:
I pushed the improved version. I used `infinity` subscription instead of `echo` so we can actually wait for the first value returned from it (since an extra `await ws.receive_json()` line) to deterministically assert the value of `active_infinity_subscriptions` (before and after as suggested by @nrbnlulu).

LMK if that looks good enough now :)

nrbnlulu (approved):

patrick91 (approved):
Nice work, thank you!

botberry:

PR #3786, opened 2025-02-17
opentelemetry: use semconv attributes

### LINKED ISSUE #3769: opentelemetry: use attributes from semconv registry
## Feature Request Type

- [ ] Core functionality
- [ ] New behavior

## Description

This blob:

https://github.com/strawberry-graphql/strawberry/blob/2d4234a70e817e726288cb30ce7d0523b337420e/strawberry/extensions/tracing/opentelemetry.py#L53-L68

should probably use `graphql.document`, `graphql.operation.name`, and `graphql.operation.type`, as suggested by upstream: https://opentelemetry.io/docs/specs/semconv/attributes-registry/graphql/

it also appears `component` has been deprecated: https://github.com/open-telemetry/opentelemetry-specification/issues/271

need to know:

* if a PR from the community to implement this is desirable (happy to do this if so)
* if the span names need to be unique (i.e., is the operation name in _both_ the span name and `graphql.operation.name`?)

guppy0130:
opentelemetry: use semconv attributes

## Description

* use attributes defined in opentelemetry's semconv
* stop setting span attribute `component`, as that's been deprecated
* fixes #3769

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist


botberry:



---

Use OTEL's semconv attribute registry for span attributes. Stops setting `component`, which has been deprecated.



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Nick Yang for the PR 👏

```

strawberry/extensions/tracing/opentelemetry.py
```diff
@@ -76,6 +82,22 @@ def on_operation(self) -> Generator[None, None, None]:
...
             span_name = f"GraphQL Query: {self.execution_context.operation_name}"
             self._span_holder[LifecycleStep.OPERATION].update_name(span_name)
+            self._span_holder[LifecycleStep.OPERATION].set_attribute(
+                graphql_attributes.GRAPHQL_OPERATION_NAME,
+                self.execution_context.operation_name,
+            )
+
+        # likewise for the operation type; we'll know it for sure after parsing.
+        # note that this means ``self.execution_context.operation_type`` must
+        # be kept in sync with ``graphql_attributes.GraphqlOperationTypeValues``.
```
erikwrede, line 92: Generally speaking I'd prefer a mapper in these situations instead of a comment. However, the operation type is actually part of the GraphQL Spec, so it should be highly unlikely to go out of sync.  
I think you can drop the comment entirely 😊

strawberry/extensions/tracing/opentelemetry.py
```diff
@@ -76,6 +82,22 @@ def on_operation(self) -> Generator[None, None, None]:
...
+        if self.execution_context.operation_type:
+            self._span_holder[LifecycleStep.OPERATION].set_attribute(
+                graphql_attributes.GRAPHQL_OPERATION_TYPE,
+                graphql_attributes.GraphqlOperationTypeValues(
+                    self.execution_context.operation_type.value.lower()
```
erikwrede, line 97:
```
class OperationType(Enum):

    QUERY = "query"
    MUTATION = "mutation"
    SUBSCRIPTION = "subscription"
```
The enum value is already lowercase.

erikwrede (changes requested):
I really like the idea of adjusting our OTEL support to the standards. However, I'd prefer a gradual transition over to a new OTEL v2 extension instead of changing the attribute names. We do not want to break any metrics or alerts in production relying on the old attributes. Putting out a breaking change for this seems excessive, so I'd favor a new OTEL extension while a deprecation warning is printed every time the old extension is used. Maybe we can already add a scheduled removal date. Pinging @patrick91 for a second opinion on this though, as usually we just release breaking changes. I just think the workaround for users relying on OTEL here is cumbersome :D

patrick91:
> I really like the idea of adjusting our OTEL support to the standards. However, I'd prefer a gradual transition over to a new OTEL v2 extension instead of changing the attribute names. We do not want to break any metrics or alerts in production relying on the old attributes. Putting out a breaking change for this seems excessive, so I'd favor a new OTEL extension while a deprecation warning is printed every time the old extension is used. Maybe we can already add a scheduled removal date. Pinging @patrick91 for a second opinion on this though, as usually we just release breaking changes. I just think the workaround for users relying on OTEL here is cumbersome :D

this makes a lot of sense, not sure if another extension or a configuration in the extension, I think I'd prefer passing a flag, I don't think I want a OpenTelemetryV2Extension :D

PR #3812, opened 2025-03-14
Change pydantic conversion to not load field data unless requested

Mark90:
Changes the Pydantic model conversion so that it doesn't retrieve data for fields that aren't going to be used.

## Description

It's been 8 months since I manually patched this in our own codebase, I wanted to create a PR here after some time when I was sure it didn't cause any regressions. 

Without this patch I observed unnecessary requests being made which originated from fields on "Pydantic strawberry types" that weren't requested in the query. Moving these lines inside the `if field.init:` section appeared sufficient to fix this.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+Change pydantic conversion to not load field data unless requested
--- a/strawberry/experimental/pydantic/conversion.py
+++ b/strawberry/experimental/pydantic/conversion.py
@@ -83,14 +83,14 @@ def convert_pydantic_model_to_strawberry_class(
         field = cast("StrawberryField", field_)
         python_name = field.python_name
 
-        data_from_extra = extra.get(python_name, None)
-        data_from_model = (
-            getattr(model_instance, python_name, None) if model_instance else None
-        )
-
         # only convert and add fields to kwargs if they are present in the `__init__`
         # method of the class
         if field.init:
+            data_from_extra = extra.get(python_name, None)
+            data_from_model = (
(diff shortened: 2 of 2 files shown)


botberry:



---

Change pydantic conversion to not load field data unless requested



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Mark Moes for the PR 👏

```

Mark90:
If tests to measure the performance difference are required I'll be happy to supply them; I haven't delved into the testsuite deep enough yet to know how you typically do this this. 

I just wanted to get this PR up before I forget to do so again. :)

patrick91:
@Mark90 thanks!

> If tests to measure the performance difference are required I'll be happy to supply them; I haven't delved into the testsuite deep enough yet to know how you typically do this.

They aren't really needed, unless you think there was a big change!

patrick91:
@Mark90 but it would be nice to have some test to check the behaviour is what we expect now 😄

Mark90:
> @Mark90 but it would be nice to have some test to check the behaviour is what we expect now 😄

Done!

The added test fails without the change in `conversion.py`, and passes with the change.

The test uses `root._original_model.name` in the resolver, which is probably not recommended usage, but this is not even necessary to reproduce the problem. E.g. if the `get_name()` resolver is changed to `return "foo"` and the change in `conversion.py` reverted, then the name computed_field is still accessed somewhere during the conversion.

botberry:

PR #3828, opened 2025-04-01
Add support for Python 3.14

patrick91:
## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

## Summary by Sourcery

Add support for Python 3.14 by updating test environments, applying compatibility fixes, and publishing a release announcement.

Bug Fixes:
- Include the required `doc` keyword for dataclass fields on Python 3.14

Enhancements:
- Add an optional codeflash dependency group in pyproject.toml
- Bump graphql-core version requirement to 3.2.6

CI:
- Update noxfile and GitHub Actions workflow to include Python 3.14 in the test matrix

Documentation:
- Add TWEET.md to announce the new release with Python 3.14 support


botberry:


<table>
	<thead>
		<tr>
			<th>Federation 1 Support</th>
			<th>Federation 2 Support</th>
		</tr>
	</thead>
	<tbody>
    <tr>
      <td>
        <table>
        </table>
      </td>
      <td>
        <table>
        </table>
      </td>
		</tr>
	</tbody>
</table>


Learn more:

bellini666:
It seems that we have 2 issues here:

1) `TypeError: Field.__init__() missing 1 required positional argument: 'doc'`

the new [field api](https://docs.python.org/3.14/library/dataclasses.html#dataclasses.field) has an optional doc keyword, which we need to add there.

Wondering if we should add `doc` and deprecate `description`? Wdyt? As `field` has `doc` now, it seems a good kw to retrieve it =P

2) `ImportError: cannot import name 'ByteString' from 'typing'`

This is an issue from graphql-core, which has been fixed for [3.3.x](https://github.com/graphql-python/graphql-core/issues/211).

Should we mark `3.14` only available with graphql-core 3.3.x+? Otherwise we might need to ask Cito to backport the fix to 3.2.x

patrick91:
> It seems that we have 2 issues here:
> 
> 1. `TypeError: Field.__init__() missing 1 required positional argument: 'doc'`
> 
> the new [field api](https://docs.python.org/3.14/library/dataclasses.html#dataclasses.field) has an optional doc keyword, which we need to add there.
> 
> Wondering if we should add `doc` and deprecate `description`? Wdyt? As `field` has `doc` now, it seems a good kw to retrieve it =P

Mmh, I don't really like doc :D I think I'd keep the same behaviour as it is now (so just passing None)

> 
> 2. `ImportError: cannot import name 'ByteString' from 'typing'`
> 
> This is an issue from graphql-core, which has been fixed for [3.3.x](https://github.com/graphql-python/graphql-core/issues/211).
> 
> Should we mark `3.14` only available with graphql-core 3.3.x+? Otherwise we might need to ask Cito to backport the fix to 3.2.x

I'll make a PR to GraphQL-core, not sure when 3.3 will be released (I think it depends on GraphQL js)

botberry:



---

This release adds support for the upcoming Python 3.14



Here's the tweet text: 
```

This release adds support for Python 3.14!

```

strawberry/types/object_type.py
```diff
@@ -51,7 +51,8 @@ def _check_field_annotations(cls: builtins.type[Any]) -> None:
 
     https://github.com/python/cpython/blob/6fed3c85402c5ca704eb3f3189ca3f5c67a08d19/Lib/dataclasses.py#L881-L884
     """
-    cls_annotations = cls.__dict__.get("__annotations__", {})
+    cls_annotations = get_annotations(cls)
+    # TODO: do we need this?
```
bellini666, line 55: If I'm not mistaken, on Python 3.9 or less, `__annotations__` will not exist in the class unless we have at least one annotation defined in it. Even if the parent classes have one

So either this is to work around that, or because we are relying on the class' annotations instead of getting from parents as well, which now that we have `get_annotations`, we can probably refactor?

Let me know what you think and I can try to refactor the codebase to use `get_annotations` only, basing on top of this branch in another PR

patrick91: we are using get_annotations, because `__annotations__` is not working anymore (or at least it is not working in the same way) in Python 3.14

bellini666: Hrmm I assumed `get_annotations` was going to work like I described, just saw [this](https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.get_annotations) and now I understand its purpose.

```python
# /// script
# dependencies = ["typing-extensions"]
# ///

from typing_extensions import get_annotations, Format


class A:
    foo: str
    bar: int


class B(A):
    bar: "float"
    baz: bool


for format in [Format.VALUE, Format.FORWARDREF, Format.STRING]:
    print(f"Format: {format}")
    print(get_annotations(A, format=format))
    print(get_annotations(B, format=format))
    print()
```

Gives:

```
Format: 1
{'foo': <class 'str'>, 'bar': <class 'int'>}
{'bar': 'float', 'baz': <class 'bool'>}

Format: 3
{'foo': <class 'str'>, 'bar': <class 'int'>}
{'bar': 'float', 'baz': <class 'bool'>}

Format: 4
{'foo': 'str', 'bar': 'int'}
{'bar': 'float', 'baz': 'bool'}
```

So it doesn't automatically get annotations from parents, we still need to go through the MRO to get the final dict

So the reason of this TODO is indeed python 3.9, but unless we are going to use `get_annotations` in other places where we retrieve the annotations from the class, we still need to have this to make sure when we do `cls.__annotations__` again, we get what `get_annotations` returned

patrick91: ah, right :) feel free to send another PR, maybe after merging this? :D

Issue #3823, opened 2025-03-28
Creating multiple schemas causes crash when using Relay node field extension


SupImDos:
## Describe the Bug

According to the [Strawberry documentation](https://strawberry.rocks/docs/types/schema#filteringcustomising-fields), a valid approach to having separate "public" and "internal" GraphQL APIs is to create two schemas.

However, if a type is shared between the two schemas that uses `strawberry.relay.node`, the schema generation fails due to this line: https://github.com/strawberry-graphql/strawberry/blob/de07ab66e215dcdfa317910dc281e3981573bb13/strawberry/relay/fields.py#L58

More technically, it appears that Strawberry field extensions are applied to fields each time a schema is created (instead of just once). This means that unless a field extension is written to be idempotent, unexpected behavior can occur when the same type is used in multiple schemas. See the following two failing unit tests for a minimal reproducible example: https://github.com/strawberry-graphql/strawberry/commit/93d627fb9d67fe014fee7c4d01197da8053e768a.

At a higher level, it seems a bit scary to me that there is some level of cross-contamination between the two schemas. I would have expected that if I create a new schema, it should be completely independent of any other schemas. Is having two schemas that share the same type(s) valid in Strawberry?

## Related Issues

* https://github.com/strawberry-graphql/strawberry/issues/3360

## System Information

 - Operating system: MacOS
 - Strawberry version: 0.262.5
 - Strawberry Django version: 0.57.1

## Additional Context

I stumbled across this when attempting to split my Strawberry Django GraphQL API into two schemas, one for public access and one for internal access. When I create the second schema, it crashes with the following traceback:

```python
Traceback (most recent call last):
  File "<redacted>/.venv/lib/python3.11/site-packages/graphql/type/definition.py", line 816, in fields
    fields = resolve_thunk(self._fields)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<redacted>/.venv/lib/python3.11/site-packages/graphql/type/definition.py", line 300, in resolve_thunk
    return thunk() if callable(thunk) else thunk
           ^^^^^^^
  File "<redacted>/.venv/lib/python3.11/site-packages/strawberry/schema/schema_converter.py", line 637, in <lambda>
    fields=lambda: self.get_graphql_fields(object_type),
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #3832, opened 2025-04-04
Ensure field extensions are only applied once

SupImDos:
## Description

Adds the `apply_field_extensions` utility function which uses `@cache` to ensure field extensions are only ever applied once, regardless of the number of schemas that use the field.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #3823

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,5 @@
+Release type: minor
+
+This release fixes an issue where field extensions were being applied multiple times when a field was used in multiple schemas. This could lead to unexpected behavior or errors if the extension's `apply` method wasn't idempotent.
+
+The issue has been resolved by introducing a caching mechanism that ensures each field extension is applied only once, regardless of how many schemas the field appears in. Test cases have been added to validate this behavior and ensure that extensions are applied correctly.
--- a/strawberry/extensions/field_extension.py
+++ b/strawberry/extensions/field_extension.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import itertools
 from collections.abc import Awaitable
-from functools import cached_property
+from functools import cache, cached_property
 from typing import TYPE_CHECKING, Any, Callable, Union
 
 if TYPE_CHECKING:
@@ -154,6 +154,17 @@ def build_field_extension_resolvers(
         f"If possible try to change the execution order so that all sync-only "
         f"extensions are executed first."
     )
+
+
+@cache
+def apply_field_extensions(field: StrawberryField) -> None:
+    """Applies the field extensions to the field.
+
+    This function is cached to avoid applying the extensions multiple times in the case
+    of multiple schema generation passes.
+    """
+    for extension in field.extensions:
+        extension.apply(field)
 
 
 __all__ = ["FieldExtension"]
--- a/strawberry/schema/schema_converter.py
+++ b/strawberry/schema/schema_converter.py
@@ -47,7 +47,10 @@ from strawberry.exceptions import (
     ScalarAlreadyRegisteredError,
     UnresolvedFieldTypeError,
 )
-from strawberry.extensions.field_extension import build_field_extension_resolvers
+from strawberry.extensions.field_extension import (
+    apply_field_extensions,
+    build_field_extension_resolvers,
+)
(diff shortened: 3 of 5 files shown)


patrick91 (commented):
Hey! Thanks for the PR! 

I wonder if we can maybe scope the cache to the schema? Or maybe when applying the extensions we make a copy of the field?

Just think about a future in which we might have default field extensions (based on the schema) and fields on different schemas will have different fields, what do you think?

bellini666:
> I wonder if we can maybe scope the cache to the schema? Or maybe when applying the extensions we make a copy of the field?

Was thinking that exactly! Scope the cache to the schema somehow (either by having it as a `dict[StrawberryField, list[Extensions]]` in the schema or `dict[Schema, list[Extension]]` in the field itself)

Copying the field is an option, but might generate other issues?

SupImDos:
> I wonder if we can maybe scope the cache to the schema? Or maybe when applying the extensions we make a copy of the field?
> 
> Just think about a future in which we might have default field extensions (based on the schema) and fields on different schemas will have different fields, what do you think?

> Was thinking that exactly! Scope the cache to the schema somehow (either by having it as a `dict[StrawberryField, list[Extensions]]` in the schema or `dict[Schema, list[Extension]]` in the field itself)
> 
> Copying the field is an option, but might generate other issues?

Hi @patrick91 & bellini666, thanks so much for your replies!

I actually considered both of these for this PR, and decided against it for my initial implementation - hopefully I can explain why below.

---

### 1. Scoping the cache to the schema

Maybe I'm missing something, but I don't think this will actually work in isolation.

The issue we're trying to solve here isn't that a single schema will apply the extensions multiple times - its that currently the same field instances are shared across multiple schemas, so each schema will apply the field extensions once. So, if we cache the extensions per-schema, we will still end up applying field extensions to the same field multiple times (once per-schema).

However, I do agree with the reasoning here - if in the future we have default field extensions that are schema-specific, we may need to cache them per-schema.

### 2. Copying the field before applying the extensions

This does seem to be a valid approach, and I think it would work (and would even work in conjunction with a schema-scoped cache).

However, I was worried about whether this might be a bit of a hack. Do we know if there are any flow-on effects of doing this? For example, is there code relying on there being only one instance of a field per schema, or code that relies on the field being the same instance after schema generation? Maybe there is even downstream code relying on this?

If we aren't concerned about that - and copying fields before applying extensions _is_ valid - then maybe this is the way to go?

---

Does my reasoning above make sense and/or give you any ideas? Or am I missing something in my understanding above? 😅

SupImDos:
@patrick91 & @bellini666

Sorry for the immediate follow-up, but I've got some extra context that might influence our decision / implementation.

I looked into copying fields in my project, and I've realised that I have custom `StrawberryField` subclasses that don't support copying properly! As a contrived example:

```python
class MyCustomStrawberryField(StrawberryField):
    def __init__(self, *args, my_custom_arg, **kwargs):
        super().__init__(*args, **kwargs)
        self.something = self.do_some_processing_with(my_custom_arg)
```

This example subclass:

1. Has a required `my_custom_arg` which is incompatible with the base `StrawberryField.__copy__` method
2. Processes `my_custom_arg` which is incompatible with the base `StrawberryField.__copy__` method
3. Doesn't implement its own `__copy__` method to copy its own custom attributes

I have also found that the base `StrawberryField.__copy__` method implementation doesn't provide a way to contribute arguments when instantiating the copy, which makes fixing this a bit tricky.

Regardless, this is probably a design issue with my subclass - but its made me wonder whether similar downstream issues might be encountered if we choose to copy the fields here?

SupImDos:
Hi @patrick91 & @bellini666 

In the short term I ended up working around this issue by avoiding the need for two schemas. However, I think think the original issue in #3823 is still quite important to fix - as it (at the very least) blocks creating multiple schemas when using Relay.

I ended up losing track of this PR, and it looks like my draft approach isn't appropriate anyway. Did you have any thoughts on the above comments, or any ideas on how we should proceed with fixing this bug?

Thanks again for all your hard work 😄

PR #3839, opened 2025-04-14
Fix graphql-ws links

johannkm:
## Summary by Sourcery

Documentation:
- Corrected the links for graphql-ws and graphql-transport-ws protocols in the subscriptions documentation, swapping their repository URLs

### DIFF
--- a/docs/general/subscriptions.md
+++ b/docs/general/subscriptions.md
@@ -306,9 +306,9 @@ class Subscription:
 ## GraphQL over WebSocket protocols
 
 Strawberry support both the legacy
-[graphql-ws](https://github.com/apollographql/subscriptions-transport-ws) and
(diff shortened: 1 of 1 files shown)


johannkm:
ah, 

> The `graphql-transport-ws` protocols repository is called `graphql-ws`. However,
`graphql-ws` is also the name of the legacy protocol. This documentation always
refers to the protocol names.

perfectly confusing.

johannkm:
Maybe could use apollo docs approach

> Note: Confusingly, the subscriptions-transport-ws library calls its WebSocket subprotocol graphql-ws, and the graphql-ws library calls its subprotocol graphql-transport-ws! In this article, we refer to the two libraries (subscriptions-transport-ws and graphql-ws), not the two subprotocols.

PR #3851, opened 2025-05-10
feat(parent): Support strawberry.Parent with future annotations

### LINKED ISSUE #3481: `strawberry.Parent` not supporting forward refs

bellini666:
Fix #3481

This also makes #3714 not required

## Summary by Sourcery

Add support for using `strawberry.Parent` with future annotations and delayed type evaluation

New Features:
- Support using strawberry.Parent with forward references and future annotations

Enhancements:
- Improve type resolution for Parent types in different annotation scenarios

Tests:
- Add test cases for Parent type with future annotations and string-based type references

Chores:
- Add a release note explaining the new Parent type annotation support

### DIFF
--- a/strawberry/types/fields/resolver.py
+++ b/strawberry/types/fields/resolver.py
@@ -118,7 +118,7 @@ class ReservedType(NamedTuple):
                 try:
                     evaled_annotation = annotation.evaluate()
                 except NameError:
-                    # If this is a strabwerry.Parent using ForwardRef, we will fail to
+                    # If this is a strawberry.Parent using ForwardRef, we will fail to
                     # evaluate at this moment, but at least knowing that it is a reserved
                     # type is enough for now
                     # We might want to revisit this in the future, maybe by postponing
--- a/strawberry/types/fields/resolver.py
+++ b/strawberry/types/fields/resolver.py
@@ -25,7 +25,7 @@ from strawberry.exceptions import (
     ConflictingArgumentsError,
     MissingArgumentsAnnotationsError,
 )
-from strawberry.parent import StrawberryParent
+from strawberry.parent import StrawberryParent, resolve_parent_forward_arg
 from strawberry.types.arguments import StrawberryArgument
 from strawberry.types.base import StrawberryType, has_object_definition
 from strawberry.types.info import Info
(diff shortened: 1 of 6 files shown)


botberry:



---

This release adds support to use `strawberry.Parent` with future annotations.

For example, the following code will now work as intended:

```python


def get_full_name(user: strawberry.Parent[User]) -> str:
    return f"{user.first_name} {user.last_name}"


@strawberry.type
class User:
    first_name: str
    last_name: str
    full_name: str = strawberry.field(resolver=get_full_name)


@strawberry.type
class Query:
    @strawberry.field
    def user(self) -> User:
        return User(first_name="John", last_name="Doe")


```

Or even when not using future annotations, but delaying the evaluation of `User`, like:


```python
# Note the User being delayed by passing it as a string
def get_full_name(user: strawberry.Parent["User"]) -> str:
    return f"{user.first_name} {user.last_name}"


@strawberry.type
class User:
    first_name: str
    last_name: str
    full_name: str = strawberry.field(resolver=get_full_name)


@strawberry.type
class Query:
    @strawberry.field
    def user(self) -> User:
        return User(first_name="John", last_name="Doe")


```



Here's the tweet text: 
```

```

strawberry/types/fields/resolver.py
```diff
@@ -118,10 +118,17 @@ def find(
                 try:
                     evaled_annotation = annotation.evaluate()
                 except NameError:
-                    continue
-                else:
-                    if self.is_reserved_type(evaled_annotation):
-                        type_parameters.append(parameter)
+                    # If this is a strabwerry.Parent using ForwardRef, we will fail to
```
patrick91, line 121:
```suggestion
                    # If this is a strawberry.Parent using ForwardRef, we will fail to
```

patrick91 (approved):

PR #3858, opened 2025-05-16
Fix graphql-ws got stuck on invalid operation name

DoctorJohn:
Currently, subscriptions using the legacy protocol get stuck if an `operationName` is selected that does not exist in the GraphQL document.

This started happening with the release that introduced extensions for subscriptions.

## Description

If a `operationName` is requested that does not exist in the GraphQL document, `get_operation_type` in `strawberry.utils.operation` will raise an `RuntimeError`.

This `RuntimeError` is explicitly handled by the newer protocol, because it supports all operation types and calls `get_operation_type` itself to figure out what to do.

The legacy protocol only supports the subscription operation type, it does not call `get_operation_type` by itself, and did not explicitly handle said `RuntimeError`. However, `get_operation_type` eventually always gets called in `Schema._parse_and_validate_(a)sync` and the `RuntimeError` bubbled all the way up without getting handled, causing the subscription handler to get stuck.

Starting with this PR, we catch the `RuntimeError` early and raise a custom exception that is then handled by the legacy protocol handler, which informs the client. 

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

### DIFF
--- a/strawberry/subscriptions/protocols/graphql_ws/handlers.py
+++ b/strawberry/subscriptions/protocols/graphql_ws/handlers.py
@@ -14,6 +14,7 @@ from typing import (
 from strawberry.exceptions import ConnectionRejectionError
 from strawberry.http.exceptions import NonTextMessageReceived, WebSocketDisconnected
 from strawberry.http.typevars import Context, RootValue
+from strawberry.schema.exceptions import CannotGetOperationTypeError
 from strawberry.subscriptions.protocols.graphql_ws.types import (
     CompleteMessage,
     ConnectionInitMessage,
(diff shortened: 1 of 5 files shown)


botberry:



---

This release resolves the issue of subscriptions started via the legacy `graphql-ws` WebSocket subprotocol getting stuck if a non-existing `operationName` was specified.



Here's the tweet text: 
```

```

strawberry/subscriptions/protocols/graphql_ws/handlers.py
```diff
@@ -193,6 +194,14 @@ async def handle_async_results(
 
             await self.send_message(CompleteMessage(type="complete", id=operation_id))
 
+        except CannotGetOperationTypeError:
+            await self.send_message(
+                ErrorMessage(
+                    type="error",
+                    id=operation_id,
+                    payload={"message": "Can't get GraphQL operation type"},
```
bellini666, line 202: **suggestion:** maybe adding the name of the operation type that could not be found would be useful? Like "Could not find GraphQL operation type XXX"

bellini666 (approved):
Left a small suggestion, but LGTM 😊

strawberry/subscriptions/protocols/graphql_ws/handlers.py
DoctorJohn: Sounds like an obvious thing to do now that you say it. Thanks Thiago! I updated the PR o7

PR #3859, opened 2025-05-16
Better implementation of GlobalID = ID

### LINKED ISSUE #3854: unable to export graphql schema or run application when using both `strawberry.ID` and `strawberry.relay.GlobalID` for `strawberry-graphql==0.268.0`
I work on a [project](https://github.com/Arize-ai/phoenix) using both `strawberry.ID` and `strawberry.relay.GlobalID`. As of `strawberry-graphql==0.268.0`, I can no longer export the schema for the project. The code snippet below reproduces the issue:

```python
import strawberry
from strawberry.relay.types import GlobalID


@strawberry.type
class Query:
    @strawberry.field
    def hello(self, id: GlobalID) -> str:
        return "Hello World"

    @strawberry.field
    def hello2(self, id: strawberry.ID) -> str:
        return "Hello World"


schema = strawberry.Schema(
    Query,
)
```

Exporting the schema results in a verbose error message including:

```
ScalarAlreadyRegisteredError: Scalar `ID` has already been registered
```

This issue is blocking our adoption of the latest fix for the strawberry Relay integration.

Related:

- https://github.com/strawberry-graphql/strawberry/issues/3551, specifically https://github.com/strawberry-graphql/strawberry/issues/3551#issuecomment-2727691899
- https://github.com/strawberry-graphql/strawberry/pull/3853

patrick91:
WIP, this could actually become a new feature once implemented properly.

Closes #3854

## Summary by Sourcery

Enable seamless use of both GlobalID and ID scalars in the same schema by unifying their conversion and registration, and validate behavior through updated tests.

Enhancements:
- Allow mixing GlobalID and ID scalars in a single schema without conflicts
- Delegate GlobalID to the standard ID scalar when relay_use_legacy_global_id is disabled
- Add argument conversion for GlobalID inputs using GlobalID.from_id
- Remove special reserved_types handling for GlobalID in the scalar registry

Documentation:
- Add release notes documenting the patch release and support for using ID and GlobalID together

Tests:
- Update tests to verify correct parsing and serialization of GlobalID and ID inputs and returns
- Add new tests for returns of GlobalID and ID fields in both legacy and new modes

### DIFF
--- a/strawberry/types/_mapped_scalar.py
+++ b/strawberry/types/_mapped_scalar.py
@@ -1,0 +1,7 @@
+class _StrawberryMappedScalar:
+    def __init__(self, scalar_type: type) -> None:
+        self.scalar_type = scalar_type
+
+
+def _mapped_scalar(cls: type) -> type:
+    return _StrawberryMappedScalar(cls)
--- a/strawberry/types/_mapped_scalar.py
+++ b/strawberry/types/_mapped_scalar.py
@@ -1,7 +1,0 @@
-class _StrawberryMappedScalar:
-    def __init__(self, scalar_type: type) -> None:
-        self.scalar_type = scalar_type
-
-
-def _mapped_scalar(cls: type) -> type:
-    return _StrawberryMappedScalar(cls)
--- a/strawberry/types/arguments.py
+++ b/strawberry/types/arguments.py
@@ -170,6 +170,11 @@ def convert_argument(
         ]
 
     if is_scalar(type_, scalar_registry):
+        from strawberry.relay.types import _GlobalID
+
+        if type_ is _GlobalID:
+            return _GlobalID.from_id(value)
+
         return value
 
     if isinstance(type_, EnumDefinition):
--- a/strawberry/types/arguments.py
+++ b/strawberry/types/arguments.py
@@ -173,7 +173,7 @@ def convert_argument(
         from strawberry.relay.types import GlobalID
 
         if type_ is GlobalID:
-            return GlobalID.from_id(value)
+            return GlobalID.from_id(value)  # type: ignore
 
         return value
 
--- a/strawberry/types/arguments.py
+++ b/strawberry/types/arguments.py
@@ -170,10 +170,10 @@ def convert_argument(
         ]
 
     if is_scalar(type_, scalar_registry):
-        from strawberry.relay.types import _GlobalID
+        from strawberry.relay.types import GlobalID
 
-        if type_ is _GlobalID:
-            return _GlobalID.from_id(value)
+        if type_ is GlobalID:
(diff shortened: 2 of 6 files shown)


strawberry/types/_mapped_scalar.py
```diff
@@ -0,0 +1,7 @@
+class _StrawberryMappedScalar:
+    def __init__(self, scalar_type: type) -> None:
+        self.scalar_type = scalar_type
+
+
+def _mapped_scalar(cls: type) -> type:
+    return _StrawberryMappedScalar(cls)
```
patrick91, line 7: The way I implemented this PR right doesn't really use this, so maybe we can just get rid of it? 🤔 

The idea was to do something like `GlobalID = Annotated[_GlobalID, _mapped_scalar(ID)]`, but we are not using that annotated value at all 😊

strawberry/types/arguments.py
```diff
@@ -170,6 +170,11 @@ def convert_argument(
         ]
 
     if is_scalar(type_, scalar_registry):
+        from strawberry.relay.types import _GlobalID
+
+        if type_ is _GlobalID:
+            return _GlobalID.from_id(value)
```
patrick91, line 176: this would check if the type is a mapped type, but maybe this is fine for now.

I think we should remove the legacy global id support as soon as we can 😊

botberry:



---

This release (finally) fixes support for using `ID` and `GlobalID`
in the same schema.



Here's the tweet text: 
```

```

patrick91:
We had this idea of implementing a mapper in order to do things like:

```python
GlobalID = Annotated[_GlobalID, _mapped_scalar(ID)]
```

or better

```python
GlobalID = Annotated[strawberry.ID, SomeNameSimilarToTransformer(_GlobalID)
```

but I think the current solution, scoped to global IDs is actually better and with less scope creep 😊

Also I think we'll remove GlobalID in schema in future, we can implement a better solution then, as we should also refactor scalars 😊

patrick91:
@axiomofjoy can you try the pre-release, so I don't push another broken release? :D

botberry:
# Pre-release
:wave:

Releasing commit [8649be87e4d6e6edfa79f0139bbcadfb91088047] to PyPi as pre-release! :package:

botberry:
# Pre-release
:wave:

Pre-release **0.268.2.dev.1747436835** [8649be87e4d6e6edfa79f0139bbcadfb91088047] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.268.2.dev.1747436835
```

axiomofjoy:
> @axiomofjoy can you try the pre-release, so I don't push another broken release? :D

@patrick91 Looks like this is working! Ready to release.

Issue #3865, opened 2025-05-20
HTTP multipart subscriptions are not terminated with closing boundary when error is present in payload


cephalization:
HTTP multipart subscriptions work by holding open an HTTP connection and streaming parts of content, terminated by a boundary string before and after each payload.

This is an example of an ideal response from a graphql subscription streamed over http multipart:

```
--graphql
Content-Type: application/json

{}

--graphql
Content-Type: application/json

{"payload": {"data": null, "errors": [{"message": "You have an error"}]}}

--graphql--
```

However, when using strawberry subscriptions with http multipart streaming and returning a response that contains an error in the payload, you receive a response like this:

```
--graphql
Content-Type: application/json

{}

--graphql
Content-Type: application/json

{"payload": {"data": null, "errors": [{"message": "You have an error"}]}}


```

The lack of terminating boundary prevents clients like the apollo js graphql client, from parsing payloads, and terminating early.

## Describe the Bug

Strawberry is not terminating error responses over subscriptions with a final boundary, preventing clients from parsing error payloads that would end the subscription.

## System Information

 - Operating system: MacOS 15.4.1 (24E263)
 - Strawberry version (if applicable): 0.266.0

## Additional Context

I don't have a replication handy just yet, the issue was encountered in https://github.com/Arize-ai/phoenix

I was able to fix locally by hacking my node_modules, and ensuring a final boundary `--graphql--` exists in the apollo multipart parser, but this is not a sustainable fix, and only works in my local environment.

I can come up with one in a day or two, but I suspect the fix is to just make sure a final boundary exists in the response before completing the subscription.

See the RFC for more details on the boundary expectations here https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html

> The body must then contain one or more "body parts," each preceded by an encapsulation boundary, and the last one followed by a closing boundary.

RogerHYang:
Could be a race condition because the task is `done()` but the queue still has the final separator.

<img width="1653" alt="Image" src="https://github.com/user-attachments/assets/bd0b30cd-bbb4-4c0c-bf83-356016e890de" />

cephalization:
Resolved by #3866

PR #3866, opened 2025-05-21
fix: always get last multipart separator from queue when streaming

RogerHYang:
This PR fixes an issue with GraphQL multipart subscriptions where the final boundary marker (`--graphql--`) could sometimes be missed due to a race condition between task completion and message handling. See screenshot below.

Screenshot 2025-05-20 at 10 10 18 PM

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #3865 

## Checklist

## Summary by Sourcery

Fix streaming of multipart GraphQL subscriptions by parameterizing the boundary separator, updating the heartbeat to use it, and draining any remaining queue entries to always emit the final multipart boundary.

Bug Fixes:
- Parameterize and propagate the multipart separator into the `_stream_with_heartbeat` method.
- Use the passed separator in the heartbeat message instead of a hardcoded value.
- Drain leftover queue items and yield the final multipart boundary at the end of the stream.

### DIFF
--- a/strawberry/http/async_base_view.py
+++ b/strawberry/http/async_base_view.py
@@ -372,7 +372,9 @@ class AsyncBaseHTTPView(
         )
 
     def _stream_with_heartbeat(
-        self, stream: Callable[[], AsyncGenerator[str, None]]
+        self,
+        stream: Callable[[], AsyncGenerator[str, None]],
+        separator: str = "graphql",
     ) -> Callable[[], AsyncGenerator[str, None]]:
         """Adds a heartbeat to the stream, to prevent the connection from closing when there are no messages being sent."""
         queue: asyncio.Queue[tuple[bool, Any]] = asyncio.Queue(1)
@@ -391,7 +393,7 @@ class AsyncBaseHTTPView(
 
         async def heartbeat() -> None:
             while True:
-                await queue.put((False, self.encode_multipart_data({}, "graphql")))
+                await queue.put((False, self.encode_multipart_data({}, separator)))
 
                 await asyncio.sleep(5)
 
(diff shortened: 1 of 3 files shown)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Roger Yang for the PR 👏

```

strawberry/http/async_base_view.py
```diff
@@ -421,6 +423,16 @@ async def cancel_tasks() -> None:
                         raise data
 
                     yield data
+
+                if not queue.empty():
+                    raised, data = queue.get_nowait()
```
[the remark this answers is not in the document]
patrick91: not sure if this can happen here, but I think the suggestion makes sense :)

strawberry/http/async_base_view.py
```diff
@@ -421,6 +423,16 @@ async def cancel_tasks() -> None:
...
+
+                    if raised:
+                        await cancel_tasks()
+                        raise data
+
+                    if data == f"\r\n--{separator}--\r\n":
+                        yield data
```
patrick91, line 435: @RogerHYang do you think there's a different way to make sure we always send the separator? not sure if I like this if 😊

patrick91:
@RogerHYang do you think we can have a test for this too? 😊

strawberry/http/async_base_view.py
RogerHYang: I changed it to a triplet data structure and added commentary. Let me know what you think.

RogerHYang:
> @RogerHYang do you think we can have a test for this too? 😊

I added a test and verified that the old implementation indeed fails. Note that the old implementation does pass if `await sleep(0.001)` is removed from inside the `async for item in` loop of the test.

patrick91 (approved):
Amazing work! Thank you so much!

RogerHYang:
Thanks, I actually just made some improvement to the test.

Screenshot below shows that it fails on main.

Screenshot 2025-05-22 at 4 52 47 AM

botberry:

PR #3869, opened 2025-05-24
Fix error message for operation selection edge-case

DoctorJohn:
## Description

Note: This PR is based on #3868 (<- let's review and merge that one first).

This PR updates both GraphQL over WS implementations to return more precise error messages when an operation cannot be found in the request's query document.

There are two cases to consider:
1. An operation name was specified in the request, but it cannot be found in the query document
2. No operation name was specified in the request, in which case we would fall back to the first operation in the query document, but the query document does not contain any operations at all

Starting with this PR, we'll send distinct error messages for both cases to clients. In case 1, we inform the client that an operation with the requested name cannot be found. In case 2, we inform the client that no operation could be found.

Previously, both cases were handled the same way which resulted in an confusing error message when the legacy GraphQL over WS protocol was used.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,5 @@
+Release type: patch
+
+In this release, we improved some GraphQL over WS error messages. More precise
+error messages are now returned if Strawberry fails to find an operation in the
+query document.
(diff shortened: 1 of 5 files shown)


bellini666 (approved):
approve-south-park-gif

PR #3885, opened 2025-06-04
Fix create type implicitly required field resolvers

### LINKED ISSUE #3421: `create_type` Throws error of field not having a name
## Describe the Bug

ValueError: Field doesn't have a name. Fields passed to `create_type` must define a name by passing the `name` argument to `strawberry.field`.

It seems like that there is an issue when using create_type and passing fields.

Looking at the code here it uses `python_name`

https://github.com/strawberry-graphql/strawberry/blob/ec4cf150f4ed33cbc24e2ecf201590cdc387b782/strawberry/tools/create_type.py#L36

which seems to be hardcoded to None here 
https://github.com/strawberry-graphql/strawberry/blob/ec4cf150f4ed33cbc24e2ecf201590cdc387b782/strawberry/field.py#L514

## System Information

 - Operating system: Mac
 - Strawberry version (if applicable): 0.221.1

DoctorJohn:
## Description

This PR addresses the issue summarized in https://github.com/strawberry-graphql/strawberry/issues/3421#issuecomment-2940066583 by restoring the ability of `create_type` to create types from fields without resolvers, which was lost when https://github.com/strawberry-graphql/strawberry/pull/798 was merged.

This is useful when creating types from fields that rely on the schema's default resolver as in the following example:

```python
import strawberry
from strawberry.tools import create_type

first_name = strawberry.field(name="firstName", graphql_type=str)
Query = create_type(f"Query", [first_name])  # This threw a ValueError because the field had no python_name (which would normally be derived from its resolver unless explicitly provided to the StrawberryField constructor)

def default_resolver(obj, field):
    return "Hi"

schema = strawberry.Schema(
    query=Query,
    config=StrawberryConfig(default_resolver=default_resolver)
)
```

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* Fixes #3421

## Summary by Sourcery

Restore create_type’s flexibility by inferring missing python_name and allowing fields with only GraphQL names or only resolvers.

Bug Fixes:
- Restore create_type to accept fields with only a GraphQL name and fields with only a resolver

Enhancements:
- Infer a field’s python_name from its GraphQL name using snake_case conversion when missing

Documentation:
- Add RELEASE.md entry describing the create_type name requirement fix

Tests:
- Add tests for create_type with resolver-only, name-only, and resolver+name scenarios
- Update error message test when a field lacks both name and resolver

### DIFF
--- a/strawberry/tools/create_type.py
+++ b/strawberry/tools/create_type.py
@@ -50,6 +50,9 @@ def create_type(
         if not isinstance(field, StrawberryField):
             raise TypeError("Field is not an instance of StrawberryField")
 
+        # Fields created using `strawberry.field` without a resolver don't have a
+        # `python_name`. In that case, we fall back to the field's `graphql_name`
+        # set via the `name` argument passed to `strawberry.field`.
         field_name = field.python_name or field.graphql_name
 
         if field_name is None:
--- a/strawberry/tools/create_type.py
+++ b/strawberry/tools/create_type.py
@@ -4,6 +4,7 @@ from typing import Optional
 
 import strawberry
 from strawberry.types.field import StrawberryField
+from strawberry.utils.str_converters import to_snake_case
 
 
 def create_type(
@@ -50,15 +51,19 @@ def create_type(
         if not isinstance(field, StrawberryField):
             raise TypeError("Field is not an instance of StrawberryField")
 
-        if field.python_name is None:
+        field_name = field.python_name or (
+            to_snake_case(field.graphql_name) if field.graphql_name else None
+        )
+
+        if field_name is None:
             raise ValueError(
                 "Field doesn't have a name. Fields passed to "
                 "`create_type` must define a name by passing the "
                 "`name` argument to `strawberry.field`."
             )
 
-        namespace[field.python_name] = field
-        annotations[field.python_name] = field.type
+        namespace[field_name] = field
+        annotations[field_name] = field.type
 
     namespace["__annotations__"] = annotations  # type: ignore
 
--- a/strawberry/tools/create_type.py
+++ b/strawberry/tools/create_type.py
@@ -4,7 +4,6 @@ from typing import Optional
 
 import strawberry
 from strawberry.types.field import StrawberryField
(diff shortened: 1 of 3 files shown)


botberry:



---

This release fixes that the `create_type` tool asked users to pass a `name` for
fields without resolvers even when a `name` was already provided.

The following code now works as expected:

```python
import strawberry
from strawberry.tools import create_type

first_name = strawberry.field(name="firstName")
Query = create_type(f"Query", [first_name])
```



Here's the tweet text: 
```

```

strawberry/tools/create_type.py
```diff
@@ -50,15 +51,19 @@ def hello(info) -> str:
         if not isinstance(field, StrawberryField):
             raise TypeError("Field is not an instance of StrawberryField")
 
-        if field.python_name is None:
+        field_name = field.python_name or (
+            to_snake_case(field.graphql_name) if field.graphql_name else None
```
patrick91: mmh, not sure I like this 🤔

DoctorJohn: I'm not sure either.

Basically, we need a unique field name for a field created like this:

```python
first_name = strawberry.field(name="firstName")
```

We could either:
1. go with the provided `name` (i.e., `firstName`)
2. convert the provided `name` to snake case (i.e., `first_name`)
3. or add another argument to `strawberry.field` (e.g., `python_name`) and require that instead of the `name` in this situation

DoctorJohn: Option 1 is how this was originally done before it broke. Maybe we'd like that better? :thinking: 

Edit: Just added https://github.com/strawberry-graphql/strawberry/pull/3885/commits/df39650a243da221cf716630e1df3628ef71db2f for option 1. Feels cleaner.

patrick91: yup, I like that! maybe we can add a comment on why we the or?

DoctorJohn: Good idea, just done that. Thanks for the review!

patrick91 (approved):

PR #3888, opened 2025-06-06
Add graphql over http test suite

DoctorJohn:
## Description

This PR tests all our HTTP integrations against the GraphQL over HTTP spec.

The GraphQL over HTTP spec used to come with a [test suite](https://github.com/graphql/graphql-over-http/blob/main/working-group/test-suite.md). However, it's outdated, and it's recommended to [audit servers](https://github.com/graphql/graphql-http/blob/main/CONTRIBUTING.md#adding-implementations) instead.

Instead of spamming their repository with requests to add audit implementations for all 11+ Strawberry integrations, I added a test suite to Strawberry itself, performing all checks of the upstream audit.
This way, we can immediately identify regressions in our testing pipeline without relying on a delayed external audit.

Currently, we don't pass all checks because either we don't support something yet, or due to a faulty internal test client.

What's next:
1. I'll remediate each xfail in a separate PR (I already identified all issues and tried to document them as xfail reasons)
2. (In parallel, I fix some issues in the upstream audit suite itself)
3. After we pass all tests, I'll add an audit implementation of our debug server to the upstream repo (our debug server uses the same HTTP base class as all our other integrations and should thus be equivalent). When the upstream audit fails for strawberry, we know we have to update the test suite added in this PR.
4. Being compliant with this spec is a prerequisite for GraphQL over SSE, which is what I was originally going to work on lol

## Summary by Sourcery

Add a comprehensive pytest suite to validate Strawberry’s HTTP integrations against the GraphQL over HTTP specification, mirroring the upstream audit tests and marking unsupported or optional behaviors as xfail to track compliance gaps.

Tests:
- Introduce GraphQL over HTTP spec compliance tests for Chalice, Django, and Sanic HTTP clients
- Use xfail markers for currently unsupported or optional behaviors to monitor spec coverage

### DIFF
--- a/tests/http/test_graphql_over_http_spec.py
+++ b/tests/http/test_graphql_over_http_spec.py
@@ -345,7 +345,7 @@ async def test_b8b3(http_client):
 )
 async def test_94b_(http_client, parameter):
     """
-    SHOULD allow null varibales/operationName/extensions parameter when accepting application/graphql-response+json
+    SHOULD allow null variables/operationName/extensions parameter when accepting application/graphql-response+json
     """
     response = await http_client.post(
         url="/graphql",
--- a/tests/http/test_graphql_over_http_spec.py
+++ b/tests/http/test_graphql_over_http_spec.py
@@ -1,4 +1,3 @@
-# ruff: noqa: N802
 import pytest
 
 from tests.http.clients.chalice import ChaliceHttpClient
@@ -10,7 +9,7 @@ from tests.http.clients.sanic import SanicHttpClient
     reason="Our integrations currently only return application/json",
     raises=AssertionError,
 )
-async def test_22EB(http_client):
+async def test_22eb(http_client):
     """
     SHOULD accept application/graphql-response+json and match the content-type
     """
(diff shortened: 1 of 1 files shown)


patrick91 (commented):
YOU'RE THE BEST! Thank you for this 😊

tests/http/test_graphql_over_http_spec.py
```diff
@@ -0,0 +1,802 @@
...
+except ImportError:
+    DjangoHttpClient = type(None)
+
+try:
+    from tests.http.clients.sanic import SanicHttpClient
+except ImportError:
+    SanicHttpClient = type(None)
+
+
+@pytest.mark.xfail(
```
bellini666, line 24: **thought:** should those `xfail` be marked with `strict=True`? That would make sure that the tests fail when they pass, so we will know that we can remove the `xfail` from here

bellini666 (approved):
YOU'RE THE BEST! Thank you for this 😊 [2]

tests/http/test_graphql_over_http_spec.py
DoctorJohn: Didn't add them purely for practical purposes: I planned to fix them one by one right after this test suite is merged. The first thing I would have to do is remove the `strict=True` again. 🤭 lazy, I know

bellini666: hahaah np then 😊

PR #3890, opened 2025-06-06
Remove unused dependency

DoctorJohn:
## Description

This PR removes `graphlib_backport` from our dependencies. It was only used for Python <3.9. However, we require Python >=3.9, so the dependency is not being installed anymore, and the standard library is used instead. 

This also resolves the following error Poetry was reporting: `Error: Cannot find dependency "graphlib_backport" for extra "cli" in main dependencies.`

I'm cleaning up our `pyproject.toml` to move us close to a migration to uv. Expect a few more follow-up PRs on separate issues. Not sending all at once to avoid lockfile conflicts.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Enhancement/optimization
- [ ] Documentation

## Summary by Sourcery

Remove the unused graphlib_backport dependency now that Python>=3.9 is required and update the project configuration accordingly.

Bug Fixes:
- Resolve Poetry error related to missing graphlib_backport dependency

Enhancements:
- Drop graphlib_backport from pyproject.toml and remove it from the lock file

PR #3891, opened 2025-06-07
Remove unwanted and never included dependency

DoctorJohn:
## Description

Note: This PR is based on #3890, merge that one first.

This PR removes `pytest-django` from the django extras. There's no need to require `pytest-django` to use Strawberry with Django. Besides that, Poetry complains that this dependency is ignored anyway because it was not listed in the main dependencies (it's only in the dev group). Running `uv init && uv add --extra=django strawberry-graphql` confirms that `pytest-django` is not getting installed even without this PR.

For reference, this is the Poetry error resolved by this PR: `Error: Cannot find dependency "pytest-django" for extra "django" in main dependencies.`

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

bellini666 (approved):
Nice!

PR #3904, opened 2025-06-13
Fix chalice did not send content type headers

DoctorJohn:
## Description

This PR fixes that the Chalice integration did not set any response headers at all. The relevant spec recommends at least setting `application/json`. That's what all our integrations currently do.

(There's more work required to pass the other GraphQL over HTTP spec tests. It's essentially two protocols: one is requested with a `application/json` accept header, and the other, stricter protocol, is requested with a `application/graphql-response+json` accept header. Our code might be cleaner if we put the HTTP protocols into handler classes similar to the GraphQL over WS protocols. We can discuss this next week <3)

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation


botberry:



---

This release fixes that the Chalice HTTP view integration did not set
appropriate content-type headers for responses, as it's recommended by the
GraphQL over HTTP specification.



Here's the tweet text: 
```

```

PR #3909, opened 2025-06-14
Remove confusing test client alias

DoctorJohn:
## Description

This PR removes a confusing alias and unused fixtures from our WS tests.

For some reason, our test client was aliased as aiohttp test client, which made it look like some tests were only run against the aiohttp integration. This alias is gone now. I also removed a few unused inclusions of pytest's request fixture.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

## Summary by Sourcery

Simplify WebSocket tests by removing the deprecated aiohttp_app_client alias and unused fixtures, and updating tests to use the http_client fixture directly.

Enhancements:
- Remove the convenience aiohttp_app_client fixture alias from WS tests
- Update all tests to reference http_client instead of the old alias
- Remove unused pytest request fixtures from WS transport tests
- Streamline test function signatures by dropping redundant parameters

### DIFF
--- a/tests/websockets/test_graphql_transport_ws.py
+++ b/tests/websockets/test_graphql_transport_ws.py
@@ -153,9 +153,7 @@ async def test_ws_message_frame_types_cannot_be_mixed(ws_raw: WebSocketClient):
     assert ws.close_reason == "WebSocket message type must be text"
 
 
-async def test_connection_init_timeout(
-    request: object, http_client_class: type[HttpClient]
-):
+async def test_connection_init_timeout(http_client_class: type[HttpClient]):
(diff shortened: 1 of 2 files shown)


patrick91:
Something happened with the type checkers, I guess another update from pyright :'D

DoctorJohn:
> Something happened with the type checkers, I guess another update from pyright :'D

Yeah, I spent some time debugging our workflows this morning. The regular lint workflow also failed, but that got resolved when I deleted the github actions caches. For pyright, I think we might want to pin the pyright version we use (just like mypy) so that tests don't start failing "randomly".

PR #3913, opened 2025-06-15
Fix some test clients set unwanted body data

DoctorJohn:
## Description

This PR fixes that the Django, Chalice, and Sanic test clients always set a body, even if no body/json data was passed to them.

This made it impossible to implement a GraphQL over HTTP test case that checks how servers react to missing POST bodies.

(Note that our test clients currently define JSON as `dict[str, object]`, while `None` indicates that no `json` argument was provided. Technically, `None` would be valid JSON too (`null`), but we currently neither reflect that in our types nor do we make use of it in tests. So, no need to replace `None` with `UNSET` in our test clients (yet)). 

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Summary by Sourcery

Fix HTTP test clients to omit the request body when no data or JSON is provided and re-enable missing-body behavior tests for Django, Chalice, and Sanic.

Bug Fixes:
- Prevent Django, Chalice, and Sanic test clients from always sending a JSON body when no body is provided
- Remove forced xfail for missing POST body tests in GraphQL over HTTP spec

Enhancements:
- Adjust body assignment logic to use dumps(json) only if json is not None

Tests:
- Re-enable and run test for missing request body on POST without client-specific skips

### DIFF
--- a/tests/http/clients/django.py
+++ b/tests/http/clients/django.py
@@ -171,7 +171,7 @@ class DjangoHttpClient(HttpClient):
 
         additional_arguments = {**headers}
 
-        body = data or dumps(json)
+        body = dumps(json) if json is not None else data
 
         if headers.get("HTTP_CONTENT_TYPE"):
             additional_arguments["content_type"] = headers["HTTP_CONTENT_TYPE"]
--- a/tests/http/clients/chalice.py
+++ b/tests/http/clients/chalice.py
@@ -144,7 +144,7 @@ class ChaliceHttpClient(HttpClient):
         json: Optional[JSON] = None,
         headers: Optional[dict[str, str]] = None,
     ) -> Response:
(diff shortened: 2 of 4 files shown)


tests/http/clients/django.py
```diff
@@ -171,7 +171,7 @@ async def post(
 
         additional_arguments = {**headers}
 
-        body = data or dumps(json)
```
bellini666, line 174: **praise:** the famous `or`/`and` gotcha 🤣

bellini666 (approved):
❤️

tests/http/clients/django.py
DoctorJohn: A classic 🤭

PR #3927, opened 2025-06-25
Fix invalid query parameters were not handled

DoctorJohn:
## Description

This PR updates our views to gracefully handle requests containing an invalid `query` parameter.

Previously, this resulted in TypeErrors internally. Now we're returning a 400 Bad Request response, conforming to the GraphQL over HTTP specification.

Bonus: pylance is now happier too

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Summary by Sourcery

Return HTTP 400 for requests with invalid `query` parameter types to conform to the GraphQL over HTTP spec.

Bug Fixes:
- Reject non-string `query` parameters with a 400 Bad Request and descriptive error message in both async and sync HTTP views.

Documentation:
- Include a release note in RELEASE.md describing the new invalid query handling.

Tests:
- Add parametrized tests to verify invalid `query` parameter values are rejected and remove the previous xfail marker for this scenario.

### DIFF
--- a/strawberry/http/async_base_view.py
+++ b/strawberry/http/async_base_view.py
@@ -533,7 +533,7 @@ class AsyncBaseHTTPView(
         if not isinstance(query, (str, type(None))):
             raise HTTPException(
                 400,
-                "GraphQL operations must contain a non-empty `query` or a `persistedQuery` extension.",
+                "The GraphQL operation's `query` must be a string or null, if provided.",
             )
 
         return GraphQLRequestData(
--- a/strawberry/http/async_base_view.py
+++ b/strawberry/http/async_base_view.py
@@ -529,8 +529,15 @@ class AsyncBaseHTTPView(
         else:
             raise HTTPException(400, "Unsupported content type")
 
+        query = data.get("query")
+        if not isinstance(query, (str, type(None))):
+            raise HTTPException(
+                400,
+                "GraphQL operations must contain a non-empty `query` or a `persistedQuery` extension.",
+            )
+
         return GraphQLRequestData(
-            query=data.get("query"),
+            query=query,
             variables=data.get("variables"),
             operation_name=data.get("operationName"),
             extensions=data.get("extensions"),
--- a/tests/http/test_query.py
+++ b/tests/http/test_query.py
@@ -196,6 +196,30 @@ async def test_missing_query(http_client: HttpClient):
     assert "No GraphQL query found in the request" in response.text
 
 
+@pytest.mark.parametrize(
+    "not_stringified_json",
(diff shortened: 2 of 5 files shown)


botberry:



---

In this release, we updated Strawberry to gracefully handle requests containing
an invalid `query` parameter. Previously, such requests could result in internal
server errors, but now they will return a 400 Bad Request response with an
appropriate error message, conforming to the GraphQL over HTTP specification.



Here's the tweet text: 
```

```

tests/http/test_query.py
```diff
@@ -196,6 +196,30 @@ async def test_missing_query(http_client: HttpClient):
     assert "No GraphQL query found in the request" in response.text
 
 
+@pytest.mark.parametrize(
+    "not_stringified_json",
+    [{"obj": "ect"}, 0, False, ["array"]],
+)
```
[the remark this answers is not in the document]
DoctorJohn: cute, but no. None would be valid

strawberry/http/async_base_view.py
```diff
@@ -529,8 +529,15 @@ async def parse_http_body(
         else:
             raise HTTPException(400, "Unsupported content type")
 
+        query = data.get("query")
+        if not isinstance(query, (str, type(None))):
```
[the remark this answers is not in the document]
DoctorJohn: okay okay sheesh. Three AI's complaining about the error message I yoinked from the Apollo server xD

strawberry/http/async_base_view.py
```diff
@@ -529,8 +529,15 @@ async def parse_http_body(
...
+            raise HTTPException(
+                400,
+                "GraphQL operations must contain a `query` string or a `persistedQuery` extension.",
```
patrick91: `persistedQuery` why this?

also is this the message coming from the spec? the query is there potentially there, but might not be valid?

DoctorJohn: The message is not part of the spec, so I took the one from Apollo, which includes the `persistedQuery` part.

I know we don't support persisted queries (yet), but searching our repo for "persisted queries" [showed](https://github.com/search?q=repo%3Astrawberry-graphql%2Fstrawberry%20persisted%20queries&type=code) that we want to enable extensions to implement them. That's why I kept that part of the error message. I can adjust the message to not mention persisted queries! :)

DoctorJohn: I'm updating it. Query must alwasy be a string anyways, even when using persisted queries

patrick91 (approved):

PR #3943, opened 2025-07-12
Fix invalid extensions parameters were not handled

DoctorJohn:
## Description

This PR does the same as #3927 and #3932, but for the `extensions` parameter. This is the final PR in this series.

(We're close to full spec compliance, I'll keep taking care of it bit by bit to keep the PRs reviewable.)

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Summary by Sourcery

Validate the `extensions` parameter in GraphQL over HTTP requests per the spec, returning a 400 error when it is not an object or null, and update related tests and client behavior.

Bug Fixes:
- Enforce that the GraphQL `extensions` parameter must be an object or null, returning a 400 Bad Request with a clear error message when invalid input is provided

Enhancements:
- Adjust the HTTP request builder to include the `extensions` field when it is explicitly set to null

Documentation:
- Add a release note outlining the patch for handling invalid `extensions` parameters gracefully

Tests:
- Add tests to verify that invalid `extensions` parameters are rejected with a 400 response
- Remove the xfail marker for invalid `extensions` in the GraphQL over HTTP spec test suite

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,6 @@
+Release type: patch
+
+In this release, we updated Strawberry to gracefully handle requests containing
+an invalid `extensions` parameter. Previously, such requests could result in
+internal server errors. Now, Strawberry will return a 400 Bad Request response
+with a clear error message, conforming to the GraphQL over HTTP specification.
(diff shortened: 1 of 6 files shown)


botberry:



---

In this release, we updated Strawberry to gracefully handle requests containing
an invalid `extensions` parameter. Previously, such requests could result in
internal server errors. Now, Strawberry will return a 400 Bad Request response
with a clear error message, conforming to the GraphQL over HTTP specification.



Here's the tweet text: 
```

```

PR #3950, opened 2025-07-19
fix: fix generics resolution when using Annotated unions

### LINKED ISSUE #3289: Strawberry cannot resolve type by inheriting a generic type with union type applied to it.
Hello! I tried to create a type inheriting a generic type with a union type applied and caught a `TypeError: Response fields cannot be resolved.`

## Describe the Bug

There's a code fragment the bug can be reproduced with:
```python
from typing import Annotated, Generic, TypeVar, Union
import strawberry

T = TypeVar("T")

@strawberry.type
class User:
    name: str
    age: int


@strawberry.type
class ProUser:
    name: str
    age: float


@strawberry.type
class GenType(Generic[T]):
    data: T


GeneralUser = Annotated[Union[User, ProUser], strawberry.union("GeneralUser")]


@strawberry.type
class Response(GenType[GeneralUser]):
    ...


@strawberry.type
class Query:
    @strawberry.field
    def user(self) -> Response:
        return Response(data=User(age=1, name="John"))

```
This code raises the following exception: `TypeError: Response fields cannot be resolved. Unexpected type 'typing.Annotated[typing.Union[__main__.User, __main__.ProUser], <strawberry.union.StrawberryUnion object at 0x14f8a90>]'`

But if you replace `GeneralUser = Annotated[Union[User, ProUser], strawberry.union("GeneralUser")]` with `GeneralUser =  strawberry.union("GeneralUser", (User, ProUser))` the code works as expected and doesn't raise any exception. 

It looks like union types cannot be applied to generic types if they're declared with `Annotated`, because this bug isn't reproducible in case the old-style approach with `strawberry.union` is used.

## System Information

 - Strawberry version (if applicable): 0.216.1
 - Python version: 3.9

bellini666:
Fix #3289

## Summary by Sourcery

Fix generic type resolution for Annotated unions by passing type definition context to annotation resolution and applying appropriate type variable mappings

Enhancements:
- Pass type_definition into annotation.resolve and apply specialized and type_var_map to StrawberryType generics
- Introduce _resolve_evaled_type helper and _is_type_generic predicate for cleaner resolution logic
- Remove old in-field generic resolution logic to centralize behavior

Documentation:
- Add release note describing the patch for generics resolution with Annotated unions

Tests:
- Add test for using Annotated Union inside a generic type to validate correct schema output

### DIFF
--- a/strawberry/annotation.py
+++ b/strawberry/annotation.py
@@ -173,7 +173,7 @@ class StrawberryAnnotation:
         if self._is_lazy_type(evaled_type):
             return evaled_type
         if self._is_streamable(evaled_type, args):
-            return self.create_list(list[evaled_type])  # type: ignore[valid-type]
+            return self.create_list(list[evaled_type])
         if self._is_list(evaled_type):
             return self.create_list(evaled_type)
         if self._is_maybe(evaled_type):
--- a/strawberry/annotation.py
+++ b/strawberry/annotation.py
@@ -125,16 +125,44 @@ class StrawberryAnnotation:
 
         return evaled_type, []
 
-    def resolve(self) -> Union[StrawberryType, type]:
+    def resolve(
+        self,
+        *,
+        type_definition: Optional[StrawberryObjectDefinition] = None,
+    ) -> Union[StrawberryType, type]:
         """Return resolved (transformed) annotation."""
-        if self.__resolve_cache__ is None:
-            self.__resolve_cache__ = self._resolve()
+        if (resolved := self.__resolve_cache__) is None:
+            resolved = self._resolve()
+            self.__resolve_cache__ = resolved
 
-        return self.__resolve_cache__
+        # If this is a generic field, try to resolve it using its origin's
+        # specialized type_var_map
+        if self._is_type_generic(resolved) and type_definition is not None:
+            from strawberry.types.base import StrawberryType
+
+            specialized_type_var_map = type_definition.specialized_type_var_map
+            if specialized_type_var_map and isinstance(resolved, StrawberryType):
+                resolved = resolved.copy_with(specialized_type_var_map)
+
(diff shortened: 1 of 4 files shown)


botberry:



---

This release fixes the resolution of `Generics` when specializing using a union
defined with `Annotated`, like in the example below:

```python
from typing import Annotated, Generic, TypeVar, Union
import strawberry

T = TypeVar("T")


@strawberry.type
class User:
    name: str
    age: int


@strawberry.type
class ProUser:
    name: str
    age: float


@strawberry.type
class GenType(Generic[T]):
    data: T


GeneralUser = Annotated[Union[User, ProUser], strawberry.union("GeneralUser")]


@strawberry.type
class Response(GenType[GeneralUser]): ...


@strawberry.type
class Query:
    @strawberry.field
    def user(self) -> Response: ...


```

Before this would raise a `TypeError`, now it works as expected.



Here's the tweet text: 
```

```

strawberry/annotation.py
```diff
@@ -292,6 +320,20 @@ def _is_enum(cls, annotation: Any) -> bool:
             return False
         return issubclass(annotation, Enum)
 
+    @classmethod
```
bellini666, line 323: We have a lot of `_is_xxx()` as private classmethods here, which could be useful in other places.

I'm wondering if we can have a new module, like `types/checks.py` and define those there? We can make use of `TypeGuard`/`TypeIs` for some cases as well for better annotation and replace the usage here and in other places in the codebase

Let me know what you think @patrick91 , @DoctorJohn and I'll do that in another PR

patrick91: not sure about the `checks` file, but I think having them outside is probably good!

bellini666: not a huge fan of `checks` name as well hahaha, but I could not think of a better name yet T_T

PR #3955, opened 2025-07-23
⚡️ Speed up function `in_async_context` by 82%

misrasaurabh1:
Saurabh's comments - This is sometimes preferred by other users because it looks cleaner. Plus since this seems to be used in important parts of the codebase, that's why i am opening it here.

### 📄 82% (0.82x) speedup for ***`in_async_context` in `strawberry/utils/inspect.py`***

⏱️ Runtime :   **`25.6 microseconds`**  **→** **`14.1 microseconds`** (best of `32` runs)
### 📝 Explanation and details

Here’s an optimized version that avoids the relatively expensive try/except path and `asyncio.get_running_loop()` call.  
Instead, it directly checks the loop using the faster `asyncio._get_running_loop()` "private" function, which is what `get_running_loop()` uses internally but without error handling overhead.

This saves a function call, exception handling, and is safe as of Python 3.7+ (including 3.12).

This is the fastest and least memory-heavy way in stock Python 3.12+.  
https://github.com/python/cpython/blob/b13a5df52fc854d1097e8b5419cb8802dc4059e0/Lib/asyncio/mixins.py#L13

✅ **Correctness verification report:**

| Test                        | Status            |
| --------------------------- | ----------------- |
| ⚙️ Existing Unit Tests | ✅ **3 Passed** |
| 🌀 Generated Regression Tests | ✅ **16 Passed** |
| ⏪ Replay Tests | ✅ **23 Passed** |
| 🔎 Concolic Coverage Tests | ✅ **1 Passed** |
|📊 Tests Coverage       | 100.0% |
<details>
<summary>⚙️ Existing Unit Tests and Runtime</summary>

| Test File::Test Function                                                                                                                                                      | Original ⏱️   | Optimized ⏱️   | Speedup   |
|:------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|:--------------|:---------------|:----------|
| `codeflash_concolic_mk522t7p/tmpwowt6j5b/test_concolic_coverage.py::test_in_async_context`                                                                                    | 1.18μs        | 684ns          | ✅71.8%   |
| `test_inspect.py::test_in_async_context_sync`                                                                                                                                 | 1.19μs        | 709ns          | ✅67.1%   |
| `test_pytest_inlinesnapshotdisable_teststest_dataloaders_py_teststest_inspect_py_teststest_deprecations_py__replay_test_0.py::test_strawberry_utils_inspect_in_async_context` | 3.49μs        | 1.77μs         | ✅97.5%   |
| `test_pytest_inlinesnapshotdisable_teststypestest_resolver_types_py_teststypestest_parent_type_future_anno__replay_test_0.py::test_strawberry_utils_inspect_in_async_context` | 9.12μs        | 4.70μs         | ✅94.0%   |

</details>

<details>
<summary>🌀 Generated Regression Tests and Runtime</summary>

```python
import asyncio
import threading
import time

# imports
import pytest  # used for our unit tests
from strawberry.utils.inspect import in_async_context

# unit tests

# 1. Basic Test Cases

def test_in_async_context_basic_sync():
    """Test that in_async_context returns False in a normal synchronous context."""
    codeflash_output = in_async_context() # 1.25μs -> 670ns (86.6% faster)

@pytest.mark.asyncio
async def test_in_async_context_basic_async():
    """Test that in_async_context returns True inside an async function with a running event loop."""
    codeflash_output = in_async_context()



def test_in_async_context_no_event_loop_set_but_created():
    """Test that in_async_context returns False if an event loop is created but not running."""
    loop = asyncio.new_event_loop()
    # Do not set as current, do not run
    try:
        codeflash_output = in_async_context()
    finally:
        loop.close()

def test_in_async_context_event_loop_set_but_not_running():
    """Test that in_async_context returns False if an event loop is set but not running."""
    loop = asyncio.new_event_loop()
    asyncio.set_event_loop(loop)
    try:
        codeflash_output = in_async_context()
    finally:
        asyncio.set_event_loop(None)
        loop.close()





def test_in_async_context_after_event_loop_closed():
    """Test that in_async_context returns False after the event loop is closed."""
    loop = asyncio.new_event_loop()
    asyncio.set_event_loop(loop)
    loop.close()
    asyncio.set_event_loop(None)
    codeflash_output = in_async_context() # 2.22μs -> 988ns (124% faster)

# 3. Large Scale Test Cases

@pytest.mark.asyncio
async def test_in_async_context_many_concurrent_tasks():
    """Test that in_async_context returns True in many concurrent async tasks."""
    results = []
    async def worker(idx):
        # Each worker should see the event loop is running
        results.append((idx, in_async_context()))
    await asyncio.gather(*(worker(i) for i in range(500)))




import asyncio
import concurrent.futures
import threading

# imports
import pytest  # used for our unit tests
from strawberry.utils.inspect import in_async_context

# ---------------------------
# unit tests for in_async_context
# ---------------------------

# 1. Basic Test Cases

def test_in_sync_context_returns_false():
    """Test that in_async_context returns False in a normal synchronous context."""
    codeflash_output = in_async_context() # 1.07μs -> 578ns (85.3% faster)

@pytest.mark.asyncio
async def test_in_async_context_returns_true():
    """Test that in_async_context returns True inside an async function."""
    codeflash_output = in_async_context()





def test_in_main_thread_with_stopped_loop():
    """Test that in_async_context returns False after the loop is closed."""
    loop = asyncio.new_event_loop()
    asyncio.set_event_loop(loop)
    loop.close()
    asyncio.set_event_loop(None)
    codeflash_output = in_async_context() # 1.55μs -> 1.02μs (52.3% faster)


def test_in_async_with_context_manager():
    """Test that in_async_context returns True inside an async context manager."""
    class DummyAsyncCM:
        async def __aenter__(self):
            return in_async_context()
        async def __aexit__(self, exc_type, exc, tb):
            return False
    async def runner():
        async with DummyAsyncCM() as val:
            return val





from strawberry.utils.inspect import in_async_context

def test_in_async_context():
    in_async_context()
```

</details>

<details>
<summary>⏪ Replay Tests and Runtime</summary>

| Test File::Test Function                                                                                                                                                      | Original ⏱️   | Optimized ⏱️   | Speedup   |
|:------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|:--------------|:---------------|:----------|
| `codeflash_concolic_mk522t7p/tmpwowt6j5b/test_concolic_coverage.py::test_in_async_context`                                                                                    | 1.18μs        | 684ns          | ✅71.8%   |
| `test_inspect.py::test_in_async_context_sync`                                                                                                                                 | 1.19μs        | 709ns          | ✅67.1%   |
| `test_pytest_inlinesnapshotdisable_teststest_dataloaders_py_teststest_inspect_py_teststest_deprecations_py__replay_test_0.py::test_strawberry_utils_inspect_in_async_context` | 3.49μs        | 1.77μs         | ✅97.5%   |
| `test_pytest_inlinesnapshotdisable_teststypestest_resolver_types_py_teststypestest_parent_type_future_anno__replay_test_0.py::test_strawberry_utils_inspect_in_async_context` | 9.12μs        | 4.70μs         | ✅94.0%   |

</details>

To edit these changes `git checkout codeflash/optimize-in_async_context-md4t2j8v` and push.

[Codeflash](https://codeflash.ai)

## Summary by Sourcery

Enhancements:
- Optimize in_async_context to use asyncio._get_running_loop for faster detection of a running event loop, eliminating exception overhead and achieving an ~82% speedup

### DIFF
--- a/strawberry/utils/inspect.py
+++ b/strawberry/utils/inspect.py
@@ -17,14 +17,9 @@ import strawberry
 
 
 def in_async_context() -> bool:
-    # Based on the way django checks if there's an event loop in the current thread
+    # Faster check for an event loop in the current thread using private API
     # https://github.com/django/django/blob/main/django/utils/asyncio.py
-    try:
-        asyncio.get_running_loop()
-    except RuntimeError:
-        return False
-    else:
-        return True
+    return asyncio._get_running_loop() is not None
 
 
 @lru_cache(maxsize=250)


botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Saurabh Misra for the PR 👏

```

strawberry/utils/inspect.py
```diff
@@ -17,14 +17,9 @@
...
-    # Based on the way django checks if there's an event loop in the current thread
+    # Faster check for an event loop in the current thread using private API
     # https://github.com/django/django/blob/main/django/utils/asyncio.py
-    try:
-        asyncio.get_running_loop()
-    except RuntimeError:
-        return False
-    else:
-        return True
+    return asyncio._get_running_loop() is not None
```
patrick91, line 22: @bellini666 what do you think?

see: https://github.com/python/cpython/blob/e41c1ce585827f92dab9b7a7fc3df2bda2f817fe/Lib/asyncio/events.py#L765-L774

bellini666: Hrmm interesting..

Yeah, apparently `get_running_loop` uses that and raises `RuntimeError` if it returns `None`

This change would prevent the `if loop is None` from there, which makes us have to except `RuntimeError`, which, at least until Python 3.11, had some overhead (3.11 introduced zero-cost exceptions)

My only concern here is to depend on a private member that could change anytime and cause issues without us knowing.

And checking the difference in times, before it was taking 0,00119 ms, with this change, it takes 0,000709 ms. Yes, it is a 67% performance improvement 🤣, but at the same time negligible, considering we don't call this that much

Sorry codeflash, I don't think this is a good change 😅

patrick91: yup, let's close it /cc @misrasaurabh1

misrasaurabh1: yep the speedup was tiny but thought that this might be on the hot path, so thought might be good to merge. I plan to sync with @patrick91 about the types of optimizations that matters the most to your project

PR #3961, opened 2025-07-29
Change `Maybe` to `Some[T] | None` instead of `Some[T | None] | None`

### LINKED ISSUE #3779: Type-safe "optional-nullable" fields
### Preface
it is a common practice in strawberry that when your data layer have 
an optional field i.e
```py
class Person:
    name: str
    phone: str | None = None
```
and you want to update it you would use `UNSET` in the mutation input
in order to check whether this field provided by the client or not like so:
```py
@strawberry.input
class UpdatePersonInput:
    id: strawberry.ID
    name: str| None
    phone: str | None = UNSET

@strawberry.mutation
def update_person(input: UpdatePersonInput) -> Person:
    inst = service.get_person(input.id)
    if name := input.name:
         inst.name = name
    if input.phone is not UNSET:
        inst.phone = input.phone  # ❌ not type safe
    
    service.save(inst)
```

Note that this is not an optimization rather a business requirement.
if the user wants to nullify the phone it won't be possible other wise
OTOH you might nullify the phone unintentionally. 

This approach can cause lots of bugs since you need to **remember** that you have
used `UNSET` and to handle this correspondingly.  

Since strawberry claims to 
> Strawberry leverages Python type hints to provide a great developer experience while creating GraphQL Libraries.

it is only natural for us to provide a typesafe way to mitigate this.

### Proposal
The `Option` type.which will require only this minimal implementation
```py
import dataclasses


@dataclasses.dataclass
class Some[T]:
    value: T
    
    def some(self) -> Some[T | None] | None:
      return self
    
@dataclasses.dataclass
class Nothing[T]:
    def some(self) -> Some[T | None] | None:
      return None 

Maybe[T] = Some[T] | Nothing[T]
```
and this is how you'd use it
```py
@strawberry.input
class UpdatePersonInput:
    id: strawberry.ID
    name: str| None
    phone: Maybe[str | None]

@strawberry.mutation
def update_person(input: UpdatePersonInput) -> Person:
    inst = service.get_person(input.id)
    if name := input.name:
         inst.name = name
    if phone := input.phone.some(): 
        inst.phone = phone.value  # ✅  type safe
    
    service.save(inst)
```

Currently if you want to know if a field was provided

### Backward compat
`UNSET` can remain as is for existing codebases.
`Option` would be handled separately.
 
### which `Option` library should we use?
1. **Don't use any library craft something minimal our own** as suggested above.
2. ** use something existing**

The sad truth is that there are no well-maintained libs in the ecosystem. 

Never the less it is not hard to maintain something just for strawberry since the implementation
is rather straight forward and not much features are needed. we can fork either
- https://github.com/rustedpy/maybe
- https://github.com/MaT1g3R/option

and just forget about it.

3. **allow users to decide**

```py
# before anything

strawberry.register_option_type((MyOptionType, NOTHING))

```
then strawberry could use that and you could use whatever you want.

- [ ] Core functionality
- [x] New behavior

patrick91:
## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* Closes #3779

## Checklist

## Summary by Sourcery

Refactor the Maybe type definition to avoid redundant nesting of None and update the GraphQL schema converter to handle the new StrawberryMaybe type accordingly

Enhancements:
- Simplify the Maybe type alias to Union[Some[T], None] by removing the nested None
- Extend schema_converter.from_maybe_optional to recognize and unwrap StrawberryMaybe instances

## Summary by Sourcery

Refine strawberry.Maybe to represent Some[T] | None by default, disallow explicit nulls for Maybe[T], and require Maybe[T | None] for nullability. Update conversion logic, schema generation, and validation rules accordingly, provide a codemod for migration, and enrich documentation and tests.

New Features:
- Introduce MaybeNullValidationRule to enforce that Maybe[T] fields cannot receive explicit null values
- Add ConvertMaybeToOptional codemod and register it in the upgrade CLI for migrating Maybe[T] to Maybe[T | None]

Enhancements:
- Simplify Maybe type alias to Union[Some[T], None] and update argument conversion to handle StrawberryMaybe correctly
- Update schema_converter.from_maybe_optional to unwrap nested Optionals and generate the proper GraphQL types for Maybe
- Expand tests to cover nuanced Maybe[T] vs Maybe[T | None] behavior across inputs, resolvers, nested types, and schema generation

Documentation:
- Add dedicated Maybe documentation in docs/types/maybe.md and link to it from resolvers and input-types guides
- Document breaking changes for v0.279.0 about the new Maybe semantics

Tests:
- Add comprehensive tests for MaybeNullValidationRule, resolver and input field validation, type checking, and codemod behavior

Chores:
- Register MaybeNullValidationRule in schema validation rules and add the maybe-optional codemod to the CLI

### DIFF
--- a/docs/breaking-changes.md
+++ b/docs/breaking-changes.md
@@ -4,6 +4,7 @@ title: List of breaking changes and deprecations
 
 # List of breaking changes and deprecations
 
+- [Version 0.279.0 - 1 August 2025](./breaking-changes/0.279.0.md)
 - [Version 0.268.0 - 10 May 2025](./breaking-changes/0.268.0.md)
 - [Version 0.249.0 - 18 November 2024](./breaking-changes/0.249.0.md)
 - [Version 0.243.0 - 25 September 2024](./breaking-changes/0.243.0.md)
--- a/docs/breaking-changes.md
+++ b/docs/breaking-changes.md
@@ -4,7 +4,7 @@ title: List of breaking changes and deprecations
 
 # List of breaking changes and deprecations
 
-- [Version 0.279.0 - 1 August 2025](./breaking-changes/0.279.0.md)
+- [Version 0.279.0 - 19 August 2025](./breaking-changes/0.279.0.md)
 - [Version 0.268.0 - 10 May 2025](./breaking-changes/0.268.0.md)
 - [Version 0.249.0 - 18 November 2024](./breaking-changes/0.249.0.md)
 - [Version 0.243.0 - 25 September 2024](./breaking-changes/0.243.0.md)
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,12 @@
+Release type: minor
+
+This release changes the `strawberry.Maybe` type to provide a more consistent and intuitive API for handling optional fields in GraphQL inputs.
+
+**Breaking Change**: The `Maybe` type definition has been changed from `Union[Some[Union[T, None]], None]` to `Union[Some[T], None]`. This means:
+
+- `Maybe[str]` now only accepts string values or absent fields (refuses explicit null)
+- `Maybe[str | None]` accepts strings, null, or absent fields (maintains previous behavior)
+
+This provides a cleaner API where `if field is not None` consistently means "field was provided" for all Maybe fields. A codemod is available to automatically migrate your code: `strawberry upgrade maybe-optional`
+
+See the [breaking changes documentation](https://strawberry.rocks/docs/breaking-changes/0.279.0) for migration details.
--- a/docs/breaking-changes/0.279.0.md
+++ b/docs/breaking-changes/0.279.0.md
@@ -1,0 +1,136 @@
+---
+title: 0.279.0 Breaking Changes
+slug: breaking-changes/0.279.0
+---
+
+# v0.279.0 Breaking Changes
+
+This release changes the `strawberry.Maybe` type definition to provide a more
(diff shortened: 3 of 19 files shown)


botberry:



---

This release changes the `strawberry.Maybe` type to provide a more consistent and intuitive API for handling optional fields in GraphQL inputs.

**Breaking Change**: The `Maybe` type definition has been changed from `Union[Some[Union[T, None]], None]` to `Union[Some[T], None]`. This means:

- `Maybe[str]` now only accepts string values or absent fields (refuses explicit null)
- `Maybe[str | None]` accepts strings, null, or absent fields (maintains previous behavior)

This provides a cleaner API where `if field is not None` consistently means "field was provided" for all Maybe fields. A codemod is available to automatically migrate your code: `strawberry upgrade maybe-optional`

See the [breaking changes documentation](https://strawberry.rocks/docs/breaking-changes/0.279.0) for migration details.



Here's the tweet text: 
```

```

docs/breaking-changes.md
```diff
@@ -4,6 +4,7 @@ title: List of breaking changes and deprecations
 
 # List of breaking changes and deprecations
 
+- [Version 0.279.0 - 1 August 2025](./breaking-changes/0.279.0.md)
```
patrick91:
```suggestion
- [Version 0.279.0 - 19 August 2025](./breaking-changes/0.279.0.md)
```

PR #3999, opened 2025-09-22
Replace EnumDefinition with StrawberryEnum

### LINKED ISSUE #3998: Resolve TODO: Replace with StrawberryEnum
Resolve the TODO's about EnumDefinition at [line 885](https://github.com/strawberry-graphql/strawberry/blob/0bdfea480db45699928fe3192bbcfaf69f00fd51/strawberry/schema/schema_converter.py#L872-L873) and [line 872](https://github.com/strawberry-graphql/strawberry/blob/0bdfea480db45699928fe3192bbcfaf69f00fd51/strawberry/schema/schema_converter.py#L872)

I'll take care of this issue.

Ckk3:
## Description

Replace `EnumDefinition` with `StrawberryEnum`

## Types of Changes

- [x] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* Fixes #3998

## Checklist

## Summary by Sourcery

Replace the legacy EnumDefinition with StrawberryEnum throughout the core codebase, ensuring all conversions, codegen paths, and tests are updated to use the new class name.

Enhancements:
- Replace all occurrences of the deprecated EnumDefinition class with the new StrawberryEnum type across schema conversion, name conversion, code generation, federation, and experimental Pydantic modules
- Update method signatures, type checks, and imports to reference StrawberryEnum instead of EnumDefinition
- Refactor tests to assert against StrawberryEnum and adjust related imports

Documentation:
- Add a patch release note in RELEASE.md documenting the renaming of EnumDefinition to StrawberryEnum

## Summary by Sourcery

Migrate all enum-related logic to use the new StrawberryEnum class, remove the legacy EnumDefinition, and introduce deprecation aliases and warnings for backward compatibility

Enhancements:
- Rename internal enum definition class from EnumDefinition to StrawberryEnum across the core codebase
- Deprecate legacy _enum_definition attribute with a backward-compatible alias and deprecation warnings
- Update schema conversion, codegen, federation, annotation, name converter, and compatibility logic to use StrawberryEnum

Documentation:
- Add a release note in RELEASE.md documenting the renaming of EnumDefinition to StrawberryEnum and deprecation of _enum_definition

Tests:
- Refactor tests to reference StrawberryEnum and assert deprecation warnings for the old enum definition

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,8 +1,21 @@
 Release type: minor
 
-## Deprecation Alert
-- **Deprecated**: Replaced `_enum_definition` with `__strawberry_definition__`. The former will continue to work but will raise a deprecation warning.
+This release changes `_enum_definition` to `__strawberry_definition__`, this is a follow up to previous
+internal changes. If you were relying on `_enum_definition` you should update your code to use `__strawberry_definition__`.
 
-### Other Changes
-- **Renamed**: Changed `EnumDefinition` to `StrawberryEnumDefinition` to standardize internal naming patterns.
-- These updates improve naming consistency and address previously identified TODOs.
+We also expose `has_enum_definition` to check if a type is a strawberry enum definition.
+
+```python
+from enum import Enum
+import strawberry
+from strawberry.types.enum import StrawberryEnumDefinition, has_enum_definition
+
+
+@strawberry.enum
+class ExampleEnum(Enum):
+    pass
+
+
+has_enum_definition(ExampleEnum)  # True
+# Now you can use ExampleEnum.__strawberry_definition__ to access the enum definition
+```


botberry:



---

This release changes `_enum_definition` to `__strawberry_definition__`, this is a follow up to previous
internal changes. If you were relying on `_enum_definition` you should update your code to use `__strawberry_definition__`.

We also expose `has_enum_definition` to check if a type is a strawberry enum definition.

```python
from enum import Enum
import strawberry
from strawberry.types.enum import StrawberryEnumDefinition, has_enum_definition


@strawberry.enum
class ExampleEnum(Enum):
    pass


has_enum_definition(ExampleEnum)  # True
# Now you can use ExampleEnum.__strawberry_definition__ to access the enum definition
```



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Luis Gustavo for the PR 👏

```

strawberry/types/enum.py
```diff
@@ -151,7 +151,7 @@ def _process_enum(
         )
         values.append(value)
 
-    cls._enum_definition = EnumDefinition(  # type: ignore
+    cls.__strawberry_definition__ = StrawberryEnum(  # type: ignore
```
bellini666: should we also keep a deprecated `._enum_definiton` for compatibility?

You can use this [DeprecatedDescriptor](https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/types/object_type.py#L174-L174), like we have for the old `_type_definition`

strawberry/types/enum.py
```diff
@@ -24,7 +24,7 @@ class EnumValue:
 
 
 @dataclasses.dataclass
-class EnumDefinition(StrawberryType):
+class StrawberryEnum(StrawberryType):
```
bellini666: Like my comment below, we also need to expose the deprecated version of this

Like in https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/types/base.py#L456-L456

RELEASE.md
```diff
@@ -0,0 +1,3 @@
+Release type: patch
```
bellini666: Would this be a minor? It does more than just fixing, it also changes variable names

strawberry/types/enum.py
Ckk3: Agree, I'll do it!

strawberry/types/enum.py
Ckk3: I agree, updating now!

RELEASE.md
Ckk3: I agree 100% with you, also will add a deprecated info on the release

strawberry/types/enum.py
Ckk3: Updated!

strawberry/types/enum.py
Ckk3: Updated, @bellini666 , I use the same implementation we done in this PR #2836 .
Please let me now if you see a better way to do it!

Ckk3:
Already fixing this CI errors.

Ckk3:
Ready to review @bellini666

Ckk3:
Ready for review again @bellini666 
I've finally fixed the CI 😅

Ckk3:
> This change makes `has_object_definition` return incorrect results. For now, the type guard is
> 
> ```python
> def has_object_definition(
>     obj: Any,
> ) -> TypeGuard[type[WithStrawberryObjectDefinition]]:
> ```
> 
> which is now true for enums, as they fulfill `if hasattr(obj, "__strawberry_definition__")`.
> 
> However, `StrawberryEnum` (formerly `EnumDefinition`) and `StrawberryObjectDefinition` don't overlap.
> 
> We should fix this problem before merging - ideally _without_ introducing additional checks in all other type guards in this file. The same would go for a rename of `_scalar_definition`. Either, we decide to rename _everything_ into `_strawberry_definition`, or keep it fanned out over the different types (scalar, enum, [input-]object).
> 
> I don't have a clear preference on how we do this as long as the code is kept clen and avoids additional checks 😊

Thanks for your review @erikwrede !
About the comment, I confess I didn't understand 100% since `has_object_definition` already returned `False` for enums before so we just keep this behavior.

Do you think this behavior is incorrect now?
To make things clearer, I've created some new tests to guarantee the `has_object_definition` current behavior.

erikwrede:
@Ckk3 I just realized I must've been on the incorrect branch when checking this locally, just double checked the diff. The tests are great though!

erikwrede (approved):
No further change requests from my side 😊

Ckk3:
> @Ckk3 I just realized I must've been on the incorrect branch when checking this locally, just double checked the diff. The tests are great though! 

Oh okay, thank you😊!

strawberry/types/base.py
```diff
@@ -202,15 +202,15 @@ def has_object_definition(
     obj: Any,
 ) -> TypeGuard[type[WithStrawberryObjectDefinition]]:
     if hasattr(obj, "__strawberry_definition__"):
-        return True
+        return isinstance(obj.__strawberry_definition__, StrawberryObjectDefinition)
```
patrick91: maybe we can make one of this for Enums too

tests/benchmarks/test_stadium.py
```diff
@@ -76,10 +76,10 @@ def create_stadium(seats_per_row: int = 250) -> Stadium:
     """Create a stadium with a configurable number of seats per row.
 
     Default configuration (250 seats/row) creates approximately 50,000 seats:
-    - North Stand: 12,500 seats (50 rows × 250 seats)
-    - South Stand: 12,500 seats (50 rows × 250 seats)
-    - East Stand: 10,000 seats (40 rows × 250 seats)
-    - West Stand: 10,000 seats (40 rows × 250 seats)
+    - North Stand: 12,500 seats (50 rows x 250 seats)
```
patrick91: revert this

PR #4000, opened 2025-09-23
Fix pydantic generics and add ability to directive json schema

Boudewijn26:
## Description

This fixes the conversion of pydantic models into strawberry models if those models include generics. It also adds the ability to pass a `SchemaDirective` to the pydantic conversion to collect the pydantic json schema details in

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

## Checklist

## Summary by Sourcery

Fix Pydantic generic model support in Strawberry conversion and introduce json_schema_directive to annotate fields with Pydantic JSON schema metadata as GraphQL directives

New Features:
- Add json_schema_directive option to Pydantic type and input decorators to map Pydantic JSON schema extras into GraphQL field directives

Bug Fixes:
- Enable correct conversion of Pydantic models with generics into Strawberry types

Enhancements:
- Extend type replacement logic to handle Pydantic generic metadata when resolving field types
- Preserve original generic bases during dataclass creation for Strawberry types

Documentation:
- Update documentation to describe json_schema_directive usage

Tests:
- Add tests for nested Pydantic generics resolution
- Add tests for applying JSON schema directives on fields

### DIFF
--- a/tests/experimental/pydantic/schema/test_json_schema_capture.py
+++ b/tests/experimental/pydantic/schema/test_json_schema_capture.py
@@ -138,3 +138,103 @@ def test_can_use_both_pydantic_1_and_2():
 
     assert not result.errors
     assert result.data == {"user": {"__typename": "LegacyUser", "name": "legacy"}}
+
+
+def test_basic_with_alias_without_using_them():
+    class UserModel(pydantic.BaseModel):
+        age: Annotated[
+            int, pydantic.Field(gt=0, json_schema_extra={"test": 0}, alias="userAge")
+        ]
+        password: pydantic.json_schema.SkipJsonSchema[Optional[str]]
+
+    @strawberry.schema_directive(locations=[Location.FIELD_DEFINITION])
+    class JsonSchema:
+        test: int
+        exclusive_minimum: Optional[int] = None
+
+    @strawberry.experimental.pydantic.type(
+        UserModel, json_schema_directive=JsonSchema, use_pydantic_alias=False
+    )
+    class User:
+        age: strawberry.auto
+        password: strawberry.auto
+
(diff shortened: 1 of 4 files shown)


botberry:



---

This release fixes the pydantic support for generics and allows capture of the
Pydantic JSON schema attributes through a schema directive.



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Boudewijn van Groos for the PR 👏

```

tests/experimental/pydantic/schema/test_json_schema_capture.py
```diff
@@ -0,0 +1,140 @@
...
+    @strawberry.type
+    class Query:
+        @strawberry.field
+        def user(self) -> User:
+            return User(age=1, password="ABC")
+
+    schema = strawberry.Schema(query=Query)
+
+    expected_schema = """
+    directive @jsonSchema(test: Int!, exclusiveMinimum: Int = null) on FIELD_DEFINITION
```
bellini666, line 35: Would this also include values from other fields on other types?

And if so, would this become too bloated in case we have N fields, each with M different extras?

I'm not much familiar with the pydantic integration here, so @patrick91 may be the best to check this :)

Boudewijn26: Only the values in the json schema which are present on the passed directive are passed to the graphQL schema.

So bloat would be limited by whichever values the user wants in their directive. They could also decide to have different directives for different types to prevent their directives from growing too much

patrick91:
@Boudewijn26 do you still need this? 

I'm not sure about adding this feature 🤔

PR #4002, opened 2025-09-27
fix: improve multipart subscription boundary and protocol validation

LouisAmon:
## Description

Improved multipart subscription header parsing to handle optional boundary parameters and properly parse quoted subscription spec values.

Current implementation was not supporting the stock-standard `Accept` header:
```
Accept: multipart/mixed;subscriptionSpec="1.0", application/json
```
Source: 
https://www.apollographql.com/docs/graphos/routing/operations/subscriptions/multipart-protocol

Key changes:

- Modified _is_multipart_subscriptions() method in BaseView to make the boundary parameter optional (previously required boundary=graphql)
- Added proper quote stripping for subscriptionSpec values to handle cases like subscriptionSpec="1.0"
- Enhanced test coverage to include scenarios with and without boundary parameters, and with quoted subscription spec values

## Types of Changes

- [x] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

## Summary by Sourcery

Improve multipart subscription boundary and protocol validation by making the boundary parameter optional and stripping quotes from the subscriptionSpec header, and update test and version skip logic accordingly.

Bug Fixes:
- Allow multipart/mixed subscriptions without an explicit boundary parameter and strip quotes around subscriptionSpec values.
- Fix Django version skip checks to use django.get_version() and correct comparison logic.

Tests:
- Parametrize multipart subscription tests for scenarios with and without boundary and with quoted subscriptionSpec values.
- Add test case for parsing content type headers with quoted subscriptionSpec.


botberry:



---

Fixed multipart subscription header parsing to properly handle optional boundary parameters and quoted subscription spec values. This improves compatibility with different GraphQL clients that may send headers in various formats.

**Key improvements:**

- Made the `boundary=graphql` parameter optional in multipart subscription detection
- Added proper quote stripping for `subscriptionSpec` values (e.g., `subscriptionSpec="1.0"`)
- Enhanced test coverage for different header format scenarios

**Example of supported headers:**

```raw
Accept: multipart/mixed;boundary=graphql;subscriptionSpec=1.0,application/json
Accept: multipart/mixed;subscriptionSpec="1.0",application/json
```



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Louis Amon for the PR 👏

```

LouisAmon:
> * Consolidate Django version checks to always compare version tuples (e.g., map int parts) rather than raw strings to avoid lexicographical comparison pitfalls.

About that : `django.VERSION` was failing the linting on my end because the version tuple is of different length and types, so using the text is actually more reliable

> * Refine the content-type parser so that subsequent media types (like application/json) aren’t merged into the subscriptionSpec parameter, ensuring each token is parsed into its own key/value pair.

There was the option of introducing `werkzeug` which has great utilities for parsing headers, but it seemed overkill for that specific purpose

It's true that we aren't validating the "application/json" at the end of the `subscriptionSpec` tho.

tests/django/test_dataloaders.py
```diff
@@ -10,15 +10,15 @@
...
+    DJANGO_VERSION: str = django.get_version()
 except ImportError:
-    DJANGO_VERSION = (0, 0, 0)
+    DJANGO_VERSION = "0.0.0"
 
 
 pytestmark = [
     pytest.mark.asyncio,
     pytest.mark.skipif(
-        DJANGO_VERSION < (3, 1),
```
bellini666, line 21: I really prefer tuples comparison, as strings would fail for `"3.10" > "3.2" for example.

Any reason for this change

bellini666 (commented):
Other than my comment, LGTM

But I'll summon @DoctorJohn as he knows more about subscriptions :)

tests/django/test_dataloaders.py
LouisAmon: You're right. I reverted it

DoctorJohn (commented):
Nice changes, I appreciate the inclusion of tests! Consider the fix for handling quotes properly in the subscriptionSpec approved.

Regarding custom boundaries, I'm a bit undecided. The Apollo docs say:

> the value for `boundary` should *always* be `graphql`

While implementing multipart subscriptions, it looks like we interpreted that as: anything not using `graphql` as boundary is not a valid multipart subscription. We should check how other implementations handle that.

LouisAmon:
> Regarding custom boundaries, I'm a bit undecided. The Apollo docs say:
> 
> > the value for `boundary` should _always_ be `graphql`
> 
> While implementing multipart subscriptions, it looks like we interpreted that as: anything not using `graphql` as boundary is not a valid multipart subscription. We should check how other implementations handle that.

Hey @DoctorJohn 👋 

I debated this "boundary" thing myself when implementing it but here's my rationale:

1. In Apollo's [docs about the multipart protocol](https://www.apollographql.com/docs/graphos/routing/operations/subscriptions/multipart-protocol) they clearly give this example

`Accept: multipart/mixed;subscriptionSpec="1.0", application/json`

...and that doesn't include a boundary, thus making it optional.

What's your source about the boundary being necessary ?

2. I think the boundary doesn't have to be in the accept headers (but if it's there it should indeed be "graphql") : it's mostly meant to be in the response so that the receiving system can properly interpret each part of the multipart package

I guess that in theory you could have multiple boundaries, hence the need to clearly specify it, and in our case it seems absolutely necessary that the boundary should indeed be "graphql" in the response

I don't have any other example of a multipart "Accept" header in which the requester demands a certain boundary


It's just one line of code tho. So if you tell me it's best to enforce the boundary I don't mind changing it!

What do you think ?

LouisAmon:
@DoctorJohn gentle nudge :)

DoctorJohn:
> What's your source about the boundary being necessary ?

After looking at this again, it turns out I don't have one and just confused protocols a bit. Thanks for mentioning the accept header example from the Apollo docs again.

This is good to go now, except we should take a look at the CI failures and get the lint workflow to pass again.

botberry:

PR #4035, opened 2025-10-26
Add custom is_awaitable to test perf

patrick91:
## Summary by Sourcery

Introduce an optimized awaitable detection helper and wire it into schema execution to improve performance on large, mostly-synchronous result sets.

Enhancements:
- Add an optimized_is_awaitable utility that short-circuits checks for common non-awaitable primitive and container types before delegating to graphql-core.
- Update async and sync schema execution paths to use the optimized awaitable checker when invoking the GraphQL execution engine.
- Expose optimized_is_awaitable via the strawberry.execution package for reuse across the codebase.

Chores:
- Add a new uv.lock file to the repository.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,5 @@
+Release type: patch
+
+Improved execution performance by up to 10% by adding an optimized `is_awaitable` check with a fast path for common synchronous types (such as int, str, list, dict, etc.).
+
+This optimization reduces overhead when processing large result sets containing mostly basic values by avoiding expensive awaitable checks for types that are known to never be awaitable.


strawberry/execution/is_awaitable.py
```diff
@@ -0,0 +1,76 @@
...
+        # check for coroutine objects
+        isinstance(value, CoroutineType)
+        # check for old-style generator based coroutine objects
+        or (
+            isinstance(value, GeneratorType)
+            and bool(value.gi_code.co_flags & CO_ITERABLE_COROUTINE)
+        )
+        # check for other awaitables (e.g. futures)
+        or hasattr(value, "__await__")
+    )
```
bellini666: **thought:** The difference between this and https://github.com/graphql-python/graphql-core/blob/main/src/graphql/pyutils/is_awaitable.py#L20-L20 is the fast path, right?

We can possibly just import it from graphql-core and `return graphql_core_is_awaitable(value)` here, so we don't need to redefine this "not so obvious" code

botberry:



---

Improved execution performance by up to 10% by adding an optimized `is_awaitable` check with a fast path for common synchronous types (such as int, str, list, dict, etc.).

This optimization reduces overhead when processing large result sets containing mostly basic values by avoiding expensive awaitable checks for types that are known to never be awaitable.



Here's the tweet text: 
```

```

bellini666 (approved):
Image

PR #4039, opened 2025-10-30
Change StrawberryConfig to be a dictionary instead of a class

patrick91:
I wanted to make it backwards compatibile, but not fully sure it is worth it, since the user changes should be pretty small


botberry:


<table>
	<thead>
		<tr>
			<th>Federation 1 Support</th>
			<th>Federation 2 Support</th>
		</tr>
	</thead>
	<tbody>
    <tr>
      <td>
        <table>
        </table>
      </td>
      <td>
        <table>
        </table>
      </td>
		</tr>
	</tbody>
</table>


Learn more:

strawberry/http/base.py
```diff
@@ -88,15 +88,15 @@ def _is_multipart_subscriptions(
...
+        if self.schema.config["batching_config"] is None:
             raise HTTPException(400, "Batching is not enabled")
 
         if protocol == "multipart-subscription":
             raise HTTPException(
                 400, "Batching is not supported for multipart subscriptions"
             )
 
-        if len(request_data) > self.schema.config.batching_config["max_operations"]:
+        if len(request_data) > self.schema.config["batching_config"]["max_operations"]:
```
erikwrede, line 99: I'm not sure about the visual noise we're adding here 😬 
But I love that the config is now a dict for the user, so I guess we've got to live with it internally :D

strawberry/schema/config.py
```diff
@@ -15,28 +14,155 @@ class BatchingConfig(TypedDict):
...
+        auto_camel_case: bool | None
+        name_converter: NameConverter
+        default_resolver: Callable[[Any, str], object]
+        relay_max_results: int
+        relay_use_legacy_global_id: bool
+        disable_field_suggestions: bool
+        info_class: type[Info]
+        enable_experimental_incremental_execution: bool
+        _unsafe_disable_same_type_validation: bool
+        batching_config: BatchingConfig | None
```
erikwrede, line 81: on the other hand, how would we make clear the default values in an accessible way? For the dataclass they were immediately visiblei n the type overlay. looking into `_complete_config` seems more complicated and less direct

strawberry/http/base.py
patrick91: yeah, it's not fun 😅

strawberry/schema/config.py
patrick91: I guess we need to make a custom language server? :P 

but probably better docs?

erikwrede: yea let's make sure everything is documented in here, maybe in a short table: https://strawberry.rocks/docs/types/schema-configurations#available-configurations

we _could_ auto generate that from the code as well with minor adjustments :P

docs/breaking-changes.md
```diff
@@ -4,6 +4,7 @@ title: List of breaking changes and deprecations
 
 # List of breaking changes and deprecations
 
+- [Version 0.285.0 - Deprecate StrawberryConfig Class - 30 October 2025](./breaking-changes/0.285.0.md)
```
bellini666, line 7: **polish:** this is not following the pattern of the ones below

PR #4047, opened 2025-11-10
Stop setting Transfer-Encoding: chunked in multipart responses (delegate to server/gateway; fixes serverless compatibility)

LouisAmon:
The motivation behind this PR is that despite https://github.com/strawberry-graphql/strawberry/commit/b8c965215d3917217c5969d1da684ac6a40e088d, Strawberry subscriptions still break on Vercel (HTTP 405) due to this extraneous header.

My own tests conclude that as it currently stands, both Starlette & Django views break on HTTP over multipart Subscriptions deployed to Vercel.

My current fix is to do this:
```py
    @override
    async def dispatch(
        self, request: HttpRequest, *args: tuple[Any, ...], **kwargs: dict[str, Any]
    ) -> HttpResponse:
        """Handle the request and strip Transfer-Encoding header from response.

        Vercel's serverless infrastructure doesn't support Transfer-Encoding: chunked.
        Django's StreamingHttpResponse adds this header automatically, so we remove it.

        Returns:
            The HTTP response with Transfer-Encoding header removed.

        """
        response = await super().dispatch(request, *args, **kwargs)

        # Strip Transfer-Encoding header if present (needed for Vercel compatibility)
        if (
            isinstance(response, StreamingHttpResponse)
            and "Transfer-Encoding" in response
        ):
            del response["Transfer-Encoding"]

        return response
```

...and it works. But since it took me several days of digging to actually understand the interaction between Django <--> Strawberry <--> Vercel and figure this out, I thought I'd contribute it directly to Strawberry

## Description

This PR removes Strawberry’s unconditional Transfer-Encoding: chunked header from multipart/mixed streaming responses (e.g., for subscriptions and incremental delivery). Transfer‑Encoding is hop‑by‑hop and should be controlled by the HTTP server/gateway. Emitting it from application code breaks on platforms that terminate connections or speak HTTP/2/HTTP/3 to clients, where the header is disallowed, while not being required by the GraphQL multipart protocol itself. The change preserves real‑time multipart streaming and improves compatibility with serverless providers (e.g., Vercel Functions).

### Background & Motivation
	•	The GraphQL‑over‑HTTP Incremental Delivery doc (used by multipart/mixed responses) states that under HTTP/1.1 a response should contain Transfer‑Encoding: chunked; under HTTP/2 it is disallowed. This is guidance tied to the transport, not an application‑level requirement.  ￼
	•	Transfer‑Encoding is a hop‑by‑hop header; proxies/CDNs/edge runtimes may add, remove, or transform it. Applications setting TE directly can create protocol mismatches when an intermediary selects a different transport (e.g., HTTP/2) or buffers/frames the response differently.  ￼
	•	Modern hosting (e.g., Vercel Python Functions) supports streaming without userland TE; streaming is enabled by default (Jan 5, 2025), so removing application‑set TE aligns with platform behavior.  ￼
	•	In traditional deployments, servers (uvicorn/gunicorn) decide whether to add TE for streaming responses when no Content‑Length is present. That’s the right layer to choose TE.  ￼

### Alternatives considered
	•	Make TE configurable (e.g., use_transfer_encoding_chunked=True/False/auto): adds complexity for a transport concern that the server should own.
	•	Set TE only when http_version == "1.1": still pushes a hop‑by‑hop header from app code and risks conflicts with intermediaries that upgrade/downgrade or re‑frame connections.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

*

## Checklist

## Summary by Sourcery

Remove explicit Transfer-Encoding: chunked header from multipart/mixed streaming responses to let the HTTP server or gateway manage chunked encoding, preserving streaming functionality while improving compatibility with HTTP/2, HTTP/3, and serverless platforms like Vercel.

Bug Fixes:
- Stop setting Transfer-Encoding: chunked on multipart streaming responses to prevent HTTP405 errors on Vercel and other environments.

Enhancements:
- Delegate hop-by-hop header management to the server or gateway to comply with HTTP transport rules and support modern hosting infrastructures.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+Removed `Transfer-Encoding: chunked` header from multipart streaming responses. This fixes HTTP 405 errors on Vercel and other serverless platforms. The server/gateway will handle chunked encoding automatically when needed.


botberry:



---

Removed `Transfer-Encoding: chunked` header from multipart streaming responses. This fixes HTTP 405 errors on Vercel and other serverless platforms. The server/gateway will handle chunked encoding automatically when needed.



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Louis Amon for the PR 👏

```

PR #4071, opened 2025-12-12
Support using enum values in the GraphQL schema

paulo-raca:
Currently, Strawberry exposes the enum name in GraphQL.

This PR adds the option to use the enum values instead.

## Why

I'm wrapping an existing OpenAPI service in GraphQL. There are tons of enums generated by the OpenAPI client that I want to reuse instead of writing again (also, they change often). They generated enums look like this:

```
class MyEnumType(str, Enum):
    FOO = 'foo'
    BAR = 'bar'
    ETC = 'etc'
```

Since the lowercase values are used in a ton of different places, including documentation and case-sensitive queries (e.g., `findEntities('myenum eq "foo"')`, I want them to stay the same everywhere.

This PR adds `@strawberry.enum(use_enum_values=True)`, so that the GraphQL mapping uses the enum values instead of names

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

## Checklist

## Summary by Sourcery

Add support for mapping GraphQL enum values from either the Python enum names or their underlying values, controlled via a new use_enum_values option.

New Features:
- Introduce a use_enum_values option on the @strawberry.enum decorator to emit GraphQL enums using their underlying Python values instead of names.

Enhancements:
- Preserve compatibility with existing enums while allowing custom GraphQL names and value-based mappings.
- Clarify enum mapping behavior in the documentation with guidance on using use_enum_values.

Tests:
- Expand enum tests to cover both default name-based behavior and the new value-based enum mapping.

### DIFF
--- a/strawberry/types/enum.py
+++ b/strawberry/types/enum.py
@@ -145,7 +145,7 @@ def _process_enum(
             if graphql_name_from == "key":
                 graphql_name = item_name
             elif graphql_name_from == "value":
-                graphql_name = item_value or "empty"
+                graphql_name = item_value
             else:
                 raise ValueError(f"Invalid mode: {graphql_name_from}")
 


botberry:



---

Support using enum values in the GraphQL schema



Here's the tweet text: 
```

```

botberry:

```markdown
Release type: patch

```

Here's the tweet text: 
```

```

docs/types/enums.md
```diff
@@ -135,7 +135,9 @@ Here is result of executed query:
 <Note>
 
 GraphQL types are not a map of name: value, like in python enums. Strawberry
-uses the name of the members of the enum to create the GraphQL type.
+defaults to using the name of the members of the enum to create the GraphQL
+type. You can use their values instead with
+`@strawberry.enum(use_enum_values=True)`
```
bellini666: **thought:** I'm thinking if maybe we should do something like `mode: 'value'`, with an actual enum or a literal like `mode: Literal['key', 'value']`, in case we end up having to support a third option in the future

@patrick91 thoughts?

strawberry/types/enum.py
```diff
@@ -135,13 +137,14 @@ def _process_enum(
...
             cls._member_map_[item_name]._value_ = item_value.value
 
-            if item_value.graphql_name:
-                item_name = item_value.graphql_name
-
+            graphql_name = item_value.graphql_name
             item_value = item_value.value
 
+        if not graphql_name:
+            graphql_name = item_value if use_enum_values else item_name
```
bellini666: **question:** is there anything that we need to validate here? i.e. can values be any type of objects? I think not, right?

docs/types/enums.md
paulo-raca: Good idea, done

strawberry/types/enum.py
paulo-raca: There is validation in place already. I've added a test

docs/types/enums.md
```diff
@@ -135,7 +135,8 @@ Here is result of executed query:
 <Note>
 
 GraphQL types are not a map of name: value, like in python enums. Strawberry
-uses the name of the members of the enum to create the GraphQL type.
+defaults to using the name of the members of the enum to create the GraphQL
+type. You can use their values instead with `@strawberry.enum(mode="value")`
```
patrick91:
```suggestion
type. You can use their values instead with `@strawberry.enum(graphql_name_from="value")`
```

I was chatting with @bellini666  about this and I think, while this is more verbose, it is clearer what it does, `mode` feels a bit off to me 😊


Also @paulo-raca would like having a global option for this (in addition to granular)?

paulo-raca: Yeah, this parameter is more clear. I've updated it.

> Also @paulo-raca would like having a global option for this (in addition to granular)?
Would be great. Any pointers on how to do that?

patrick91: > Any pointers on how to do that?

not really, maybe we can merge this and it later?

it's a bit annoying because we'd need to defer the creation of the enum to later on (when we have access to the schema config)

paulo-raca: Yeah, maybe later

paulo-raca:
@patrick91, @bellini666: Anything holding this?

PR #4080, opened 2025-12-28
docs: extend Maybe documentation and add a new page about UNSET

### LINKED ISSUE #4069: Use case overlap between `UNSET` and `Maybe`
## Feature Request Type

- [x] Core functionality
- [ ] New behavior

# Description

First of all, thanks a lot for this amazing library. We make great use of 🍓  for several of our `django` projects. 🍓 offers an abundance of features, and its extensibility is great.

## problem statement
With that said, we have run into some confusing behaviour around the `UNSET` scalar and `Maybe` type. It is unclear to us what unique functionality `UNSET` provides that `Maybe` does not cover. We regularly make use of the `Maybe` type, but parsing input types is complicated by the existence of `UNSET` as one of 🍓 's fallback value for some types.

### our thought process
There are three possible 'states' identified by the docs on `Maybe`: https://strawberry.rocks/docs/types/maybe#maybe
1. Field present with a value
2. Field present but explicitly null
3. Field completely absent.

In other words, `Maybe` is aimed at enabling (but not forcing) differentation between _nullable_ and _optional_.

Let's exhaustively list the possible types that we can construct since `Maybe` has been introduced:

| field type               | provided value | absent   | null |
|--------------------|----------------|------------|------------|
| int                | 1              | PROHIBITED | PROHIBITED |
| int \| None        | 1              | None       | None       |
| Maybe[int]         | Some(1)        | None       | PROHIBITED |
| Maybe[int \| None] | Some(1)        | None       | Some(None) |

We can see here that `absent` and `null` are considered separate states in row 3 and 4, but this is not forced, since row 2 also exists. Nowhere in this table do we see a need for `UNSET`.

The [docs for `UNSET`](https://strawberry.rocks/api-reference/strawberry.UNSET) descibe it as:

> A special value that can be used to represent an unset value in a field or argument. Similar to undefined in JavaScript, this value can be used to differentiate between a field that was not set and a field that was set to None or null . 

This seems like it does not cover any use case that is not covered by `Maybe`.

In contrast, we do find that `Maybe` covers use cases that are not covered by `UNSET`, such as https://github.com/strawberry-graphql/strawberry/issues/2144

## conclusion
Unless we are missing an important use case we did not mention above, we would request removal/deprecation of `UNSET`.

bellini666:
Fix #4069

## Summary by Sourcery

Document Maybe as the recommended pattern for optional input fields and introduce dedicated documentation for the legacy UNSET sentinel and migration guidance.

Documentation:
- Clarify Maybe usage patterns, including differences between Maybe[T] and Maybe[T | None], schema behavior, and update examples to use explicit None checks.
- Add a migration guide from UNSET to Maybe with side‑by‑side examples and notes on type safety and nullability semantics.
- Create a new UNSET documentation page describing its legacy status, usage, limitations, and comparison with Maybe.

### DIFF
--- a/docs/types/maybe.md
+++ b/docs/types/maybe.md
@@ -130,10 +130,15 @@ input UpdateUserInput {
 ```python
 @strawberry.input
 class UpdatePostInput:
-    title: strawberry.Maybe[str]  # Can be set to new value or omitted
+    # Maybe[T] - value or absent, null is INVALID
+    # Use when the field must have a value if provided
+    title: strawberry.Maybe[str]
     content: strawberry.Maybe[str]
     published: strawberry.Maybe[bool]
-    tags: strawberry.Maybe[list[str] | None]  # Can be set, cleared, or omitted
+
+    # Maybe[T | None] - value, null, or absent
+    # Use when null is a valid value (e.g., to clear/remove the field)
+    tags: strawberry.Maybe[list[str] | None]
 
 
 @strawberry.type
@@ -143,13 +148,13 @@ class Mutation:
         post = get_post(post_id)
 
         # Only update fields that were explicitly provided
-        if input.title:
+        if input.title is not None:
             post.title = input.title.value
-        if input.content:
+        if input.content is not None:
             post.content = input.content.value
-        if input.published:
+        if input.published is not None:
             post.published = input.published.value
-        if input.tags:
+        if input.tags is not None:
             post.tags = input.tags.value  # Could be None to clear tags
 
         return post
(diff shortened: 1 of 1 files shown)


docs/types/maybe.md
```diff
@@ -207,6 +220,89 @@ update_if_provided(user, "phone", input.phone)
...
+clearer semantics.
+
+### Why Migrate?
+
+`Maybe` provides several advantages over `UNSET`:
+
+1. **Type Safety**: `Maybe[T]` is a proper generic type that works well with
+   type checkers
+2. **Clearer Semantics**: `Some(value)` vs `None` is more explicit than checking
+   `value is UNSET`
```
patrick91: is it? 👀

docs/types/maybe.md
```diff
@@ -207,6 +220,89 @@ update_if_provided(user, "phone", input.phone)
...
+3. **Better Null Handling**: `Maybe[T | None]` explicitly supports the
+   three-state pattern (value, null, absent)
```
patrick91: same here, unset did that too

I think the main advantaged are type checking and better API design?

docs/types/maybe.md
bellini666: no 🤣

bellini666:
@sourcery-ai review

PR #4094, opened 2026-01-09
Add AGENTS.md

magicmark:
https://agents.md/ is now a standard supported by most coding agents.

Seems worth adding  - e.g. my instance of claude keeps trying to run things with uv, so this would teach it we're using poetry atm.

wdyt?

## Summary by Sourcery

Documentation:
- Add AGENTS.md documenting repository setup, development commands, project structure, and pull request requirements for automated coding agents and contributors.


AGENTS.md
```diff
@@ -0,0 +1,42 @@
...
+  - `types/` - GraphQL type definitions
+  - `extensions/` - Built-in extensions
+  - `django/`, `fastapi/`, `flask/`, etc. - Framework integrations
+- `tests/` - pytest suite with integration markers (`django`, `fastapi`, `pydantic`, etc.)
+
+## PR Requirements
+
+- Include `RELEASE.md` file describing changes
+- Release types: patch/minor/major
+- Tests required for all code changes with full coverage
```
patrick91: shall we add an example of release file too? 😊

AGENTS.md
```diff
@@ -0,0 +1,42 @@
...
+
+Strawberry GraphQL
+
+- Docs: https://strawberry.rocks/docs
+- Repo: https://github.com/strawberry-graphql/strawberry
+
+## Setup
+
+```shell
+$ poetry install                    # Install dependencies
```
bellini666: maybe?

```suggestion
$ poetry install --with integrations                  # Install dependencies
```

AGENTS.md
patrick91: let's also add a CLAUDE.md with `@AGENTS.md` 😊

XChikuX:
I'd recommend making the AGENTS.md significantly more detailed Eg. Almost 90% of the file tree. This helps CLAUDE and other agents gather all the needed context right out of the bat. 

It's important not to go overboard and use up too much context at the same time by being too wordy. Including the `pre-commit` checks would help the agent avoid reading the README.md and using up too much context for every single commit.

I'd also recommend a [.serena](https://github.com/oraios/serena) folder with memories of important incomplete tasks and possibly  a `skills` folder for Claude with python concepts that you'd want to be used in strawberry.

patrick91:
@XChikuX I'd avoid serena, I haven't found it to be that useful, do you have an example CLAUDE.md?

XChikuX:
[Full Stack Application Example](https://gist.github.com/XChikuX/9f233a940bfe90b619d85ae579116108) You may want to adjust for tools available to opensource as well as pure python project structure like strawberry. **Caveat:** this is only one way in a vast amount of ways you can structure your `CLAUDE.md` I do not claim it is the best way, I've just seen good things come out of this general structure.

>  I'd avoid serena

It aids LLMs that are not as good with tool calling as Claude. 

I've found it noticeably reduces context bloat when I prompt agents as well in exchange for some accuracy loss. Of course, your mileage may vary.

magicmark:
@XChikuX i'm strongly in favour of the most minimal AGENTS.md/CLAUDE.md file possible. These instructions get added to the context window and added to every prompt. Token space is at a premium.

Information that is already easily discernible is redundant (such as the directory layout, general intent of certain files). I would reserve the token count for common stumbling blocks on the most common actions. (and I would expect the need for such instructions to diminish over time too as the models improve!)

AGENTS.md
magicmark: > shall we add an example of release file too? 😊

would be cool to make this a [skill](https://agentskills.io/home)

XChikuX:
> These instructions get added to the context window and added to every prompt

I'd argue more context get's used in searching for files, understanding file structure and repeat API calls that lack much needed context. I assumed the same as you in the beginning, but my real world use has been counter intuitive.

The only time CLAUDE didn't waste precious tokens by creating new unnecessary files was when it already knew the file existed and exactly what that file did.

> and I would expect the need for such instructions to diminish over time too as the models improve!

We can always edit it when that time comes. In fact, I recommend we do. :)

EDIT:

Keeping AGENTS.md small would sacrifice input token context in exchange for lower quality output tokens (which cost x10+) because of the extra tool calls and the unnecessary *Now I will figure out what this file does* outputs.

Although it will help free tiers, it will not help Claude from my experience. Instead of pointing CLAUDE.md - > AGENTS.md consider keeping both separate since Claude more or less ignores the latter, you can find the issue on their github that proves this. 

Larger input context for Claude. Smaller context for free tier LLMs would be the way to go. 

Just a suggestion @patrick91

erikwrede:
I'm with @magicmark and @patrick91 on this one. However, we should think about adding _some_ knowledge on structure:

>  I would reserve the token count for common stumbling blocks on the most common actions.

This is what we need. however, that's something we need to add piece by piece when we actually identify the stumbling blocks. So maybe we all add whatever we frequently see misinterpreted by claude code. For the rest, claude's plan / explore agents do an exceptional job at understanding the codebase nowadays.

AGENTS.md
erikwrede: same thing goes for schema tests for a new skill, think of a new `/generate-tests` command  which generates the test file structure, schema, query inputs and expected outputs.

erikwrede (approved):
awesome!

bellini666:
@patrick91 @magicmark what are we missing to get this merged?

XChikuX:
Seems really minimal and non-intrusive. We should merge this as it would provide at least bare minimal context for AI to work with.

AI always seems to skip the lint, ruff and mypy checks for last, or ignores it until prompted.  So this looks good to me.

I'd still recommend that at least `CLAUDE.md` should be significantly bigger and cover the full working tree with file summaries, due to what I mentioned above:

> sacrifice input token context in exchange for lower quality output tokens (which cost x10+) because of the extra tool calls

But I'll leave that up to you folks..

botberry:

```markdown
---
release type: patch
---

```

PR #3331, opened 2024-01-10
fix #3129, simply provide a default

devkral:
## Description

Quite simple solution for a nasty bug:

because a default or default_factory is  missing, dataclasses requires for relay.node annotations on subtypes (see bugreport) an initial value.

The reason why I didn't use default_factory in the tests is, that there is a bug preventing this (not scope of the PR)

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement/optimization
- [ ] Documentation

## Issues Fixed or Closed by This PR

* #3129

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the CONTRIBUTING document.
- [x] I have added tests to cover my changes.
- [x] I have tested the changes and verified that they work and don't break anything (as well as I can manage).

PR #4140, opened 2026-01-14
Fix relay.node() requiring explicit initialization on nested types

### LINKED ISSUE #3129: relay.node on subtype needs initialization
Consider the code

```python

@strawberry.type
class Sub:
  node: relay.Node =  relay.node()
  nodes: list[relay.Node] =  relay.node()

@strawberry.type
class Query:
  @strawberry.field
  @staticmethod
  def sub():
    return Sub

schema = Schema(query=Query)
```
This code should work but doesn't, it requires an initialization for node

Note there is a workaround (initialization with None or empty list, or (I think whatever))

``` python
...
  @strawberry.field
  @staticmethod
  def sub():
    return Sub(node=None, nodes=None)
```

## System Information

 - Operating system:
 - Strawberry version (if applicable): since strawberry-django-plus times after a rewrite, there is a bugreport in the old repo

patrick91:
## Summary

This PR fixes an issue where `relay.node()` fields on nested types required explicit initialization, even though the field values are resolved dynamically.

Fixes #3129

**This is a fresh implementation of the fix from #3331**, which was a year old and had merge conflicts. The original PR by @devkral had the right approach but:
1. The branch was 621 commits behind main
2. The RELEASE.md filename had a typo (`:` instead of `.`)
3. The release type was `minor` instead of `patch`

### The Problem

Previously, this code would fail at runtime:

```python
@strawberry.type
class Sub:
    node: relay.Node = relay.node()

@strawberry.type
class Query:
    @strawberry.field
    def sub(self) -> Sub:
        return Sub()  # Error: missing required argument 'node'
```

The workaround was to explicitly pass `node=None`:
```python
return Sub(node=None, nodes=None)
```

### The Fix

`relay.node()` now provides a default value (`None`), which serves as a stub for dataclasses. The actual value is still resolved by the `NodeExtension` resolver at runtime.

## Test Plan

- [x] Added regression test `test_relay_node_on_nested_type`
- [x] All 232 relay tests pass

### DIFF
--- a/strawberry/relay/fields.py
+++ b/strawberry/relay/fields.py
@@ -41,6 +41,7 @@ from strawberry.types.cast import cast as strawberry_cast
 from strawberry.types.field import _RESOLVER_TYPE, StrawberryField, field
 from strawberry.types.fields.resolver import StrawberryResolver
 from strawberry.types.lazy_type import LazyType
+from strawberry.types.unset import UNSET
 from strawberry.utils.aio import asyncgen_to_list
 from strawberry.utils.typing import eval_type, is_generic_alias, is_optional, is_union
 
@@ -375,7 +376,7 @@ if TYPE_CHECKING:
     node = field
 else:
 
-    def node(*args: Any, default: Any = None, **kwargs: Any) -> StrawberryField:
+    def node(*args: Any, default: Any = UNSET, **kwargs: Any) -> StrawberryField:
         kwargs["extensions"] = [*kwargs.get("extensions", []), NodeExtension()]
         # The default value is a stub for dataclasses so you can instantiate
         # types with relay.node() fields without explicit initialization.


botberry:



---

This release fixes an issue where `relay.node()` fields on nested types required
explicit initialization, even though the field values are resolved dynamically.

Previously, this code would fail at runtime:

```python
@strawberry.type
class Sub:
    node: relay.Node = relay.node()


@strawberry.type
class Query:
    @strawberry.field
    def sub(self) -> Sub:
        return Sub()  # Error: missing required argument 'node'
```

Now `relay.node()` provides a default value, allowing nested types with relay
node fields to be instantiated without explicit initialization.



Here's the tweet text: 
```

```

patrick91:
Closing in favor of pushing directly to #3331

PR #4156, opened 2026-01-20
Fix lazy union types using new Annotated syntax (issue #4066)

### LINKED ISSUE #4066: ValueError when updating union types syntax per deprecation warning
## Describe the Bug

I have a piece of code that looks like this:

```python
MyType = strawberry.union(name="MyType", types=(A, B, C))
```

This yields a deprecation warning and the suggested syntax is the following:

```python
MyType = Annotated[Union[A, B, C], strawberry.union("MyType")]
```

But changing to that syntax gives me the following error:

`Union type MyType must define one or more member types.`

I have tried many different variants on lazy-loading the types instead, using `strawberry.union()` vs `typing.Union` and many other things, but I cannot get past this error.

## System Information

 - Operating system: macOS 15.6.1 (24G90)
 - Python version: 3.12
 - Strawberry version (if applicable): 0.284.2

patrick91:
When using the new recommended union syntax without explicit types argument:
  `MyUnion = Annotated[Union[A, B], strawberry.union("MyUnion")]`

The union would fail with "must define one or more member types" error
when used with lazy loading:
  `LazyUnion = Annotated["MyUnion", strawberry.lazy("module")]`

The issue was in `from_type()` which directly extracted the StrawberryUnion
from Annotated args. When created via the new syntax (without types=),
the StrawberryUnion has empty type_annotations. The fix now properly
extracts types from the Union[...] in the first Annotated argument and
creates a properly populated StrawberryUnion.

Added tests for:
- New syntax with pipe operator (TypeA | TypeB)
- New syntax with typing.Union
- Single-type unions (Union[TypeA])

## Summary by Sourcery

Fix lazy-loaded unions defined with the new Annotated union syntax so they resolve correctly through LazyType and can be converted to GraphQL types.

Bug Fixes:
- Resolve lazy unions created with the new Annotated union syntax (pipe and typing.Union forms) without requiring an explicit types= argument.

Tests:
- Add regression tests for lazy unions using the new Annotated union syntax with pipe operator and typing.Union to verify schema generation.

Closes #4066

## Summary by Sourcery

Fix resolution of lazy-loaded unions defined with the new Annotated union syntax so they can be correctly converted to GraphQL types.

Bug Fixes:
- Ensure lazy unions defined via Annotated union syntax (including single-type and typing.Union forms) no longer fail with missing member types during schema conversion.

Tests:
- Add regression tests for lazy unions using the new Annotated union syntax with pipe and typing.Union forms to verify correct schema generation.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,17 @@
+Release type: patch
+
+This release fixes an issue where lazy union types using the new `Annotated` syntax were not being resolved correctly.
+
+Example that now works:
+
+```python
+from typing import Annotated
+
+import strawberry
+
+
+@strawberry.type
+class Query:
+    @strawberry.field
+    def example(self) -> Annotated["SomeUnion", strawberry.lazy("module")]: ...
+```


patrick91:
@bellini666 just a release file :D I'll do it later

botberry:



---

This release fixes an issue where lazy union types using the new `Annotated` syntax were not being resolved correctly.

Example that now works:

```python
from typing import Annotated

import strawberry


@strawberry.type
class Query:
    @strawberry.field
    def example(self) -> Annotated["SomeUnion", strawberry.lazy("module")]: ...
```



Here's the tweet text: 
```

```

PR #4179, opened 2026-01-29
Allow strawberry.auto to override all_fields=True | Allow Private Fields to work with strawberry.auto (all_fields=True is pending)

XChikuX:
## Description

Make `all_fields=True` respect explicit field definitions. Allow initial support for `strawberry.Private` types combined with `strawberry.auto`

Currently if you want to use private types it with `all_fields=True` you must provide a default value.

```py
@pyd_type(model=RedisUser, all_fields=True)  # If you remove all_fields=True, it does not error
class User:
    pk: strawberry.auto
    fname: strawberry.auto
    auth: strawberry.Private[identity | None] = None  # WORKS
    bio: strawberry.Private[str] = "Welcome to my App"  # WORKS
    verified: strawberry.Private[
        bool
    ]  # Causes User.__init__() missing 1 required keyword-only argument: 'verified'"
```

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

Unreported Issue

## Checklist

## Summary by Sourcery

Respect explicit field definitions and private fields when generating experimental Pydantic-based Strawberry types, and ensure private values are correctly populated and extensions honored.

New Features:
- Support handling of strawberry.Private fields when converting from Pydantic models, including population from model instances and extra data.
- Support field extensions and explicit type overrides on fields in experimental Pydantic types, even when using all_fields=True.

Bug Fixes:
- Ensure all_fields=True no longer overrides explicit field definitions and no longer emits warnings for their combined use.
- Prevent fields explicitly marked as strawberry.Private from being exposed in the GraphQL schema even if present on the underlying Pydantic model.

Enhancements:
- Preserve extensions, metadata, and explicit typing for fields derived from Pydantic models in experimental Strawberry types.

Documentation:
- Add release notes documenting the new behaviour for all_fields=True, private fields, and field extensions with experimental Pydantic types.

Tests:
- Add comprehensive tests covering private fields with from_pydantic, interaction with all_fields=True, field extensions (sync and async), and explicit field overrides in experimental Pydantic types.


botberry:



---

This release fixes issues with explicit field definitions in experimental pydantic types:

**`all_fields=True` now respects explicit field definitions**: Previously, using `all_fields=True` would override any explicitly defined fields in the strawberry type. Now, explicit definitions take precedence, allowing you to:

- Mark fields as `strawberry.Private` to hide them from the GraphQL schema
- Add field extensions (e.g., for authentication, caching, transformations)
- Override field types
- Add custom resolvers

The warning "Using all_fields overrides any explicitly defined fields" has been removed since combining `all_fields=True` with explicit definitions is now a valid and useful pattern.

**Private fields are auto-populated from pydantic models**: When a `strawberry.Private` field has the same name as a pydantic model field, `from_pydantic()` will automatically populate it from the model (can be overridden via the `extra` dict).

Example:

```python
import strawberry
from strawberry.experimental.pydantic import type as pyd_type
from strawberry.extensions.field_extension import FieldExtension


class MaskExtension(FieldExtension):
    def resolve(self, next_, source, info, **kwargs):
        result = next_(source, info, **kwargs)
        return result[:3] + "****" if result else result


class UserModel(BaseModel):
    name: str
    email: str
    password: str


@pyd_type(model=UserModel, all_fields=True)
class User:
    # Add extension to mask email in responses
    email: str = strawberry.field(extensions=[MaskExtension()])
    # Hide password from GraphQL schema entirely
    password: strawberry.Private[str]


# name and email are exposed in schema, password is not
# email will be masked by the extension
pydantic_user = UserModel(name="Alice", email="alice@example.com", password="secret")
strawberry_user = User.from_pydantic(pydantic_user)

# Private field is still accessible internally
print(strawberry_user.password)  # "secret"
```



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to @Seetku for the PR 👏

```

XChikuX:
@patrick91 Would love to know your thoughts.

Several areas of complexity arise cause we differentiate how `strawberry.auto` and `all_fields=True` do.
We should unify them in effect just making `all_fields` perform `strawberry.auto` operations on each fields if possible.

I already have a fix for the bug I've induced in the what I've specified above. And I'd like to add it in another PR since it is very minor and already has a workaround, plus I'm not happy with the solution and would like to see if the unification way is possible first.

PR #4180, opened 2026-01-31
 fix: Ensure closing multipart boundary emitted in Multipart Subscription Stream

ymilki:
## Description

Addresses a race condition in `AsyncBaseHTTPView._stream_with_heartbeat` that causes the final multipart boundary to be lost, resulting in malformed responses and unpredictable client behavior.

I discovered this race condition when testing some added additional behaviour to the streaming response. Tests were inconsistently seeing proper streams, revealing the race condition.
```
   def _generate_subscription_stream(
        self,
        stream: Callable[[], AsyncGenerator[str, None]],
    ) -> Generator[bytes, None, None]:
        loop = asyncio.new_event_loop()
        asyncio.set_event_loop(loop)
        async_gen = stream()

        try:
            while True:
                try:
                    chunk = loop.run_until_complete(async_gen.__anext__())
                   <.... do stuff ...>
                    yield chunk.encode("utf-8")
                except StopAsyncIteration:
                    break
        finally:
            loop.run_until_complete(async_gen.aclose())
            loop.close()

    async def create_streaming_response(
        self,
        request: Request,
        stream: Callable[[], AsyncGenerator[str, None]],
        sub_response: Response,
    ) -> Response:
        """Create streaming response for GraphQL subscriptions."""
        <.... do stuff ...>
        sub_response.app_iter = self._generate_subscription_stream(stream=stream)
        <.... do stuff ...>
        return sub_response
```

🤖 PR created with the help of [Cursor](https://cursor.com/)

### The Problem

The race occurs in the `merged()` async generator within `_stream_with_heartbeat` (lines 570-580 of `async_base_view.py`):

**Race window:**

1. The drain task queues the final boundary: `await queue.put((False, False, final_boundary))`
2. The drain task queues completion: `await queue.put((False, True, None))`
3. The drain task completes, making `task.done()` return `True`
4. **RACE:** If `task.done()` is checked before `await queue.get()` is called, the loop exits
5. The final boundary remains in the queue, never yielded

The existing code (#3866) comments acknowledge this race ([lines 499-507](https://github.com/strawberry-graphql/strawberry/blob/414e8fd82f0ca73bde911cc166341fd26c649bfe/strawberry/http/async_base_view.py#L499-L507), [576-581](https://github.com/strawberry-graphql/strawberry/blob/414e8fd82f0ca73bde911cc166341fd26c649bfe/strawberry/http/async_base_view.py#L576-L581)) but assume the queue has been fully consumed when `task.done()` is True - this assumption doesn't hold when the message is still queued at the time of the check.

### Impact

- **Clients wait until timeout** - The incomplete response causes clients to wait until timeout or connection closure
- **Malformed multipart response** - Violates RFC 2046 multipart/mixed specification
- **Unpredictable client behavior** - Different clients handle the missing boundary differently (errors, warnings, or silent failures)

The race is more likely to occur under:
- High system load or CPU contention
- Rapid subscription completions (fast-yielding generators)

### The Fix

The fix tracks whether the closing boundary has been yielded and manually appends it if missing:

**Implementation Details:**
- Added `closing_boundary` and `seen_closing_boundary` flag at the start of `merged()`
- Check each chunk for the closing boundary pattern as it's yielded
- After the loop exits (including via the race condition), check if boundary was seen
- If not seen, manually append the closing boundary to terminate the response

This approach:
- ✅ Detects when the final boundary was queued but not yielded due to the race
- ✅ Manually appends the boundary when missing
- ✅ Ensures multipart responses are always properly terminated
- ✅ Has no impact when the boundary was already included (idempotent)
- ✅ Maintains backward compatibility

#### New Test (`tests/http/incremental/test_subscription_race_condition.py`)

**`test_subscription_race_condition_missing_boundary`** - Uses `MockRacyAsyncGenerator` to simulate the race condition by intercepting the `stream()` function passed to `_stream_with_heartbeat` and dropping the final boundary chunk before it enters the internal queue. This deterministically reproduces the race that's difficult to trigger naturally in tests.

The test:
- Patches `_stream_with_heartbeat` to wrap the input `stream()` function with the mock
- The mock intercepts chunks from the stream and raises `StopAsyncIteration` when it sees the closing boundary
- This simulates the boundary being lost before reaching the internal queue in `merged()`
- Runs subscriptions via GET and POST methods across multiple HTTP client implementations
- Verifies the final boundary is present in the response (will fail without the fix)
- Confirms the response ends with the proper closing boundary `--graphql--`

#### Why Mocking is Necessary

The race condition is non-deterministic and depends on asyncio task scheduling, making it nearly impossible to trigger reliably in test environments. The `MockRacyAsyncGenerator` wrapper simulates the exact timing where the stream yields the closing boundary but it gets lost before being added to the queue, which is functionally equivalent to the race where it's queued but not retrieved before `task.done()` returns True.

Normal subscription behavior (without the race) is already covered by existing tests in `test_multipart_subscription.py`.

## Types of Changes

- [x] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

Fixes a race condition in subscription responses that causes clients to timeout waiting for the closing multipart boundary.

## Checklist

## Summary by Sourcery

Ensure multipart subscription HTTP responses are always terminated with the closing boundary to avoid malformed responses caused by a race condition in AsyncBaseHTTPView._stream_with_heartbeat.

Bug Fixes:
- Guard against a race condition in multipart subscription streaming that could drop the final boundary and produce malformed HTTP responses.

Documentation:
- Add release notes describing the multipart subscription boundary race condition and its fix.

Tests:
- Add a regression test that deterministically simulates the subscription boundary race condition across GET and POST incremental HTTP clients to verify the closing boundary is always present.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,6 +2,6 @@ Release type: patch
 
 Fixed a race condition in `AsyncBaseHTTPView._stream_with_heartbeat` that could cause the final multipart boundary to be lost in subscription responses. This race condition resulted in malformed HTTP responses where clients would wait until timeout, as the response was never properly terminated per RFC 2046 multipart/mixed specification.
 
-The fix ensures that the closing boundary (`--graphql--`) is always present in multipart responses, even when the race condition causes the final boundary chunk to remain in the queue when the drain task completes.
+The fix ensures that the closing boundary (`--graphql--`) is always present in multipart responses, even when the race condition leaves the final boundary chunk unprocessed in the queue when the drain task completes.
 
 This particularly affected subscriptions under high system load or with fast-yielding generators, where the timing of asyncio task scheduling could cause `task.done()` to return `True` before the final boundary was retrieved from the queue.


botberry:



---

Fixed a race condition in `AsyncBaseHTTPView._stream_with_heartbeat` that could cause the final multipart boundary to be lost in subscription responses. This race condition resulted in malformed HTTP responses where clients would wait until timeout, as the response was never properly terminated per RFC 2046 multipart/mixed specification.

The fix ensures that the closing boundary (`--graphql--`) is always present in multipart responses, even when the race condition leaves the final boundary chunk unprocessed in the queue when the drain task completes.

This particularly affected subscriptions under high system load or with fast-yielding generators, where the timing of asyncio task scheduling could cause `task.done()` to return `True` before the final boundary was retrieved from the queue.



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to milki for the PR 👏

```

strawberry/http/async_base_view.py
```diff
@@ -583,10 +585,19 @@ async def cancel_tasks() -> None:
                         await cancel_tasks()
                         raise data
 
+                    if closing_boundary in data:
```
magicmark, line 591: what if a field in the graphql response contains the string "--graphql--"? (from user submitted content or whatever)

this wouldn't be escaped here i don't think?

maybe we can bound it by newlines

ymilki: Using the encoded boundary line in the same style as `encode_multipart_data`

magicmark (commented):
This PR observes the race condition and patches behaviour

Would it be possible to prevent this happening in the first place?

tests/http/incremental/test_subscription_race_condition.py
```diff
@@ -0,0 +1,160 @@
...
+@pytest.fixture
+def subscription_schema():
+    """Schema with a simple subscription for race condition testing."""
+    return strawberry.Schema(
+        query=Query,
+        subscription=Subscription,
+    )
+
+
+class MockRacyAsyncGenerator:
```
magicmark, line 66: Rather than _simulating_ the race condition (which may or may not actually occur in real life) could we write a regression test that actually _triggers_ the race condition with some theoretically valid usage of the api?

ymilki:
I looked further into this. We are running on an older version of strawberry (0.262.6). The mentioned PR with the multipart fix was introduced in 0.270.1. Applying the changes to `_stream_with_heartbeat` to our local version did fix our tests.

The key is the introduction of the `done` signal` and setting `maxsize=1`.

So, this patch isn't necessary as the existing fix solved this issue.

PR #4181, opened 2026-02-01
fix: ApolloTracingExtension resolvers field not populated on subsequent requests

### LINKED ISSUE #3656: `ApolloTracingExtension` regression populating `resolvers` field.
## Describe the Bug
```python
import asyncio
import strawberry
from strawberry.extensions import tracing

@strawberry.type
class Query:
    @strawberry.field
    def node(self) -> str:
        return ''

schema = strawberry.Schema(Query, extensions=[tracing.ApolloTracingExtension])

for i in range(2):
    result = asyncio.run(schema.execute('{ node }'))
    assert result.extensions['tracing']['execution']['resolvers'], i  # second time fails
```
outputs
```console
    assert result.extensions['tracing']['execution']['resolvers'], i  # second time fails
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
AssertionError: 1
```
Only the first request populates `resolvers` as expected. This worked in previous version.

## System Information
 - Strawberry version (if applicable): >=0.240

bellini666:
Fix async execute to use cached _async_extensions instead of get_extensions(), which was creating new extension instances while the middleware manager retained the old ones. Also clear _resolver_stats at the start of each operation since extensions are now reused across requests.

Closes #3656

## Summary by Sourcery

Ensure Apollo tracing consistently records resolver data across repeated schema executions.

Bug Fixes:
- Fix ApolloTracingExtension so resolver metrics are correctly populated on every async execution when extensions are reused.
- Reset resolver statistics at the start of each traced operation to avoid leaking resolver data between requests.

Tests:
- Add async and sync tests verifying that Apollo tracing resolvers are populated on multiple consecutive executions of the same schema.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,8 @@
 Release type: patch
 
-Fix `ApolloTracingExtension` not populating the `resolvers` field after the first request.
+This release fixes a bug where the async `execute` method was creating
+new extension instances on every request (via `get_extensions()`),
+instead of reusing cached instances like the sync `execute_sync` method
+already did. This caused extensions that accumulate state across the
+execution lifecycle (such as `ApolloTracingExtension`) to lose their
+state between requests when using async execution.


botberry:



---

This release fixes a bug where the async `execute` method was creating
new extension instances on every request (via `get_extensions()`),
instead of reusing cached instances like the sync `execute_sync` method
already did. This caused extensions that accumulate state across the
execution lifecycle (such as `ApolloTracingExtension`) to lose their
state between requests when using async execution.



Here's the tweet text: 
```

```

strawberry/schema/schema.py
```diff
@@ -570,7 +570,7 @@ async def execute(
             operation_name=operation_name,
             operation_extensions=operation_extensions,
         )
-        extensions = self.get_extensions()
```
patrick91, line 573: This is a change in behaviour right? we might want to put this in the release notes

RELEASE.md
```diff
@@ -0,0 +1,3 @@
+Release type: patch
+
+Fix `ApolloTracingExtension` not populating the `resolvers` field after the first request.
```
patrick91: Can we update this with something like:

```
This release fixes a bug where extensions...
```

strawberry/schema/schema.py
bellini666: Yes, but at the same time, it is a "fix" in a sense that the sync version was using the `cached_property`, but the async version wasn't, probably due to an oversight

But let me add to the release notes

PR #4186, opened 2026-02-02
chore(test): looser types for test client responses

millar:
To make it easier to make assertions against test data I propose loosening the types of `Response`'s `data` and `extensions` fields to prevent type issues such as this when accessing nested fields:

```
error[not-subscriptable]: Cannot subscript object of type `object` with no `__getitem__` method
    |
345 |     assert response.data is not None
346 |     edges = response.data["viewer"]["me"]["organisationInvitations"]["edges"]
```

## Summary by Sourcery

Enhancements:
- Relax type annotations for Response.data and Response.extensions in the test client to use Any for contained values, avoiding restrictive object typing in test code.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+Release type: patch
+
+This update relaxes type annotations for `Response.data` and `Response.extensions` from `dict[str, JsonValue]` to `dict[str, JsonValue]` in the test client, making assertions in tests easier to write.


botberry:



---

This update relaxes type annotations for `Response.data` and `Response.extensions` from `dict[str, JsonValue]` to `dict[str, JsonValue]` in the test client, making assertions in tests easier to write.



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Sam Millar for the PR 👏

```

bellini666:
Should we just maybe define a JsonValue (i.e. `JsonValue: TypeAlias = str | int | float | bool | list["JsonValue"] | dict[str, "JsonValue"]`) and use it instead of `Any`?

bellini666 (commented):
We are just missing a `RELEASE.md` file, otherwise this will not get released:

https://github.com/strawberry-graphql/strawberry/pull/4186#issuecomment-3834480364

millar:
Thanks for merging! I think we may have to go back to `Any` over `JsonValue` because otherwise subscripting still fails type checks:

```
error[not-subscriptable]: Cannot subscript object of type `float` with no `__getitem__` method
```

😞

bellini666:
@millar that's weird. Can you share what code caused that?

millar:
It's if you want access to nested values, e.g. `response.data["viewer"]["me"]`, where the JsonValue type says `response.data["viewer"]` could be a float, int, etc.

bellini666:
@millar ohh I see. Yep, lets go with `Any`

Would you like to open a new "patch"/"fix" PR to fix this?

PR #4192, opened 2026-02-05
fix(schema-codegen): ensure types are generated before their references

### LINKED ISSUE #4191: schema-codegen: Types implementing interfaces generated at end of file cause NameError when referenced by other types
## Describe the Bug

When using `strawberry schema-codegen` to generate Python code from a GraphQL schema, types that **implement an interface** are emitted at the **end of the generated file** (after the `schema = strawberry.Schema(...)` line), regardless of where they appear in the schema or whether other types reference them.

If another type in the schema has a field that references such an implementing type, that referencing type is generated in schema order, i.e. _before_ the implementing type. At runtime, Python then tries to resolve the type hint when the referencing class is defined, but the implementing type is not defined yet, leading to a **NameError**.

### Minimal reproduction

**GraphQL schema:**

```graphql
interface Bar {
    bar: String!
}

type Foo implements Bar {
    bar: String!
}

type FooContainer {
    foo: Foo!
}
```

**Generated Python (simplified):**

- `Bar` (interface) is generated in order.
- `FooContainer` is generated next (because it follows schema order), with `foo: Foo`.
- Later in the file, after `schema = strawberry.Schema(...)`, `Foo` is generated.

So the generated file looks conceptually like:

```python
@strawberry.interface
class Bar:
    bar: str

@strawberry.type
class FooContainer:
    foo: Foo   # NameError: name 'Foo' is not defined

# ... schema = strawberry.Schema(...) ...

@strawberry.type
class Foo(Bar):
    bar: str
```

**Result:**

```
NameError: name 'Foo' is not defined
```

when importing the generated module (e.g. when `FooContainer` is defined and Python evaluates the annotation `foo: Foo`).

### Expected behaviour

Implementing types (e.g. `Foo`) should be generated **before** any type that references them (e.g. `FooContainer`), so that all names are defined when the module is loaded. In other words, the codegen should topologically sort generated types so that dependencies (including interface implementers) are defined before their dependents.

## System Information

- Operating system: macOS 26.2
- Python version: 3.13
- Strawberry version: strawberry-graphql[cli,fastapi]>=0.291.0
- Command: `strawberry schema-codegen schema.graphqls -o schema.py`

sanlil:
### Problem

`schema-codegen` can produce output that raises `NameError` at import time when types are referenced before they are defined. This PR addresses two such cases:

1. **Forward references in field annotations:** A type like `FooContainer` with a field `foo: Foo` can be emitted before `Foo` is defined.
2. **Union definitions:** A union like `FooOrBar = Annotated[Foo | Bar, strawberry.union(...)]` can be emitted before its member types `Foo` and `Bar` are defined.

### Solution

For field annotations, the generated code now includes `from __future__ import annotations` ([PEP 563](https://peps.python.org/pep-0563/)), which stores annotations as strings instead of evaluating them at class definition time, so referenced names don't need to exist yet.

For unions, `from __future__ import annotations` cannot help since `FooOrBar = Annotated[Foo | Bar, ...]` is a runtime expression, not an annotation. Union member types are now declared as dependencies in the topological sort, so unions are always emitted after their members.

Additionally, the schema assignment (`schema = strawberry.Schema(...)`) is now emitted last to ensure all types are defined before strawberry resolves their annotations.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* fixes #4191

## Checklist

## Summary by Sourcery

Ensure schema-codegen emits generated types in a dependency-safe order so referenced types are always defined before the types that reference them, and the schema assignment is emitted last.

Bug Fixes:
- Prevent NameError during import of generated schema-codegen modules by ordering interface implementations and referenced types before their dependents.
- Ensure union types are generated after their member types to avoid forward-reference issues.

Documentation:
- Add release notes describing the schema-codegen emit-order fix and its impact.

Tests:
- Add tests verifying that implementing types are emitted before referencing types and that the generated module imports without NameError.
- Adjust union codegen test expectations to reflect the new ordering where union definitions follow their member types.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,4 +1,4 @@
-Release type: patch
+Release type: minor
 
 Fix two `NameError` issues in schema-codegen output when types are referenced before they are defined.
 


botberry:



---

Fix two `NameError` issues in schema-codegen output when types are referenced before they are defined.

First, forward references in field annotations (e.g. `foo: Foo` appearing before `Foo` is defined) are now handled by emitting `from __future__ import annotations` at the top of the generated file. Per PEP 563, this stores all annotations as strings instead of evaluating them at class definition time, so the referenced names don't need to exist yet.

Second, union definitions like `FooOrBar = Annotated[Foo | Bar, strawberry.union(...)]` are runtime expressions that `from __future__ import annotations` cannot defer. These are now correctly ordered by declaring union member types as dependencies, so unions are always emitted after their members.

Changes:
- Emit `from __future__ import annotations` in generated code to handle forward references in field annotations.
- Add member-type dependencies for union definitions so unions are emitted after their member types.
- Ensure the schema assignment is emitted last by giving it dependencies on all other definitions.



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Sandra Liljeqvist for the PR 👏

```

RELEASE.md
```diff
@@ -0,0 +1,9 @@
+Release type: patch
+
+Fix schema-codegen emit order so types that implement interfaces (or are referenced by other types) are generated before types that reference them. Previously, types like `Foo implements Bar` were emitted at the end of the file after `schema = strawberry.Schema(...)`, so types such as `FooContainer { foo: Foo! }` were defined before `Foo`, causing `NameError: name 'Foo' is not defined` when importing the generated module.
```
patrick91: wouldn't this be fixed with `from __future__ import annotations` (and automatically in the latest python version?)

sanlil: Great point! I've updated the fix to use `from __future__ import annotations` instead. 

However, the issue that originally triggered this also involved union definitions, which are evaluated immediately regardless of annotation semantics. I've updated the PR description, RELEASE.md and code to cover this case separately.

sanlil: @patrick91 please let me know if you still have concerns or comments!

sanlil: @patrick91 my team has an ongoing issue with the ordering of unions and we currently solve it with a custom script on top of the codegen. It would be great to get this fixed in strawberry instead though! Is there anything you'd like to discuss about my suggested approach? I'm of course open to other solutions!

bellini666:
@sanlil can you just rebase on top of the latest `main` and fix breaking tests?

sanlil:
Thanks @bellini666! Rebasing and fixing test is done!

RELEASE.md
```diff
@@ -0,0 +1,12 @@
+Release type: patch
```
bellini666:
```suggestion
Release type: minor
```

botberry:

PR #4193, opened 2026-02-06
fix: false-positive DuplicatedTypeName for same-origin types

### LINKED ISSUE #4142: Duplicated type name error when running a single test
## Describe the Bug

In the project I'm working on we have a lots of types and we use strawberry.lazy very often. The application is functioning correctly, but sometimes there is a problem with running a single test (running all the tests works fine). An error is raised that some type is defined multiple times in the schema, but the error message indicates the same type twice. 

Simplified error message:

```
  error: Type `UserFilter` is defined multiple times in the schema              
                                                                                
       @ core_api/core/schema/users/users_types.py:133                          
                                                                                
    132 | @strawberry_django.filter_type(User, lookups=True)                    
  ❱ 133 | class UserFilter:                              
                ^^^^^^^^^^ first class defined here                             
    134 |     email: auto                                                                                                               
                                                                                
       @ core_api/core/schema/users/users_types.py:133                          
                                                                                
    132 | @strawberry_django.filter_type(User, lookups=True)                    
  ❱ 133 | class UserFilter:                              
                ^^^^^^^^^^ second class defined here                            
    134 |     email: auto                                                                                                                  
                                                                                
                                                                                
  To fix this error you should either rename the type or remove the duplicated  
  definition.                                                                   
                                                                                
  Read more about this error on                                                 
  https://errors.strawberry.rocks/duplicated-type-name     
```

## System Information

 - Operating system: MacOS 14.8
 - Python version: 3.12.8
 - Strawberry version (if applicable): 0.288.2

bellini666:
- Add `origin is origin` identity check in `is_same_type_definition` so that two different `StrawberryObjectDefinition` instances sharing the same origin class are recognized as the same type
- Prevents false-positive `DuplicatedTypeName` when third-party decorators (e.g. strawberry-django's `filter_type`) re-process a type, creating a new definition while keeping the same Python class as `origin`

Fix #4142

## Summary by Sourcery

Avoid treating multiple Strawberry object definitions with the same origin class as duplicate types and extend tests to cover this behavior.

Bug Fixes:
- Prevent false-positive DuplicatedTypeName errors when separate StrawberryObjectDefinition instances share the same origin class.

Tests:
- Add regression tests ensuring same-origin type definitions are treated as the same type while genuinely different types with the same name still raise duplication errors.

### DIFF
--- a/strawberry/schema/schema_converter.py
+++ b/strawberry/schema/schema_converter.py
@@ -1053,6 +1053,18 @@ class GraphQLCoreConverter:
         if first_type_definition.origin is second_type_definition.origin:
             return True
 
+        # When sys.modules is cleared (e.g. by test runners or Django reloaders)
+        # and a module is reimported, Python creates brand-new class objects.
+        # The identity check above fails, so fall back to comparing the
+        # fully-qualified class name which survives reimports.
+        first_origin = first_type_definition.origin
+        second_origin = second_type_definition.origin
+        if (
+            first_origin.__qualname__ == second_origin.__qualname__
+            and first_origin.__module__ == second_origin.__module__
+        ):
+            return True
+
         if (
             first_type_definition.concrete_of is None
             or first_type_definition.concrete_of != second_type_definition.concrete_of
--- a/tests/schema/test_same_origin_reprocess/test_same_origin.py
+++ b/tests/schema/test_same_origin_reprocess/test_same_origin.py
@@ -1,3 +1,4 @@
+import sys
 from typing import TYPE_CHECKING, Annotated
 
 import strawberry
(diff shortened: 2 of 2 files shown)


botberry:



---

Fix false-positive `DuplicatedTypeName` error when two different `StrawberryObjectDefinition` instances share the same `origin` class. This can happen when third-party decorators (e.g. strawberry-django's `filter_type`) re-process a type, creating a new definition while keeping the same Python class as origin.



Here's the tweet text: 
```

```

tests/schema/test_duplicated_types.py
```diff
@@ -261,3 +262,57 @@ class Query:
...
+
+    @strawberry.type(name="Shared")
+    class B:
+        b: int
+
+    @strawberry.type
+    class Query:
+        field: int
+
+    strawberry.Schema(query=Query, types=[A, B])
```
patrick91: Don't we have a test like this already?😊

tests/schema/test_duplicated_types.py
```diff
@@ -261,3 +262,57 @@ class Query:
...
+    class Query:
+        foo: Foo
+
+    schema = strawberry.Schema(query=Query)
+    converter = schema.schema_converter
+
+    # "Foo" is already cached from schema construction via Query.foo.
+    # Calling from_object with a *different* definition instance (same origin)
+    # must not raise DuplicatedTypeName.
+    converter.from_object(duplicate_def)
```
patrick91: I'd prefer to have a real test that exercises this edge case, might be more complicated, but I want to avoid these kinds of tests 😊

botberry:
# Pre-release
:wave:

Pre-release **0.291.2.dev1771437961** [fc0004eeb079bb099f6c22f88fcbdf2e053a9e8a] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.291.2.dev1771437961
```

botberry:
# Pre-release
:wave:

Pre-release **0.291.2.dev1770456508** [a98d9df2bea5d6d2264e579242b161c4a69d64cc] has been released on PyPi! :rocket:
You can try it by doing:

```shell
poetry add strawberry-graphql==0.291.2.dev1770456508
```

PR #4196, opened 2026-02-07
Refactor federation functions to share federation params TypedDicts and processing functions

bellini666:
## Summary

- Add `strawberry.federation.params` module with `FederationFieldParams`, `FederationInterfaceParams`, and `FederationTypeParams` TypedDicts consumable via `Unpack[...]`
- Extract `process_federation_field_directives()` and `process_federation_type_directives()` from inline logic in `field.py` and `object_type.py`
- Fix `inaccessible` bug in `_impl_type` where `inaccessible=False` incorrectly added the `Inaccessible` directive

## Summary by Sourcery

Introduce shared federation parameter definitions and directive-processing helpers and refactor field and type decorators to use them while correcting directive handling for inaccessible types.

New Features:
- Add federation parameter TypedDicts for fields, interfaces, and types that can be reused via Unpack[...] across decorators.
- Provide shared helper functions to translate federation parameters into directive instances for fields and types/interfaces.

Bug Fixes:
- Fix incorrect addition of the Inaccessible directive when the inaccessible parameter is explicitly set to False.

Enhancements:
- Refactor federation field and type decorators to delegate directive construction to shared processing helpers, reducing duplication and centralizing logic.

## Summary by Sourcery

Introduce shared federation parameter definitions and directive-processing helpers and refactor federation decorators to use them while correcting inaccessible directive handling.

New Features:
- Add strawberry.federation.params module with reusable TypedDicts for federation field, interface, and type parameters.
- Provide shared helper functions to convert federation parameters into directive instances for fields and types/interfaces.

Bug Fixes:
- Fix incorrect addition of the Inaccessible directive when inaccessible is explicitly set to False on types and interfaces.

Enhancements:
- Refactor federation field, type, and interface decorators to delegate directive construction to shared processing helpers, reducing duplicated logic.

Documentation:
- Document the new federation params module and bug fix in RELEASE.md.

### DIFF
--- a/strawberry/federation/field.py
+++ b/strawberry/federation/field.py
@@ -7,14 +7,14 @@ from typing import (
     TypeVar,
     overload,
 )
+from typing_extensions import Unpack
 
 from strawberry.types.field import field as base_field
-from strawberry.types.unset import UNSET
 
-from .types import FieldSet
+from .params import FederationFieldParams, process_federation_field_directives
 
 if TYPE_CHECKING:
-    from collections.abc import Callable, Iterable, Mapping, Sequence
+    from collections.abc import Callable, Mapping, Sequence
     from typing import Literal
 
     from strawberry.extensions.field_extension import FieldExtension
@@ -23,10 +23,7 @@ if TYPE_CHECKING:
         _RESOLVER_TYPE,
         _RESOLVER_TYPE_ASYNC,
         _RESOLVER_TYPE_SYNC,
-        StrawberryField,
     )
-
-    from .schema_directives import Override
 
 T = TypeVar("T")
 
@@ -41,16 +38,6 @@ def field(
     name: str | None = None,
     is_subscription: bool = False,
     description: str | None = None,
-    authenticated: bool = False,
-    external: bool = False,
-    inaccessible: bool = False,
-    policy: list[list[str]] | None = None,
-    provides: list[str] | None = None,
-    override: Override | str | None = None,
-    requires: list[str] | None = None,
-    requires_scopes: list[list[str]] | None = None,
-    tags: Iterable[str] | None = (),
-    shareable: bool = False,
     init: Literal[False] = False,
     permission_classes: list[type[BasePermission]] | None = None,
     deprecation_reason: str | None = None,
@@ -60,6 +47,7 @@ def field(
     directives: Sequence[object] | None = (),
     extensions: list[FieldExtension] | None = None,
     graphql_type: Any | None = None,
+    **federation_kwargs: Unpack[FederationFieldParams],
 ) -> T: ...
 
 
@@ -70,16 +58,6 @@ def field(
     name: str | None = None,
     is_subscription: bool = False,
     description: str | None = None,
-    authenticated: bool = False,
-    external: bool = False,
-    inaccessible: bool = False,
-    policy: list[list[str]] | None = None,
-    provides: list[str] | None = None,
-    override: Override | str | None = None,
-    requires: list[str] | None = None,
-    requires_scopes: list[list[str]] | None = None,
-    tags: Iterable[str] | None = (),
-    shareable: bool = False,
     init: Literal[False] = False,
     permission_classes: list[type[BasePermission]] | None = None,
     deprecation_reason: str | None = None,
@@ -89,6 +67,7 @@ def field(
     directives: Sequence[object] | None = (),
     extensions: list[FieldExtension] | None = None,
     graphql_type: Any | None = None,
+    **federation_kwargs: Unpack[FederationFieldParams],
 ) -> T: ...
 
 
@@ -98,16 +77,6 @@ def field(
     name: str | None = None,
     is_subscription: bool = False,
     description: str | None = None,
-    authenticated: bool = False,
-    external: bool = False,
-    inaccessible: bool = False,
-    policy: list[list[str]] | None = None,
-    provides: list[str] | None = None,
-    override: Override | str | None = None,
-    requires: list[str] | None = None,
-    requires_scopes: list[list[str]] | None = None,
-    tags: Iterable[str] | None = (),
-    shareable: bool = False,
     init: Literal[True] = True,
     permission_classes: list[type[BasePermission]] | None = None,
     deprecation_reason: str | None = None,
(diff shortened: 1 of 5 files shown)


botberry:



---

Add `strawberry.federation.params` module with shared TypedDicts (`FederationFieldParams`, `FederationInterfaceParams`, `FederationTypeParams`) and processing functions (`process_federation_field_directives`, `process_federation_type_directives`) for federation directives.

These TypedDicts can be consumed via `Unpack[...]` to avoid duplicating federation parameter lists across packages. The processing functions are extracted from inline logic previously in `field.py` and `object_type.py`.

Also fixes a bug where `inaccessible=False` incorrectly added the `Inaccessible` directive on types/interfaces.



Here's the tweet text: 
```

```

botberry:


<table>
	<thead>
		<tr>
			<th>Federation 1 Support</th>
			<th>Federation 2 Support</th>
		</tr>
	</thead>
	<tbody>
    <tr>
      <td>
        <table>
        </table>
      </td>
      <td>
        <table>
        </table>
      </td>
		</tr>
	</tbody>
</table>


Learn more:

strawberry/federation/params.py
```diff
@@ -0,0 +1,180 @@
...
+        result.append(
+            Override(override_from=override, label=UNSET)
+            if isinstance(override, str)
+            else override
+        )
+
+    if policy:
+        result.append(Policy(policies=[list(p) for p in policy]))
+
+    if provides:
```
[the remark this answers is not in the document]
bellini666: No behavior change here — the original code in both `field.py` and `object_type.py` already used `if policy:` (truthiness check), not `if policy is not None:`.

strawberry/federation/params.py
```diff
@@ -0,0 +1,180 @@
...
+) -> tuple[list[object], bool]:
+    """Convert federation type/interface parameters into directive instances.
+
+    Returns ``(directives, extend)`` where *directives* is a new list
+    starting with the input directives, extended with any directive objects
+    implied by the keyword arguments.
+    """
+    result = list(directives or [])
+
+    result.extend(
```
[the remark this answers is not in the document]
bellini666: `extend` needs to be returned for `Unpack` consumers. When strawberry-django calls `process_federation_type_directives(directives, **federation_kwargs)`, `extend` lives inside the unpacked kwargs dict — the caller has no other way to retrieve it. That's why the return tuple exists.

strawberry/federation/object_type.py
```diff
@@ -42,43 +41,23 @@ def _impl_type(
...
+    directives, extend = process_federation_type_directives(
+        directives,
+        keys=keys,
+        extend=extend,
+        shareable=shareable,
+        inaccessible=inaccessible is True,
+        authenticated=authenticated,
+        policy=policy,
+        requires_scopes=requires_scopes,
+        tags=tags,
```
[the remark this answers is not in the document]
bellini666: Good call — added a regression test in 8833fc29.

strawberry/federation/params.py
```diff
@@ -0,0 +1,180 @@
...
+    """Federation parameters accepted by field decorators."""
+
+    authenticated: bool
+    external: bool
+    inaccessible: bool
+    policy: Sequence[Sequence[str]]
+    provides: Sequence[str]
+    override: Override | str
+    requires: Sequence[str]
+    requires_scopes: Sequence[Sequence[str]]
```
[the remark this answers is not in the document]
bellini666: The TypedDicts intentionally use `total=False`, making all keys optional (can be omitted). Omitting a key is the replacement for passing `None` — the processing function defaults handle it. There's no need for `| None` in the TypedDict values.

bellini666:
@sourcery-ai review

strawberry/federation/params.py
```diff
@@ -0,0 +1,200 @@
...
+    shareable: bool
+    tags: Sequence[str]
```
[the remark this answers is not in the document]
bellini666: Good catch — standardized the processing function params from `Iterable[str]` to `Sequence[str]` for `tags` to match the TypedDicts.

strawberry/federation/object_type.py
```diff
@@ -276,25 +198,15 @@ def interface(
...
         directives=directives,
-        authenticated=authenticated,
-        keys=keys,
-        inaccessible=inaccessible,
-        policy=policy,
-        requires_scopes=requires_scopes,
-        tags=tags,
         is_interface=True,
+        **federation_kwargs,
     )
```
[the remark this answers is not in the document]
bellini666: mypy doesn't actually flag this. Since `FederationTypeParams` extends `FederationInterfaceParams` and all keys are `total=False`, a dict conforming to the parent structurally satisfies the child — the extra keys (`extend`/`shareable`) are simply absent, which is valid for optional keys.

strawberry/federation/field.py
```diff
@@ -145,26 +107,18 @@ def field(
...
     extensions: list[FieldExtension] | None = None,
     graphql_type: Any | None = None,
+    **federation_kwargs: Unpack[FederationFieldParams],
 ) -> StrawberryField: ...
 
 
 @overload
 def field(
     resolver: _RESOLVER_TYPE_SYNC[T],
+    /,
```
patrick91: let's remove this 😊 (everywhere), I think we should still allow doing `field(resolver=x)` :)

strawberry/federation/params.py
```diff
@@ -0,0 +1,200 @@
+"""Shared TypedDicts and processing functions for federation directives.
+
+Provides ``FederationFieldParams``, ``FederationInterfaceParams``, and
+``FederationTypeParams`` TypedDicts that can be consumed via
+``Unpack[...]`` to avoid duplicating federation parameter lists.
+"""
```
patrick91: let's remove this 😊

strawberry/federation/params.py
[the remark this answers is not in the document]
patrick91: why sequence? 😊

strawberry/federation/params.py
```diff
@@ -0,0 +1,200 @@
...
+    policy: Sequence[Sequence[str]]
+    requires_scopes: Sequence[Sequence[str]]
+    tags: Sequence[str]
+
+
+class FederationTypeParams(FederationInterfaceParams, total=False):
+    """Federation parameters accepted by type decorators.
+
+    Extends :class:`FederationInterfaceParams` with ``extend`` and ``shareable``.
+    """
```
patrick91: all these dataclasses can be removed, claude really likes writing them :D

strawberry/federation/params.py
```diff
@@ -0,0 +1,200 @@
...
+        Provides,
+        Requires,
+        RequiresScopes,
+        Shareable,
+        Tag,
+    )
+    from .schema_directives import (
+        Override as OverrideDirective,
+    )
+    from .types import FieldSet
```
patrick91: mmh, can we avoid inline imports?

strawberry/federation/field.py
bellini666: This was here to solve a mypy issue with overloads signatures not matching. Lets see if we can solve it without the `/`

strawberry/federation/params.py
bellini666: I thought it was ok 🤣, but lets remove them

strawberry/federation/params.py
bellini666: Nice catch, I missed this one. Noticed other inline imports that made no sense, this one I thought was to avoid circular imports, but it shouldn't based on what it is importing

bellini666: Ok, we can't avoid all inline imports, but we can move `UNSET` and `FieldSet` to the top. The others causes circular imports

strawberry/federation/field.py
bellini666: So, the `/` was here in the first 2 overloads to allow overload differentiation when doing:

```python
@strawberry.federation.field
def xxx(...):
    ...
```

Without any argument

But I guess if you are not passing anything, you can just use `@strawberry.field` directly, right? The only fix to keep mypy happy is to drop those 2 first overloads, wich will prevent the example code above from being used (which, as I mentioned, is probably ok)

strawberry/federation/params.py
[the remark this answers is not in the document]
bellini666: @patrick91 I actually asked it to do it, because `list` is invariant, meaning you are forced to actually pass a list and can't do it as a `tuple` or something like that. In the end, what we need is a sequence, so I was trying to loosen it a bit

Issue #3265, opened 2023-11-30
strawberry.asdict doesn't recognize strawberry.UNSET


sanfilippopablo:
## Describe the Bug

When a type has field with the value `strawberry.UNSET`, calling `strawberry.asdict` in the value gets you a dict where that value has `None`. The expected behavior is that the field is not present in the resulting dict.

## Additional Context

This is code that I wrote to circunvent the issue:
```py
def transform_items_recursively(
    input_dict: Dict,
    fn: Callable[[str, Any], Optional[Dict]],
) -> Dict:
    """
    Apply fn to every key-value in dict, recursively.
    """
    modified_dict = {}
    for key, value in input_dict.items():
        result: Optional[Dict] = None

        if isinstance(value, dict):
            # If it is a dict, first transform it and then call fn on the key with the new value
            result = fn(key, transform_items_recursively(value, fn))

        elif isinstance(value, list) and any([isinstance(v, dict) for v in value]):
            # If it is a list of dicts, transform each dict in it and build a new list. Finally,
            # call fn on the new value.
            result = fn(key, [transform_items_recursively(val, fn) for val in value])

        else:
            result = fn(key, value)

        if result is not None:
            key, value = result
            modified_dict[key] = value

    return modified_dict


def remove_unset_values(input_dict):
    """
    Recursively remove key-value pairs with strawberry.UNSET values from a dictionary and its nested dictionaries.
    """

    def fn(key, value):
        if value is strawberry.UNSET:
            return None
        else:
            return key, value

    return transform_items_recursively(input_dict, fn)

```

patrick91:
Fixed here https://github.com/strawberry-graphql/strawberry/pull/4320

PR #4231, opened 2026-02-14
fix asdict to handle Some and UNSET values

### LINKED ISSUE #4141: `strawberry.asdict` does not handle `Some`
## Describe the Bug

`strawberry.asdict` only wraps `dataclasses.asdict`, so it does not handle `Some` - dataclass inside the object is not serialized.

## Additional Context

UNSET aren't handled as well
https://github.com/strawberry-graphql/strawberry/issues/3265

Mr-Neutr0n:
`strawberry.asdict` was just delegating to `dataclasses.asdict`, which doesn't know about `Some` or `UNSET`. This meant `Some(value)` would end up as-is in the dict instead of being unwrapped, and `UNSET` fields would leak through instead of being excluded.

Fixed by replacing the plain `dataclasses.asdict` call with a custom recursive implementation that:
- unwraps `Some(value)` to just `value`
- skips fields that are `UNSET`
- still recursively handles nested dataclasses, lists, tuples, and dicts the same way `dataclasses.asdict` does

Fixes #4141
Related: #3265

## Summary by Sourcery

Update strawberry.asdict to correctly handle framework-specific sentinel and wrapper types while maintaining recursive conversion behavior.

Bug Fixes:
- Ensure asdict unwraps Some(value) instances to their inner values so they serialize as plain data instead of wrapper objects.
- Exclude fields set to UNSET from asdict output to prevent leaking unset sentinel values in serialized dictionaries.

Documentation:
- Document the updated asdict behavior in the public API docstring and release notes, including handling of Some and UNSET values.

Tests:
- Add tests covering Some unwrapping, UNSET field exclusion, None preservation inside Some, and nested dataclass handling for Maybe fields.

### DIFF
--- a/strawberry/types/object_type.py
+++ b/strawberry/types/object_type.py
@@ -508,9 +508,7 @@ def asdict(obj: Any) -> dict[str, object]:
             cls = builtins.type(obj)
             return cls(_asdict_inner(v) for v in obj)
         if isinstance(obj, dict):
-            return {
-                _asdict_inner(k): _asdict_inner(v) for k, v in obj.items()
-            }
+            return {_asdict_inner(k): _asdict_inner(v) for k, v in obj.items()}
         return obj
 
     return _asdict_inner(obj)


botberry:



---

`strawberry.asdict` now correctly handles `Some` and `UNSET` values.
`Some(value)` is unwrapped to its inner value, and fields set to `UNSET`
are excluded from the resulting dictionary.



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Harikrishna KP for the PR 👏

```

botberry:



---

`strawberry.asdict` now correctly handles `Some` and `UNSET` values.
`Some(value)` is unwrapped to its inner value, and fields set to `UNSET`
are excluded from the resulting dictionary.



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Harikrishna KP for the PR 👏

```

strawberry/types/object_type.py
```diff
@@ -486,7 +490,28 @@ class User:
...
+
+    def _asdict_inner(obj: Any) -> Any:
+        if isinstance(obj, Some):
+            return _asdict_inner(obj.value)
+        if dataclasses.is_dataclass(obj) and not isinstance(obj, builtins.type):
+            result = {}
+            for f in dataclasses.fields(obj):
+                value = getattr(obj, f.name)
+                if value is UNSET:
+                    continue
```
rcybulski1122012, line 504: When you use `Maybe` annotation, and value is not provided it's going to be `None`  - https://strawberry.rocks/docs/types/maybe

You need to check if the field annotation is `maybe` - AFAIR there is such a function already in strawberry. If so and the value is None, exclude it from the result.

This is the implementation I've been using in my project - it's not perfect, but may come in handy 

```python
from dataclasses import fields, is_dataclass

def strawberry_to_dict(obj) -> dict[str, Any]:
    result: dict[str, Any] = {}
    for field in fields(obj):
        value = getattr(obj, field.name)
        is_maybe = _annotation_is_maybe(field.type)

        if isinstance(value, Some):
            if is_dataclass(value.value):
                result[field.name] = _strawberry_to_dict(value.value)
            elif isinstance(value.value, list):
                result[field.name] = [_strawberry_to_dict(item) if is_dataclass(item) else item for item in value.value]
            else:
                result[field.name] = value.value
        elif value is None and not is_maybe:
            result[field.name] = None
        elif value is not UNSET and not is_maybe:
            if is_dataclass(value):
                result[field.name] = _strawberry_to_dict(value)
            elif isinstance(value, list):
                result[field.name] = [_strawberry_to_dict(item) if is_dataclass(item) else item for item in value]
            else:
                result[field.name] = value

    return result


_maybe_re = re.compile(r"^(?:strawberry\.)?Maybe\[(.+)\]$")


def _annotation_is_maybe(annotation: Any) -> bool:
    # copied from strawberry
    if isinstance(annotation, str):
        # Ideally we would try to evaluate the annotation, but the args inside
        # may still not be available, as the module is still being constructed.
        # Checking for the pattern should be good enough for now.
        return _maybe_re.match(annotation) is not None

    orig = typing.get_origin(annotation)
    if orig is typing.Annotated:
        return _annotation_is_maybe(typing.get_args(annotation)[0])
    return orig is Maybe
```

strawberry/types/object_type.py
```diff
@@ -486,7 +490,28 @@ class User:
     # {"name": "Lorem", "age": 25}
     ```
     """
-    return dataclasses.asdict(obj)
+    from strawberry.types.maybe import Some
+    from strawberry.types.unset import UNSET
```
bellini666, line 494: Are those necessary for preventing import error? If not, we can move then to module level

strawberry/types/object_type.py
```diff
@@ -486,7 +490,28 @@ class User:
...
+
+    def _asdict_inner(obj: Any) -> Any:
+        if isinstance(obj, Some):
+            return _asdict_inner(obj.value)
+        if dataclasses.is_dataclass(obj) and not isinstance(obj, builtins.type):
```
bellini666, line 499: what happens when `dataclasses.is_dataclass(obj)` and also `isinstance(obj, builtins.type)`?

bellini666:
This got supersed by https://github.com/strawberry-graphql/strawberry/pull/4320, which solved it in a cleaner way.

Due to the inspiration in this PR, I added you as co-author of it :)

PR #4256, opened 2026-02-20
Fix: ExecutionResult not always present in execution context during concurrent requests

### LINKED ISSUE #4255: ExecutionResult not always present in execution context in extensions
## Describe the Bug

We have an extension that reads the ExecutionResult so we can log some details about it. Since version 0.291.3 we have noticed that the `result` field is not always present. This seems to be when making concurrent requests.

## System Information

 - Operating system: Mac & linux
 - Python version: 3.13
 - Strawberry version (if applicable): 0.291.3

## Additional Context

Repro here
```python
import asyncio

import strawberry
from strawberry.extensions import SchemaExtension


class Extension(SchemaExtension):
    async def on_operation(self):
        print("Before operation")
        yield
        print("After operation")
        print(f"Result: {self.execution_context.result}")


@strawberry.type
class Query:
    @strawberry.field
    async def node(self) -> str:
        return ""


async def main() -> None:
    await asyncio.gather(
        schema.execute("query { node }"),
        schema.execute("query { node }"),
    )


if __name__ == "__main__":
    schema = strawberry.Schema(Query, extensions=[Extension])
    asyncio.run(
        main(),
    )
```

bellini666:
## Summary

- Fix `execution_context.result` being `None` or belonging to a wrong request when multiple async requests execute concurrently (e.g. via `asyncio.gather`)
- Create fresh extension instances and middleware manager per async `execute()` and `subscribe()` call instead of reusing cached ones
- Add regression test that runs concurrent requests and asserts `execution_context.result` is non-None for all of them

Fixes #4255

## Root Cause

Since v0.291.3 (commit `4c29e399`), `_async_extensions` uses `@cached_property` — the same extension **instances** are shared across all async requests. Each call to `execute()` mutates `extension.execution_context` on these shared instances. With concurrent requests, Request B overwrites the `execution_context` that Request A's extensions are still using, so after `yield` in `on_operation`, `self.execution_context.result` may be `None` or belong to a different request.

## Why this doesn't regress the original fix (4c29e399)

That commit fixed a **sequential** bug: `get_extensions()` created fresh instances per request, but `_get_middleware_manager()` cached references to the first request's instances. So `resolve()` ran on old instances while lifecycle hooks ran on new ones — causing `ApolloTracingExtension._resolver_stats` to be empty on subsequent executions.

This PR creates **both** fresh extensions **and** a fresh middleware manager per async request, so `resolve()` and lifecycle hooks always target the same per-request instances. The sequential bug stays fixed because each request's middleware manager references that request's own extension instances.

Sync execution (`execute_sync`) keeps cached instances since it cannot run concurrently.

## Test plan

- [x] New `test_concurrent_execution_context_result` passes
- [x] Existing `test_tracing_resolvers_populated_on_multiple_executions` still passes (sequential correctness)
- [x] Full `tests/schema/extensions/` suite passes (140 passed, 6 skipped)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Ensure async GraphQL executions create per-request extensions and middleware to keep execution context results isolated and correct during concurrent requests.

Bug Fixes:
- Fix async execution_context.result sometimes being None or associated with the wrong request when multiple operations run concurrently.

Tests:
- Add regression test asserting execution_context.result is populated for all requests when running concurrent async executions.

### DIFF
--- a/strawberry/schema/schema.py
+++ b/strawberry/schema/schema.py
@@ -383,6 +383,12 @@ class Schema(BaseSchema):
     def _sync_extensions(self) -> list[SchemaExtension]:
         return self.get_extensions(sync=True)
 
+    @property
+    def _async_extensions(self) -> list[SchemaExtension]:
+        # Fresh instances per access: concurrent async requests must not share
+        # extension state. See _sync_extensions for the cached counterpart.
+        return self.get_extensions(sync=False)
+
     def create_extensions_runner(
         self, execution_context: ExecutionContext, extensions: list[SchemaExtension]
     ) -> SchemaExtensionsRunner:
@@ -572,7 +578,7 @@ class Schema(BaseSchema):
             operation_name=operation_name,
             operation_extensions=operation_extensions,
         )
-        extensions = self.get_extensions()
+        extensions = self._async_extensions
         # TODO (#3571): remove this when we implement execution context as parameter.
         for extension in extensions:
             extension.execution_context = execution_context
@@ -902,7 +908,7 @@ class Schema(BaseSchema):
             root_value=root_value,
             operation_name=operation_name,
         )
-        extensions = self.get_extensions()
+        extensions = self._async_extensions
         # TODO (#3571): remove this when we implement execution context as parameter.
         for extension in extensions:
             extension.execution_context = execution_context


botberry:



---

Fix `execution_context.result` being `None` or belonging to a wrong request when multiple async requests execute concurrently (e.g. via `asyncio.gather`). Shared cached extension instances are no longer reused across concurrent async requests.



Here's the tweet text: 
```

```

PR #4267, opened 2026-02-24
fix: prevent async iterable detection for sync execution in gql-core 3.3

bellini666:
## Summary

graphql-core 3.3 added an `is_async_iterable` parameter to `execute()`. When not provided, it defaults to a check that returns `True` for any object with `__aiter__` — including Django QuerySets.

This causes `execute()` to return a coroutine instead of an `ExecutionResult` during sync execution, which then crashes in `execute_sync()` with:

```
RuntimeError: There is no current event loop in thread 'MainThread'.
```

**Fix:** When `execute_sync()` calls `execute()`, pass `is_async_iterable=lambda _x: False` to prevent graphql-core 3.3 from treating sync iterables as async. This is gated behind the `IS_GQL_33` version flag so it only applies to 3.3+, and a runtime version check ensures users on older 3.3 alphas (before `is_async_iterable` existed) get a clear error directing them to upgrade to `>= 3.3.0a12`.

## Test plan

- strawberry-django CI with gql-core 3.3 matrix (249 tests) — currently all failing, should pass with this fix
- gql-core 3.2 matrix unaffected (parameter not passed)
- Added regression test verifying `execute_sync` works with resolvers returning objects that have `__aiter__`

### DIFF
--- a/tests/schema/test_basic.py
+++ b/tests/schema/test_basic.py
@@ -581,8 +581,12 @@ def test_kw_only():
 
 
 def test_execute_sync_with_aiter_object():
-    """Regression: graphql-core 3.3's is_async_iterable default treats objects
-    with __aiter__ as async, causing execute_sync to return a coroutine."""
+    """Sync execution works with objects that define __aiter__.
+
+    graphql-core 3.3's is_async_iterable default treats any object with
+    __aiter__ (e.g. Django QuerySets) as async, causing execute_sync to
+    return a coroutine instead of an ExecutionResult.
+    """
 
     class FakeQuerySet:
         def __aiter__(self):


botberry:



---

Fix sync execution crash with graphql-core 3.3 where `execute_sync()` would return a coroutine
instead of an `ExecutionResult`, causing `RuntimeError: There is no current event loop`,
because graphql-core 3.3's `is_async_iterable` default treats objects with `__aiter__`
(like Django QuerySets) as async iterables.

Now passes `is_async_iterable=lambda _x: False` during sync execution to prevent this.

Note: graphql-core >= 3.3.0a12 is now the minimum required version for the 3.3.x series.



Here's the tweet text: 
```

```

strawberry/schema/schema.py
```diff
@@ -95,6 +96,10 @@
     from strawberry.types.scalar import ScalarDefinition, ScalarWrapper
     from strawberry.types.union import StrawberryUnion
 
+_EXECUTE_HAS_IS_ASYNC_ITERABLE = (
+    IS_GQL_33 and "is_async_iterable" in _inspect.signature(execute).parameters
+)
```
patrick91: can we avoid this? we can make the assumption that the users have the latest 3.3 if they use the alpha 😊

I guess we also need to wait for a new alpha release

bellini666: Yep! Just need a new alpha release for that: https://github.com/graphql-python/graphql-core/issues/258

bellini666: For now, used this to make https://github.com/strawberry-graphql/strawberry-django/pull/850 pass :)

botberry:


<table>
	<thead>
		<tr>
			<th>Federation 1 Support</th>
			<th>Federation 2 Support</th>
		</tr>
	</thead>
	<tbody>
    <tr>
      <td>
        <table>
        </table>
      </td>
      <td>
        <table>
        </table>
      </td>
		</tr>
	</tbody>
</table>


Learn more:

strawberry/schema/schema.py
bellini666: Ready!

bellini666:
@patrick91 should be ready now

edit: not really, some tests are failing. Will check why =P

edit2: there was another "breaking" change, fixed now

PR #4319, opened 2026-03-19
fix: await cancelled subscription tasks on ws shutdown

### LINKED ISSUE #4284: graphql-transport-ws: cancelled subscription tasks are not awaited during shutdown, allowing zombie cleanup to corrupt shared state
## Describe the Bug

When a WebSocket client disconnects while a subscription is active, `shutdown()` cancels the `run_operation` task via `cleanup_operation()` but never awaits it. The cancelled task's `finally` block (e.g. releasing DB resources) runs asynchronously at an unpredictable time — potentially after the caller has already torn down shared state.

This is a regression of the fix in PR #2141 / issue #2125. That PR added `await task` in `cleanup_operation`. A later refactoring to `Operation` objects removed the await with the comment:

```python
# do not await the task here, lest we block the main websocket handler Task.
```

This makes sense during message processing (called from `handle_complete`), but `shutdown()` has no more messages to process — awaiting is both safe and necessary.

## Current behavior

```python
async def shutdown(self) -> None:
    # ...
    for operation_id in list(self.operations.keys()):
        await self.cleanup_operation(operation_id)   # cancel() without await
    await self.reap_completed_tasks()                 # nothing to reap yet — task hasn't finished
```

The cancelled task becomes a zombie. Its `finally` block runs later, at whatever `await` point the event loop happens to schedule it.

## Impact

In our case (pytest with `loop_scope="function"`, psycopg connection pools):

1. Test 1 subscription creates `run_operation` task on event loop L1
2. Client disconnects → `shutdown()` cancels task but doesn't await it
3. Fixture teardown closes and reinitializes DB pools (still on L1)
4. Zombie's `finally` runs, opens the **new** pool on L1
5. L1 closes, test 2 starts on L2
6. Test 2 teardown tries to close the pool → `RuntimeError: Event loop is closed` (scheduler waiter bound to dead L1)

The zombie is visible as `async_generator_asend` with `cancelling=1` in `asyncio.all_tasks()`.

## Proposed fix

Collect cancelled operation tasks in `shutdown()` and await them before reaping:

```python
async def shutdown(self) -> None:
    if self.connection_init_timeout_task:
        with suppress(asyncio.CancelledError):
            await self.connection_init_timeout_task

    cancelled_tasks: list[asyncio.Task] = []
    for operation_id in list(self.operations.keys()):
        op = self.operations.get(operation_id)
        if op and op.task:
            cancelled_tasks.append(op.task)
        await self.cleanup_operation(operation_id)

    for task in cancelled_tasks:
        with suppress(BaseException):
            await task

    await self.reap_completed_tasks()
```

This preserves the "don't block the message loop" property of `cleanup_operation` while ensuring `shutdown()` waits for all cleanup to complete.

## System Information

- Python 3.13
- strawberry-graphql 0.262.5
- psycopg / psycopg_pool
- pytest-asyncio with `loop_scope="function"`

## Related

- #2125 / PR #2141 — original fix (await in cleanup), lost in refactoring
- #2720 — "Task was destroyed but it is pending" (same root cause)
- #3991 — "Task exception was never retrieved" (related pattern)

Flamefork:
## Description

`cleanup_operation` cancels subscription tasks but does not await them (to avoid blocking the message loop). During WebSocket shutdown this meant tasks' `finally` blocks could run after shared state (DB pools, event loop) was already torn down.

This fix collects cancelled tasks during shutdown and awaits them via `asyncio.gather` before returning.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* Fixes https://github.com/strawberry-graphql/strawberry/issues/4284

## Checklist

## Summary by Sourcery

Ensure WebSocket shutdown waits for cancelled subscription tasks so their cleanup completes before shared resources are torn down.

Bug Fixes:
- Fix WebSocket shutdown to await previously cancelled subscription tasks, preventing subscription cleanup from running after shared resources are closed.

Documentation:
- Document this change as a patch release in RELEASE.md.

Tests:
- Add a regression test verifying that WebSocket shutdown leaves no active infinity subscriptions after awaiting cancelled subscription tasks.


botberry:



---

Await cancelled subscription tasks during WebSocket shutdown so their `finally` blocks run before shared state (DB pools, event loop) is torn down.



Here's the tweet text: 
```
🆕 Release (next) is out! Thanks to Ilia Ablamonov for the PR 👏

```

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -92,9 +92,17 @@ async def shutdown(self) -> None:
             with suppress(asyncio.CancelledError):
                 await self.connection_init_timeout_task
 
+        cancelled_tasks: list[asyncio.Task] = []
         for operation_id in list(self.operations.keys()):
+            op = self.operations[operation_id]
+            if op.task:
+                cancelled_tasks.append(op.task)
             await self.cleanup_operation(operation_id)
```
bellini666: Wondering here if `cleanup_operation` should do that, and then this would be called in a gather, like:

```python
await asyncio.gather(self.cleanup_operation(operation_id) for operation_id in list(self.operations))
```

cc @patrick91 @DoctorJohn

Flamefork: @bellini666 Yeah, makes sense to have the logic in `cleanup_operation`. The constraint is that it can't await the task without blocking the message loop in `handle_complete`. I worked around this in https://github.com/strawberry-graphql/strawberry/pull/4319/commits/82594029d5ee3073c8229d827ce0ae463098a9e6 —    
  `cleanup_operation` now collects cancelled tasks, and shutdown reaps them via a symmetrical `reap_cancelled_tasks`. Let me know what you think.

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -329,6 +332,7 @@ async def cleanup_operation(self, operation_id: str) -> None:
         operation = self.operations.pop(operation_id)
         assert operation.task
         operation.task.cancel()
+        self.cancelled_tasks.append(operation.task)
         # do not await the task here, lest we block the main
```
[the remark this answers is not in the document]
bellini666: This is something that I noticed with the current code. It is not an issue with `cancelled_tasks` only, but `completed_tasks` as well.

Maybe we should consider creating a periodic task that cleans up those?

Will discuss this tomorrow with @patrick91, but maybe it can be done in a follow-up PR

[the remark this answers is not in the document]
patrick91: we could use `task.add_done_callback` (let's do this in a new PR)

bellini666:
Hi @Flamefork ,

Based on https://github.com/strawberry-graphql/strawberry/pull/4319#discussion_r2976024050, I ended up making this change to fix the memory leak: https://github.com/strawberry-graphql/strawberry/pull/4345

Maybe it also fixes the issue here? If not, could you rebase this PR to adjust it to the new code?

botberry:

---

Await cancelled subscription tasks during WebSocket shutdown so their `finally` blocks run before shared state (DB pools, event loop) is torn down.

This release was contributed by @Flamefork in https://github.com/strawberry-graphql/strawberry/pull/4319

Flamefork:
@bellini666 I've [updated](https://gist.github.com/Flamefork/3d2abd1dcee83988be752c198359b089/revisions) the repro gist and the PR. Not sure if the solution is idiomatic though, would be happy to get advice on it

strawberry/subscriptions/protocols/graphql_transport_ws/handlers.py
```diff
@@ -93,8 +93,14 @@ async def shutdown(self) -> None:
...
                 await self.connection_init_timeout_task
 
+        cancelled_tasks = []
         for operation_id in list(self.operations.keys()):
-            await self.cleanup_operation(operation_id)
+            task = await self.cleanup_operation(operation_id)
+            if task:
+                cancelled_tasks.append(task)
+        for task in cancelled_tasks:
+            with suppress(BaseException):
```
bellini666: I don't think we should supress `BaseException`. At maximum `Exception`

tests/websockets/test_graphql_transport_ws.py
```diff
@@ -1570,3 +1570,55 @@ async def failing_run(operation):
...
+
+async def test_shutdown_awaits_cancelled_subscription_tasks(
+    http_client: HttpClient,
+):
+    """Shutdown must await cancelled subscription tasks so their finally-blocks
+    complete before shared resources are torn down.
+
+    Regression test for https://github.com/strawberry-graphql/strawberry/issues/4284
+    """
+    cleanup_done_at_shutdown_end: bool | None = None
```
bellini666:
```suggestion
    cleanup_done_at_shutdown_end: bool = False
```

tests/websockets/test_graphql_transport_ws.py
```diff
@@ -1570,3 +1570,55 @@ async def failing_run(operation):
 
     assert len(failed_task_exception) == 1
     assert isinstance(failed_task_exception[0], RuntimeError)
+
+
+async def test_shutdown_awaits_cancelled_subscription_tasks(
```
bellini666, line 1575: question: I assume that without the change here, this fails, right?

bellini666 (changes requested):
Looks good! :)

Some small comments, almost ready for merge

tests/websockets/test_graphql_transport_ws.py
Flamefork: Yes:
```
FAILED 😰  tests/websockets/test_graphql_transport_ws.py::test_shutdown_awaits_cancelled_subscription_tasks[AioHttpClient] - assert False is True
FAILED 😰  tests/websockets/test_graphql_transport_ws.py::test_shutdown_awaits_cancelled_subscription_tasks[AsgiHttpClient] - assert False is True
FAILED 😰  tests/websockets/test_graphql_transport_ws.py::test_shutdown_awaits_cancelled_subscription_tasks[ChannelsHttpClient] - assert False is True
FAILED 😰  tests/websockets/test_graphql_transport_ws.py::test_shutdown_awaits_cancelled_subscription_tasks[FastAPIHttpClient] - assert False is True
FAILED 😰  tests/websockets/test_graphql_transport_ws.py::test_shutdown_awaits_cancelled_subscription_tasks[LitestarHttpClient] - assert False is True
FAILED 😰  tests/websockets/test_graphql_transport_ws.py::test_shutdown_awaits_cancelled_subscription_tasks[QuartHttpClient] - assert False is True
```

Flamefork:
@bellini666 thank you for the feedback! I've [updated the code](https://github.com/strawberry-graphql/strawberry/pull/4319/changes/1f2ef2e96fc2c9a61e9e4817a0bea49a83b8438d) accordingly.

After it, I've tried [alternative simpler approach](https://github.com/strawberry-graphql/strawberry/pull/4319/commits/4cde997d9c11bb207f8495472c8a173fee3710d2), that felt more natural after your PR landed, and which lead to (imho) cleaner implementation and [final diff](https://github.com/strawberry-graphql/strawberry/pull/4319/changes). Please take a look. If it's less desirable I'd just revert it, but imho it's worth checking.

Thanks again for your time!

bellini666 (approved):
Looks great, thanks again :)

botberry:
This PR was published as [0.315.2](https://github.com/strawberry-graphql/strawberry/releases/tag/0.315.2). Thank you for contributing!

PR #4359, opened 2026-04-12
feat: allow passing config as a TypedDict to Schema

### LINKED ISSUE #2942: Move StrawberryConfig to a typed dict

Deepak8858:
This PR allows passing the schema configuration as a dictionary (TypedDict) to the `strawberry.Schema` constructor, as suggested in #2942.

### Changes:
- Added `StrawberryConfigDict` TypedDict in `strawberry/schema/config.py`.
- Updated `strawberry.Schema` to accept `StrawberryConfigDict` and convert it to `StrawberryConfig` internally.

Example usage:
```python
import strawberry

schema = strawberry.Schema(query=Query, config={"auto_camel_case": False})
```

Fixes #2942

## Summary by Sourcery

Allow passing schema configuration as a TypedDict to the Strawberry Schema constructor.

New Features:
- Support providing schema configuration as a StrawberryConfigDict TypedDict when constructing a Schema.

Enhancements:
- Expose the new StrawberryConfigDict TypedDict in the schema config module exports.

### DIFF
--- a/strawberry/federation/schema.py
+++ b/strawberry/federation/schema.py
@@ -7,7 +7,6 @@ from typing import (
     Any,
     Literal,
     NewType,
-    Optional,
     Union,
     cast,
 )
--- a/strawberry/federation/schema.py
+++ b/strawberry/federation/schema.py
@@ -35,7 +35,7 @@ if TYPE_CHECKING:
 
     from strawberry.extensions import SchemaExtension
     from strawberry.federation.schema_directives import ComposeDirective
-    from strawberry.schema.config import StrawberryConfig
+    from strawberry.schema.config import StrawberryConfig, StrawberryConfigDict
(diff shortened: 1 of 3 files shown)


botberry:

```markdown
---
release type: patch
---

```

PR #4364, opened 2026-04-13
fix: handle PEP 649 NameError in Signature.from_callable for forward refs

Kobusvdwalt:
## Description

On Python 3.14+ (PEP 649), function annotations are lazily evaluated via `__annotate__`. When `inspect.Signature.from_callable()` is called on a resolver function, it triggers this deferred evaluation. If the annotation references a name that does not exist yet (e.g. a forward reference to a class currently being defined), this raises `NameError` **before** Strawberry's own error handling in `ReservedType.find()` can intervene.

### Example that fails on Python 3.14

```python
import strawberry

def get_full_name(user: strawberry.Parent[User]) -> str:
    return f"{user.first_name} {user.last_name}"

@strawberry.type
class User:
    first_name: str
    last_name: str
    full_name: str = strawberry.field(resolver=get_full_name)
```

Error:
```
NameError: name 'User' is not defined
```

This pattern works fine on Python ≤3.13 with `from __future__ import annotations` (thanks to #3851), but PEP 649 introduces a different annotation evaluation path where the `NameError` occurs inside `inspect.Signature.from_callable()` rather than during `StrawberryAnnotation.evaluate()`.

### Fix

Catch `NameError` from `Signature.from_callable()` in the `StrawberryResolver.signature` property and fall back to `inspect.Format.FORWARDREF` (new in Python 3.14), which wraps unresolvable names in `ForwardRef` objects instead of raising.

### Changes

- `strawberry/types/fields/resolver.py`: Added `NameError` handling with `Format.FORWARDREF` fallback in `StrawberryResolver.signature`
- `tests/types/test_parent_type_pep649.py`: New test file for PEP 649 forward reference scenarios
- `RELEASE.md`: Patch release note

Related to #3481

## Summary by Sourcery

Handle PEP 649-related NameError in resolver signature inspection and add coverage for strawberry.Parent forward references on Python 3.14+.

Bug Fixes:
- Prevent NameError when inspecting resolver signatures that use strawberry.Parent with forward-referenced types under Python 3.14+ deferred annotation evaluation.

Documentation:
- Add patch release notes describing the PEP 649-related fix for strawberry.Parent forward references and the new Signature.from_callable fallback behaviour.

Tests:
- Add Python 3.14+ integration-style tests validating strawberry.Parent with forward and non-forward references for sync and async resolvers.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,26 @@
+Release type: patch
+
+Fix `NameError` when using `strawberry.Parent` with forward references on
+Python 3.14+ (PEP 649).
+
+On Python 3.14, annotations are lazily evaluated via `__annotate__` (PEP 649).
+When a resolver references a type via `strawberry.Parent[SomeType]` and
+`SomeType` is defined after the resolver, `inspect.Signature.from_callable()`
+triggers annotation evaluation which raises `NameError`. Strawberry now falls
+back to `inspect.Format.FORWARDREF` to handle this gracefully.
+
+For example, the following code now works on Python 3.14+:
+
+```python
+import strawberry
+
+# Resolver defined before the class it references
+def get_full_name(user: strawberry.Parent[User]) -> str:
+    return f"{user.first_name} {user.last_name}"
+
+@strawberry.type
+class User:
+    first_name: str
+    last_name: str
+    full_name: str = strawberry.field(resolver=get_full_name)
+```
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -14,9 +14,11 @@ For example, the following code now works on Python 3.14+:
 ```python
 import strawberry
 
+
 # Resolver defined before the class it references
 def get_full_name(user: strawberry.Parent[User]) -> str:
     return f"{user.first_name} {user.last_name}"
+
 
 @strawberry.type
 class User:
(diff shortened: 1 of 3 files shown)


botberry:

---

Fix `NameError` when using `strawberry.Parent` with forward references on
Python 3.14+ (PEP 649).

On Python 3.14, annotations are lazily evaluated via `__annotate__` (PEP 649).
When a resolver references a type via `strawberry.Parent[SomeType]` and
`SomeType` is defined after the resolver, `inspect.Signature.from_callable()`
triggers annotation evaluation which raises `NameError`. Strawberry now falls
back to `inspect.Format.FORWARDREF` to handle this gracefully.

For example, the following code now works on Python 3.14+:

```python
import strawberry


# Resolver defined before the class it references
def get_full_name(user: strawberry.Parent[User]) -> str:
    return f"{user.first_name} {user.last_name}"


@strawberry.type
class User:
    first_name: str
    last_name: str
    full_name: str = strawberry.field(resolver=get_full_name)
```

This release was contributed by @Kobusvdwalt in https://github.com/strawberry-graphql/strawberry/pull/4364

Additional contributors: @pre-commit-ci[bot]

bellini666:
Hi @Kobusvdwalt ,

This looks great! :)

Can you check the lint unit test error? Also, worth addressing the comment from sourcery-ai, I think it is a good one

PR #4379, opened 2026-04-23
fix(dataloader): cancel dispatch task when all futures are cancelled

### LINKED ISSUE #3606: Cancelling `DataLoader.load` futures doesn't cancel the underlying task
## Describe the Bug

I'm working in a codebase where we use DataLoaders quite heavily with an SQLAlchemy-based database client. Often, we'd like to await two or more `load()` futures. We use FastAPI and rely on it's dependency injection to handle SQLAlchemy session, forcibly closing it at the end of a request. Unfortunately, currently we're hitting a problem where using `asyncio.gather()` or even `asyncio.TaskGroup()` results in hard to debug errors, whereby dataloader task continues to run even if the `load()` future is cancelled. 

```py
with AsyncSession(db_engine) as session:
    async def user_loader_fn(ids: List[int]) -> List[User]:
        users = await session.select(User).where(User.id.in_(ids)).all()
        return reorder_by_ids(users, ids)
    
    user_dataloader = DataLoader(user_loader_fn)
    async def load_user(id):
        return await user_dataloader.load(id)
    async def load_post():
        raise ValueError("Post not found")

    with asyncio.TaskGroup as tg:
        t1 = tg.create_task(load_post(1)) # this will raise before t2 is done or even properly started, leading to cancellation of t2
        t2 = tg.create_task(load_user(2)) 
# the error from `load_post` will bubble up, closing the session, but user_dataloader will still try to use the session somewhere in the background 
```

To summarise, there's no way to cancel or wait for the underlying dataloader task, making it impossible to cleanly close the resources used by the dataloader. 

I've got a [branch](https://github.com/oinopion/strawberry/tree/dataloader-future-cancellation-issue) ([diff](https://github.com/strawberry-graphql/strawberry/compare/main...oinopion:strawberry:dataloader-future-cancellation-issue?expand=1)) with a test case that shows this problem.

I've also got a very speculative [branch](https://github.com/strawberry-graphql/strawberry/compare/main...oinopion:strawberry:keep-track-of-dispatch-tasks?expand=1) that addresses this issue and improves general handling of task cancellations in the DataLoader. It's not 100% done yet. I'll try to finish it & open a PR in the next few days.

## System Information

 - Strawberry version (if applicable): 0.237.3

## Additional Context

A little bit like https://github.com/strawberry-graphql/strawberry/issues/3414, this issue has to do with keeping track of tasks/futures and a proper resource cleanup.

ChihebBENCHEIKH1:
## Description

Closes #3606.

When all futures in a DataLoader batch are cancelled (e.g. via `asyncio.TaskGroup` cancellation cascade), the dispatch task now gets cancelled too. Previously, the background task created by `dispatch()` was fire-and-forget — its reference was never stored, so it couldn't be cancelled even when no caller was waiting for the results.

This caused real-world issues where cancelled requests would still trigger database queries, API calls, or other expensive operations via `load_fn`, wasting resources and causing errors when those operations used already-closed resources (like SQLAlchemy sessions).

### Changes

**`strawberry/dataloader.py`:**
- Store dispatch `Task` reference in `Batch._dispatch_task`
- Add `done_callback` on each future to cancel the dispatch task when all futures in the batch are cancelled
- Skip `load_fn` call entirely if all futures are already cancelled before dispatch runs

**`tests/test_dataloaders.py`:**
- `test_all_futures_cancelled_skips_load_fn` — verifies load_fn is never called when all futures are cancelled
- `test_cancelled_futures_cancel_dispatch_task` — verifies the dispatch task itself gets cancelled
- `test_partial_cancellation_still_dispatches` — verifies partial cancellation still delivers results to non-cancelled futures

### Example scenario fixed

```python
async with AsyncSession(engine) as session:
    loader = DataLoader(lambda ids: session.execute(...))

    async with asyncio.TaskGroup() as tg:
        tg.create_task(failing_operation())  # raises, cancels group
        tg.create_task(loader.load(42))      # future cancelled by group
    # Previously: loader's dispatch task still ran against closed session
    # Now: dispatch task is cancelled, no wasted query
```

## Types of Changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] My code follows the code style of this project
- [x] My change requires a change to the documentation (N/A)
- [x] I have updated the documentation accordingly (N/A)
- [x] I have read the CONTRIBUTING document
- [x] I have added tests to cover my changes
- [x] All new and existing tests passed (locally verified)

## Summary by Sourcery

Ensure DataLoader dispatch tasks are cancelled when all associated futures are cancelled to avoid unnecessary work.

Bug Fixes:
- Cancel the DataLoader batch dispatch task when all futures in the batch have been cancelled.
- Skip invoking the batch load function when every future in the batch has already been cancelled before dispatch.

Tests:
- Add tests to verify load_fn is skipped when all futures are cancelled, the dispatch task itself is cancelled, and partially cancelled batches still dispatch for remaining futures.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,5 +1,5 @@
 ---
-release type: patch
+release type: minor
 ---
 
 Cancel DataLoader dispatch task when all futures are cancelled.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,14 @@
+---
+release type: patch
+---
+
+Cancel DataLoader dispatch task when all futures are cancelled.
+
+Previously, the background task created by `dispatch()` was fire-and-forget — its
+reference was never stored, so it couldn't be cancelled even when no caller was
+waiting for the results. This caused wasted work (e.g. database queries against
+closed sessions) when using `asyncio.TaskGroup` or similar structured concurrency
+patterns that cancel futures on failure.
+
+The dispatch task is now tracked in `Batch._dispatch_task` and automatically
+cancelled when all futures in the batch are cancelled.
--- a/strawberry/dataloader.py
+++ b/strawberry/dataloader.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import dataclasses
 from abc import ABC, abstractmethod
 from asyncio import create_task, gather, get_event_loop
@@ -41,10 +42,24 @@ class LoaderTask(Generic[K, T]):
 class Batch(Generic[K, T]):
     tasks: list[LoaderTask] = dataclasses.field(default_factory=list)
     dispatched: bool = False
+    _dispatch_task: asyncio.Task[None] | None = dataclasses.field(
+        default=None, repr=False
+    )
 
     def add_task(self, key: Any, future: Future) -> None:
         task = LoaderTask[K, T](key, future)
         self.tasks.append(task)
+        future.add_done_callback(self._on_future_done)
+
+    def _on_future_done(self, future: Future) -> None:
+        """Cancel the dispatch task if all futures in the batch are cancelled."""
+        if (
+            future.cancelled()
+            and self._dispatch_task is not None
+            and not self._dispatch_task.done()
+            and all(t.future.cancelled() for t in self.tasks)
+        ):
+            self._dispatch_task.cancel()
 
     def __len__(self) -> int:
         return len(self.tasks)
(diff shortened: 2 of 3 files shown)


botberry:

---

Cancel DataLoader dispatch task when all futures are cancelled.

Previously, the background task created by `dispatch()` was fire-and-forget — its
reference was never stored, so it couldn't be cancelled even when no caller was
waiting for the results. This caused wasted work (e.g. database queries against
closed sessions) when using `asyncio.TaskGroup` or similar structured concurrency
patterns that cancel futures on failure.

The dispatch task is now tracked in `Batch._dispatch_task` and automatically
cancelled when all futures in the batch are cancelled.

This release was contributed by @ChihebBENCHEIKH1 in https://github.com/strawberry-graphql/strawberry/pull/4379

RELEASE.md
```diff
@@ -0,0 +1,14 @@
+---
+release type: patch
```
bellini666: I think this deserves to be a `minor`, as it is not actually "fixing a bug", but changing the behavior


```suggestion
release type: minor
```

bellini666 (approved):

ChihebBENCHEIKH1:
Updated to release type: minor, makes sense since this changes behavior rather than fixing a crash. Thanks for the review!

bellini666 (approved):

botberry:
This PR was published as [0.315.0](https://github.com/strawberry-graphql/strawberry/releases/tag/0.315.0). Thank you for contributing!

PR #4380, opened 2026-04-25
feat(schema-codegen): pyright-clean output + --config for scalar overrides

### LINKED ISSUE #4378: `schema-codegen` emits deprecated `ScalarWrapper` output for user-defined scalars, fails pyright
## Describe the Bug

`strawberry schema-codegen` produces Python that fails pyright (default strictness, not just strict mode) when the SDL contains a **user-defined custom scalar** (one not in Strawberry's built-in known-scalar registry). The generated `X = strawberry.scalar(NewType("X", object), ...)` pattern both:

1. Fails every `strawberry.scalar(...)` overload because pyright infers inline `NewType("X", object)` as `FunctionType`, not as `type | NewType`.
2. Produces a `ScalarWrapper` runtime value that can't be used as a type annotation — every downstream field (`field_name: X`) reports `Variable not allowed in type expression`.

This is the deprecated `ScalarWrapper` pattern (deprecated in 0.288.0, per #4227). `strawberry upgrade replace-scalar-wrappers` on the generated file skips it ("1 files skipped"), so there's no built-in migration path for codegen output.

## Scope: unknown scalars only

Scalars that Strawberry's codegen recognizes (e.g. `UUID`, `Date`, `DateTime`, `Decimal`) are emitted as real imports from the stdlib and type-check cleanly. Only scalars that aren't in the registry fall through to the broken `ScalarWrapper` pattern — so any project with a domain-specific custom scalar (`JSONObject`, `BigInt`, `Email`, `URL`, etc.) trips over this.

Mixed SDL demonstrating the split:

```graphql
scalar UUID
scalar Foo
scalar Date

type Query {
  id: UUID!
  data: Foo!
  when: Date!
}
```

Generated output:

```python
import strawberry
from datetime import date       # Date -> real type
from typing import NewType
from uuid import UUID           # UUID -> real type

Foo = strawberry.scalar(NewType("Foo", object), serialize=lambda v: v, parse_value=lambda v: v)

@strawberry.type
class Query:
    id: UUID     # clean
    data: Foo    # "Variable not allowed in type expression"
    when: date   # clean
```

Only the `Foo` field triggers pyright errors.

## Note on `JSONObject`

`JSONObject` is a de facto community convention (emitted by Hasura, Prisma, `graphql-scalars`, and many other tools) for "arbitrary JSON value" — but it's not in Strawberry's `SCALAR_MAPPING` in `strawberry/schema_codegen/__init__.py`, so every project using it hits the `ScalarWrapper` fallback. Adding `JSONObject` (mapped to `strawberry.scalars.JSON`) alongside the existing `JSON` entry would cover the common case.

## Bonus: the existing `JSON` entry is also broken

While investigating, I found that the `JSON` entry already in `SCALAR_MAPPING` (lines 61–64) emits the wrong path:

```python
"JSON": cst.Attribute(
    value=cst.Name("strawberry"),
    attr=cst.Name("JSON"),
),
```

SDL:
```graphql
scalar JSON

type Query {
  data: JSON!
}
```

Generated output:
```python
import strawberry

@strawberry.type
class Query:
    data: strawberry.JSON

```

pyright:
```
error: "JSON" is not a known attribute of module "strawberry" (reportAttributeAccessIssue)
```

This fails at runtime too — importing the generated module (or any code that triggers Strawberry's forward-reference evaluation, e.g. `strawberry.Schema(query=Query)`) raises:
```
AttributeError: module 'strawberry' has no attribute 'JSON'
```

Confirmed at the REPL:
```python
>>> import strawberry
>>> hasattr(strawberry, "JSON")
False
```

The correct reference is `strawberry.scalars.JSON` (i.e. `from strawberry.scalars import JSON`). As is, any SDL with `scalar JSON` produces codegen that both fails pyright and raises `AttributeError` during schema construction.

## Reproduction (minimal unknown-scalar case)

```graphql
scalar JSONObject

type Query {
  data: JSONObject!
}
```

```bash
$ strawberry schema-codegen schema.graphqls -o types.py
$ cat types.py
```

```python
import strawberry
from typing import NewType

JSONObject = strawberry.scalar(NewType("JSONObject", object), serialize=lambda v: v, parse_value=lambda v: v)

@strawberry.type
class Query:
    data: JSONObject

```

```bash
$ pyright types.py
types.py:5:14 - error: No overloads for "scalar" match the provided arguments (reportCallIssue)
types.py:5:32 - error: Argument of type "type[JSONObject]" cannot be assigned to parameter "cls" of type "_T@scalar"
  Type "FunctionType" is not assignable to type "type | NewType" (reportArgumentType)
types.py:9:11 - error: Variable not allowed in type expression (reportInvalidTypeForm)
3 errors, 0 warnings, 0 informations
```

Same three errors reproduce with any unknown scalar name (verified by replacing `JSONObject` with `Foo`) — it's the pattern, not the name.

## Expected Behavior

Codegen should emit output that type-checks cleanly under pyright's default ruleset for unknown scalars too, matching Strawberry's own built-in scalar convention (e.g. `strawberry.scalars.JSON = NewType("JSON", object)`), where the name exported is a plain `NewType` usable as both a field annotation and an argument to `strawberry.scalar(...)`.

Concretely, three orthogonal fixes would resolve the current state:

1. **Fix the existing `JSON` entry** to point at `strawberry.scalars.JSON` instead of the non-existent `strawberry.JSON` — one-line change, independent of the larger refactor.
2. **Add `JSONObject` to `SCALAR_MAPPING`** (also pointing at `strawberry.scalars.JSON`) so the common community convention works out-of-the-box.
3. **Fix the fallback for arbitrary unknown scalars** so the emitted code type-checks under pyright — this is what #4227 appears to be doing via `scalar_map`.

A workaround we use today is a post-processing script that rewrites the codegen output to:
```python
from strawberry.scalars import JSON as JSONObject
```
…which resolves all three errors but changes the runtime scalar name from `JSONObject` to `JSON`. A direct fix from codegen would be preferable.

## Related

- #4227 (open PR) removes `ScalarWrapper` entirely and moves to `scalar_map` in `StrawberryConfig`. Once that lands, codegen will need to emit the new pattern for user-defined scalars — this issue tracks the specific codegen correctness gap under pyright, plus the two smaller registry bugs that are independent of #4227.

## System Info

- Python: **3.13.3**
- Strawberry: verified reproducing on **0.312.3**, **0.314.2**, and **0.314.3** (each installed in a fresh venv; codegen output is byte-identical across the three)
- pyright: **1.1.408**
- OS: **macOS 26.4.1**
- Reliability: **always reproducible** with any user-defined scalar in the SDL (and with `scalar JSON` specifically)

bellini666:
Fix #4378

## Summary

`strawberry schema-codegen` previously produced output that failed pyright's
default ruleset whenever the SDL contained a user-defined custom scalar, and
referenced a non-existent `strawberry.JSON` attribute (raising
`AttributeError` at schema construction). This PR fixes both problems and
adds a `--config / -c` flag for telling codegen how specific scalars should
be emitted.

### 1. Pyright-clean output for custom scalars

Replaced the deprecated `Foo = strawberry.scalar(NewType("Foo", object), ...)`
pattern with a module-level `Foo = NewType("Foo", object)` binding registered
through `StrawberryConfig.scalar_map` — the recommended Strawberry API for
custom scalars. The `scalar_map` is emitted as a module-level constant so
descriptions and `@specifiedBy` URLs survive even on scalars-only SDL
fragments.

### 2. Fix the `JSON` registry entry

`_SCALAR_MAP["JSON"]` previously emitted `strawberry.JSON`, which does not
exist. Now emits `from strawberry.scalars import JSON` plus a bare `JSON`
field annotation.

### 3. New: `--config / -c <file>` for scalar overrides

The single section today is `scalars:`, a mapping from a GraphQL scalar name
to a `<module>:<object>` Python target. For each entry, codegen imports the
target instead of generating a `NewType` plus `scalar_map` entry.

```yaml
# codegen.yaml
scalars:
  JSONObject: strawberry.scalars:JSON
  MyDecimal: my_app.scalars:MyDecimal
  Date: my_app.scalars:UnixDate
```

```shell
strawberry schema-codegen schema.graphql -c codegen.yaml
```

`JSONObject` is the canonical example: Strawberry doesn't ship a `JSONObject`
scalar, but several schemas use that name as a synonym for `JSON`. Pointing
it at `strawberry.scalars:JSON` via the config is the recommended way to
wire that convention up — it is intentionally not a built-in. Overrides win
over the built-in scalar mappings, so the same mechanism can redirect
`Date`, `JSON`, `UUID`, etc. to custom implementations.

## Test plan

- [x] `poetry run pytest tests/schema_codegen/ tests/cli/test_schema_codegen.py` — 90/90 pass
- [x] `poetry run mypy --config-file mypy.ini` — clean (239 files)
- [x] `poetry run pyright strawberry/schema_codegen strawberry/cli/commands/schema_codegen.py` — 0 errors
- [x] `poetry run ruff check` + `ruff format --check` — clean
- [x] End-to-end smoke: generated module imports successfully and `strawberry.Schema` builds for both fall-through (`scalar_map`) and override paths

## Summary by Sourcery

Update schema code generation to produce Pyright-clean custom scalar output and support configurable scalar overrides via a YAML-driven CLI flag.

New Features:
- Add YAML-based config support to schema codegen, including a -c/--config CLI option for specifying scalar overrides.
- Allow schema codegen to import custom scalar implementations instead of generating NewType-based placeholders, including support for aliased imports.

Bug Fixes:
- Correct the generated annotation for the JSON scalar to use the actual strawberry.scalars.JSON implementation instead of a non-existent strawberry.JSON attribute.
- Change custom scalar codegen to use top-level NewType bindings plus a scalar_map configuration so generated modules type-check under Pyright.

Enhancements:
- Emit a shared scalar_map constant and wire it into StrawberryConfig when generating schemas with non-built-in scalars, preserving descriptions and specifiedBy URLs across fragments.
- Extend import generation to handle aliased imports and ensure deterministic sorting of imports that may include aliases.
- Document schema-codegen output options, custom scalar handling, and config-file usage in the user guide.
- Add tests covering scalar_map registration, scalar override behavior, malformed config handling, and multiple-unknown-scalar aggregation.

Build:
- Add pyyaml as a CLI dependency for loading schema-codegen configuration files.

Documentation:
- Document writing schema-codegen output to files, the behavior of built-in and custom scalars, and the new YAML config mechanism for scalar overrides.

Tests:
- Expand schema-codegen unit and CLI tests to cover scalar_map generation, override precedence over built-ins, config parsing validation, and the new --config flag behavior.

Chores:
- Add a release note describing the improved Pyright compatibility for custom scalar codegen and the new config-based override mechanism.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,60 @@
+Release type: patch
+
+`strawberry schema-codegen` now produces output that type-checks cleanly
+under pyright's default ruleset for user-defined custom scalars, and the
+`JSON` / `JSONObject` registry entries have been repaired.
+
+Previously, an SDL like:
+
+```graphql
+scalar JSON
+scalar JSONObject
+scalar Foo
+
+type Query {
+  a: JSON!
+  b: JSONObject!
+  c: Foo!
+}
+```
+
+generated code that referenced the non-existent `strawberry.JSON`
+attribute (raising `AttributeError` at schema construction) and emitted
+the deprecated `ScalarWrapper` pattern for `Foo` (`Foo = strawberry.scalar(NewType("Foo", object), ...)`),
+which fails three pyright checks per scalar.
+
+The generated output now uses `from strawberry.scalars import JSON`,
+`from strawberry.scalars import JSON as JSONObject` for the
+community-standard `JSONObject` convention, and a module-level `NewType`
+binding registered through `StrawberryConfig.scalar_map` (the recommended
+Strawberry API for custom scalars) for arbitrary user-defined scalars:
+
+```python
+from __future__ import annotations
+import strawberry
+from strawberry.scalars import JSON
+from strawberry.scalars import JSON as JSONObject
+from strawberry.schema.config import StrawberryConfig
+from typing import NewType
+
+Foo = NewType("Foo", object)
+
+
+@strawberry.type
+class Query:
+    a: JSON
+    b: JSONObject
+    c: Foo
+
+
+schema = strawberry.Schema(
+    query=Query,
+    config=StrawberryConfig(
+        scalar_map={
+            Foo: strawberry.scalar(
+                name="Foo", serialize=lambda v: v, parse_value=lambda v: v
+            ),
+        }
+    ),
+)
+```
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,85 +1,39 @@
 Release type: minor
 
-`strawberry schema-codegen` now produces output that type-checks cleanly under
-Pyright's default ruleset for user-defined custom scalars, and adds a
-`-c / --config` option for telling codegen how specific scalars should be
-emitted.
+Two improvements to `strawberry schema-codegen`:
 
-## Pyright-clean custom scalar output
-
-Previously, an SDL like:
-
-```graphql
-scalar Foo
-
-type Query {
-  c: Foo!
-}
-```
-
-generated a `Foo = strawberry.scalar(NewType("Foo", object), ...)` binding,
-which fails three Pyright checks per scalar and uses the deprecated
-`ScalarWrapper` pattern.
-
-The generated output now uses a module-level `NewType` binding plus a
-`scalar_map` registered through `StrawberryConfig` (the recommended Strawberry
-API):
+**Pyright-clean output for custom scalars.** Previously, `scalar Foo` generated
+`Foo = strawberry.scalar(NewType("Foo", object), ...)` — the deprecated
+`ScalarWrapper` pattern, which fails three Pyright checks per scalar. Output
+now uses a module-level `NewType` binding plus a `scalar_map` registered
+through `StrawberryConfig` (the recommended API), preserving descriptions and
+`@specifiedBy` URLs:
 
 ```python
-from __future__ import annotations
-import strawberry
-from strawberry.schema.config import StrawberryConfig
-from strawberry.types.scalar import ScalarDefinition
-from typing import NewType
-
 Foo = NewType("Foo", object)
-
-
-@strawberry.type
-class Query:
-    c: Foo
-
 
 scalar_map: dict[object, ScalarDefinition] = {
     Foo: strawberry.scalar(name="Foo", serialize=lambda v: v, parse_value=lambda v: v),
 }
 
-schema = strawberry.Schema(
-    query=Query,
-    config=StrawberryConfig(scalar_map=scalar_map),
-)
+schema = strawberry.Schema(query=Query, config=StrawberryConfig(scalar_map=scalar_map))
 ```
 
-The `scalar_map` is emitted as a module-level constant so it is preserved even
-when codegen runs against a scalars-only SDL fragment, and so descriptions and
-`@specifiedBy` URLs are never silently dropped.
-
-## Config file
-
-`strawberry schema-codegen` now accepts `-c / --config <file>` pointing at a
-YAML config. The first available section is `scalars:`, a mapping from a
-GraphQL scalar name to a `<module>:<object>` Python target. For each entry,
-codegen imports the target instead of generating a `NewType` plus `scalar_map`
-entry.
+**New `-c / --config <file>` flag.** Point codegen at a YAML config to control
+how specific scalars are emitted. Today the only section is `scalars:`, a
+mapping from a GraphQL scalar name to a `<module>:<object>` Python target —
+codegen imports the target instead of generating a `NewType` for it.
 
 ```yaml
 # codegen.yaml
 scalars:
   JSONObject: strawberry.scalars:JSON
   MyDecimal: my_app.scalars:MyDecimal
-  Date: my_app.scalars:UnixDate
 ```
 
 ```shell
 strawberry schema-codegen schema.graphql -c codegen.yaml
 ```
 
-For a scalar named `JSONObject` mapped to `strawberry.scalars:JSON` the
-generated code now contains `from strawberry.scalars import JSON as JSONObject`
-and skips the `NewType` / `scalar_map` machinery entirely. Overrides win over
-the built-in scalar mappings, so the same mechanism can redirect `Date`,
-`JSON`, `UUID`, etc. to custom implementations.
-
-Note: previous pre-release builds of this branch auto-mapped `JSONObject` to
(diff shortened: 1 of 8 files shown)


botberry:

---

Two improvements to `strawberry schema-codegen`:

**Pyright-clean output for custom scalars.** Previously, `scalar Foo` generated
`Foo = strawberry.scalar(NewType("Foo", object), ...)` — the deprecated
`ScalarWrapper` pattern, which fails three Pyright checks per scalar. Output
now uses a module-level `NewType` binding plus a `scalar_map` registered
through `StrawberryConfig` (the recommended API), preserving descriptions and
`@specifiedBy` URLs:

```python
Foo = NewType("Foo", object)

scalar_map: dict[object, ScalarDefinition] = {
    Foo: strawberry.scalar(name="Foo", serialize=lambda v: v, parse_value=lambda v: v),
}

schema = strawberry.Schema(query=Query, config=StrawberryConfig(scalar_map=scalar_map))
```

**New `-c / --config <file>` flag.** Point codegen at a YAML config to control
how specific scalars are emitted. Today the only section is `scalars:`, a
mapping from a GraphQL scalar name to a `<module>:<object>` Python target —
codegen imports the target instead of generating a `NewType` for it.

```yaml
# codegen.yaml
scalars:
  JSONObject: strawberry.scalars:JSON
  MyDecimal: my_app.scalars:MyDecimal
```

```shell
strawberry schema-codegen schema.graphql -c codegen.yaml
```

Overrides win over the built-in scalar mappings, so `Date`, `JSON`, `UUID`,
etc. can also be redirected to custom implementations.

This release was contributed by @bellini666 in https://github.com/strawberry-graphql/strawberry/pull/4380

botberry:


<table>
	<thead>
		<tr>
			<th>Federation 1 Support</th>
			<th>Federation 2 Support</th>
		</tr>
	</thead>
	<tbody>
    <tr>
      <td>
        <table>
          <tr><th><code>federated tracing</code></th><td>🟢</td></tr>
        </table>
      </td>
      <td>
        <table>
        </table>
      </td>
		</tr>
	</tbody>
</table>


Learn more:

bellini666:
@sourcery-ai summary

bellini666:
@sourcery-ai review

PR #4381, opened 2026-04-25
Isolate extension state per request to fix concurrency leaks

### LINKED ISSUE #4369: Extension instances are not isolated per request, race condition with concurrent execution
## Describe the Bug

Hi there 😁🖐️

When using extensions with concurrent requests (threads for sync, asyncio.gather for async), extension state leaks between requests. The execution_context assigned to an extension instance belongs to whichever request set it last, not to the request currently being processed.

## System Information

 - Operating system: macOS
 - Python version: 3.12
 - Strawberry version (if applicable): 0.314.3

## Additional Context

https://github.com/strawberry-graphql/strawberry/issues/3172

> I think we should maybe disallow passing extension instances, you should this:

```
schema = strawberry.Schema(
    Query,
    extensions=[
-        MaskErrors(),
+        MaskErrors,
    ],
)
```

That works for extensions with no required arguments, but many built-in extensions require configuration

```python
extensions=[
    MaxTokensLimiter(max_token_count=1000),
    QueryDepthLimiter(max_depth=10),
]
```

A callable factory could work in theory:

```python
extensions=[
    lambda execution_context: MaxTokensLimiter(max_token_count=1000),
]
```

But there are two problems with this approach:

- It doesn't fix the sync path, the cache invalidation was only applied to async (https://github.com/strawberry-graphql/strawberry/pull/4256), so sync concurrent requests would still race
- It breaks the existing type annotation `Iterable[Union[type[SchemaExtension], SchemaExtension]]`

Examples:

2 requests

```
{ helloAsync }
{ test: helloAsync }
```

<details>
<summary>as instnce CustomMaxTokensLimiter()</summary>

```python
import asyncio
import time
from concurrent.futures import ThreadPoolExecutor

import strawberry
from strawberry.extensions import MaxTokensLimiter


@strawberry.type
class Query:

    @strawberry.field
    def hello() -> str:
        time.sleep(0.1)
        return 'world'

    @strawberry.field
    async def hello_async() -> str:
        await asyncio.sleep(0.1)
        return 'world'



class CustomMaxTokensLimiter(MaxTokensLimiter):

    def get_results(self) -> dict[str, str | bool]:
        return {
            "query": str(self.execution_context.query)
        }


schema = strawberry.Schema(
    query=Query,
    extensions=[
        CustomMaxTokensLimiter(max_token_count=100),
    ],
)


async def run_async():
    print("-- async --")
    results = await asyncio.gather(
        schema.execute("{ helloAsync }"),
        schema.execute("{ test: helloAsync }"),
    )
    for i, result in enumerate(results):
        print(f"Request {i + 1}: {result}")


def run_sync():
    print("-- sync (threads) --")
    queries = ["{ hello }", "{ test: hello }"]
    with ThreadPoolExecutor(max_workers=2) as executor:
        futures = {executor.submit(schema.execute_sync, q): i for i, q in enumerate(queries)}
        for future, i in futures.items():
            result = future.result()
            print(f"Request {i + 1}: {result}")


if __name__ == "__main__":
    run_sync()
    asyncio.run(run_async())

```

</details>

```
-- sync (threads) --
Request 1: ExecutionResult(data={'hello': 'world'}, errors=None, extensions={'query': '{ test: hello }'})
Request 2: ExecutionResult(data={'test': 'world'}, errors=None, extensions={'query': '{ test: hello }'})
-- async --
Request 1: ExecutionResult(data={'helloAsync': 'world'}, errors=None, extensions={'query': '{ test: helloAsync }'})
Request 2: ExecutionResult(data={'test': 'world'}, errors=None, extensions={'query': '{ test: helloAsync }'})
```
Passing instances is fundamentally broken. Both sync and async concurrent requests end up sharing the same execution_context instead of getting their own.

<details>
<summary>as class CustomMaxTokensLimiter</summary>

```python
import asyncio
import time
from concurrent.futures import ThreadPoolExecutor

import strawberry
from strawberry.extensions import MaxTokensLimiter


@strawberry.type
class Query:

    @strawberry.field
    def hello() -> str:
        time.sleep(0.1)
        return 'world'

    @strawberry.field
    async def hello_async() -> str:
        await asyncio.sleep(0.1)
        return 'world'



class CustomMaxTokensLimiter(MaxTokensLimiter):
    def __init__(
        self,
        execution_context,
        max_token_count: int = 100,
    ) -> None:
        self.max_token_count = max_token_count

    def get_results(self) -> dict[str, str | bool]:
        return {
            "query": str(self.execution_context.query)
        }


schema = strawberry.Schema(
    query=Query,
    extensions=[
        CustomMaxTokensLimiter,
    ],
)


async def run_async():
    print("-- async --")
    results = await asyncio.gather(
        schema.execute("{ helloAsync }"),
        schema.execute("{ test: helloAsync }"),
    )
    for i, result in enumerate(results):
        print(f"Request {i + 1}: {result}")


def run_sync():
    print("-- sync (threads) --")
    queries = ["{ hello }", "{ test: hello }"]
    with ThreadPoolExecutor(max_workers=2) as executor:
        futures = {executor.submit(schema.execute_sync, q): i for i, q in enumerate(queries)}
        for future, i in futures.items():
            result = future.result()
            print(f"Request {i + 1}: {result}")


if __name__ == "__main__":
    run_sync()
    asyncio.run(run_async())

```

</details>

```
-- sync (threads) --
Request 1: ExecutionResult(data={'hello': 'world'}, errors=None, extensions={'query': '{ test: hello }'})
Request 2: ExecutionResult(data={'test': 'world'}, errors=None, extensions={'query': '{ test: hello }'})
-- async --
Request 1: ExecutionResult(data={'helloAsync': 'world'}, errors=None, extensions={'query': '{ helloAsync }'})
Request 2: ExecutionResult(data={'test': 'world'}, errors=None, extensions={'query': '{ test: helloAsync }'})
```

Sync is broken. Both requests return the same `execution_context` ({ test: hello }) instead of their own. Async works correctly here since the cache was already fixed for that path.

# Solution

1. Remove the cache from the sync path (same fix already applied to async)
2. Fix documentation
3. Ensure a fresh extension instance is created for each request

as options:

- fix type annotation 

```python
Iterable[type[SchemaExtension] | Callable[[], SchemaExtension]] = (),
```
And guide users toward using a lambda instead of passing instances directly

```python
# before
extensions=[MaxTokensLimiter(max_token_count=1000)]

# after
extensions=[lambda: MaxTokensLimiter(max_token_count=1000)]
```

- recreating under the hood

```python
class MaxTokensLimiter(SchemaExtension):
    def __call__(self, execution_context) -> "MaxTokensLimiter":
        extension = MaxTokensLimiter(max_token_count=self.max_token_count)
        extension.execution_context = execution_context
        return extension
```

I don't see a clean way to generalize this, but the option is still on the table.

- And last but not least (© Snoop Dogg) store execution_context as a ContextVar instead of an instance attribute.

I can help with the implementation, but we need to agree on the approach first. What's your preference? 😌

link https://github.com/strawberry-graphql/strawberry/issues/3148

bellini666:
## Summary

- Replace the per-extension `execution_context` instance attribute with a `contextvars.ContextVar`. Concurrent requests sharing an extension instance now each see their own request's context.
- Move the per-request mutable state of `ApolloTracingExtension`, `DatadogTracingExtension`, and `OpenTelemetryExtension` (and their `*Sync` variants) into per-extension `ContextVar`s, so passing them as instances to `extensions=[…]` is safe under `asyncio.gather` and threaded `execute_sync`.
- Reading `self.execution_context` outside a lifecycle hook now raises `RuntimeError` instead of returning a stale value.

Fix #4369

## Test plan

- [x] `poetry run pytest tests/schema/extensions/ tests/extensions/`
- [x] `poetry run pytest tests/schema/`
- [x] `poetry run mypy --config-file mypy.ini` on changed files
- [x] `poetry run ruff check strawberry/`
- [x] New regression suite at `tests/schema/extensions/test_isolation.py` covers the reporter's exact repro (sync threaded + async gathered, instance- and class-passed) plus concurrent `ApolloTracingExtension` instance under both paths.

## Summary by Sourcery

Isolate per-request execution and tracing state for GraphQL extensions using context variables to prevent concurrency leaks, and add regression tests and documentation for the new behavior.

Bug Fixes:
- Prevent leakage of ExecutionContext and tracing state between concurrent async and threaded executions when extensions are passed as instances.
- Fix Apollo, Datadog, and OpenTelemetry tracing extensions so concurrent requests no longer share resolver stats or span objects.

Enhancements:
- Back SchemaExtension.execution_context with a ContextVar and make access outside lifecycle hooks fail fast with RuntimeError.
- Refine extension context manager and runner plumbing so execution contexts are bound and reset correctly for execute, execute_sync, and subscribe flows.
- Adjust GraphQL WS subscription cleanup to close result generators in the originating task context to keep context variables consistent.

Documentation:
- Document ContextVar-backed execution_context and guidance for storing request-scoped state in custom extensions.

Tests:
- Add a regression suite ensuring extension state is isolated across concurrent async, sync, and subscription executions, including built-in tracing extensions and nested executions.

### DIFF
--- a/docs/guides/custom-extensions.md
+++ b/docs/guides/custom-extensions.md
@@ -291,3 +291,20 @@ class MyExtension(SchemaExtension):
         yield
         self.execution_context.context["db"].close()
 ```
+
+`execution_context` is backed by a
+[context variable](https://docs.python.org/3/library/contextvars.html), so
+concurrent requests sharing the same extension instance each see their own
+request's context. Reading `self.execution_context` outside an extension
+lifecycle hook (`on_operation`, `on_validate`, `on_parse`, `on_execute`,
+`resolve`, `get_results`) raises `RuntimeError`.
+
+### Request isolation
+
+If your extension needs to store request-scoped state, prefer holding it on
+`self.execution_context.extensions_results` or in a
+[`contextvars.ContextVar`](https://docs.python.org/3/library/contextvars.html)
+rather than on `self` directly. Plain `self` attributes are only safe when the
+extension is passed as a class (so the schema instantiates one per request); if
+you pass an extension as an instance, two concurrent requests share the same
+object and any mutable attribute on `self` will race.
--- a/strawberry/extensions/base_extension.py
+++ b/strawberry/extensions/base_extension.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import contextvars
 from collections.abc import Callable
 from enum import Enum
 from typing import TYPE_CHECKING, Any
@@ -19,14 +20,49 @@ class LifecycleStep(Enum):
     RESOLVE = "resolve"
 
 
-class SchemaExtension:
-    execution_context: ExecutionContext
+# The currently-running request's ``ExecutionContext``. Set by the schema at
+# the start of every ``execute`` / ``execute_sync`` / ``subscribe`` call and
+# read via :attr:`SchemaExtension.execution_context`. Using a context variable
+# keeps a single shared extension instance safe under concurrent execution
+# (asyncio.gather, threaded ``execute_sync``) — each task/thread observes its
+# own request's context. See issue #4369.
+execution_context_var: contextvars.ContextVar[ExecutionContext | None] = (
+    contextvars.ContextVar("strawberry.extension.execution_context", default=None)
+)
 
+
+class SchemaExtension:
     # to support extensions that still use the old signature
     # we have an optional argument here for ease of initialization.
     def __init__(
         self, *, execution_context: ExecutionContext | None = None
     ) -> None: ...
+
+    @property
+    def execution_context(self) -> ExecutionContext:
+        """The currently-running request's :class:`ExecutionContext`.
+
+        Backed by a context variable, so concurrent requests sharing this
+        extension instance each see their own context. Accessing this outside
+        an extension lifecycle hook raises ``RuntimeError``.
+        """
+        ec = execution_context_var.get()
+        if ec is None:
+            raise RuntimeError(
+                "ExecutionContext is only available inside an extension "
+                "lifecycle hook (on_operation, on_validate, on_parse, "
+                "on_execute, resolve, get_results)."
+            )
+        return ec
+
+    @execution_context.setter
+    def execution_context(self, value: ExecutionContext) -> None:
+        # Backwards-compat setter for code that does
+        # ``self.execution_context = ec`` (notably older third-party
+        # extensions). Updates the context variable so reads see the new value
+        # within the current task/thread.
+        execution_context_var.set(value)
+
     def on_operation(  # type: ignore
         self,
     ) -> AsyncIteratorOrIterator[None]:  # pragma: no cover
@@ -79,4 +115,10 @@ HOOK_METHODS: set[str] = {
     SchemaExtension.on_execute.__name__,
 }
 
-__all__ = ["HOOK_METHODS", "Hook", "LifecycleStep", "SchemaExtension"]
+__all__ = [
+    "HOOK_METHODS",
+    "Hook",
+    "LifecycleStep",
+    "SchemaExtension",
+    "execution_context_var",
+]
--- a/strawberry/extensions/tracing/apollo.py
+++ b/strawberry/extensions/tracing/apollo.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import contextvars
 import dataclasses
 import time
 from datetime import datetime, timezone
@@ -17,9 +18,6 @@ if TYPE_CHECKING:
     from graphql import GraphQLResolveInfo
 
 DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
-
-if TYPE_CHECKING:
-    from strawberry.types.execution import ExecutionContext
 
 
 @dataclasses.dataclass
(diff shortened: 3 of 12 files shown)


botberry:

---

Fixes a concurrency bug where extension state could leak between
requests running in parallel.

Two cases were affected:

- Custom extensions that read `self.execution_context` (e.g. inside
  `get_results`) and were passed as instances to `extensions=[…]` —
  concurrent requests could observe each other's `ExecutionContext`.
- The built-in `ApolloTracingExtension`, `DatadogTracingExtension`,
  and `OpenTelemetryExtension` (plus their `*Sync` variants) — when
  passed as instances, concurrent requests' tracing data could be
  mixed.

Both are now isolated per request, so passing tracing extensions as
instances is safe.

If you write custom extensions, note that reading
`self.execution_context` outside an extension lifecycle hook
(`on_operation`, `on_validate`, `on_parse`, `on_execute`, `resolve`,
`get_results`) now raises `RuntimeError` instead of returning a stale
value.

This release was contributed by @bellini666 in https://github.com/strawberry-graphql/strawberry/pull/4381

strawberry/extensions/tracing/apollo.py
```diff
@@ -81,61 +79,102 @@ def to_json(self) -> dict[str, Any]:
...
+)
+
+
+def _get_state() -> _ApolloTracingState:
+    state = _apollo_state_var.get()
+    if state is None:
+        raise RuntimeError(
+            "ApolloTracingExtension state is not initialised: this method "
+            "must be called inside the on_operation lifecycle hook."
+        )
```
bellini666, line 113: Wondering if this should instead lazily initialize when state is `None`, then on `on_operation` we just do a simple `_get_state()`. wdyt @patrick91 ? Or is this way better?

bellini666: Actually, no, because then we lose the `token` to reset it after

strawberry/schema/schema.py
```diff
@@ -695,12 +694,10 @@ def execute_sync(
...
         extensions = self._sync_extensions
-        # TODO (#3571): remove this when we implement execution context as parameter.
-        for extension in extensions:
-            extension.execution_context = execution_context
-
         extensions_runner = self.create_extensions_runner(execution_context, extensions)
-        middleware_manager = self._get_middleware_manager(extensions)
+        # Uncached: extensions are fresh per request, so the middleware
+        # manager wrapping them must be too.
+        middleware_manager = self._get_middleware_manager(extensions, cached=False)
```
Arfey, line 700: The `_cached_middleware_manager` parameter and its related logic can be removed. It's unused.

strawberry/schema/schema.py
```diff
@@ -380,14 +380,12 @@ def get_extensions(self, sync: bool = False) -> list[SchemaExtension]:
             for ext in extensions
         ]
 
-    @cached_property
+    @property
```
Arfey, line 383: `_sync_extensions` and `_async_extensions` can be removed. They're just aliases for `self.get_extensions`, and were originally added for caching.

strawberry/extensions/base_extension.py
```diff
@@ -19,14 +20,49 @@ class LifecycleStep(Enum):
...
+        return ec
+
+    @execution_context.setter
+    def execution_context(self, value: ExecutionContext) -> None:
+        # Backwards-compat setter for code that does
+        # ``self.execution_context = ec`` (notably older third-party
+        # extensions). Updates the context variable so reads see the new value
+        # within the current task/thread.
+        execution_context_var.set(value)
+
```
Arfey, line 65: I don't follow why that's not enough. There are only 3 places where `execution_context` is set:

- subscribe
- execute_sync
- execute
```
extension.execution_context = execution_context
```

Nothing overwrites it during execution, so `get_extensions_results_sync` should be safe to use as-is without passing `execution_context` explicitly. No need to touch `ExtensionContextManagerBase` or `OperationContextManager`.

Could you elaborate? 🙂

docs/guides/custom-extensions.md
```diff
@@ -291,3 +291,20 @@ class MyExtension(SchemaExtension):
         yield
         self.execution_context.context["db"].close()
 ```
+
```
Arfey, line 294: The `ExecutionCache` extension example still uses `self.cache_key`. Worth fixing

Arfey (commented):
It's better than before, but `ContextVar` doesn't solve everything. Extensions still feel fragile on the user side:

- Users shouldn't rely on `self.<property_name>` because it's unsafe, but nothing stops them if they skip the docs.
- Users have to manage context manually and set it in the right place.

Overall it overcomplicates the extension API. 😕

If you have time and think it's worth exploring, I'm happy to discuss different approaches, compare the tradeoffs, and figure out what's better 😌

---
 
Solution: new extension instance per request

```python
from collections.abc import Callable
from dataclasses import field
from typing import Any

@dataclass(frozen=True, slots=True)
class SchemaExtension:
    execution_context: Callable[..., Any] | None = field(default=None)
    
    def on_operation(self): ...
    def on_validate(self): ...

    ...
```

`MaskErrors` would look like this:

```python
@dataclass(frozen=True, slots=True)
class MaskErrors(SchemaExtension):
    should_mask_error: Callable[[GraphQLError], bool] = field(default=default_should_mask_error)
    error_message: str = "Unexpected error."
```
changes:

- remove `__init__` method
- add `dataclass` decorator

Advantages:

- Properties are read-only, so should_mask_error and error_message can't be mutated at runtime. No conflicts.
- Safe to use self, since we create a fresh instance per request via replace(ext, execution_context=execution_context).

Breaking changes:

1. Shared data. Extensions that rely on shared state in `__init__` will break. Two built-in extensions already do this: `ValidationCache` and `ParserCache`.

```
def __init__(self, maxsize: int | None = None) -> None:
        self.cached_parse_document = lru_cache(maxsize=maxsize)(parse)
```

The cache won't survive across requests. For built-in extensions we can work around it, but existing user extensions are a real concern.

2. `frozen=True`. Right now users can freely add dynamic attributes or mutate class variables. It's unsafe, but it works. `frozen=True` fixes the safety issue but breaks any existing extension that does this.

Thoughts?

bellini666:
Discussed this with @patrick91 , and we decided that contextvars here would overcomplicate extensions too much.

I opened a new PR in https://github.com/strawberry-graphql/strawberry/pull/4392, which suggests a factory pattern for extensions that have config, while copying and emitting a warning for legacy instance-passing ones.

PR #4396, opened 2026-05-04
Fix example in strawberry.argument docstring

tsauerwein:
## Description

The docstring for `strawberry.argument` has the following [example](https://github.com/strawberry-graphql/strawberry/blob/d15c2d45e4bf8434fb22e332600ffdef3ec04624/strawberry/types/arguments.py#L326-L338):

```
import strawberry


@strawberry.type
class Query:
    @strawberry.field
    def example(
        self, info, value: int = strawberry.argument(description="The value")
    ) -> int:
        return value
```

When trying to print the SDL for a schema with this type, the following error is raised:

```
>>> schema = strawberry.Schema(query=Query)
>>> schema.as_str()
  File "/Users/.../.venv/lib/python3.13/site-packages/graphql/type/scalars.py", line 56, in serialize_int
    num = int(output_value)  # raises ValueError if not an integer
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'StrawberryArgumentAnnotation'
```

The problem is that the `StrawberryArgumentAnnotation` instance returned by `strawberry.argument` is interpreted as default for the argument. And then it fails because an integer is expected.

This PR fixes the example, so that `typing.Annotated` is used.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation

*

## Checklist

## Summary by Sourcery

Documentation:
- Fix the strawberry.argument docstring example to use typing.Annotated and a typed info parameter so the example reflects the correct usage.

### DIFF
--- a/strawberry/types/arguments.py
+++ b/strawberry/types/arguments.py
@@ -335,7 +335,7 @@ def argument(
         def example(
             self,
             info: strawberry.Info,
-            value: Annotated[int, strawberry.argument(description="The value")]
+            value: Annotated[int, strawberry.argument(description="The value")],
         ) -> int:
             return value
     ```
--- a/strawberry/types/arguments.py
+++ b/strawberry/types/arguments.py
@@ -325,6 +325,7 @@ def argument(
 
     Example:
     ```python
+    from typing import Annotated
     import strawberry
 
 
@@ -332,7 +333,9 @@ def argument(
     class Query:
         @strawberry.field
         def example(
-            self, info, value: int = strawberry.argument(description="The value")
+            self,
+            info: strawberry.Info,
+            value: Annotated[int, strawberry.argument(description="The value")]
         ) -> int:
             return value
     ```


botberry:

```markdown
---
release type: patch
---

```

tsauerwein:
Could it make sense to add a schema validation rule to check that `StrawberryArgumentAnnotation` is not used as default for arguments?

It's even worse when the type of the argument is a string. Because in that case, Strawberry silently accepts the default and generates a schema like this:

```
>>> from typing import Annotated
... import strawberry
...
...
... @strawberry.type
... class Query:
...     @strawberry.field
...     def example(
...         self,
...         info: strawberry.Info,
...         value: str = strawberry.argument(description="The value")
...     ) -> int:
...         return value
KeyboardInterrupt
>>> schema = strawberry.Schema(query=Query)
>>> schema.as_str()
'type Query {\n  example(value: String! = "<strawberry.types.arguments.StrawberryArgumentAnnotation object at 0x10b7f06e0>"): Int!\n}'
```

This would be caught by a type checker. But probably not everyone using Strawberry is also using a type checker.

bellini666 (approved):
ty! :)

PR #4415, opened 2026-05-14
fix: resolve module-level lazy alias under future annotations

### LINKED ISSUE #3568: From future import annotations breaks lazy types
## Describe the Bug

`from __future__ import annotations` breaks type resolver which results in `TypeError: Model fields cannot be resolved. Unexpected type 'typing.Any'`

I did not manage to track down what exactly causes this issue, but whenever I add `from __future__ import annotations` to the top of the file with strawberry types - strawberry type resolver breaks

## System Information

 - Operating system:
 - Strawberry version (if applicable): 0.235.2

bellini666:
Fix `UnresolvedFieldTypeError` triggered by combining `from __future__ import annotations` with a module-level `Annotated[..., strawberry.lazy(...)]` alias. Inlining `Annotated[...]` directly on the field already worked; the aliased form did not.

In `eval_type`, `_eval_type` resolves `ForwardRef('LazyUser')` to its bound value `Annotated[ForwardRef('User'), StrawberryLazyReference(...)]`, but the inner `ForwardRef('User')` stays unresolved because `User` is only imported under `TYPE_CHECKING`. The fix calls `StrawberryLazyReference.resolve_forward_ref` right after `_eval_type` to convert it to a `LazyType`, keeping the `Annotated` wrapper intact.

Regression tests cover the bare alias and a `list[Alias]` variant. One existing test inlined `Annotated` under future annotations, another used a module-level alias without future annotations; neither combined the two.

Fix #3568

## Summary by Sourcery

Fix resolution of module-level lazy Annotated aliases when using future annotations so that aliased lazy types behave like inlined annotations.

Bug Fixes:
- Resolve UnresolvedFieldTypeError when a module-level Annotated[..., strawberry.lazy(...)] alias is used under from __future__ import annotations.
- Ensure lazy Annotated aliases inside list[...] are correctly resolved to their underlying types under future annotations.

Documentation:
- Add release notes describing the fix for lazy Annotated aliases with future annotations.

Tests:
- Add regression tests for module-level lazy Annotated aliases and their list-wrapped variants under from __future__ import annotations.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,19 @@
+Release type: patch
+
+Fix `UnresolvedFieldTypeError` when using `from __future__ import annotations` together with a module-level `Annotated[..., strawberry.lazy(...)]` alias. The aliased form now resolves the same as inlining `Annotated[...]` on the field.
+
+```python
+from __future__ import annotations
+from typing import TYPE_CHECKING, Annotated
+import strawberry
+
+if TYPE_CHECKING:
+    from .user import User
+
+LazyUser = Annotated["User", strawberry.lazy(".user")]
+
+
+@strawberry.type
+class Post:
+    user: LazyUser  # previously failed
+```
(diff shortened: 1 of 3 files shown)


bellini666:
@sourcery-ai summary

patrick91 (approved):

PR #4416, opened 2026-05-14
Add offset argument to relay connections

### LINKED ISSUE #3053: Add offset argument to relay connection
It would be very useful to add an `offset` argument to relay connection. It's a feature that is implemented in graphene. See https://github.com/graphql-python/graphene-django/blob/v3.1.5/graphene_django/fields.py#L139-L145

## Feature Request Type

- [ ] Core functionality
- [ ] New behavior

## Description
When using this library's `limit/offset` pagination, `pageInfo` doesn't update `hasPreviousPage` and `hasNextPage`, they are always false. Currently, I am using a custom connection as a workaround:
```python
import strawberry
from strawberry.relay.utils import from_base64, to_base64

import strawberry_django
from strawberry_django import relay

@strawberry.type
class CustomConnection(relay.ListConnectionWithTotalCount[strawberry.relay.NodeType]):
    """
    A strawberry connection to count the number of query results
    """

    # Adding offset argument to custom connection
    @classmethod
    def resolve_connection(
        cls,
        nodes: NodeIterableType[strawberry.relay.NodeType],
        *,
        info: Info,
        before: Optional[str] = None,
        after: Optional[str] = None,
        first: Optional[int] = None,
        last: Optional[int] = None,
        offset: Optional[int] = None,
        **kwargs: Any,
    ) -> AwaitableOrValue[Self]:

        # This implemntation is based on the graphene
        # implementation of first/offset pagination
        if offset:
            if after:
                offset += from_base64(after) + 1
            # input offset starts at 1 while the offset starts at 0
            after = to_base64("arrayconnection", offset - 1)

        conn = super().resolve_connection(
            nodes,
            info=info,
            before=before,
            after=after,
            first=first,
            last=last,
            **kwargs,
        )

        if inspect.isawaitable(conn):

            async def wrapper():
                resolved = await conn
                resolved.nodes = nodes
                return resolved

            return wrapper()

        conn = cast(Self, conn)
        conn.nodes = nodes
        return conn

```

But that means a custom resolver would have to look like this:
```python
    @strawberry.field
    def foos(
        self,
        info: strawberry.types.Info,
        before: Optional[str] = None,
        after: Optional[str] = None,
        first: Optional[int] = None,
        last: Optional[int] = None,
        offset: Optional[int] = None,
    ) -> Optional[CustomConnection[FooType]]:
        # Resolver logic here returning queryset qs
        return CustomConnection[FooType].resolve_connection(info=info, nodes=qs, offset=offset, first=first, last=last, after=after, before=before)
```

iamdinhthuan:
## Summary

- add an `offset` argument to relay connection fields
- apply `offset` in `SliceMetadata` so `ListConnection` skips the requested number of items and keeps `pageInfo` accurate
- update relay docs, release note, schema snapshots, and regression coverage

Fixes #3053

## Tests

- `uv run pytest tests/relay -q`
- `uv run pytest -q`
- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run mypy`
- `git diff --check`

## Summary by Sourcery

Add support for an offset argument on Relay list connections to allow skipping items before applying cursor-based pagination.

New Features:
- Introduce an offset argument on Relay connection fields and ListConnection resolvers to skip a specified number of items before pagination.

Enhancements:
- Integrate offset into SliceMetadata calculation so pagination ranges and pageInfo remain accurate when skipping items.
- Validate offset as a non-negative integer and propagate it through connection resolution utilities.

Documentation:
- Update Relay documentation and schema examples to describe the new offset argument and its behavior.

Tests:
- Extend Relay connection, slice metadata, and custom edge tests and schema snapshots to cover the new offset behavior and validation.

Chores:
- Add a release note announcing offset support for relay.ListConnection.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,6 @@
+Release type: minor
+
+`relay.ListConnection` now supports an `offset` argument in addition to
+`before`, `after`, `first`, and `last`. This lets clients skip a known number of
+items without first converting that offset into a cursor, while keeping
+`pageInfo` consistent with the skipped range.
--- a/docs/guides/relay.md
+++ b/docs/guides/relay.md
@@ -138,6 +138,7 @@ type Query {
     after: String = null
     first: Int = null
     last: Int = null
+    offset: Int = null
   ): FruitConnection!
 }
 ```
@@ -335,9 +336,10 @@ By default the connection will automatically insert some arguments for it to be
 able to paginate the results. Those are:
 
 - `before`: Returns the items in the list that come before the specified cursor
-- `after`: Returns the items in the list that come after the " "specified cursor
+- `after`: Returns the items in the list that come after the specified cursor
 - `first`: Returns the first n items from the list
-- `last`: Returns the items in the list that come after the " "specified cursor
+- `last`: Returns the last n items from the list
+- `offset`: Skips the first n items from the list
 
 You can still define extra arguments to be used by your own resolver or custom
 pagination logic. For example, suppose we want to return the pagination of all
@@ -367,6 +369,7 @@ type Query {
     after: String = null
     first: Int = null
     last: Int = null
+    offset: Int = null
   ): FruitConnection!
 }
 ```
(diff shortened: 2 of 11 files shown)


iamdinhthuan:
I checked the bot review notes against the current diff. The `last + offset` path is now covered by `test_connection_with_last_and_offset_argument`, and the Relay docs describe `last` as "Returns the last n items from the list" next to the new `offset` entry.

Focused check: `uv run pytest tests/relay/test_schema.py::test_connection_with_last_and_offset_argument -q` passes locally.

PR #4431, opened 2026-05-27
Feat: Support for TypedDicts

GabrielTDS-dev:
Add native support for `typing.TypedDict` GraphQL object and input types via the new `@strawberry.typed_dict` and `@strawberry.typed_dict_input` decorators.

## Description

TypedDicts can now be used to define GraphQL schemas backed directly by Python dictionaries, including support for nested TypedDicts, `Required`, `NotRequired`, `total=False`, and `typing.Annotated` metadata.

Example:

```python
from typing import TypedDict

import strawberry


@strawberry.typed_dict
class User(TypedDict):
    id: int
    name: str


@strawberry.type
class Query:
    @strawberry.field
    def get_user(self) -> User:
        return {"id": 1, "name": "Alice"}
```

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

* #2187

## Checklist

## Summary by Sourcery

Add native support for using typing.TypedDict to define Strawberry GraphQL object and input types backed by plain Python dictionaries.

New Features:
- Introduce @strawberry.typed_dict and @strawberry.typed_dict_input decorators to map TypedDicts to GraphQL object and input types.
- Add runtime validation utilities for TypedDict-backed data, including TypedDictValidationError and validate_typed_dict.
- Support nested TypedDicts, Required/NotRequired keys, total=False, Optional/PEP 604 unions, and metadata via typing.Annotated when generating GraphQL schemas.

Enhancements:
- Expose typed_dict and typed_dict_input from the main strawberry package and extend Strawberry object definitions to track TypedDict-backed types.

Documentation:
- Document TypedDict support, including usage for object and input types, nullability and required keys, metadata/directives, and runtime validation, with new examples in the type docs and a dedicated TypedDicts guide.

Tests:
- Add comprehensive tests covering TypedDict output and input handling, nullability semantics, nesting, lists, metadata via Annotated, validation behavior, and error cases for invalid usage.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,29 @@
+Release type: patch
+
+Add native support for `typing.TypedDict` GraphQL object and input types via
+the new `@strawberry.typed_dict` and `@strawberry.typed_dict_input` decorators.
+
+TypedDicts can now be used to define GraphQL schemas backed directly by Python
+dictionaries, including support for nested TypedDicts, `Required`,
+`NotRequired`, `total=False`, and `typing.Annotated` metadata.
+
+Example:
+
+```python
+from typing import TypedDict
+
+import strawberry
+
+
+@strawberry.typed_dict
+class User(TypedDict):
+    id: int
+    name: str
+
+
+@strawberry.type
+class Query:
+    @strawberry.field
+    def get_user(self) -> User:
+        return {"id": 1, "name": "Alice"}
+```
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,4 +1,4 @@
-Release type: patch
+Release type: minor
 
 Add native support for `typing.TypedDict` GraphQL object and input types via
 the new `@strawberry.typed_dict` and `@strawberry.typed_dict_input` decorators.
--- a/docs/types/input-types.md
+++ b/docs/types/input-types.md
@@ -89,6 +89,47 @@ completely absent (common in update operations), you can use `strawberry.Maybe`.
 See the [Maybe documentation](./maybe.md) for comprehensive examples and usage
 patterns.
 
+## Using TypedDicts for Inputs
+
+If you prefer your resolvers to receive standard Python dictionaries rather than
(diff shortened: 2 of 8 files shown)


Ladol:
@greptileai

erikwrede (changes requested):
Hey @GabrielTDS-dev, thank you for working on this. It's a long-standing request. I want to be honest with you though: I don't think this is close to mergeable, and the gap is in the design rather than anything you can patch in a few commits. Summarizing the major issues in a few bullets:

- **Separate decorators** Every type decorator in strawberry is named for the **GraphQL** type it produces — `type`, `input`, `interface`, `enum`. There's no TypedDict in GraphQL, so `typed_dict` / `typed_dict_input` name the Python *source* and tack the GraphQL role on as a `_input` suffix. That gives users a second, parallel way to declare an object/input type, named after an implementation detail. A TypedDict isn't a new *kind* of type, at most it's an alternative way to *back* a normal `@strawberry.type` / `@strawberry.input`. All additions should follow this opinionated pattern.

- **Runtime identity for unions** In several code paths, we do direct checks on incumbents of an interface / union to determine the type to resolve. A resolver here returns a bare `dict`, which carries no type information. In order to stay efficient, we need to think about solutions like specific discriminator fields like `__typename`. This could also be solved in a stacked pr, but the base design has to allow for it, so we need to spec it now.

- **Global registries** your PR introduces a global registry while strawberry is generally schema-scoped. an application can have multiple (incompatible) strawberry schemas running at the same time - this will clash with your approach.

Happy to provide feedback or guidance on any alternative solutions you come up with!

GabrielTDS-dev:
Hello @erikwrede. I'm developing with some support from @Ladol (who was my co-author on this PR) an alternative solution that targets fixing the problems pointed out by you about this current implementation. The PR for this will be added soon (we'll try to make it this week) after some final refinements, hoping that it satisfies Strawberry structural requirements better.

GabrielTDS-dev:
The new PR has been added at #4465

PR #4440, opened 2026-05-31
Match scalar_overrides generic containers by origin (#4433)

### LINKED ISSUE #4433: scalar_overrides should match generic containers by origin, not exact key
# Strawberry: `scalar_overrides` should match generic containers by origin, not exact key

**Repository:** https://github.com/strawberry-graphql/strawberry

## Summary

`strawberry.Schema(scalar_overrides={...})` uses **exact-key dictionary lookup** (`strawberry/schema/schema_converter.py:813`). It does not unwrap generic containers. Registering `dict[str, Any] → JSON` does not match `dict[str, int]`, `dict[str, Optional[int]]`, `dict[str, list[X]]`, or any other parameterization — each variant requires its own separate entry in the override map.

For the common case of "I want every `dict[*, *]` annotation to surface as the conventional `JSON` scalar," users either:

1. Enumerate every variant their codebase happens to produce, growing the list every time the schema crashes on a new shape (whack-a-mole), or
2. Monkey-patch `GraphQLCoreConverter.from_type` to short-circuit `get_origin(type_) is dict` to `JSON`.

Both are bad. The library should let users say "any `dict[*, *]` resolves to `JSON`" once.

## Why this matters

This bites hardest when integrating with libraries that emit typed `dict[K, V]` field annotations on Strawberry-rendered types — most notably **`strawberry.experimental.pydantic.type`**. A Pydantic model with a `dict[str, Optional[int]]` field gets wrapped into a Strawberry type, and schema construction fails with:

```
TypeError: SomeModelType fields cannot be resolved. Unexpected type 'dict[str, typing.Optional[int]]'
```

Users now have to enumerate every concrete parameterization their Pydantic models happen to use, even though the intent ("treat it as JSON") is uniform.

## Proposed change

Lookup in `GraphQLCoreConverter.from_type` (and `StrawberryAnnotation` where applicable) should:

1. First try exact match in `scalar_registry` (current behavior — preserves backward compatibility).
2. If no exact match, try `get_origin(type_)` as a fallback key.

So registering `{dict: JSON}` matches any `dict[K, V]`. Registering `{list: SomeScalar}` matches any `list[X]`. Etc.

Concrete sketch:

```python
# strawberry/schema/schema_converter.py, around line 901
if compat.is_scalar(type_, self.scalar_registry):
    return self.from_scalar(type_)

# NEW — generic container fallback
origin = typing.get_origin(type_)
if origin is not None and compat.is_scalar(origin, self.scalar_registry):
    return self.from_scalar(origin)

raise TypeError(f"Unexpected type '{type_}'")
```

This is conservative: users still opt in by registering the unparameterized type (`dict`, `list`, etc.); the only change is that one registration covers every parameterization.

## Workaround being used today

```python
from strawberry.schema.schema_converter import GraphQLCoreConverter

_original_from_type = GraphQLCoreConverter.from_type

def _from_type_with_dict_fallback(self, type_):
    if get_origin(type_) is dict:
        return _original_from_type(self, strawberry.scalars.JSON)
    return _original_from_type(self, type_)

GraphQLCoreConverter.from_type = _from_type_with_dict_fallback
```

Works, but it's a monkey-patch and only covers `dict`. The right fix is library-level.

## Repro

```python
import strawberry
from typing import Optional

class Config(BaseModel):
    settings: dict[str, Optional[int]]

@strawberry.experimental.pydantic.type(Config, all_fields=True)
class ConfigType:
    pass

@strawberry.type
class Query:
    cfg: ConfigType

# Registering dict[str, Any] does NOT cover dict[str, Optional[int]]:
schema = strawberry.Schema(
    query=Query,
    scalar_overrides={dict[str, Any]: strawberry.scalars.JSON},
)
# TypeError: ConfigType fields cannot be resolved.
# Unexpected type 'dict[str, typing.Optional[int]]'
```

## Environment

- strawberry-graphql 0.x
- Python 3.11+

Vansh-Sharma27:
## Description

Fixes #4433.

`scalar_overrides` currently does an exact-key dictionary lookup, so registering `dict[str, Any] -> JSON` does **not** match `dict[str, int]`, `dict[str, Optional[int]]`, or any other parameterization. You end up having to add a new entry every time a different `dict[...]` shape shows up in the schema, which is easy to miss and annoying to maintain (it comes up a lot with `strawberry.experimental.pydantic` types that have `dict[K, V]` fields).

This PR adds an origin fallback in `GraphQLCoreConverter.from_type`: if there's no exact scalar match, it tries `typing.get_origin(type_)`. So registering the bare `dict` once covers every `dict[K, V]`.

```python
@strawberry.type
class Query:
    settings: dict[str, int]
    metadata: dict[str, list[str]]

schema = strawberry.Schema(
    query=Query,
    scalar_overrides={dict: JSON},   # one entry covers both fields now
)
```

## Implementation

The change is in `from_type` (`strawberry/schema/schema_converter.py`):

```python
if compat.is_scalar(type_, self.scalar_registry):
    return self.from_scalar(type_)

# NEW: fall back to the generic container's origin
origin = typing.get_origin(type_)
if origin is not None and compat.is_scalar(origin, self.scalar_registry):
    return self.from_scalar(origin)

raise TypeError(f"Unexpected type '{type_}'")
```

It's intentionally conservative:
- The exact match runs first, so existing overrides behave exactly as before.
- The fallback only kicks in when the user has registered the unparameterized type (`dict`, `list`, ...), so it's opt-in.
- If nothing matches, the same `TypeError` is raised as before.

This is basically the approach suggested in the issue body.

## Tests

Added two tests in `tests/schema/test_scalars.py`:
- `test_override_generic_container_by_origin` – a bare `dict` override covers different `dict[K, V]` fields, including a nested value type.
- `test_override_exact_generic_key_still_matches` – the existing exact-key override still works (backward-compat guard).

Ran the scalar test suite locally (all passing) and checked `ruff` / `ruff format` / `mypy` on the touched files. Also included a `RELEASE.md` (minor) as per CONTRIBUTING.

## Summary by Sourcery

Allow scalar_overrides to match parameterized generic containers by their origin type while preserving existing exact-match behavior.

New Features:
- Support registering scalar_overrides for unparameterized generic containers (e.g., dict) that automatically apply to all their parameterizations.

Bug Fixes:
- Prevent Unexpected type errors when scalar_overrides are defined only for the origin of a generic container but used with parameterized variants.

Documentation:
- Add a minor release note documenting the new scalar_overrides behavior for generic containers.

Tests:
- Add tests to verify generic container overrides apply by origin and that exact generic keys still take precedence.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,30 +1,35 @@
 Release type: minor
 
-`scalar_overrides` now matches generic containers by their origin as a
+`scalar_overrides` now matches generic `dict` annotations by their origin as a
 fallback, so registering a bare `dict` covers every `dict[K, V]`
 parameterization instead of needing one entry per variant.
 
-Before, a schema like this only worked if you registered the exact
-annotation (`dict[str, Any]`), and any other shape such as
-`dict[str, Optional[int]]` raised `Unexpected type`:
+Before, registering only the unparameterized origin (`dict`) did not match
+parameterized annotations like `dict[str, int]`. You had to register the exact
+annotation, and any other shape such as `dict[str, list[int]]` raised
+`Unexpected type`:
 
 ```python
+from typing import Any
+
 import strawberry
 from strawberry.scalars import JSON
 
 
 @strawberry.type
 class Query:
-    settings: dict[str, int]
-    metadata: dict[str, list[str]]
+    settings: dict[str, Any]
+    metadata: dict[str, list[int]]
 
 
 schema = strawberry.Schema(
     query=Query,
-    scalar_overrides={dict: JSON},
+    scalar_overrides={dict[str, Any]: JSON},
 )
 ```
 
-Now registering the unparameterized `dict` (or `list`, etc.) covers all
-of its parameterizations. Lookup still tries an exact match first, so
-existing overrides keep working unchanged.
+Now registering the unparameterized `dict` covers all of its parameterizations,
+including resolver and mutation arguments typed with `dict[...]`. This is
+especially useful for `strawberry.experimental.pydantic` models with typed
+dictionary fields. Lookup still tries an exact match first, so existing overrides
+keep working unchanged.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,28 @@
+Release type: minor
+
+`scalar_overrides` now matches generic containers by their origin as a
+fallback, so registering a bare `dict` covers every `dict[K, V]`
+parameterization instead of needing one entry per variant.
+
+Before, a schema like this only worked if you registered the exact
+annotation (`dict[str, Any]`), and any other shape such as
+`dict[str, Optional[int]]` raised `Unexpected type`:
+
+```python
+import strawberry
+from strawberry.scalars import JSON
+
+@strawberry.type
+class Query:
+    settings: dict[str, int]
+    metadata: dict[str, list[str]]
+
+schema = strawberry.Schema(
+    query=Query,
+    scalar_overrides={dict: JSON},
+)
+```
+
+Now registering the unparameterized `dict` (or `list`, etc.) covers all
+of its parameterizations. Lookup still tries an exact match first, so
+existing overrides keep working unchanged.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -12,10 +12,12 @@ annotation (`dict[str, Any]`), and any other shape such as
 import strawberry
 from strawberry.scalars import JSON
 
+
 @strawberry.type
 class Query:
     settings: dict[str, int]
     metadata: dict[str, list[str]]
+
 
 schema = strawberry.Schema(
     query=Query,
--- a/strawberry/scalars.py
+++ b/strawberry/scalars.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, NewType
+from typing import TYPE_CHECKING, Any, NewType, get_origin
 
 if TYPE_CHECKING:
     from collections.abc import Mapping
@@ -29,6 +29,11 @@ def is_scalar(
     scalar_registry: Mapping[object, ScalarWrapper | ScalarDefinition],
 ) -> bool:
     if annotation in scalar_registry:
+        return True
+
+    # Keep scalar checks consistent for schema generation and argument conversion.
+    origin = get_origin(annotation)
+    if origin is not None and origin in scalar_registry:
         return True
 
     return hasattr(annotation, "_scalar_definition")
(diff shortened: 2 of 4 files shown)


bellini666 (commented):
Copilot review has a possible valid point on runtime. Can we make the test do an actual query (or an extra test) to check that?

Vansh-Sharma27:
Updated this PR to address the runtime concern from review and keep the scalar override behavior consistent.

What changed:

- `scalar_overrides={dict: JSON}` now works for typed dictionary annotations like `dict[str, int]` by falling back to the registered `dict` origin when there is no exact scalar override for the parameterized annotation.
- The fallback is implemented in the two shared scalar paths:
  - `strawberry/scalars.py`: `is_scalar()` now checks `typing.get_origin(annotation)` after the exact registry lookup. This covers runtime argument conversion because `_is_leaf_type()` uses this predicate.
  - `strawberry/schema/schema_converter.py`: `from_scalar()` now resolves a parameterized generic annotation through its registered origin before reading the scalar definition. This covers schema construction and avoids passing `dict[str, int]` to the scalar resolver without a matching registry key.
- The older bespoke fallback in `from_type()` was removed because it only handled schema conversion. With the predicate and resolver aligned, schema generation and runtime argument conversion now use the same matching rule.
- Exact overrides still take priority. For example, an override registered for `dict[str, Any]` still matches only that exact annotation and does not accidentally apply to `dict[str, int]`.

Why this fixes the reported issue:

- The original PR behavior fixed the schema construction side, but runtime argument conversion had a separate scalar check that still depended on exact registry lookup.
- This version fixes the shared scalar predicate used by runtime conversion and the scalar resolver used by schema construction.
- That avoids registry mutation or ordering assumptions, and it keeps the behavior predictable: exact match first, origin fallback second.

Tests added:

- Query execution with fields typed as `dict[str, int]` and nested typed dictionaries.
- Mutation argument execution for `dict[str, int]`.
- Optional mutation argument execution for `dict[str, int] | None`.
- List argument execution for `list[dict[str, int]]`.
- Negative coverage showing `scalar_overrides={dict[str, Any]: JSON}` does not match `dict[str, int]`.

Release note update:

- Scoped the wording to generic `dict` annotations.
- Fixed the old example so it describes the previous exact-annotation requirement correctly.
- Mentioned resolver/mutation argument support.
- Added the Pydantic context because this helps typed dictionary fields coming from `strawberry.experimental.pydantic` models.

bellini666 (approved):
Thank you! :)

botberry:
This PR was published as [0.317.0](https://github.com/strawberry-graphql/strawberry/releases/tag/0.317.0). Thank you for contributing!

PR #4451, opened 2026-06-11
feat: add lexicographic_sort_schema config option

wjdwl002:
## Summary by Sourcery

Add optional lexicographic schema sorting while preserving definition order by default.

New Features:
- Add a `sort_schema` configuration option that alphabetically orders schema types, fields, and arguments in introspection and exported SDL.

Documentation:
- Document the `sort_schema` option, its default behavior, and its effect on generated SDL.

Tests:
- Add coverage for the default definition order, lexicographic sorting, and schema execution with sorting enabled.

### DIFF
--- a/strawberry/schema/config.py
+++ b/strawberry/schema/config.py
@@ -34,6 +34,10 @@ class StrawberryConfig:
             any type (including NewType) to be used as a GraphQL scalar with
             proper type checking support.
         batching_config: Configuration for operation batching.
+        lexicographic_sort_schema: Whether to sort the schema's types, fields and
+            arguments lexicographically. This affects both the introspection
+            result and the exported SDL, making it easier to find related
+            fields. Defaults to False, which preserves definition order.
     """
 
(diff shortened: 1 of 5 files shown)


strawberry/schema/config.py
```diff
@@ -34,6 +34,10 @@ class StrawberryConfig:
             any type (including NewType) to be used as a GraphQL scalar with
             proper type checking support.
         batching_config: Configuration for operation batching.
+        lexicographic_sort_schema: Whether to sort the schema's types, fields and
```
bellini666, line 37: Maybe just call this `sort_schema`?

bellini666 (commented):
I like the change! :)

Just made a small suggestion for the name (@patrick91 wtyt?) and should be good to merge

strawberry/schema/config.py
wjdwl002: Addressed [here](https://github.com/strawberry-graphql/strawberry/pull/4451/commits/2ddf9ecaecdd2cd2dc5fad726227a70543bb21da)!

PR #4357, opened 2026-04-09
feat: Add GraphQL-SSE (Server-Sent Events) support for subscriptions

XChikuX:
## Description

This PR introduces native support for GraphQL-SSE (Server-Sent Events) as a subscription transport mechanism in Strawberry GraphQL. This enables real-time subscription updates over HTTP using the SSE protocol, providing a more efficient and persistent connection compared to polling.
Features Added:
1. GraphQL-SSE Subscription Support
   - Full support for streaming subscriptions over HTTP using Server-Sent Events
   - Compatible with the graphql-transport-ws subprotocol
2. Supported HTTP Methods
   - GET: SSE over GET requests for subscriptions
   - POST: SSE over POST requests with JSON body
3. Response Formats
   - SSE (text/event-stream): Streams individual events as they occur
   - Multipart (multipart/mixed): Supports the graphql-transport-ws multipart format
4. Event Types Emitted
   - next: Sent when a new subscription result is available
   - complete: Sent when the subscription stream has finished
   - error: Sent when an error occurs during subscription execution

Technical Details:

Explicitly requires setting `GRAPHQL_SSE_PROTOCOL` to activate

```
    subscription_protocols=[
        GRAPHQL_TRANSPORT_WS_PROTOCOL,
        GRAPHQL_SSE_PROTOCOL,
    ]
```

- Uses AsyncBaseHTTPView._get_sse_stream() and _stream_sse_with_heartbeat() for SSE streaming
- Heartbeat interval of 5 seconds to prevent connection timeouts
- Automatic format selection based on Accept header (text/event-stream vs multipart/mixed)
- Full backward compatibility - existing subscriptions continue to work without modification

```py
    @strawberry.subscription
    async def check(self, info: Info) -> AsyncGenerator[str, None]:
        while True:
            yield "Hello"
            await anyio.sleep(2.5)
```

# Before

## SSE with GET

```sh
06:53:06 root@MSI ~ → curl -N -k -G 'https://localhost:8888/graphql' \
  -H 'Accept: text/event-stream' \
  --data-urlencode 'query=subscription { check }'
{"data":null,"errors":[{"message":"Cannot return null for non-nullable field Subscription.check.","locations":[{"line":1,"column":16}],"path":["check"]}]}06:53:08 
```

## SSE with POST

```sh
06:53:12 root@MSI ~ → curl -N -k -X POST https://localhost:8888/graphql \
  -H 'Accept: text/event-stream' \
  -H 'Content-Type: application/json' \
[...]

PR #4460, opened 2026-06-17
Add Schema.stream

patrick91:
## Stack [🍰](https://shortcake.patrick.wtf)

- #4466 (`2026-06-14-add-support-for-sse`)
- #4474 (`2026-06-20-return-pre-execution-errors-from-execute-and-execu`)
- **#4460** (`2026-06-14-unify-streaming-execution-behind-schema-stream`) <-- this PR
- #4469 (merged) (`2026-06-14-extract-http-stream-transports`)

Shortcake-Parent: main

## Summary by Sourcery

Add a unified streaming execution API that supports queries, mutations, subscriptions, and incremental delivery, and refactor transports to use it while centralising parsing/validation and improving error handling.

New Features:
- Introduce Schema.stream as a shared async streaming entry point for queries, mutations, subscriptions, and incremental delivery frames.
- Expose streaming execution at the abstract schema level so custom integrations can use a single API for all operation types.

Bug Fixes:
- Prevent double-processing of errors during streaming execution and ensure extensions see the correct error state.
- Ensure extension `operation_extensions` are available for subscription and streaming operations.
- Fix multipart subscription handling of non-subscription operations by returning a proper error response instead of executing disallowed operations.
- Avoid graphql-transport-ws crashes or malformed payloads when incremental delivery frames are produced by explicitly rejecting such operations.

Enhancements:
- Refactor parsing, operation-type checking, validation, and execution into shared helpers used by both async and sync execution paths.
- Ensure GraphQL documents can be supplied pre-parsed and that parsing honours schema extensions such as parser caching and max-token limits.
- Unify selection of the graphql-core execution function (including experimental incremental execution) in a single helper.
- Expand incremental execution containers into flat streams of frames so streaming transports receive a uniform sequence of results.
- Improve error coercion for missing or disallowed operations and reuse it across execution and streaming paths.

Documentation:
- Add release notes documenting the new Schema.stream API, its semantics, and how transports integrate with it.

Tests:
- Add comprehensive tests for Schema.stream covering queries, mutations, subscriptions, validation, error handling, extensions, and incremental delivery.
- Extend websocket and HTTP multipart tests to cover single-parse behaviour, parser-cache usage, incremental frame rejection, and error responses for disallowed operations.
- Add async max-tokens tests to verify that async execution respects parser token limits.

PR #4466, opened 2026-06-19
Add support for SSE

patrick91:
Closes #4357 #2091 

## Stack [🍰](https://shortcake.patrick.wtf)

- **#4466** (`2026-06-14-add-support-for-sse`) <-- this PR
- #4460 (merged) (`2026-06-14-unify-streaming-execution-behind-schema-stream`)
- #4469 (merged) (`2026-06-14-extract-http-stream-transports`)

## Summary by Sourcery

Add Server-Sent Events (SSE) as a first-class streaming transport for GraphQL operations alongside existing HTTP multipart and websocket support.

New Features:
- Introduce SSE transport in the HTTP async view for streaming GraphQL queries, mutations, subscriptions, and incremental delivery frames using text/event-stream responses.
- Add SSE-based Apollo and Relay client configurations plus a dedicated SSE UI page and components in the end-to-end test app.

Enhancements:
- Refactor streaming heartbeat handling into a reusable helper and add SSE-specific heartbeat comments to keep connections alive without interfering with protocol semantics.
- Extend request protocol detection and validation to distinguish between HTTP, multipart subscriptions, and SSE, rejecting batched SSE requests and aligning allowed operation types accordingly.
- Enhance HTTP context construction to propagate Authorization headers so SSE requests can share the same auth path as regular HTTP operations.

Build:
- Add the graphql-sse dependency to the e2e project for client-side SSE support.

Documentation:
- Document how to authenticate SSE subscriptions, explain the lack of an SSE connection hook, and describe deployment considerations such as HTTP/2 and connection limits.

Tests:
- Add extensive SSE transport tests across integrations, including subscription, query, authentication, heartbeat, and @stream directive coverage, plus new e2e tests for Apollo and Relay SSE flows.


XChikuX:
@patrick91 Could you please include the connection_params and custom headers changes from my PR ?

This also allows me to use my subscriptions completly unchanged. 

I'd really find it useful of that was enabled along with SSE via `POST` 

Native EventSource is extremely limiting, I don't really prefer using MultiPart requests as they are apollo specific.

patrick91:
> @patrick91 Could you please include the connection_params and custom headers changes from my PR ?
> 
> This also allows me to use my subscriptions completly unchanged.

I'll think about it, but I'm not sure (we can also add it later)

> 
> I'd really find it useful of that was enabled along with SSE via `POST`
> 
> Native EventSource is extremely limiting, I don't really prefer using MultiPart requests as they are apollo specific.

Can you tell me more about this?

XChikuX:
Native EventSource protocol is completely http `GET` based.

This is extremely limiting, especially in the AI era.
In general `GET` requests are not expected to carry custom headers as they are for retrieving websites and some other minor content. SSE was built for that purpose.

However, with the evolution of AI, we now use SSE for various things, including MCP servers.
But, even before that change, several people have proposed Polyfills

1. https://www.npmjs.com/package/@microsoft/fetch-event-source
2. https://github.com/mpetazzoni/sse.js/

Also the `graphql-sse` package allows for using

```
import { createClient as createSseClient } from "graphql-sse";

async function createSseClientWithToken(token: string) {
    const client = createSseClient({
        url: 'https://example.com/graphql',

        headers: () => ({
            Authorization: token,
            'User-Agent': 'chrome',
        }),
        fetchFn: fetch,
    });

    return client;
}
```

So it literally makes no sense not to include custom headers, when the ability for SSE to be more secure already exists.

> we can also add it later

Considering, it took over 2 months to get to this feature, I'd like it, if we can finish it now. Especially since the reference implementation is there (live and tested in a prod environment)

XChikuX:
Directly from the `graphql-sse` docstring:

```
    /**
     * HTTP headers to pass along the request.
     *
     * If the option is a function, it will be called on each connection attempt.
     * Returning a Promise is supported too and the connection phase will stall until it
     * resolves with the headers.
     *
     * A good use-case for having a function is when using the headers for authentication,
     * where subsequent reconnects (due to auth) may have a refreshed identity token in
     * the header.
     *
     * The request is passed for distinct connections mode only.
     */
    headers?: Record<string, string> | ((request: SingleConnection extends true ? undefined : RequestParams) => Promise<Record<string, string>> | Record<string, string>);
```

**Distinct Connection Mode** is what we are building.

XChikuX:
We also need `Last-Event-ID` support

patrick91:
> Considering, it took over 2 months to get to this feature, I'd like it, if we can finish it now. Especially since the reference implementation is there (live and tested in a prod environment)

Remember that every feature needs to be maintained by us, I'd love to avoid adding things that are probably not needed (or without understanding exactly why they are needed)

You can see all the work that went into adding this PR, having a reference implementation doesn't really make it easier to add the code. Even with AI being there 😊

XChikuX:
Understood!

I've provided several reasons though :)

XChikuX:
@patrick91 Can we add benchmark tests for WS and SSE if not done already?

I'm not sure benchmarks have been updated for these paths, or done in the first place.

XChikuX:
Since `uvicorn` is default http1.1 and standard with FastAPI

This warning would be useful.

https://github.com/strawberry-graphql/strawberry/pull/4357/changes/f983203af4fbc34bc6af92ce66a1417204bd8832

patrick91:
@XChikuX I opted for updating the docs instead 😊

botberry:
This PR was published as [0.320.0](https://github.com/strawberry-graphql/strawberry/releases/tag/0.320.0). Thank you for contributing!

PR #4483, opened 2026-06-24
fix: preserve NewType in pydantic type resolution for scalar_map

### LINKED ISSUE #4482: scalar_map not being considered for pydantic types in schema generation
Hi everyone,

I have problems mapping types from pydantic models to the GQL types in the schema generation. I tried to use `scalar_map` for this but it is being ignored. For example mapping the `id` field of a model from my python ID class to the GQL `ID` type.

## Describe the Bug
Using `scalar_map` does not work when trying to map a type from a pydantic class to a schema type when building the schema.

Let's say I have this pydantic class:
```
ID = NewType("ID", str)

class PydanticRemoveUserInput(BaseModel):
    """Specify which user to remove."""

    id: ID = Field(..., description="Unique identifier of the user to remove")
```

and I have this strawberry mapping class:
```
@strawberry.experimental.pydantic.input(
    model=PydanticRemoveUserInput,
    all_fields=True,
)
class RemoveUserInput:
    pass
```
and using the scalar_map like this:
```
schema = strawberry.Schema(
    Query,
    Mutation,
    Subscription,
    config=StrawberryConfig(
        scalar_map={
            ID: strawberry.ID,
        }
    ),
)
```

I would expect the resulting schema having a field `id` of type `ID`. But the generated schema has `String` as a type:
```
input RemoveUserInput {
  """Unique identifier of the user to remove"""
  id: String!
}
```
I am using the `scalar_map` successfully for runtime conversion but it seems to fail for the schema generation. 

## System Information
 - Operating system: Linux
 - Python version: 3.12
 - Strawberry version: 0.319.0

## Additional Context
The only way I found for having the type conversion is to add it manually to the strawberry type:
```
@strawberry.experimental.pydantic.input(
    model=users.RemoveUserInput,
    all_fields=True,
)
class RemoveUserInput:
    id: strawberry.ID
```
This gives me:
```
input RemoveUserInput {
  """Unique identifier of the user to remove"""
  id: ID!
}
```
But this is quite error prone when having to do that for a larger number of classes.

gaoflow:
Fixes #4482 — `scalar_map` entries keyed by `typing.NewType` were silently ignored for pydantic model fields during schema generation.

## Problem

The pydantic compat layer (`PydanticV1Compat.get_basic_type` / `PydanticV2Compat.get_basic_type` in `strawberry/experimental/pydantic/_compat.py`) unconditionally unwrapped `typing.NewType` annotations to their underlying supertype (e.g. `ID = NewType("ID", str)` → `str`) during field type resolution. This happens at decorator time, so by the time the schema converter consults `scalar_registry` (which includes `scalar_map` entries), the original NewType identity has already been destroyed — the field type is `str`, not the NewType the user keyed in `scalar_map`.

## Fix

Three coordinated changes:
1. **_compat.py**: Remove the `if is_new_type(type_): return new_type_supertype(type_)` lines from both compat classes (4 lines)
2. **scalars.py**: Add `__supertype__` fallback to `is_scalar()` — if a NewType is not in `scalar_registry`, recursively check its supertype
3. **schema_converter.py**: Add `__supertype__` fallback to `from_scalar()` — unwrap a NewType to its supertype as a last resort

This ensures:
- NewTypes in `scalar_map` are matched correctly (the fix)
- NewTypes without a `scalar_map` entry still resolve via their supertype (backward compatible)
- No behavioral change for existing code that does not use `scalar_map` with NewTypes

This pull request was prepared with the assistance of AI, under my direction and review.

## Summary by Sourcery

Preserve typing.NewType identities during scalar resolution so pydantic-backed fields correctly respect scalar mappings while maintaining fallback behavior via supertypes.

Bug Fixes:
- Ensure scalar_map entries keyed by typing.NewType are honored for pydantic model fields during schema generation.
- Prevent NewType-based scalar annotations from being silently treated as their underlying builtin types when resolving schema scalars.

Enhancements:
- Add recursive supertype fallback in scalar and schema conversion logic so NewTypes without explicit scalar mappings still resolve via their supertypes.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,17 +1,20 @@
 ---
 release type: patch
+social_messages:
+  x: >-
+    {project_name} {version} is out! Pydantic fields now keep their NewType
+    scalar mappings when Strawberry builds the GraphQL schema. 🍓
+    https://strawberry.rocks/release/{version}
+  linkedin: >-
+    {project_name} {version} is out. Pydantic fields annotated with NewType now
+    respect StrawberryConfig scalar mappings during schema generation, keeping
+    the generated GraphQL scalar consistent with runtime conversion.
 ---
 
-Fix `scalar_map` not being applied to pydantic model fields during schema
-generation. Previously, `NewType` wrappers on pydantic fields were
-unconditionally unwrapped to their underlying type (e.g. `str`) during
-pydantic type resolution, which destroyed the type identity before the
-`scalar_map` from `StrawberryConfig` could intercept it at schema
-construction time.
+This release fixes `StrawberryConfig.scalar_map` being ignored for Pydantic
+fields annotated with `NewType`.
 
-This change removes the early `NewType` unwrapping in the pydantic compat
-layer and instead adds `__supertype__` fallback logic to `is_scalar()` and
-`from_scalar()`, so that:
-- `NewType` annotations in `scalar_map` are matched correctly (the fix)
-- `NewType` fields without a `scalar_map` entry still resolve to their
-  underlying type via the new fallback (backward compatible)
+Strawberry now preserves the annotation until schema scalar resolution, so the
+generated GraphQL type and runtime scalar conversion use the same configured
+scalar. Unmapped `NewType` annotations continue to resolve through their
+underlying type.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,17 @@
+---
+release type: patch
+---
+
+Fix `scalar_map` not being applied to pydantic model fields during schema
+generation. Previously, `NewType` wrappers on pydantic fields were
+unconditionally unwrapped to their underlying type (e.g. `str`) during
+pydantic type resolution, which destroyed the type identity before the
+`scalar_map` from `StrawberryConfig` could intercept it at schema
+construction time.
+
+This change removes the early `NewType` unwrapping in the pydantic compat
+layer and instead adds `__supertype__` fallback logic to `is_scalar()` and
+`from_scalar()`, so that:
+- `NewType` annotations in `scalar_map` are matched correctly (the fix)
+- `NewType` fields without a `scalar_map` entry still resolve to their
+  underlying type via the new fallback (backward compatible)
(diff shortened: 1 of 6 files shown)


binggao1230:
Thanks for the review @greptile-apps! Both issues are valid — I've pushed a fix:

1. **NewType wrapping pydantic BaseModel** — `get_basic_type` now conditionally unwraps NewType only when the supertype is a pydantic model (using `is_model_class(supertype)`). Scalar NewTypes (str/int/etc) preserve their identity for `scalar_map` interception.

2. **Chained NewType `from_scalar` inconsistency** — `from_scalar` now walks the full `__supertype__` chain looking for a registry match, matching the recursive behavior of `is_scalar()`. Falls back to single-level unwrap if nothing in the chain is registered.

binggao1230:
Addressed the chained NewType review points in `9ff8baef`.

Changes:
- `get_basic_type` now walks the NewType chain when looking for wrapped pydantic models, so `NewType(NewType(BaseModel))` still resolves through `replace_pydantic_types` correctly while preserving scalar NewType identities.
- `from_scalar` now continues the NewType fallback to the resolved terminal scalar instead of falling back to only the immediate supertype. It also stops when it reaches a custom scalar definition, so chained aliases of `strawberry.scalar(NewType(...))` keep the custom scalar rather than degrading to the builtin supertype.
- Added regression coverage for both cases: a pydantic nested model wrapped through two NewTypes, and a chained NewType alias to a custom scalar covering schema output plus serialize/parse behavior.

Verification:
- `uv run pytest tests/experimental/pydantic/schema/test_basic.py::test_basic_type_with_newtype_chain_nested_model tests/schema/test_scalars.py::test_newtype_chain_to_custom_scalar -q`
- `uv run pytest tests/experimental/pydantic/schema/test_basic.py tests/schema/test_scalars.py -q`
- `uv run ruff format --check strawberry/experimental/pydantic/_compat.py strawberry/schema/schema_converter.py tests/experimental/pydantic/schema/test_basic.py tests/schema/test_scalars.py`
- `uv run ruff check strawberry/experimental/pydantic/_compat.py strawberry/schema/schema_converter.py tests/experimental/pydantic/schema/test_basic.py tests/schema/test_scalars.py`
- `uv run mypy strawberry/experimental/pydantic/_compat.py strawberry/schema/schema_converter.py`
- `git diff --check`

PR #4487, opened 2026-06-28
Allow easy swapping between SSE and WS connections.

XChikuX:
SSE protocol authenticates via authorization headers similar to normal HTTP requests. However, websockets don't have this and use `connection_params` instead.

This PR introduces the ability to easily swap between different protocol providers, for the following reasons:

* Allows incremental upgrade to websocket protocol and vice versa.
* Convenience to pick the protocol of choice without frontend refactor.

## Description

Restores SSE auth convenience features that were dropped during the SSE overhaul.

SSE subscriptions now auto-populate `context["connection_params"]["authorization"]` from the request's `Authorization` header, so SSE resolvers can reuse the same auth code patterns as WebSocket subscriptions without needing a custom `get_context` override solely for auth.

Additionally, a new `on_sse_connect(context)` hook is called before the subscription stream begins. Override it to perform transport-level connection validation. Raise `ConnectionRejectionError` to reject the connection; the client receives an SSE `error` event followed by `complete`.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [x] Documentation

*

## Checklist

## Summary by Sourcery

Align SSE subscriptions with WebSocket authentication and connection handling for easier protocol swapping.

New Features:
- Mirror SSE request Authorization headers into context connection_params for subscriptions.
- Introduce an on_sse_connect hook to allow transport-level acceptance or rejection of SSE connections, returning structured error events on rejection.

Enhancements:
- Add helpers to inject SSE connection params and generate SSE error responses in the async HTTP view.
- Simplify subscription protocol configuration examples in the documentation to focus on the modern WebSocket and SSE transports.

Documentation:
- Update subscription docs to describe SSE Authorization mirroring, unified resolver auth patterns, and the new on_sse_connect hook.
- Refresh protocol configuration examples across frameworks and multipart subscription docs to remove legacy GRAPHQL_WS_PROTOCOL usage.

Tests:
- Add SSE tests covering Authorization header mirroring into connection_params and behavior of on_sse_connect for accepted and rejected connections.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,5 @@
+Release type: minor
+
+Add SSE `connection_params` mirroring and `on_sse_connect` to drop SSE connections at the transport level.
+
+SSE subscriptions mirror the `Authorization` header into `context["connection_params"]["authorization"]`, matching WebSocket behaviour. Override `on_sse_connect` to accept or reject connections at the transport level. Before now, there was no equivalent to `on_ws_connect` for SSE — the only way to reject an unauthorized connection was to throw inside the subscription resolver after the stream had already started.
--- a/docs/general/multipart-subscriptions.md
+++ b/docs/general/multipart-subscriptions.md
@@ -30,7 +30,6 @@ Multipart subscriptions are opt-in. Enable them by including
 from strawberry.fastapi import GraphQLRouter
 from strawberry.subscriptions import (
     GRAPHQL_TRANSPORT_WS_PROTOCOL,
-    GRAPHQL_WS_PROTOCOL,
     MULTIPART_SUBSCRIPTION_PROTOCOL,
 )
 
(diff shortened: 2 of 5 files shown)


XChikuX:
@patrick91 Would be nice to get this in.

patrick91:
@XChikuX I'm not sure we need this, can you open an issue instead describing how you would use this?

I mean, something like this could work for you:

```python
class AuthenticatedGraphQL(GraphQL):
    async def get_context(self, request, response):
        context = await super().get_context(request, response)

        if isinstance(request, Request):
            authorization = request.headers.get("authorization")
            context["connection_params"] = (
                {"authorization": authorization} if authorization else {}
            )

        return context
```

XChikuX:
What about the `on_sse_connect` optimization?

Security wise, I think it's better to allow the denial of the SSE request before it has the opportunity to respond with data. 

The authorization pass through to connection params seemed like a minimal non-invasive addon. But I can live without it. 

@patrick91

PR #4495, opened 2026-06-29
Add Claude release copy review

patrick91:
## Summary

This replaces the broad `@claude` workflow with a focused, label-triggered release-copy reviewer.

- Runs only on `pull_request_target` when the PR has the `release-copy-review` label.
- Uses the existing `CLAUDE_CODE_OAUTH_TOKEN` secret path.
- Checks out only the base branch and limits Claude to read-only tools.
- Prompts Claude to review `RELEASE.md` and social copy, then leave one concise sticky PR comment.
- Pins `anthropics/claude-code-action` to the current `v1.0.159` commit.

## Why

The existing Claude workflow was a general `@claude` responder, but we do not really use it. The release-copy workflow gives us a narrower path that matches the social announcement work: maintainers can add one label when they want suggestions, without making AI review a required or blocking gate.

## Safety

The workflow uses `pull_request_target`, so it intentionally checks out the base branch rather than the PR head. Claude can inspect PR metadata and diffs through read-only `gh` commands, but it cannot edit files, create commits, push branches, or open PRs.

## Validation

- `go run github.com/rhysd/actionlint/cmd/actionlint@latest .github/workflows/claude.yml`
- `uvx zizmor .github/workflows/claude.yml`
- `git diff --check`
- commit hooks passed, with cached Node-version warnings for skipped unhealthy hook installs

## Summary by Sourcery

Introduce a label-triggered Claude workflow that performs focused release copy reviews on pull requests targeting main.

New Features:
- Add a Claude-powered release copy review job that runs on pull_request_target when the release-copy-review label is applied and comments with a concise summary.
- Configure Claude to review RELEASE.md and social announcement copy using read-only GitHub tools and a tailored prompt.
- Pin the anthropics/claude-code-action to a specific v1.0.159 commit for deterministic behavior.

Enhancements:
- Tighten workflow permissions and scope so the Claude action runs only against the base branch with restricted, read-only tooling.

PR #4496, opened 2026-06-29
Use GITHUB_TOKEN for release copy review comment

patrick91:
## Summary

The `Claude release copy review` workflow (added in #4495) fails on every run with:

```
Exchanging OIDC token for app token...
App token exchange failed: 401 Unauthorized - Invalid OIDC token
```

The action was minting a GitHub App token via the OIDC → `api.anthropic.com/api/github/github-app-token-exchange` flow. That endpoint 401s here because the org's **pre-existing** Claude app installation isn't linked to the Anthropic account behind `CLAUDE_CODE_OAUTH_TOKEN` — re-running `/install-github-app` doesn't fix it (the install short-circuits when the app is already present).

## Fix

Pass the workflow's built-in `GITHUB_TOKEN` via the action's `github_token` input. Per `claude-code-action`'s `token.ts`, providing it sets `OVERRIDE_GITHUB_TOKEN` and the action returns the token directly ("Using provided GITHUB_TOKEN for authentication") — skipping the OIDC exchange entirely. The job already grants `pull-requests: write` and `issues: write`, so it can post the sticky comment.

## Notes

- LLM auth is unchanged — the review still runs on `CLAUDE_CODE_OAUTH_TOKEN` (subscription billing).
- Cosmetic: the review comment now posts as `github-actions[bot]` instead of `claude[bot]`.
- `id-token: write` in the job permissions is now unused but left in place; harmless.
- Must land on `main` to take effect — `pull_request_target` always runs the workflow definition from the default branch.

## Summary by Sourcery

CI:
- Update claude-code-action configuration in the release copy review workflow to pass GITHUB_TOKEN via the github_token input instead of relying on GitHub App token exchange.

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -36,6 +36,11 @@ jobs:
         uses: anthropics/claude-code-action@a92e7c70a4da9793dc164451d829089dc057a464 # v1.0.159
         with:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
+          # Use the workflow's built-in token to post the review comment.
+          # Avoids the OIDC -> Claude-app-token exchange, which 401s here
+          # because the org's pre-existing Claude app install isn't linked
+          # to the account behind CLAUDE_CODE_OAUTH_TOKEN.
+          github_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


botberry:

```markdown
---
release type: patch
---

```

Issue #4502, opened 2026-07-06
`print_schema` duplicates enum types referenced by schema directive fields


jonfinerty:
## Describe the Bug

When a `@strawberry.enum` type is used both as a regular schema type (e.g. a mutation argument) and as a field type in a `@schema_directive`, `print_schema` emits the enum definition twice in the SDL output. This produces invalid GraphQL that violates the spec ("Each named type must be defined exactly once within a document") and is rejected by `graphql-core`'s `build_schema`.

## System Information

- strawberry-graphql version: 0.316.0
- Python version: 3.14.3
- graphql-core version: 3.2.6

## Minimal Reproduction

```python
import enum

import strawberry
from strawberry.printer import print_schema
from strawberry.schema_directive import Location


@strawberry.enum
class Role(enum.Enum):
    EDITOR = "editor"
    VIEWER = "viewer"


@strawberry.schema_directive(locations=[Location.FIELD_DEFINITION])
class RequiresRole:
    roles: list[Role]


@strawberry.type
class Item:
    name: str

    @strawberry.field(directives=[RequiresRole(roles=[Role.EDITOR])])
    def secret(self) -> str:
        return "hidden"


@strawberry.type
class Query:
    @strawberry.field
    def item(self) -> Item:
        return Item(name="test")


@strawberry.type
class Mutation:
    @strawberry.mutation
    def assign_role(self, role: Role) -> bool:
        return True


schema = strawberry.Schema(
    query=Query,
    mutation=Mutation,
    schema_directives=[RequiresRole],
)

sdl = print_schema(schema)
print(sdl)

from graphql import build_schema
build_schema(sdl)  # raises TypeError
```

## Actual Output

```graphql
directive @requiresRole(roles: [Role!]!) on FIELD_DEFINITION

schema {
  query: Query
  mutation: Mutation
}

type Item {
  name: String!
  secret: String! @requiresRole(roles: [EDITOR])
}

type Mutation {
  assignRole(role: Role!): Boolean!
}

type Query {
  item: Item!
}

enum Role {
  EDITOR
  VIEWER
}

enum Role {
  EDITOR
  VIEWER
}
```

`graphql-core` rejects this with:

```
TypeError: There can be only one type named 'Role'.
Enum value 'Role.EDITOR' can only be defined once.
Enum value 'Role.VIEWER' can only be defined once.
```

## Expected Output

`enum Role` should appear exactly once.

## Possible Root Cause?

[...]

PR #4503, opened 2026-07-06
Fix print_schema duplicating types referenced by schema directives

jonfinerty:
When a type (e.g. an enum) was used both in a schema directive field and elsewhere in the schema, print_schema would emit its definition twice — once from the type_map and once from extras.types — producing invalid SDL.

Skip extras.types entries that already exist in the schema's type_map.

## Description

See issue: #4502

## Types of Changes

- [ ] Core
- [x] Bugfix
- [ ] New feature
- [ ] Documentation

* #4502

## Checklist

## Summary by Sourcery

Prevent schema printing from duplicating type definitions when types are used both in schema directives and elsewhere in the schema.

Bug Fixes:
- Avoid emitting duplicate SDL definitions for types referenced by schema directives and also present in the schema type map.

Tests:
- Add regression test ensuring enums used in both schema directives and fields are printed once without duplication.

### DIFF
--- a/strawberry/printer/printer.py
+++ b/strawberry/printer/printer.py
@@ -647,6 +647,7 @@ def print_schema(schema: BaseSchema) -> str:
                 )
                 # Make sure extra types are ordered for predictive printing
                 for type_ in sorted(extras.types, key=_name_getter)
+                if _name_getter(type_) not in type_map
             ),
         )
     )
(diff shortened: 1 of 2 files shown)


jonfinerty:
@patrick91 I also you'll want to press ahead with #4504 instead of this? shall I close?

patrick91:
> @patrick91 I also you'll want to press ahead with #4504 instead of this? shall I close?

yes, sorry, I'm adding a new test to make sure renamed types are also supported!

jonfinerty:
No worries, thanks for fixing it so quick

PR #4505, opened 2026-07-06
Fix pydantic SkipJsonSchema[None] in unions being treated as a union instead of optional

### LINKED ISSUE #3992: Pydantic Union with SkipJsonSchema breaks schema generation
## Describe the Bug

When using `SkipJsonSchema` annotation in a type union, the schema generation fails with `Type `str` cannot be used in a GraphQL Union`

Minimal replicating example:

```python
import strawberry
from pydantic.json_schema import SkipJsonSchema


class ModelA(BaseModel):
    field_a: str | SkipJsonSchema[None] = None


    model=ModelA,
    all_fields=True,
)
class TypeA:
    pass


@strawberry.type
class Query:
    @strawberry.field
    def example(self) -> TypeA:
        return TypeA()


print(schema.as_str())
```

This pattern is useful for denoting the field as not required but without explicit `type: none` in JSON schema generated from Pydantic

## System Information

 - Operating system: Ubuntu
 - Python version: 3.13
 - Strawberry version (if applicable): 0.282.0

Pablosinyores:
Fixes #3992

## Description

Schema generation raised `InvalidUnionTypeError` when a Pydantic model field used `SkipJsonSchema[None]` in a union:

```python
import strawberry
from pydantic.json_schema import SkipJsonSchema

class ModelA(BaseModel):
    field_a: str | SkipJsonSchema[None] = None

@strawberry.experimental.pydantic.type(model=ModelA, all_fields=True)
class TypeA:
    pass
```

```
strawberry.exceptions.invalid_union_type.InvalidUnionTypeError: Type `str` cannot be used in a GraphQL Union
```

`SkipJsonSchema[None]` expands to `Annotated[None, SkipJsonSchema()]`, so the field annotation is `str | Annotated[None, SkipJsonSchema()]`. In `replace_types_recursively` (`strawberry/experimental/pydantic/fields.py`), the `Annotated` branch reduced `converted` to just the underlying type but then called `copy_with(...)`, which re-wraps it back into `Annotated[None, SkipJsonSchema()]`. The leftover annotation meant the wrapped `None` was not recognised as `None`, so `str | Annotated[None, ...]` was treated as a real (invalid) GraphQL union instead of an optional `str`.

## Fix

Return the unwrapped underlying type for the `Annotated` branch instead of re-wrapping it. This drops the annotation metadata (which has no meaning in the GraphQL schema) and matches the behaviour the existing `test_annotated`/`test_nested_annotated` tests already expect (`Annotated[int, "metadata"]` resolves to bare `int`). The field now correctly becomes a nullable `String`.

## Tests

Added `test_annotated_none_in_union_is_optional` (guarded by `needs_pydantic_v2`) asserting the field resolves to `StrawberryOptional[str]` and that schema generation succeeds. It fails on `main` (field resolves to `StrawberryUnion`) and passes with the fix. Full `tests/experimental/pydantic/` suite passes (153 passed).

## Summary by Sourcery

Treat Pydantic fields using Annotated None (e.g. SkipJsonSchema[None]) as optional rather than GraphQL unions during schema generation.

Bug Fixes:
- Fix InvalidUnionTypeError when Pydantic model fields use SkipJsonSchema[None] in a union type by correctly recognising annotated None as optional.

Documentation:
- Add RELEASE notes describing the fix for SkipJsonSchema[None] unions being handled as optional fields.

Tests:
- Add a Pydantic v2-only test verifying that fields with SkipJsonSchema[None] in a union become StrawberryOptional and that schema generation succeeds.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,27 @@
+---
+release type: patch
+social_messages:
+  x: >-
+    {project_name} {version} fixes schema generation crashing when a Pydantic
+    field uses `SkipJsonSchema[None]` in a union — the field is now correctly an
+    optional.
+  linkedin: >-
+    {project_name} {version} fixes a bug where a Pydantic model field using
+    `SkipJsonSchema[None]` in a union (e.g. `str | SkipJsonSchema[None]`)
+    crashed schema generation with `InvalidUnionTypeError`. The annotated `None`
+    is now treated as null, so the field becomes an optional as expected.
+---
+
+This release fixes schema generation raising `InvalidUnionTypeError` when a
+Pydantic model field uses `SkipJsonSchema[None]` in a union type, for example:
+
+```python
+class ModelA(BaseModel):
+    field_a: str | SkipJsonSchema[None] = None
+```
+
+`SkipJsonSchema[None]` expands to `Annotated[None, SkipJsonSchema()]`, and the
+annotated `None` was not recognised as `None`, so the field was treated as a
+(invalid) GraphQL union of `str` and `None` instead of an optional `str`. The
+annotation metadata is now stripped during conversion, so the field correctly
+becomes a nullable `String`.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -4,7 +4,7 @@ social_messages:
   x: >-
     {project_name} {version} fixes schema generation crashing when a Pydantic
     field uses `SkipJsonSchema[None]` in a union — the field is now correctly an
-    optional.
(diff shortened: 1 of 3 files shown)


Pablosinyores:
Tightened the schema assertion per Greptile's note — `"fieldA: String"` was a substring of `"fieldA: String!"`, so it wouldn't have caught a regression to a non-null field. Now asserts the exact nullable line (`fieldA: String\n`) and that `fieldA: String!` is absent, so the test actually proves the field is optional. CI green.

Issue #4507, opened 2026-07-07
@dataclass_transform on overloads is not picked up by PyCharm for @strawberry.type


Speedy1991:
PyCharm does not infer the generated dataclass-style `__init__` for classes decorated with `@strawberry.type`.

Example:

```python
@strawberry.type
class SuccessType:
    message: str
    obj_id: strawberry.ID | None = None

SuccessType(message="ok")
```

PyCharm reports:

```text
Unexpected argument
```

Runtime behavior is correct because Strawberry wraps the class with `dataclasses.dataclass(kw_only=True)` internally.

In `strawberry/types/object_type.py`, `@dataclass_transform(...)` is currently applied to the overload declarations of `type`/`input`/`interface`, but PyCharm appears to ignore it there when resolving the public decorator from `strawberry.type`.

Moving or duplicating the same `@dataclass_transform(...)` annotation onto the concrete implementation fixes the IDE warning:

```python
@overload
def type(...) -> T: ...

@overload
def type(...) -> Callable[[T], T]: ...

@dataclass_transform(
    order_default=True,
    kw_only_default=True,
    field_specifiers=(field, StrawberryField),
)
def type(...):
    ...
```

Same applies to `input`; `interface` already had the implementation-level transform in this installed version.

This keeps runtime behavior unchanged and improves PyCharm/static-analysis support for generated keyword constructors.

patrick91:
Fixed in https://github.com/strawberry-graphql/strawberry/pull/4508

PR #4508, opened 2026-07-07
fix for #4507

Speedy1991:
PyCharm does not infer the generated dataclass-style `__init__` for classes decorated with `@strawberry.type`.

Example:

```python
@strawberry.type
class SuccessType:
    message: str
    obj_id: strawberry.ID | None = None

SuccessType(message="ok")
```

PyCharm reports:

```text
Unexpected argument
```

Runtime behavior is correct because Strawberry wraps the class with `dataclasses.dataclass(kw_only=True)` internally.

In `strawberry/types/object_type.py`, `@dataclass_transform(...)` is currently applied to the overload declarations of `type`/`input`/`interface`, but PyCharm appears to ignore it there when resolving the public decorator from `strawberry.type`.

Moving or duplicating the same `@dataclass_transform(...)` annotation onto the concrete implementation fixes the IDE warning:

```python
@overload
def type(...) -> T: ...

@overload
def type(...) -> Callable[[T], T]: ...

@dataclass_transform(
    order_default=True,
    kw_only_default=True,
    field_specifiers=(field, StrawberryField),
)
def type(...):
    ...
```

Same applies to `input`; `interface` already had the implementation-level transform in this installed version.

This keeps runtime behavior unchanged and improves PyCharm/static-analysis support for generated keyword constructors.
## Types of Changes

- [ ] Core
- [ ] Bugfix
- [ ] New feature
- [ ] Documentation

* https://github.com/strawberry-graphql/strawberry/issues/4507

## Checklist

## Summary by Sourcery

Ensure Strawberry's public decorators expose dataclass transformation metadata for better static analysis support.

Bug Fixes:
- Expose dataclass transform metadata on the concrete implementations of strawberry.type and strawberry.input so IDEs can infer generated __init__ signatures.

Tests:
- Add tests to verify that strawberry.type, strawberry.input, and strawberry.interface expose the expected __dataclass_transform__ metadata.

### DIFF
--- a/tests/test_type.py
+++ b/tests/test_type.py
@@ -47,18 +47,3 @@ def test_get_object_definition_strict():
         match=r".* does not have a StrawberryObjectDefinition",
     ):
         get_object_definition(OtherFruit, strict=True)
-
-
-def test_public_decorators_have_dataclass_transform():
-    decorators = (strawberry.type, strawberry.input, strawberry.interface)
-
-    for decorator in decorators:
-        transform = decorator.__dataclass_transform__
-
-        assert transform["order_default"] is True
-        assert transform["kw_only_default"] is True
-
-        field_specifiers = transform["field_specifiers"]
-
-        assert field_specifiers
-        assert all(callable(field_specifier) for field_specifier in field_specifiers)
--- a/tests/test_type.py
+++ b/tests/test_type.py
@@ -6,6 +6,8 @@ import pytest
 
 import strawberry
 from strawberry.types.base import StrawberryObjectDefinition, get_object_definition
+from strawberry.types.field import StrawberryField
+from strawberry.types.object_type import field
 
 
 def test_get_object_definition():
@@ -47,3 +49,18 @@ def test_get_object_definition_strict():
         match=r".* does not have a StrawberryObjectDefinition",
     ):
         get_object_definition(OtherFruit, strict=True)
+
+
+def test_public_decorators_have_dataclass_transform():
+    expected = {
+        "eq_default": True,
+        "order_default": True,
+        "kw_only_default": True,
+        "frozen_default": False,
+        "field_specifiers": (field, StrawberryField),
+        "kwargs": {},
+    }
+
+    assert strawberry.type.__dataclass_transform__ == expected
+    assert strawberry.input.__dataclass_transform__ == expected
+    assert strawberry.interface.__dataclass_transform__ == expected
--- a/tests/test_type.py
+++ b/tests/test_type.py
@@ -6,8 +6,6 @@ import pytest
 
 import strawberry
 from strawberry.types.base import StrawberryObjectDefinition, get_object_definition
-from strawberry.types.field import StrawberryField
-from strawberry.types.object_type import field
 
 
 def test_get_object_definition():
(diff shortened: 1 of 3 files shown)


tests/test_type.py
```diff
@@ -47,3 +47,18 @@ class OtherFruit:
...
+    for decorator in decorators:
+        transform = decorator.__dataclass_transform__
+
+        assert transform["order_default"] is True
+        assert transform["kw_only_default"] is True
+
+        field_specifiers = transform["field_specifiers"]
+
+        assert field_specifiers
+        assert all(callable(field_specifier) for field_specifier in field_specifiers)
```
patrick91: was this failing before?

Speedy1991: Good point, I think you're right. This shouldn't have been failing at runtime. I originally added this to verify the typing metadata, but this test doesn't actually validate the type-checking behavior - the real issue only shows up in type checkers/IDEs. I think this test is obsolete. Does it make sense to just remove it?

Speedy1991:
I removed the test @patrick91

patrick91 (commented):
I'm not opposed to this, I would be nice if we could make sure this is actually work for jetbrains' IDEs

also do we need to do it for all type, input, etc? (also for federation)?

Speedy1991:
> I’m not opposed to this, but it would be nice if we could make sure this actually works in JetBrains IDEs.

I tested my branch locally, and the warnings/errors for types are gone.

> Also, do we need to do it for all types, inputs, etc. (including federation)?

This PR also includes a fix for inputs. I'm not sure about federation, though - I haven't worked with it in about two years, so my knowledge there is a bit outdated. If you have an example of a federation setup that should be supported, feel free to share it and I'll see if we need to handle it as well.

€dit: Interfaces are already working as expected

patrick91:
@Speedy1991 we use the same decorators on the federation decorators, if you do `strawberry.federation.type` you should have the same behaviour, see https://github.com/strawberry-graphql/strawberry/blob/main/strawberry/federation/object_type.py#L59-L81

Speedy1991:
@patrick91 Federation type looks fine in pycharm:
image

Looking at the source code, this makes sense because dataclass_transform is duplicated in that location.

Speedy1991:
Maybe it doesn't make sense.

Why is `@dataclass_transform` required on the **actual** `strawberry.type` function, but not on `federation.type`? 🤔

patrick91:
> Maybe it doesn't make sense.
> 
> Why is `@dataclass_transform` required on the **actual** `strawberry.type` function, but not on `federation.type`? 🤔

ahahah no idea! I haven't checked if the federation decorator is already doing what your code does, maybe that's it?

Or maybe it's pycharm

Speedy1991:
I think this is a PyCharm issue. The main question is why it doesn't honor the overloaded `dataclass_transform` in the first place - that should work as expected. But after waiting two years for a fix in PyCharm, I think it's probably better to add those transforms to the actual function definition as well. :/

The release file still needs `social_messages`. @patrick91, could you finalize that? I'm not sure what should go there.

Is there anything else missing before this can be included in a release? I'm happy to make any additional changes needed to get this merged into `master`. 🙂

patrick91 (approved):

botberry:
This PR was published as [0.320.4](https://github.com/strawberry-graphql/strawberry/releases/tag/0.320.4). Thank you for contributing!

PR #4509, opened 2026-07-07
Preserve user-defined methods on experimental pydantic types

### LINKED ISSUE #3601: strawberry.pydantic.experimental.type strips all staticmethods from type definition besides `from_pydantic()` and `to_pydantic()`
When a staticmethod not named `from_pydantic` or `to_pydantic` is added to a type decorated with `strawberry.pydantic.experimental.type` it is not available at runtime. Attempting to call the staticmethod will result in an AttributeError.

## Describe the Bug

given an example type like this:
```
import strawberry


class Foo(BaseModel):
    bar: int
    

@strawberry.experimental.pydantic.type(model=Foo)
class FooGQLType:
    bar: strawberry.auto

    @staticmethod
    def do_something() -> None:
        # impl doesn't matter
        pass



if __name__ == "__main__":
    FooGQLType.do_something()
```

Running this code will result in `AttributeError: type object 'FooGQLType' has no attribute 'do_something'

In the pydantic type wrapper the `cls` object is redefined here using `dataclasses.make_dataclass()`: https://github.com/strawberry-graphql/strawberry/blob/cafc388f8221781507a53f00c02d3913679ae5be/strawberry/experimental/pydantic/object_type.py#L248

The `do_something` method is available as an attribute on `cls` before this redefinition but is unavailable afterwards.

Both `to_pydantic` and `from_pydantic` are handled as special cases. If they exist on the type definition they are saved to the namespace dict before this redefinition and re-added after the fact. if they do not exist, the default definitions are supplied

## System Information

 - Operating system: MacOS Sonoma 14.4.1
 - Strawberry version: 0.237.3
 - Python version: 3.11.2

## Additional Context

I ran into this issue attempting to find a workaround for a question I posted in the discord titled "Conditionally override `from_pydantic()`"

Pablosinyores:
## Description

Closes #3601.

`strawberry.experimental.pydantic.type` rebuilds the decorated class with `dataclasses.make_dataclass(..., bases=cls.__bases__, namespace=namespace)`. Because it builds from the original *bases* rather than from `cls` itself, only the entries explicitly put into `namespace` (`is_type_of`, and custom `from_pydantic` / `to_pydantic` / `resolve_reference`) survive. Any other attribute defined on the class body — a `staticmethod`, `classmethod`, regular method, or `property` — was silently dropped, so accessing it raised `AttributeError`:

```python
@strawberry.experimental.pydantic.type(model=Foo)
class FooType:
    bar: strawberry.auto

    @staticmethod
    def do_something() -> str:
        return "done"


FooType.do_something()  # AttributeError: type object 'FooType' has no attribute 'do_something'
```

(`@strawberry.field` resolvers were unaffected because they are collected as fields; only non-field members were lost.)

## Fix

Before calling `make_dataclass`, carry the decorated class's own non-field methods/properties into `namespace`, skipping dunders, existing fields, and anything already handled (`is_type_of` / custom `from_pydantic` / `to_pydantic` / `resolve_reference`). `.input` and `.interface` delegate to the same wrapper, so they get the fix too.

## Testing

Added `test_preserves_user_defined_methods` covering a staticmethod (the reported case) plus classmethod, instance method, and property. Verified it fails on `main` with the exact `AttributeError` and passes after the change. Full `tests/experimental/pydantic/` suite: 154 passed, 12 skipped, 1 xfailed. `ruff format` / `ruff check` clean.

## Summary by Sourcery

Preserve user-defined methods and properties on experimental Pydantic-based Strawberry types so they remain accessible at runtime.

Bug Fixes:
- Ensure static, class, instance methods, and properties defined on experimental Pydantic type classes are retained instead of being dropped during dataclass reconstruction.

Documentation:
- Add release notes documenting the fix for user-defined methods being stripped from experimental Pydantic types.

Tests:
- Add regression test verifying static, class, instance methods, and properties remain callable on experimental Pydantic types.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,32 @@
+---
+release type: patch
+social_messages:
+  x: >-
+    {project_name} {version} keeps user-defined methods on experimental Pydantic types.
+  linkedin: >-
+    {project_name} {version} fixes experimental Pydantic types dropping user-defined
+    static, class, and instance methods (and properties) from the decorated class body.
+---
+
+Fix `strawberry.experimental.pydantic.type` (and `.input` / `.interface`) stripping
+user-defined methods from the decorated class.
+
+Previously any `staticmethod`, `classmethod`, regular method, or `property` defined on
+the class body — other than `from_pydantic` / `to_pydantic` — was dropped, because the
+type is rebuilt with `make_dataclass` from the original bases rather than from the class
+itself, so calling it raised `AttributeError`:
+
+```python
+@strawberry.experimental.pydantic.type(model=Foo)
+class FooType:
+    bar: strawberry.auto
+
+    @staticmethod
+    def do_something() -> str:
+        return "done"
+
+
+FooType.do_something()  # used to raise AttributeError
+```
+
+These members are now carried over and remain accessible at runtime.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -29,4 +29,4 @@ class FooType:
 FooType.do_something()  # used to raise AttributeError
 ```
 
-These members are now carried over and remain accessible at runtime.
+These members are now carried over and keep working at runtime.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -3,6 +3,7 @@ release type: patch
 social_messages:
   x: >-
     {project_name} {version} keeps user-defined methods on experimental Pydantic types.
+    🍓 https://strawberry.rocks/release/{version}
   linkedin: >-
     {project_name} {version} fixes experimental Pydantic types dropping user-defined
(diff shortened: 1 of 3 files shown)


Pablosinyores:
Good catch on the edge. The dunder skip here is deliberate — copying arbitrary `__…__` attributes into the `make_dataclass` namespace would drag in `__init__`/`__dict__`/`__weakref__` and fight the rebuild, so I limited the carry-over to non-dunder methods/properties, which is the scope of #3601. `__post_init__` specifically is the one supported hook worth preserving; happy to add a narrow exception for it (plus a test) if you'd prefer that in this PR, or leave it as a follow-up.

PR #3965, opened 2025-08-02
First-class Pydantic v2+ support

patrick91:
## Stack [🍰](https://shortcake.patrick.wtf)

- **#3965** (`feature/pydantic-first-class`) <-- this PR
- #4492 (merged) (`2026-06-28-add-generic-exception-handlers`)
- #4510 (merged) (`2026-07-06-unpack-input-mutation-arguments`)

## Summary

Add first-class support for Pydantic v2+ models in Strawberry GraphQL.

This PR introduces a new `strawberry.pydantic` module that allows you to directly decorate Pydantic `BaseModel` classes to create GraphQL types, inputs, and interfaces without requiring separate wrapper classes.

## Basic Usage

```python
import strawberry
from pydantic import BaseModel

@strawberry.pydantic.type
class User(BaseModel):
    name: str
    age: int

@strawberry.pydantic.input
class CreateUserInput(BaseModel):
    name: str
    age: int

@strawberry.pydantic.interface
class Node(BaseModel):
    id: str
```

### Core Features
- ✅ `@strawberry.pydantic.type` - Convert Pydantic models to GraphQL types
- ✅ `@strawberry.pydantic.input` - Convert Pydantic models to GraphQL input types
- ✅ `@strawberry.pydantic.interface` - Convert Pydantic models to GraphQL interfaces
- ✅ Automatic field extraction from Pydantic models
- ✅ Pydantic field descriptions preserved in GraphQL schema
- ✅ Pydantic field aliases used as GraphQL field names
- ✅ Support for `strawberry.Private` to exclude fields from schema
- ✅ Support for `strawberry.field()` with `Annotated` for directives, permissions, deprecation
- ✅ Generic Pydantic model support
- ✅ Nested Pydantic types
- ✅ `strawberry.pydantic.Error` type for validation error handling
- ✅ Computed fields via `include_computed=True`

### Pydantic v2 Features (130 Tests Passing)
- ✅ **Functional Validators** - `BeforeValidator`, `AfterValidator`, `WrapValidator` with `Annotated` types
- ✅ **@model_validator Support** - Cross-field validation with `mode='before'|'after'|'wrap'`
- ✅ **Validation Context** - Strawberry `Info` automatically passed to Pydantic validators
- ✅ **model_config Settings** - `strict=True`, `extra='forbid'`, `from_attributes=True` all respected
- ✅ **Strict Mode Per-Field** - `Field(strict=True)` for individual field validation
[...]

PR #4510, opened 2026-07-07
Unpack input-mutation arguments during argument conversion

patrick91:
## Stack [🍰](https://shortcake.patrick.wtf)

- #3965 (`feature/pydantic-first-class`)
- #4492 (`2026-06-28-add-generic-exception-handlers`)
- **#4510** (`2026-07-06-unpack-input-mutation-arguments`) <-- this PR

## Summary by Sourcery

Ensure input-mutation fields unpack their synthetic input objects into individual resolver arguments during argument conversion so permissions and field extensions see the original arguments, and treat extensions without resolve hooks as transparent to the resolve chain.

Bug Fixes:
- Fix InputMutationExtension so permission classes and other field extensions receive the resolver’s individual arguments instead of the wrapping input object.

Enhancements:
- Introduce FieldExtension.map_arguments to allow extensions to reshape converted resolver arguments before execution.
- Treat extensions that do not implement sync or async resolve methods as transparent, excluding them from the resolve chain while still running their apply and argument-mapping hooks.

Documentation:
- Add a release note describing the InputMutationExtension fix and how permissions now receive unpacked arguments.

Tests:
- Add tests verifying transparent behavior of extensions without resolve methods on sync and async fields.
- Add tests confirming that input-mutation arguments are unpacked before permissions and other field extensions run.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,15 +1,19 @@
 ---
 release type: patch
+social_messages:
+  x: >-
+    {project_name} {version} fixes `InputMutationExtension` — permission classes and other extensions on input-mutation fields now receive individual arguments instead of the wrapping input object.
+  linkedin: >-
+    {project_name} {version} fixes `InputMutationExtension`: permission classes and field extensions on input-mutation fields now receive the resolver's individual arguments (e.g. `name`, `color`) instead of the wrapping `input` object.
 ---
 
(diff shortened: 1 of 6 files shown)


botberry:
This PR was published as [0.320.3](https://github.com/strawberry-graphql/strawberry/releases/tag/0.320.3). Thank you for contributing!

PR #4526, opened 2026-07-16
fix(printer): convert list-valued custom-scalar defaults instead of crashing

chuenchen309:
## Description

`ast_from_leaf_type` (in `strawberry/printer/ast_from_value.py`) converts serialized custom-scalar values — the `JSON` scalar feature it exists for — into AST nodes. It has branches for `bool`, `int`, `float`, `str`, and `dict` (recursing into dict values), but **no branch for `list`/`tuple`**, so any list value falls through to the `raise TypeError(...)` that is even marked `# pragma: no cover` (assumed unreachable).

It is reachable: a `JSON`-typed input field/argument whose default is — or contains — a list. Printing such a schema raises instead of emitting SDL.

### Reproduction

```python
import strawberry

@strawberry.input
class MyInput:
    j: JSON = strawberry.field(default_factory=lambda: [1, 2, 3])

@strawberry.type
class Query:
    @strawberry.field
    def search(self, input: MyInput) -> JSON:
        return {}

print(strawberry.Schema(query=Query))
# TypeError: Cannot convert value to AST: [1, 2, 3].
```

A list nested inside a dict default (e.g. `{"tags": ["a", "b"]}`) breaks the same way, while plain dict defaults already print fine.

## Fix

Add a `list`/`tuple` branch that recurses via `ast_from_leaf_type` into a `ListValueNode` (already imported), mirroring the existing `dict` branch and matching graphql-core’s own `ast_from_value`, which handles iterables.

## Test

Added `test_input_defaults_scalars_with_lists` (top-level list default + list nested in a dict default + list of dicts). It asserts on list-literal substrings rather than an exact SDL match, since `ObjectValueNode` spacing varies across graphql-core versions (the reason the sibling `test_input_defaults_scalars` is `skip_if_gql_32`). Full `tests/test_printer/` passes (32), and `ruff format/check` is clean on the touched files.

## Summary by Sourcery

Fix schema printing for custom-scalar input defaults that are lists or contain lists by converting these values to AST list nodes and document the patch release.

Bug Fixes:
- Handle list- and tuple-valued custom scalar (e.g. JSON) input defaults when converting values to AST, avoiding printer crashes on schema export.

Documentation:
- Add RELEASE notes describing the fix for crashes when printing schemas with list-valued custom-scalar defaults.

Tests:
- Add regression test covering list-valued JSON input defaults, including nested lists and lists of dicts, to ensure schema printing remains stable.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,21 @@
+---
+release type: patch
+social_messages:
+  x: >-
+    🍓 Strawberry {version} is out! This release fixes a crash when printing a
+    schema whose custom-scalar (e.g. JSON) input default is a list.
+    https://strawberry.rocks/release/{version}
+  linkedin: >-
+    Strawberry {version} is out. This release fixes a crash when printing a
+    schema whose custom-scalar (e.g. JSON) input default is or contains a list,
+    so SDL export and snapshot tests keep working for list-valued defaults.
+---
+
+This release fixes a crash when printing a schema whose custom-scalar (e.g.
+`JSON`) input field has a list-valued default.
+
+`ast_from_leaf_type` handled `dict`/`str`/number defaults but had no branch for
+lists, so a field like
+`j: JSON = strawberry.field(default_factory=lambda: [1, 2, 3])` — or a list
+nested inside a dict default — raised `TypeError: Cannot convert value to AST`
+when the schema was printed. Lists (and tuples) are now converted correctly.
--- a/strawberry/printer/ast_from_value.py
+++ b/strawberry/printer/ast_from_value.py
@@ -82,6 +82,11 @@ def ast_from_leaf_type(serialized: object, type_: GraphQLInputType | None) -> Va
             )
         )
 
+    if isinstance(serialized, (list, tuple)):
+        return ListValueNode(
+            values=tuple(ast_from_leaf_type(item, None) for item in serialized)
+        )
+
     raise TypeError(
         f"Cannot convert value to AST: {inspect(serialized)}."
     )  # pragma: no cover
--- a/tests/test_printer/test_basic.py
+++ b/tests/test_printer/test_basic.py
@@ -265,6 +265,33 @@ def test_input_defaults_scalars():
     assert print_schema(schema) == textwrap.dedent(expected_type).strip()
 
 
+def test_input_defaults_scalars_with_lists():
+    # Regression: a list-valued custom-scalar (JSON) default — at the top level or
+    # nested inside a dict — used to raise TypeError when printing the schema, because
+    # ast_from_leaf_type had branches for dict/str/int/... but none for list/tuple.
+    # (List *literal* formatting is stable across graphql-core versions, unlike the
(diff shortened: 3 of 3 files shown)


Vansh-Sharma27 (commented):
I've been working around the JSON scalar recently (#4440, #4529), so I checked this one out at `6feb3a40` and ran it.

### The regression test is sound

I reverted only `strawberry/printer/ast_from_value.py` back to `main` and kept your test file:

```
E       TypeError: Cannot convert value to AST: [1, 2, 3].
strawberry\printer\ast_from_value.py:85: TypeError
```

So it genuinely fails without the fix rather than just exercising the new lines.

### Your cross-version note checks out

The comment explains that list formatting is stable, unlike the `ObjectValueNode` spacing that gates `test_input_defaults_scalars` behind `skip_if_gql_32`. I ran `tests/test_printer/` at both ends of the supported range (`graphql-core>=3.2.0,<3.4.0`):

- graphql-core 3.2.11 — 15 passed, 1 skipped
- graphql-core 3.3.0rc0 — 15 passed, 1 skipped

Leaving the new test ungated looks correct.

### One gap: `None` inside the default still crashes the same way

`ast_from_leaf_type` has no branch for `None`, so a JSON default containing `null` falls through to the same `raise TypeError` this PR is fixing:

```
list of ints                     OK   j: JSON! = [1, 2, 3]
list containing null             FAIL TypeError: Cannot convert value to AST: None.
dict with null value             FAIL TypeError: Cannot convert value to AST: None.
nested list with null            FAIL TypeError: Cannot convert value to AST: None.
empty list                       OK   j: JSON! = []
```

The dict-with-null case predates this PR, so it isn't something you introduced. But `null` is part of JSON, and you're already extending exactly this function, so it seems worth covering here instead of waiting for the follow-up issue. `NullValueNode` is already imported at line 16, so it's a one-liner near the top:

```python
if serialized is None:
    return NullValueNode()
```

I patched that in locally to confirm it does the right thing:

```
list containing null     OK   j: JSON! = [1, null, 3]
dict with null value     OK   j: JSON! = {a: null}
nested list w/ null      OK   j: JSON! = {tags: ["a", null]}
```

### Smaller things

The `# pragma: no cover` on that final `raise` is what kept this class of bug invisible to coverage — the branch is plainly reachable, which is why this PR exists. If the `None` case gets handled too, it might be worth dropping the pragma so the next unhandled type shows up as a coverage gap instead of a user traceback.

Also minor: `ast_from_value` further down uses `is_iterable(value)` (already imported from `graphql.pyutils`) for its list branch, while this uses `isinstance(serialized, (list, tuple))`. Same behaviour for the cases that matter here — only noting it in case you'd prefer consistency within the file.

### What I ran

- `pytest tests/test_printer/ tests/schema/test_scalars.py` — 53 passed, 12 skipped (graphql-core 3.2.11)
- `pytest tests/test_printer/` on graphql-core 3.3.0rc0 — 15 passed, 1 skipped
- `ruff check strawberry/printer/ast_from_value.py` — clean

The fix itself is right and mirrors the existing `dict` branch cleanly.

PR #4528, opened 2026-07-18
Report non-string scalar variables as invalid input

### LINKED ISSUE #4524: Non-string variables for built-in UUID/Date/DateTime/Time scalars raise unhandled AttributeError/TypeError instead of a clean coercion error
## Describe the Bug

`wrap_parser` in `strawberry/schema/types/base_scalars.py` only converts `ValueError` into a clean `GraphQLError`. When a client sends a **non-string** variable into a `UUID`, `Date`, `DateTime`, or `Time` position, the underlying parser raises `AttributeError` or `TypeError` instead, which escapes `wrap_parser`. graphql-core then wraps it with the original exception and full stack trace attached, so what is plain invalid client input gets reported by error trackers (Sentry, Datadog Error Tracking, ...) as a server-side crash fingerprinted on stdlib frames like `uuid.py __init__`.

Minimal repro (strawberry 0.320.2):

```python
import uuid
import strawberry

@strawberry.type
class Query:
    @strawberry.field
    def user(self, id: uuid.UUID) -> str:
        return str(id)

result = schema.execute_sync(
    "query ($id: UUID!) { user(id: $id) }",
    variable_values={"id": 469610.0},
)
print(result.errors[0])
```

Output (note the chained `AttributeError` traceback from inside `uuid.UUID`):

```
Variable '$id' got invalid value 469610.0; Expected type 'UUID'. 'float' object has no attribute 'replace'

  File ".../graphql/utilities/coerce_input_value.py", line 137, in coerce_input_value
    parse_result = type_.parse_value(input_value)
  File ".../strawberry/schema/types/base_scalars.py", line 16, in inner
    return parser(value)
  File ".../python3.12/uuid.py", line 175, in __init__
    hex = hex.replace('urn:', '').replace('uuid:', '')
graphql.error.graphql_error.GraphQLError: Expected type 'UUID'. 'float' object has no attribute 'replace'
```

The same happens for `Date`/`Time` (`fromisoformat: argument must be str`, a `TypeError`) and `DateTime`.

For comparison, a malformed *string* (`{"id": "not-a-uuid"}`) takes the `ValueError` path and produces a clean, traceback-free coercion error. Non-string input should behave the same: it is a client error, not a server crash.

This is easy to hit in practice: any REST-ish client sending a numeric id, and especially LLM/MCP clients, which occasionally emit unquoted ids that JSON parsing turns into ints/floats.

## Suggested fix

`Decimal` already handles this correctly since #1507: `parse_decimal` routes the value through `str(value)` first, so every invalid input ends up on the clean exception path. Applying the same one-liner to the remaining built-in scalars fixes it, e.g.:

```python
def wrap_parser(parser: Callable, type_: str) -> Callable:
    def inner(value: object) -> object:
        try:
            return parser(str(value))
        except ValueError as e:
            raise GraphQLError(f'Value cannot represent a {type_}: "{value}". {e}') from None
    return inner
```

(`uuid.UUID`, `datetime.date.fromisoformat`, `datetime.time.fromisoformat`, and `dateutil.parser.isoparse` all accept strings, so serialization output is unchanged and the exported SDL stays identical.)

Happy to send a PR if this approach is acceptable.

## Workaround

Override the built-in scalar per schema with a parser that stringifies first:

```python
import uuid
import strawberry
from graphql import GraphQLError

def _parse_uuid(value: object) -> uuid.UUID:
    try:
        return uuid.UUID(str(value))
    except ValueError as e:
        raise GraphQLError(f'Value cannot represent a UUID: "{value}". {e}') from None

UUIDScalar = strawberry.scalar(uuid.UUID, name="UUID", serialize=str, parse_value=_parse_uuid)

schema = strawberry.Schema(query=Query, scalar_overrides={uuid.UUID: UUIDScalar})
```

The exported SDL is byte-identical, so this is a drop-in mitigation until fixed upstream.

## System Information

 - Operating system: macOS (also reproduced on Linux in production)
 - Python version: 3.12
 - Strawberry version (if applicable): 0.320.2 (bug still present on current `main`)

## Additional Context

Related prior art: #1507 fixed the `original_error` leakage for `ValueError` on `Decimal`/`UUID` and introduced the `str(value)` pattern for `Decimal`, but the non-string case for `UUID`/`Date`/`DateTime`/`Time` was left out.

dchaudhari7177:
## Summary

Fixes #4524. `wrap_parser` only converts `ValueError` into a `GraphQLError`, but the `UUID`, `Date`, `DateTime` and `Time` parsers accept a string and raise `AttributeError`/`TypeError` on anything else. Those escaped the wrapper, so plain invalid client input surfaced as the standard library's own message.

@patrick91 — you asked for "the same thing as decimal" and wondered if it could be generic rather than lots of small similar functions. I think `parse_decimal` already had the answer: it does `decimal.Decimal(str(value))`, and that leading `str()` is precisely why `Decimal` never had this bug.

So the fix is one line in `wrap_parser` — coerce before parsing — which covers all four scalars with no per-type functions:

```python
return parser(str(value))
```

I verified that `str()` normalises every non-string case into a `ValueError` the existing handler already reports cleanly:

| parser | direct | after `str()` |
| --- | --- | --- |
| `uuid.UUID` | AttributeError / TypeError | **ValueError** |
| `date.fromisoformat` | TypeError | **ValueError** |
| `time.fromisoformat` | TypeError | **ValueError** |
| `dateutil.parser.isoparse` | TypeError | **ValueError** |

(tested against `469610.0`, `123`, `True`, `None`, `['a']`)

## Before / after

Running the issue's exact repro:

```
before: Variable '$id' got invalid value 469610.0; Expected type 'UUID'. 'float' object has no attribute 'replace'
after:  Variable '$id' got invalid value 469610.0; Value cannot represent a UUID: "469610.0". badly formed hexadecimal UUID string
```

Being precise about the traceback claim, since it's the point of the issue: `original_error` is a `GraphQLError` in both cases (graphql-core wraps coercion failures either way). What changes is that the error is now **raised deliberately by strawberry** with the parser's own message, instead of an unhandled `AttributeError` propagating out — so the client-facing text no longer contains the stdlib's complaint:

```
before  deepest frames: coerce_input_value.py → base_scalars.py:inner → uuid.py:__init__
        chain: GraphQLError                        (AttributeError escaped, never wrapped)
after   chain: GraphQLError <- ValueError          (raised by wrap_parser)
```

## Compatibility

- A value that is already a `str` is completely unaffected — every existing error message is byte-identical, and the existing per-scalar tests pass untouched.
- A value that *stringifies* to something valid now parses instead of raising: passing a real `uuid.UUID` or `date` object programmatically via `variable_values` used to be an `AttributeError`/`TypeError`. That's a strict improvement, but flagging it as the one behavioural change.
- I left `parse_decimal` alone. It could now be folded into `wrap_parser`, but its message deliberately omits the `{e}` suffix that the others include, so unifying them would change `Decimal`'s public error text. Happy to do that consolidation as a follow-up if you want the two to share one code path as well as one approach.

## Tests

Added `tests/schema/types/test_base_scalars.py` — one parametrized test over all four scalars × five non-string inputs, matching the generic shape of the fix rather than adding a near-duplicate to each per-type file. It asserts a clean coercion message, that `has no attribute` never appears in a client-facing message, and that `original_error` is not an `AttributeError`/`TypeError`. A second test pins that valid strings still parse.

Reverting only `strawberry/` fails **18 of the 20** non-string cases (the 2 that pass are `dateutil.isoparse` cases that already raised `ValueError`).

## Verification

- `pytest tests/schema/types/` → **65 passed**
- `pytest tests/schema/` → **696 passed, 12 skipped, 2 xfailed**
- `ruff check` / `ruff format --check` on both files → clean
- Added `RELEASE.md` (patch)

The 2 failures in `tests/schema/test_pydantic.py` and the collection error in `test_opentelemetry.py` are missing optional dependencies in my environment (`pydantic`, `opentelemetry`), confirmed on a clean tree by stashing.

## Summary by Sourcery

Ensure non-string variables for built-in UUID, Date, DateTime, and Time scalars are treated as invalid input with consistent coercion errors instead of surfacing as server errors.

Bug Fixes:
- Normalize non-string values before scalar parsing so UUID, Date, DateTime, and Time coercion failures always surface as clean GraphQL input errors rather than leaking stdlib AttributeError/TypeError details.

Enhancements:
- Add shared tests covering scalar behavior for non-string inputs and valid string inputs across UUID, Date, DateTime, and Time.
- Document the change in a new patch-level RELEASE.md entry highlighting the improved error reporting for these scalars.

Documentation:
- Add a release note describing the fix to error handling for non-string variables passed to the built-in date/time and UUID scalars.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -20,6 +20,8 @@ standard library's own message, such as `'float' object has no attribute
 'replace'`. Error trackers then recorded plain invalid input as a crash
 fingerprinted on stdlib frames.
 
-Values are now coerced with `str()` before parsing, the same way `Decimal`
-already did, so an unparseable value raises a `ValueError` that the wrapper
-reports as a normal coercion error.
+A non-string value is now reported as an ordinary coercion error:
+
+```
+Value cannot represent a UUID: "469610.0".
+```
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,25 @@
+---
+release type: patch
+social_messages:
+  x: >-
+    {project_name} {version} is out! Non-string variables for UUID, Date,
+    DateTime and Time now return a clean coercion error instead of a server
+    error. 🍓 https://strawberry.rocks/release/{version}
+  linkedin: >-
+    {project_name} {version} is out. Non-string variables for the built-in
+    UUID, Date, DateTime and Time scalars now return a clean coercion error
+    rather than surfacing as a server error in your error tracker.
+---
+
+This release fixes non-string variables for the built-in `UUID`, `Date`,
+`DateTime` and `Time` scalars being reported as server errors.
+
+These parsers accept only strings and raise `AttributeError` or `TypeError` on
+anything else, which escaped the scalar wrapper and reached the client as the
+standard library's own message, such as `'float' object has no attribute
+'replace'`. Error trackers then recorded plain invalid input as a crash
+fingerprinted on stdlib frames.
+
+Values are now coerced with `str()` before parsing, the same way `Decimal`
+already did, so an unparseable value raises a `ValueError` that the wrapper
+reports as a normal coercion error.
--- a/strawberry/schema/types/base_scalars.py
+++ b/strawberry/schema/types/base_scalars.py
@@ -14,11 +14,18 @@ def wrap_parser(parser: Callable, type_: str) -> Callable:
     def inner(value: object) -> object:
         # These parsers only accept a string, and raise AttributeError or
         # TypeError on anything else. That escapes the except below and reaches
-        # the client as a server error complete with a stdlib traceback, even
-        # though it is just invalid input. Coercing first, the way parse_decimal
-        # does, keeps every rejection a ValueError we can report cleanly.
+        # the client as a server error carrying a stdlib traceback, even though
+        # it is only invalid input. Reject it here instead.
+        #
+        # Deliberately not coerced with str(): these scalars are serialised as
+        # strings, and stringifying would start accepting values the grammar was
+        # never meant to take, such as the number 20230517 as a Date or 1200 as
+        # a Time.
+        if not isinstance(value, str):
+            raise GraphQLError(f'Value cannot represent a {type_}: "{value}".')
+
         try:
-            return parser(str(value))
+            return parser(value)
         except ValueError as e:
             raise GraphQLError(  # noqa: B904
                 f'Value cannot represent a {type_}: "{value}". {e}'
--- a/strawberry/schema/types/base_scalars.py
+++ b/strawberry/schema/types/base_scalars.py
@@ -11,9 +11,14 @@ from strawberry.types.scalar import ScalarDefinition
 
 
 def wrap_parser(parser: Callable, type_: str) -> Callable:
-    def inner(value: str) -> object:
+    def inner(value: object) -> object:
+        # These parsers only accept a string, and raise AttributeError or
+        # TypeError on anything else. That escapes the except below and reaches
+        # the client as a server error complete with a stdlib traceback, even
+        # though it is just invalid input. Coercing first, the way parse_decimal
+        # does, keeps every rejection a ValueError we can report cleanly.
         try:
-            return parser(value)
+            return parser(str(value))
         except ValueError as e:
             raise GraphQLError(  # noqa: B904
                 f'Value cannot represent a {type_}: "{value}". {e}'
--- a/tests/schema/types/test_base_scalars.py
+++ b/tests/schema/types/test_base_scalars.py
@@ -64,6 +64,52 @@ def test_non_string_variable_is_reported_as_invalid_input(field, type_name, valu
 
 
 @pytest.mark.parametrize(
+    ("field", "type_name", "value"),
+    [
+        # Each of these stringifies to something the parser would happily
+        # accept, so coercing rather than rejecting would quietly widen the
+        # scalar: 20230517 as a Date, 1200 and even 12 as a Time.
+        ("dateField", "Date", 20230517),
+        ("dateField", "Date", 20230517.0),
+        ("timeField", "Time", 1200),
+        ("timeField", "Time", 12),
+    ],
+)
+def test_number_is_not_accepted_as_a_temporal_scalar(field, type_name, value):
+    result = schema.execute_sync(
+        f"query ($value: {type_name}!) {{ {field}(value: $value) }}",
+        variable_values={"value": value},
+    )
+
+    assert result.errors
+    assert (
+        f'Value cannot represent a {type_name}: "{value}".' in result.errors[0].message
+    )
+
+
+@pytest.mark.parametrize(
+    ("field", "type_name"),
+    [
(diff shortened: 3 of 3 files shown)


dchaudhari7177:
Thanks both — the greptile point is a real flaw in my first approach, so I've reworked it.

**On numeric inputs becoming valid scalars (P1):** this was worse than the comment suggests. Checking each case, `str()` coercion would have accepted all of these:

```
date.fromisoformat('20230517')   -> 2023-05-17   ACCEPTED
date.fromisoformat('20230517.0') -> 2023-05-17   ACCEPTED
time.fromisoformat('1200')       -> 12:00:00     ACCEPTED
time.fromisoformat('12')         -> 12:00:00     ACCEPTED
```

So a client sending the JSON number `12` for a `Time` would have silently got midday. Fixing a crash by widening what the scalar accepts is the wrong trade, and my "a value that stringifies to something valid now parses" note in the description was underselling it.

`wrap_parser` now rejects a non-string outright rather than coercing:

```python
if not isinstance(value, str):
    raise GraphQLError(f'Value cannot represent a {type_}: "{value}".')
```

That still keeps the crash away from the client — `AttributeError`/`TypeError` no longer escape the wrapper — without letting any new value through. These scalars are serialised as strings, so a non-string was never valid input.

@patrick91 this does step away from the literal "same thing as decimal", since `parse_decimal` coerces with `Decimal(str(value))`. That reads as deliberate for a numeric scalar, where accepting `1.5` is natural, but the same move on `Date`/`Time`/`UUID` is what opened the hole above. The fix is still generic — one guard in `wrap_parser`, no per-type functions. Happy to switch to coercion if you'd rather match `Decimal` exactly.

**On the `None` case:** good catch that the description mentioned it while the params didn't. `None` can't reach the parser though — `graphql-core` rejects it against the non-null wrapper first — so asserting `Value cannot represent ...` for it would fail. I've added `test_null_is_rejected_by_the_non_null_wrapper` to pin that behaviour explicitly instead, asserting the `must not be null` message and that no `AttributeError`/`TypeError` leaks.

Also added `test_number_is_not_accepted_as_a_temporal_scalar` covering all four numeric cases above. I verified it catches the regression by restoring the coercion version — all four fail, and pass again with the guard.

The original issue is still fixed:

```
message       : Variable '$id' got invalid value 469610.0; Value cannot represent a UUID: "469610.0".
original_error: GraphQLError
leaks stdlib? : False
```

`pytest tests/schema/types/` → 73 passed. `pytest tests/schema/` → 704 passed, 12 skipped, 2 xfailed; the 2 failures are pre-existing `test_pydantic.py` cases from `pydantic` not being installed in my env. `ruff check` and `ruff format --check` are clean.

dchaudhari7177:
Closing as superseded — this was fixed in main by #4525 (`ac604687`, "coerce non-string input for built-in scalars into clean errors"), which also added per-scalar tests (test_uuid/date/time/datetime/decimal) covering non-string rejection. After rebasing, my implementation change is identical to main and the test coverage overlaps #4525's, so there's nothing left worth landing here. Thanks!

PR #4330, opened 2026-03-26
Feat #4329: Add `on_stream_result` hook to SchemaExtension

Ladol:
## Description

Previously, `SchemaExtension` hooks only wrapped the initial setup phase of a GraphQL subscription, leaving extensions completely disconnected from the actual stream of yielded events.

This commit introduces the `on_subscription_result` hook to the base `SchemaExtension` class and triggers it inside the `schema._subscribe` generator. This allows extensions to safely mutate streamed data before it reaches the transport layer.

Additionally, the `MaskErrors` extension has been updated to use this new hook, fixing an issue where sensitive errors were leaking unmasked over WebSocket connections.

## Types of Changes

- [ ] Core
- [x] Bugfix
- [x] New feature
- [x] Enhancement/optimization
- [ ] Documentation

## Issues Fixed or Closed by This PR

#3680
#4329

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the CONTRIBUTING document.
- [x] I have added tests to cover my changes.
- [x] I have tested the changes and verified that they work and don't break anything (as well as I can manage).

## Summary by Sourcery

Add a subscription-result lifecycle hook to schema extensions and ensure subscription errors are properly masked.

New Features:
- Introduce an `on_subscription_result` hook on `SchemaExtension` that runs for every yielded subscription event, allowing extensions to inspect or mutate each result.

Bug Fixes:
- Ensure the `MaskErrors` extension masks exceptions emitted from subscription streams so sensitive error details are not leaked over WebSocket connections.

Enhancements:
- Wire the new subscription-result hook through the extension runner and subscription execution path so all active extensions can participate in streaming results.

Documentation:
- Document the new subscription-result hook and behavior change in the release notes.

Tests:
- Add subscription-focused tests covering custom extensions mutating streamed data and `MaskErrors` correctly masking subscription errors.

PR #3968, opened 2025-08-05
Fix MaskErrors for synchronous pre-execution errors

dextermb:
## Description

This fixes `MaskErrors` so parsing and validation errors are also masked during synchronous execution, including when `ValidationCache` supplies the validation result. 🍓

The sync path previously returned its pre-execution result from inside the operation lifecycle. After-yield extension hooks could update `execution_context.pre_execution_errors`, but the already-created return object retained the original errors.

This update:

- lets `MaskErrors` process `pre_execution_errors` when no execution result exists
- finalises the synchronous pre-execution result after operation hooks complete
- preserves the existing execution and streamed-result masking paths
- adds regression coverage for syntax and cached validation failures

Fixes #3844.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation-only change

## Validation

- `uv run pytest tests/schema/extensions/test_mask_errors.py -q` — 9 passed
- `uv run pytest tests/schema/extensions/ -q` — 174 passed, 10 skipped
- `uv run mypy --config-file mypy.ini` — 240 source files clean
- `uv run ruff check .` — passed
- `uv run ruff format --check .` — 647 files formatted

## Summary by Sourcery

Fix masking of pre-execution errors during synchronous execution so error details are consistently hidden, and add regression coverage and release notes for this behavior.

Bug Fixes:
- Ensure MaskErrors also masks parsing and validation errors for synchronous execution, including when ValidationCache is used.

Enhancements:
- Refine MaskErrors error processing to operate on a typed result-with-errors protocol shared across sync, async, and streamed executions.

Documentation:
- Add release notes describing the fix for synchronous masking of parsing and validation errors.

Tests:
- Add sync and async tests verifying that pre-execution parsing and validation errors are masked when using ValidationCache with MaskErrors.

PR #4577, opened 2026-08-10
feat: add `mask_pre_execution_errors` option to `MaskErrors`

### LINKED ISSUE #3326: `MaskErrors` does not mask validation errors
The `MaskErrors` extension does not seem to mask validation errors. This is because the `.on_operation()` handler only looks at `self.execution_context.result.errors` while validation errors (such as misspelled field names, which is how I stumbled upon this) are in `self.execution_context.errors`.

I have subclassed `MaskErrors` in my project and added a `.on_validate()` handler that does the following:

```python
def on_validate(self) -> Iterator[None]:
    yield
    if self.execution_context.errors:
        processed_errors: List[GraphQLError] = []
        for error in self.execution_context.errors:
            if self.should_mask_error(error):
                processed_errors.append(self.anonymise_error(error))
            else:
                processed_errors.append(error)

        self.execution_context.errors = processed_errors
```

This will correctly mask such validation errors as well.

Since I'm not very familiar with the internals of this project it was not immediately obvious how to write a test for this, since the existing tests in `test_mask_errors.py` all look at execution results. With some pointers on that, I could turn this ticket into a PR.

alimony:
## Description

The original report – that `MaskErrors` does not mask validation errors – was already fixed by #3968 and #4330. `MaskErrors` masks parse errors and validation errors in every execution path today.

What is still open is the option that @erikwrede and @patrick91 asked for in that thread. There are two separate concerns: hide what the resolvers raise, and hide the shape of the schema. You cannot ask for the first one alone, so a client that sends a malformed document only gets `Unexpected error.` and cannot correct its query.

This PR adds `mask_pre_execution_errors` to the `MaskErrors` constructor. It defaults to `True`, so the behaviour does not change for anybody.

```python
schema = strawberry.Schema(
    Query,
    extensions=[lambda: MaskErrors(mask_pre_execution_errors=False)],
)

# "Cannot query field 'helloo' on type 'Query'. Did you mean 'hello'?"
schema.execute_sync("{ helloo }")

# "Unexpected error."
schema.execute_sync("{ hiddenError }")
```

### How it works

The extension masks the errors of `execution_context.result` and of each streamed frame. At that point it cannot tell which phase produced them. It now tracks the phase with the existing lifecycle hooks: `on_parse` sets a private flag and `on_execute` clears it. Every execution path enters `parsing()` before `executing()`, so an operation that reaches `on_execute` did not fail to parse or to validate. `on_operation` and `on_stream_result` both consult the flag, which keeps the sync, async, streaming, subscription and incremental paths consistent.

Errors that occur before the parse step, a missing query for example, stay masked. `Schema.process_errors` still receives the original errors, so logs do not change.

### Validation

- `uv run pytest` – 5122 passed. The single failure, `tests/typecheckers/test_pydantic.py::test_pydantic_type`, also fails on a clean checkout here, because a local binary is missing.
- `uv run mypy --config-file mypy.ini` – clean, 240 source files.
- `uv run ruff check .` and `uv run ruff format --check .` – clean.
- The 8 new test cases fail before the change and pass after it.
- I then broke the new logic seven ways – dropped each hook, dropped each guard, and forced the phase check to a constant – and confirmed that the tests catch every one.
- Manual checks of the sync, async and streaming paths, and of the example in the documentation.
- `RELEASE.md` included, release type `minor`.

## Types of Changes

- [ ] Core
- [ ] Bugfix
- [x] New feature
- [x] Documentation

* Closes #3326

## Checklist

## Summary by Sourcery

Add a configurable option to the MaskErrors extension to control whether syntax and validation errors are masked before execution while keeping resolver-raised errors masked by default.

New Features:
- Introduce the mask_pre_execution_errors option on the MaskErrors extension to let clients see pre-execution syntax and validation errors when desired.

Enhancements:
- Track execution phase within MaskErrors so masking behavior can differ between pre-execution and execution errors without affecting logging.
- Extend MaskErrors to handle phase-aware masking for streaming operations and multiple operations sharing the same extension instance.

Documentation:
- Document the new mask_pre_execution_errors option in the MaskErrors extension guide, including examples and guidance on schema visibility implications.

Tests:
- Add sync, async, and streaming tests verifying mask_pre_execution_errors behavior for parse, validation, and resolver errors, and phase handling across multiple operations.

Chores:
- Add a minor release note describing the new mask_pre_execution_errors option and its default behavior.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,49 @@
+---
+release type: minor
+social_messages:
+  x: >-
+    {project_name} {version} is out! This release adds a
+    `mask_pre_execution_errors` option to `MaskErrors`, so clients can still see
+    why their query is invalid. 🍓 https://strawberry.rocks/release/{version}
+  linkedin: >-
+    {project_name} {version} is out. This release adds a
+    `mask_pre_execution_errors` option to the `MaskErrors` extension, so syntax
+    and validation errors can reach clients while the errors your resolvers
+    raise stay masked.
+---
+
+This release adds a `mask_pre_execution_errors` option to the `MaskErrors` extension.
+
+`MaskErrors` masks every error. This includes the syntax errors of a document and the validation errors against the schema, so a client that sends an invalid query gets only the generic message. Set the new option to `False` to send these errors to the client. The errors that your resolvers raise stay masked:
+
+```python
+import strawberry
+from strawberry.extensions import MaskErrors
+
+
+@strawberry.type
+class Query:
+    @strawberry.field
+    def hello(self) -> str:
+        return "world"
+
+    @strawberry.field
+    def hidden_error(self) -> str:
+        raise KeyError("This error will not be visible")
+
+
+schema = strawberry.Schema(
+    Query,
+    extensions=[
+        MaskErrors(mask_pre_execution_errors=False),
+    ],
+)
+
+# "Cannot query field 'helloo' on type 'Query'. Did you mean 'hello'?"
+schema.execute_sync("{ helloo }")
+
+# "Unexpected error."
+schema.execute_sync("{ hiddenError }")
+```
+
+The default value is `True`, which keeps the current behaviour. Validation errors give the names of the fields, the arguments and the types of your schema. Thus disable the option only if the clients can know the shape of the schema.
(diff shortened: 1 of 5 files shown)


alimony:
I will resolve the comments from Sourcery/Greptile.

alimony:
@patrick91 Still looking for a review here.

PR #4591, opened 2026-08-27
fix: correct dataclass transform ordering metadata

### LINKED ISSUE #3790: Incorrect typing for the `type` decorator
## Describe the Bug

The type function is decorated with 
```
@dataclass_transform(
    order_default=True, kw_only_default=True, field_specifiers=(field, StrawberryField)
)
```

Therefore mypy treats classes decorated with type as being dataclasses with ordering functions.

In particular, defining `__gt__` on such a class will be treated as an error by mypy.

However, type (that is the underlying `_wrap_dataclass` function) do not do anything to ensure the dataclass actually has order functions defined.

I see multiple solutions:
- Removing the `order_default=True` part of the `dataclass_transform` decorating `type`
- Enforcing the `order=True` in `_wrap_dataclass`
- Allowing the caller to pass dataclass kwargs (as per [my previous issue](https://github.com/strawberry-graphql/strawberry/issues/2688))

## System Information

 - Operating system: Ubuntu 24.04
 - Strawberry version (if applicable): 0.256.1

## Additional Context

Code samples to be clear on the issue
```
@strawberry.type
class MyClass:
    attr: str
 
k = MyClass(attr="abc")
j = MyClass(attr="def")
j > k # TypeError: '<' not supported between instances of 'MyClass' and 'MyClass'
```

``` 
@strawberry.type
class MyClass:
    attr: str

    def __gt__(self, other):
        return self.attr > other.attr
 
k = MyClass(attr="abc")
j = MyClass(attr="def")
j > k # True

# When running mypy
error: You may not have a custom "__gt__" method when "order" is True  [misc]
```

subham-hq:
## Summary

Align Strawberry's `dataclass_transform` ordering metadata with its actual runtime dataclass behavior.

Strawberry ultimately wraps these decorated classes with `dataclasses.dataclass(kw_only=True)`, which uses Python's default `order=False`. The corresponding `dataclass_transform` declarations, however, advertised `order_default=True`.

That mismatch causes type checkers such as mypy and ty to model Strawberry-decorated classes as if ordering methods were generated by default, incorrectly rejecting valid user-defined comparison methods such as `__gt__`.

## Root cause

At runtime, Strawberry's dataclass wrapping is equivalent to:

```python
@dataclasses.dataclass(kw_only=True)
class Example:
    ...
```

Because `dataclasses.dataclass` defaults to `order=False`, defining a custom ordering method is valid.

The static typing metadata instead declared:

```python
@dataclass_transform(
    order_default=True,
    kw_only_default=True,
    ...,
)
```

Under PEP 681, this tells type checkers to assume `order=True` when interpreting classes produced by the decorator. That does not match Strawberry's runtime behavior.

For example:

```python
@strawberry.type
class User:
    name: str

    def __gt__(self, other: "User") -> bool:
        return self.name > other.name
```

Before this change, mypy rejects the custom `__gt__` because it assumes ordering is enabled, and ty reports the equivalent conflict. The class is valid at runtime because Strawberry does not enable dataclass ordering.

## Changes

- Set `order_default=False` in Strawberry's dataclass-transform metadata so static typing matches runtime behavior.
- Apply the correction consistently to related decorators that use the same dataclass semantics:
  - core `type`, `input`, and `interface`
  - federation `type`, `input`, `interface`, and `interface_object`
  - schema directives
  - federation schema directives
- Add regression coverage for custom ordering methods on:
  - `@strawberry.type`
  - `@strawberry.federation.type`
  - `@schema_directive`
  - federation `@schema_directive`
- Add the required `RELEASE.md` entry.

This is a typing-metadata correction only; it does not change Strawberry's runtime dataclass behavior.

## Validation

The issue was reproduced before the fix with a custom `__gt__` method:

- mypy reported that a custom `__gt__` is not allowed when `order` is `True`
- ty reported the equivalent ordering conflict
- pyright did not report an ordering error

After correcting the metadata, the same examples type-check without ordering errors.

Focused type-checker tests:

```bash
uv run pytest \
  tests/typecheckers/test_type.py \
  tests/typecheckers/test_federation.py \
  tests/typecheckers/test_directives.py \
  -q
```

Result:

```text
10 passed
```

Formatting and repository checks also passed:

```bash
uv run pre-commit run --files \
  strawberry/types/object_type.py \
  strawberry/federation/object_type.py \
  strawberry/schema_directive.py \
  strawberry/federation/schema_directive.py \
  tests/typecheckers/test_type.py \
  tests/typecheckers/test_federation.py \
  tests/typecheckers/test_directives.py \
  RELEASE.md

git diff --check
```

Fixes #3790

## Summary by Sourcery

Align Strawberry's dataclass transform metadata with its runtime behavior by disabling default ordering in all applicable decorators.

Bug Fixes:
- Correct dataclass transform ordering metadata so Strawberry-decorated classes are accurately modeled by type checkers and can define custom comparison methods such as `__gt__`.
- Apply the metadata correction consistently across core, federation, and schema directive decorators.

Tests:
- Add type-checker regression coverage for custom ordering methods on core and federation types and schema directives.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,17 @@
+---
+release type: patch
+social_messages:
+  x: >-
+    {project_name} {version} is out! This release fixes dataclass transform
+    metadata so custom ordering methods are correctly accepted by type checkers. 🍓
+    https://strawberry.rocks/release/{version}
+  linkedin: >-
+    {project_name} {version} is out. This release fixes dataclass transform
+    metadata so custom ordering methods are correctly accepted by type checkers.
+---
+
+This release fixes incorrect dataclass transform ordering metadata.
+
+Strawberry decorators now correctly declare that ordering methods are not generated
+by default, matching their runtime dataclass behavior and allowing custom ordering
+methods such as `__gt__` to be used without type-checking errors.
--- a/strawberry/federation/object_type.py
+++ b/strawberry/federation/object_type.py
@@ -58,7 +58,7 @@ def _impl_type(
 
 @overload
 @dataclass_transform(
-    order_default=True,
+    order_default=False,
     kw_only_default=True,
     field_specifiers=(base_field, field, StrawberryField),
 )
@@ -74,7 +74,7 @@ def type(
 
 @overload
 @dataclass_transform(
-    order_default=True,
+    order_default=False,
     kw_only_default=True,
     field_specifiers=(base_field, field, StrawberryField),
 )
@@ -106,7 +106,7 @@ def type(
 
 @overload
 @dataclass_transform(
-    order_default=True,
+    order_default=False,
     kw_only_default=True,
     field_specifiers=(base_field, field, StrawberryField),
 )
@@ -124,7 +124,7 @@ def input(
 
 @overload
 @dataclass_transform(
-    order_default=True,
+    order_default=False,
     kw_only_default=True,
     field_specifiers=(base_field, field, StrawberryField),
 )
@@ -163,7 +163,7 @@ def input(
 
 @overload
 @dataclass_transform(
-    order_default=True,
+    order_default=False,
     kw_only_default=True,
     field_specifiers=(base_field, field, StrawberryField),
 )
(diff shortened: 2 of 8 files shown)


bellini666 (approved):
Oh, really nice fix. Thank you! :)

botberry:
This PR was published as [0.324.5](https://github.com/strawberry-graphql/strawberry/releases/tag/0.324.5). Thank you for contributing!

PR #4606, opened 2026-08-31
Use a bounded LRU cache by default in ParserCache and ValidationCache

patrick91:
## Summary by Sourcery

Bound parser and validation caches by default to protect publicly exposed servers from unbounded memory growth.

Bug Fixes:
- Prevent unbounded memory growth from unique query texts by making ParserCache and ValidationCache bounded by default.

Enhancements:
- Set both caches to a 128-entry least-recently-used limit by default while preserving explicitly configurable, including unbounded, cache sizes.

Documentation:
- Document the new 128-entry defaults, eviction behavior, and risks of opting into unbounded caches.

Tests:
- Add coverage verifying that both caches retain at most 128 entries by default.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,42 @@
+---
+release type: minor
+social_messages:
+  x: >-
+    {project_name} {version} is out! This release bounds the ParserCache and
+    ValidationCache extensions by default, protecting servers from unbounded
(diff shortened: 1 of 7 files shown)


botberry:
This PR was published as [0.327.0](https://github.com/strawberry-graphql/strawberry/releases/tag/0.327.0). Thank you for contributing!

PR #4611, opened 2026-09-06
Reduce DataLoader allocation and enqueue overhead

patrick91:
DataLoader currently constructs a parameterized generic task object for every
uncached key. Use slotted batch/task records and plain, statically annotated task
construction to reduce that overhead. Keep the shared enqueue helpers and trim
batch-selection lookups without changing dispatch timing.

Add coverage for batching across event-loop turns, priming an entire pending
batch, and partial cancellation during a running load; exercise existing batch
limits and cancellation tests with both cache modes. Include small/large batch
benchmarks and a patch release note.

Validation: 786 tests passed across schema, Django, DataLoader, and the new
benchmark cases (14 skipped, 2 xfailed). All 36 DataLoader tests also pass against
main. Repository Ruff lint/format checks and mypy pass.

Against main f9bbe8e9 on CPython 3.14.1/macOS arm64, local median time for 10,000
keys falls from 30.07 to 23.65 ms without caching and 31.47 to 25.06 ms with an
empty enabled cache. Ten-key batches improve by 6–8%; single-key changes are small.
These measurements isolate DataLoader overhead with one async yield, not database
latency. The full CI version matrix remains to be run.

Review note: slots remove dynamic attributes and weak references from LoaderTask
and Batch, implementation records that are exported from the module.

## Summary by Sourcery

Optimize DataLoader’s internal batch and task handling to improve throughput without changing dispatch semantics.

Enhancements:
- Reduce DataLoader allocation and enqueue overhead while preserving batching, caching, priming, and cancellation behavior.

Tests:
- Expand DataLoader coverage across event-loop batching, batch priming, partial cancellation, batch limits, and both cache modes.
- Add small and large DataLoader performance benchmarks.

Chores:
- Add patch release notes for the DataLoader performance improvements.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,17 @@
+---
+release type: patch
+social_messages:
+  x: >-
+    {project_name} {version} is out! DataLoader now spends less time allocating
+    and enqueueing batch entries. https://strawberry.rocks/release/{version}
+  linkedin: >-
+    {project_name} {version} is out! We've reduced DataLoader's allocation and
+    enqueue overhead while preserving batching, caching, and cancellation
+    behavior. No application changes are needed.
+---
+
+This release fixes unnecessary allocation and enqueue overhead in DataLoader.
+
+Batch entries now use slotted dataclasses and avoid runtime generic construction.
+Batch selection also avoids repeated attribute lookups and length-method calls.
+Existing batching, caching, priming, and cancellation behavior is preserved.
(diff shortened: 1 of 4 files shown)


botberry:
This PR was published as [0.327.4](https://github.com/strawberry-graphql/strawberry/releases/tag/0.327.4). Thank you for contributing!

PR #4613, opened 2026-09-06
Fix Channels HTTP encode_json overrides

patrick91:
Channels HTTP responses previously called `json.dumps` directly, bypassing custom `encode_json` overrides. Both synchronous and asynchronous consumers now use the established hook, passing bytes through unchanged and encoding strings as UTF-8. Status, headers, content type, and single/batch error handling are preserved.

Adds 16 regression cases covering both consumers, string and bytes overrides, single and batch responses, GraphQL errors, and custom status and headers. Includes a patch release note; the existing documentation already describes the hook.

Validation:
- Channels and shared HTTP suite (`pytest tests/channels tests/http -m channels -q`): 352 passed, 55 skipped, 16 xfailed, 2 xpassed.
- All 16 new cases fail with the previous response implementation restored in memory.
- Repository-wide Ruff lint, formatting, and mypy checks passed.
- `git diff --check` passed.

## Summary by Sourcery

Ensure Channels HTTP consumers use the established JSON encoding hook for all GraphQL responses.

Bug Fixes:
- Fix Channels HTTP responses so synchronous and asynchronous consumers respect custom `encode_json` overrides for single and batched responses, including GraphQL errors.
- Preserve custom JSON output returned as bytes or strings while retaining response status and headers.

Tests:
- Add regression coverage for both Channels consumers, string and bytes encoders, single and batched responses, GraphQL errors, and custom response metadata.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,19 @@
+---
+release type: patch
+social_messages:
+  x: >-
+    {project_name} {version} is out! Channels HTTP responses now respect custom
+    encode_json overrides returning str or bytes.
+    https://strawberry.rocks/release/{version}
+  linkedin: >-
+    {project_name} {version} is out! This release fixes custom JSON encoding for
+    Channels HTTP responses. Both synchronous and asynchronous consumers now
+    respect encode_json overrides returning str or bytes.
+---
+
+This release fixes Channels HTTP responses bypassing custom `encode_json`
+overrides.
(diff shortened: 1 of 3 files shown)


botberry:
This PR was published as [0.327.5](https://github.com/strawberry-graphql/strawberry/releases/tag/0.327.5). Thank you for contributing!

